PR #1, opened 2024-10-02
PRSD-388: Adds initial build and test pipeline

Travis-Softwire:
Note: The Build & Test pipeline will be altered to only trigger on raising a PR once we have a deployment pipeline. At that point the slack reporting step will also move to the deployment pipeline.

Also, TODO: Create a PR template!


gradlew
Travis-Softwire, line 1: This is a permissions change to make `gradlew` executable

rowanhill: Yep, I've noticed was necessarily locally (on WSL), too.

rowanhill (approved):
Thanks, this looks like a sensible start!

PR #2, opened 2024-10-07
PRSD-NONE: Prevents error on hot reload from re-seeding data

Travis-Softwire:
Updates hibernate config when running locally to avoid crash from incorrectly re-seeding the database when hot-reloading


Kiran-R-K (approved):
👍

PR #3, opened 2024-10-07
PRSD-NONE: Sets up Playwright

isobel-softwire:
- Adds Playwright as a dependency
- Adds Playwright and TestContainer set-up for integration tests
- Replaces Selenium integration tests with Playwright ones
- **Note TODO: Figure out how to mock roles with Playwright**


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/IntegrationTest.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    @MockBean
+    lateinit var mockDefaultAuthorizationCodeTokenResponseClient: DefaultAuthorizationCodeTokenResponseClient
+
+    @MockBean
+    lateinit var jwtDecoderFactory: JwtDecoderFactory<ClientRegistration?>
+
+    @MockBean
+    lateinit var oneLoginConfig: OneLoginConfig
+}
```
Travis-Softwire, line 33: I think on reflection let's disable security for the Playwright tests until we've got the One Login mocking working. Otherwise we'll be pretty limited in what we can do in these tests. I think that just means adding a new `MockBean` here of type `SecurityFilterChain` (and updating the assertions in some of the Playwright tests).

Travis-Softwire (commented):
Thanks for picking this up - I reflected a bit on this since we chatted about it and I think for now we should just disable security for the Playwright tests until we can mock the auth flow (which will hopefully be soon!). I've left a comment about the config change we'll need. Otherwise LGTM :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/IntegrationTest.kt
isobel-softwire: This works but throws the following error: An Authentication object was not found in the SecurityContext

Travis-Softwire: Ah of course, this is because we have method security active which will want the Authentication object. Eventually this will go away once we get One Login mocking working, but for now let's disable method security when running the integration tests (we briefly chatted about an approach for this).

Travis-Softwire (approved):
LGTM :)

PR #7, opened 2024-10-10
Feat/prsd 358 create controller and service

JasminConterioSW:
This isn't quite there but I find that PRs are a useful place to ask questions!

In this PR:
- Add LA_ADMIN or LA_USER roles if a user logs in who is in the local_authority_users database
- Add a ManageLocalAuthorityUsersController which loads a placeholder page if the logged in user is an LA_ADMIN
- Get a list of Local Authority Users who are in the same local authority as the logged in user, show their userNames and access levels (isManager)
- Added tests for the controller to check that the page only checks shown to authorised LA_ADMIN users

To do - add some tests for the services:
- test the UserRolesService.getRolesForSubjectId function - are the landlord, la_user and la_admin roles getting added correctly?
- test the LocalAuthorityDataService.getLocalAuthorityUsersForLocalAuthority - this is doing some mapping so check it's being returned correctly
- test the getLocalAuthorityForUser service - this should return a Local Authority if the logged in user is in the local_authority_users database and should return null if they are not.


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/LocalAuthorityUserRepository.kt
```diff
@@ -3,4 +3,11 @@ package uk.gov.communities.prsdb.webapp.database.repository
...
 import uk.gov.communities.prsdb.webapp.database.entity.LocalAuthorityUser
 
-interface LocalAuthorityUserRepository : JpaRepository<LocalAuthorityUser?, Long?>
+interface LocalAuthorityUserRepository : JpaRepository<LocalAuthorityUser?, Long?> {
+    // The underscore tells JPA to access fields relating to the referenced table
+    @Suppress("ktlint:standard:function-naming")
+    fun findByLocalAuthority_Id(localAuthorityId: Int): List<LocalAuthorityUser>
+
+    @Suppress("ktlint:standard:function-naming")
+    fun findByBaseUser_Id(userName: String): LocalAuthorityUser?
```
JasminConterioSW, line 12: This version of findByBaseUser_Id is different to what is currently in the LandlordUserRepositry - the landlord vesion returns a list instead of a single user.

The DB enforces that subject_identifier is unique, so it makes sense to me to return one user instead of a list, but what do we prefer? I can made the LandlordUser and LocalAuthorityUser versions work the same way, it's just a question of which!

src/main/resources/data-local.sql
```diff
@@ -1,14 +1,19 @@
 INSERT INTO one_login_user (id, name, email, created_date, last_modified_date)
 VALUES ('urn:fdc:gov.uk:2022:ABCDE', 'Bob T Builder', 'bobthebuilder@gmail.com', '09/13/24', '09/13/24'),
        ('urn:fdc:gov.uk:2022:FGHIJ', 'Anne Other', 'Anne.Other@hotmail.com', '09/13/24', '09/13/24'),
-       ('urn:fdc:gov.uk:2022:KLMNO', 'Ford Prefect', 'Ford.Prefect@hotmail.com', '10/07/24', '10/07/24');
+       ('urn:fdc:gov.uk:2022:KLMNO', 'Ford Prefect', 'Ford.Prefect@hotmail.com', '10/07/24', '10/07/24'),
+       ('urn:fdc:gov.uk:2022:PQRST', 'Arthur Dent', 'Arthur.Dent@hotmail.com', '10/09/24', '10/09/24'),
+       ('urn:fdc:gov.uk:2022:07lXHJeQwE0k5PZO7w_PQF425vT8T7e63MrvyPYNSoI', 'Jasmin Conterio', 'jasmin.conterio@softwire.com', '10/07/24', '10/07/24');
```
JasminConterioSW, line 6: I've given myself some roles so I can see the pages when I am logged in... should I add something to the README about how to do this?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/LocalAuthorityUserRepository.kt
Travis-Softwire: Your approach sounds v. sensible to me - happy for you to update the Landlord version to match

src/main/resources/data-local.sql
Travis-Softwire: Yes please on updating the README. On One Login, we will eventually need a set of dummy accounts with different roles, as when we turn on identity verification there are only a predetermined set of answers that will pass verification in the integration environment. Could you set up those accounts as part of this ticket (using the team DL and plus encoding for the email addresses - shout if that doesn't make sense!)?

JasminConterioSW: Do you mean something like teamDL+landlord@softwire.com?
What is the team DL I should be using?

MFA is going to be annoying for dummy accounts, I think it will work though (just whenever someone logs in they might have to change where the verification code is going to).

Also is there a way to log out and change account? Thanks!

Travis-Softwire: Yes, exactly that. It should be Team-PRSDB. 

Good point about MFA - I think a partial answer will be to use a script to generate the OTOP, e.g. https://docs.sign-in.service.gov.uk/test-your-integration/using-integration-for-testing/#generate-a-one-time-code-using-a-scripting-language but it will require some thought about how to make that as smooth a process as possible

For logging out - there is, but we haven't implemented it yet. I have a few One Login - related tasks that I'm going to add to next sprint so I'll include this as well.

JasminConterioSW:
I've now added some service tests, updated the landlord repository and updated the readme.
I haven't added the teamDL dummy accounts yet (mainly because I need to wait for it to log me out...) but can put that in the next PR

src/main/resources/data-local.sql
JasminConterioSW: Is sorting out the MFA going to be a different ticket?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.ui.Model
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import uk.gov.communities.prsdb.webapp.constants.SERVICE_NAME
+import uk.gov.communities.prsdb.webapp.services.LocalAuthorityDataService
+import java.security.Principal
+
+@PreAuthorize("hasRole('LA_ADMIN')")
+@Controller
+@RequestMapping("/manage-users")
```
Travis-Softwire, line 16: Not one to worry about right now, but eventually there will be multiple pages for managing users for different user-types (e.g. agents, possibly company landlords tbc). So I expect we will want to group LA-specific endpoints under a common path - I'm open to suggestions on what that should be :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
```diff
@@ -0,0 +1,39 @@
...
+    val localAuthorityDataService: LocalAuthorityDataService,
+) {
+    @GetMapping
+    fun index(
+        model: Model,
+        principal: Principal,
+    ): String {
+        val currentUserLocalAuthority = localAuthorityDataService.getLocalAuthorityForUser(principal.name)
+        if (currentUserLocalAuthority?.id != null) {
+            // We should always get to here as only LA_ADMINs can access this page
```
Travis-Softwire, line 27: If we're 100% confidence that it can't be null you can coerce it with the `!!` operator (but this will throw an exception if you access a property on it and it's not null). If we're not 100% confident then we should probably handle the case where it is null.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthority.kt
```diff
@@ -10,7 +10,7 @@ import jakarta.persistence.Id
 class LocalAuthority : AuditableEntity() {
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
-    private val id: Int? = null
+    val id: Int? = null
```
Travis-Softwire, line 13: Did the id being private cause an issue?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityDataService.kt
```diff
@@ -0,0 +1,25 @@
...
+) {
+    fun getLocalAuthorityUsersForLocalAuthority(localAuthorityId: Int): List<LocalAuthorityUserDataModel> {
+        val usersInThisLocalAuthority = localAuthorityUserRepository.findByLocalAuthority_Id(localAuthorityId)
+        return usersInThisLocalAuthority.map { LocalAuthorityUserDataModel(it.baseUser.name, it.isManager) }
+    }
+
+    fun getLocalAuthorityForUser(subjectId: String): LocalAuthority? {
+        val localAuthorityUser = localAuthorityUserRepository.findByBaseUser_Id(subjectId)
+        if (localAuthorityUser == null) {
+            return null
```
Travis-Softwire, line 20: I think we could just return `localAuthorityUser?.localAuthority` here?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthority.kt
JasminConterioSW: It wasn't letting me access it in the ManageLocalAuthorityUsersController - I needed currentUserLocalAuthority.id

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/UserRolesService.kt
```diff
@@ -2,17 +2,30 @@ package uk.gov.communities.prsdb.webapp.services
...
+        if (matchingLandlordUser != null) {
             roles.add("ROLE_LANDLORD")
         }
+
+        val matchingLocalAuthorityUser = localAuthorityUserRepository.findByBaseUser_Id(subjectId)
+        if (matchingLocalAuthorityUser != null) {
+            if (matchingLocalAuthorityUser.isManager) {
+                roles.add("ROLE_LA_ADMIN")
+            } else {
+                roles.add("ROLE_LA_USER")
```
Travis-Softwire, line 25: I think we probably want LA Admins to also be LA Users? But open to being persuaded if you think it's better to keep them as distinct groups

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
JasminConterioSW: The id gets initialised to null in the entity class, but presumably there was always be an Id generated for any data added to the table? As long as that is true, I think this can't be null:

- The PreAuthorize tag means that only users with the LA_ADMIN role can access the page
- The user must be in the LocalAuthorityUser database to get the LA_ADMIN role. The local_authority_id column has the `@OneToOne(optional=false)` annotation so must have a non-null local_authority_id

image

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/UserRolesService.kt
JasminConterioSW: That's fine, I wasn't sure! But I guess there won't be anything the user can do that the admin can't so that probably does make sense

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
JasminConterioSW: I guess these are all going to come in an admin section as well, so it might end up being something like

admin/manage-la-users
admin/manage-landlords

etc?

Travis-Softwire: They are, but because of them potentially needing different URLs we might well end up having the endpoints for specific 'service levels' (i.e. landlords, LAs, and the public) sitting underneath common routes so that we can easily route requests to the correct endpoint. So I _think_ it's more likely to end up being something like `la/admin/manage-users` and `landlords/agent/manage-users`, or something along those lines

Travis-Softwire (approved):
Subject to the point about looking up by entity rather than Id I'm happy with this :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthority.kt
JasminConterioSW: Ah ok, thanks!

PR #8, opened 2024-10-14
Feat/PRSD-364 integrate with notify

AEPR:
Initial pull request for integrating with notify. This adds:
* Configuration to use a .env file for secrets in the code (I'll make a usable one available separately)
* A type-parameterised email sending service
* An email template for the test email in notify already
* A page at '/send-test-email' that allows you to send a test email to a specified email address.
* A single very simple test for the email sending service

Worth noting:
* I've not added any controller tests as the controller is only to demo notify integration and won't be used by the time we reach release 1. Let me know if you disagree.* 
* I've deliberately made the template for the email page as simple as possible, not trying to avoid repitition, for the same reason.
* I've left a TODO in the tests file mentioning the need for tests integrating with Notify to explain a slight gap currently left in the testing. Let me know if we should follow a particular convention here.
* I've not covered any of the error handling we talked about - I'll catch up with you and then add unhappy paths. As it is any errors will just result in reaching the "Sorry there's a problem" page. 
* The NotifyEmailNotificationService<TestEmail> appears to Just Work (tm) with dependency injection without any additional boilerplate.

Edit: Also, I've not checked with other branches where we're planning to put viewmodels in the file structure - very happy to move them to match on merge.

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/viewmodel/EmailTemplate.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/viewmodel/EmailTemplate.kt
@@ -1,0 +1,7 @@
+package uk.gov.communities.prsdb.webapp.viewmodel
+
+interface EmailTemplate {
+    val templateId: String
+
+    fun asHashMap(): HashMap<String, String>
+}
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/viewmodel/TestEmail.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/viewmodel/TestEmail.kt
@@ -2,10 +2,10 @@ package uk.gov.communities.prsdb.webapp.viewmodel
 
 data class TestEmail(
     var firstName: String,
-) {
+) : EmailTemplate {
     private val firstNameName = "first name"
 
-    val templateId = "71551da6-f616-45c7-a4e0-23d6a8434561"
+    override val templateId = "71551da6-f616-45c7-a4e0-23d6a8434561"
 
-    fun asHashMap(): HashMap<String, String> = hashMapOf(firstNameName to firstName)
+    override fun asHashMap(): HashMap<String, String> = hashMapOf(firstNameName to firstName)
 }
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/viewmodel/TestEmail.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/viewmodel/TestEmail.kt
@@ -1,0 +1,11 @@
+package uk.gov.communities.prsdb.webapp.viewmodel
+
+data class TestEmail(
+    var firstName: String,
+) {
+    private val firstNameName = "first name"
+
+    val templateId = "71551da6-f616-45c7-a4e0-23d6a8434561"
+
+    fun asHashMap(): HashMap<String, String> = hashMapOf(firstNameName to firstName)
+}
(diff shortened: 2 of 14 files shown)


Travis-Softwire:
> * I've not added any controller tests as the controller is only to demo notify integration and won't be used by the time we reach release 1. Let me know if you disagree.*

Yep, agree. Could you add a TODO referencing ticket PRSD-404 which will be the first to use the new service, at which point this endpoint can be deleted. I've added a comment on that ticket as a reminder to look for todos referencing it.

> * I've deliberately made the template for the email page as simple as possible, not trying to avoid repitition, for the same reason.

Sounds good

> * I've left a TODO in the tests file mentioning the need for tests integrating with Notify to explain a slight gap currently left in the testing. Let me know if we should follow a particular convention here.

I wasn't quite sure about this as a TODO - I think in my mind a TODO is something that will be done rather than an ongoing reminder (and ideally references the ticket where it will be done). We do need to document this need though - I think for now let's add it to the ReadMe and I'll also create a task for myself to create a PR checklist that includes this

> * I've not covered any of the error handling we talked about - I'll catch up with you and then add unhappy paths. As it is any errors will just result in reaching the "Sorry there's a problem" page. 

👍

> * The NotifyEmailNotificationService appears to Just Work (tm) with dependency injection without any additional boilerplate.

Ah, nice :) 

I've also added a couple of super super minor comments, but otherwise this all looks great

src/main/kotlin/uk/gov/communities/prsdb/webapp/viewmodel/EmailTemplate.kt
```diff
@@ -0,0 +1,7 @@
+package uk.gov.communities.prsdb.webapp.viewmodel
+
+interface EmailTemplate {
```
Travis-Softwire, line 3: I paused over this as the name just because we presumably will also have html 'email templates' elsewhere once you've worked out the best way to include those in source control. Ofc naming things is _hard_ so I don't have any amazing suggestions for this, the best I could come up with was `EmailTemplateModel` but very open to suggestions

src/main/kotlin/uk/gov/communities/prsdb/webapp/viewmodel/EmailTemplate.kt
```diff
@@ -0,0 +1,7 @@
...
+    val templateId: String
+
+    fun asHashMap(): HashMap<String, String>
```
Travis-Softwire, line 6: Extremely minor, but I feel like the convention for returning a different representation of a class's data is `toX` rather than `asX`

src/main/kotlin/uk/gov/communities/prsdb/webapp/viewmodel/TestEmail.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.communities.prsdb.webapp.viewmodel
+
+data class TestEmail(
+    var firstName: String,
+) : EmailTemplate {
+    private val firstNameName = "first name"
```
Travis-Softwire, line 6: No need to change it here given this is temporary, but I bounced off of the double name a bit - perhaps in this context as there'll be a few fields containing 'names' we could instead go with `firstNameKey` or similar? But again with naming things being hard!

src/main/kotlin/uk/gov/communities/prsdb/webapp/viewmodel/EmailTemplate.kt
AEPR: I think I like "EmailTemplateModel" because it's more descriptive - this is the data model for a template not actually a template itself. Also why the actual subclass I've made doesn't have template in its name (but could have model!)

src/main/kotlin/uk/gov/communities/prsdb/webapp/viewmodel/TestEmail.kt
AEPR: I am actually going to update it to key now as when someone comes to do PRSD-404 they'll probably just follow the convention already here!

Travis-Softwire (approved):

PR #9, opened 2024-10-14
Chore/PRSD-NONE: mock one login endpoint

Kiran-R-K:
# Description

This PR creates a mock of the One Login Oauth2 API that will automatically run when the project is started in the `local` profile. It will log the user in as a user who has all possible roles.

## Testing
This has been manually tested to see that it works but there are no automated tests.

## Possible concerns

I have not mocked the `logout` endpoint because it requires our web app to have a logout journey/route which we don't yet. If we implement one we will need to also add a `logout` endpoint to the mock.

Error handling - I have not added error handling to these endpoints - I'm not sure it's necessary for our purpose

## Documentation
I've added information to the `README` about this mock, including: how to use it, how to change the user you're logged in as and a reminder to add any new roles to the `userId` we are using to login.


isobel-softwire (commented):
Mostly LGTM - left a few comments

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/mockOneLogin/MockOneLoginController.kt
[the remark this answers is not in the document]
Kiran-R-K: Good idea

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/mockOneLogin/MockOneLoginController.kt
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.http.MediaType
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
+import java.net.URI
+
+@Profile("local")
```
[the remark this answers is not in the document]
Kiran-R-K: I don't know if there's a convention already, the `local` profile had already been set up previously. Let's ask the rest of the dev team and see if there is a convention/decide one?

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/mockOneLogin/MockOneLoginHelper.kt
```diff
@@ -0,0 +1,196 @@
...
+
+class MockOneLoginHelper {
+    companion object {
+        val ecKey = generateSecretKey()
+
+        private fun generateSecretKey(): ECKey {
+            val ecKey =
+                ECKeyGenerator(Curve.P_256)
+                    .keyUse(KeyUse.SIGNATURE)
+                    .keyID("6a4bc1e3-9530-4d5b-90c5-10dcf3ffccd0")
```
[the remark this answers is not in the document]
Kiran-R-K: they're both random but the `keyId` needs to be used in 2 places.
I've changed the keyId to a random UUID saved to a variable so it can be re-used.

I've left the authorization code as is for now, do you think it could benefit from a name change or something to make it clearer that it's random/changeable?

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/mockOneLogin/MockOneLoginController.kt
```diff
@@ -0,0 +1,63 @@
...
+@RestController
+@RequestMapping("/one-login-local")
+class MockOneLoginController(
+    private var helper: MockOneLoginHelper = MockOneLoginHelper(),
```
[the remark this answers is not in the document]
Kiran-R-K: yeah, I tried out calling it `mockOneLoginHelper` but it ended up feeling like too much given there's only one helper in the file.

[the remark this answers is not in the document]
isobel-softwire: Fair enough, that makes sense

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/mockOneLogin/MockOneLoginHelper.kt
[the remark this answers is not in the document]
isobel-softwire: It's probably fine if we'll never need to change it?

ReadMe.md
```diff
@@ -80,4 +80,13 @@ Database migrations <will be/ are> run at deployment time for all non-local depl
...
 application start up. If you are using the `local` launch profile in IntelliJ, this will also run the `flywayClean` task
 before running the migrations. After the migrations have run Spring Boot will then run the SQL in `data-local.sql` to
-populate the database with seed data. 
\ No newline at end of file
+populate the database with seed data. 
+
+### Mock One Login Oauth2
+
+For development, we've mocked elements of the governments one login system (that the web app will be using in deployment).
+When you start the app using the `local` profile this will be available, when you attempt to login, This will automatically log you in as a user that has every role - and therefore can access all pages.
```
Travis-Softwire, line 88: Thinking about it, I wonder if this should be behind a `local-auth` profile or similar that we can then activate independently of `local`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/controller/MockOneLoginController.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
+import uk.gov.communities.prsdb.webapp.local.api.helper.MockOneLoginHelper
+import java.net.URI
+
+@Profile("local")
+@RestController
+@RequestMapping("/one-login-local")
+class MockOneLoginController(
+    private var helper: MockOneLoginHelper = MockOneLoginHelper(),
```
Travis-Softwire, line 18: Minor, but I wonder if this is more of a delegate than a helper? As it's broadly just handling the response in each case?

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/helper/MockOneLoginHelper.kt
```diff
@@ -0,0 +1,199 @@
...
+
+            return ecKey
+        }
+    }
+
+    private val userId = "urn:fdc:gov.uk:2022:PQRST"
+    private val userEmail = "julia.jones@hotmail.com"
+    private val userNumber = "07123456789"
+
+    val authorizationCode = "SplxlOBeZQQYbYS6WxSbIA"
```
Travis-Softwire, line 40: To avoid any concern from the casual reader, can we add a clarifying comment here somewhere explaining that these are all example values from the One Login docs (I think that's where they came from?)

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/helper/MockOneLoginHelper.kt
```diff
@@ -0,0 +1,199 @@
...
+
+        val signedJwt = SignedJWT(headerBuilder.build(), claimSetBBuilder.build())
+
+        signedJwt.sign(ECDSASigner(ecKey))
+
+        return signedJwt.serialize()
+    }
+
+    fun getOpenidConfigurationResponse(): String =
+        "{\n" +
```
Travis-Softwire, line 72: How much of a pain would it be to move these into files and parse them instead of having them inline?

Travis-Softwire (commented):
Have left a few Qs and suggestions, but otherwise this is looking good :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/controller/MockOneLoginController.kt
```diff
@@ -0,0 +1,64 @@
...
+    fun authorize(
+        @RequestParam response_type: String,
+        @RequestParam client_id: String,
+        @RequestParam scope: String,
+        @RequestParam state: String,
+        @RequestParam redirect_uri: String,
+        @RequestParam nonce: String,
+    ): ResponseEntity<Unit> {
+        helper.lastReceivedNonce = nonce
+        val locationURI: URI = URI.create("$redirect_uri?code=${helper.authorizationCode}&state=$state")
```
rowanhill, line 36: Don't we need to escape these parameter values? And if so, should we use something like UriComponentsBuilder?

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/controller/MockOneLoginController.kt
rowanhill: I wonder if there's much benefit having it as a separate class? I think all the code could just live on the controller, and it'd be okay?

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/helper/MockOneLoginHelper.kt
Kiran-R-K: changed some details and added this

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/controller/MockOneLoginController.kt
Kiran-R-K: I'm inclined to keep at least the `get{X}Response` methods separate for readability (especially considering we might be adding more endpoints in future), but I'm happy to be suggested otherwise as well

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/controller/MockOneLoginController.kt
Kiran-R-K: I don't think so, the URI is being built with them as expected, unless there are other things to consider in escaping?

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/helper/MockOneLoginHelper.kt
Kiran-R-K: I haven't done this yet, given there's also a thread about how many classes this should all have and that may affect the need for this. I have also shortened one of the responses by a lot which means there's now only one long one sand it's at the bottom of the file which I thinks helps readability/organisation in general

Travis-Softwire: Thinking about it, if these were being parsed in from files then we probably could just collapse this class into the controller as Rowan suggests given this is all just a local utility?

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/controller/MockOneLoginController.kt
Travis-Softwire: I've also commented below, but I think if we move the JSON into files that we parse in then the rest of the utility class probably can be inlined without the controller becoming too bloated

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/controller/MockOneLoginController.kt
rowanhill: Well, what if the authorizationCode or state had some special value (e.g. a & or a =) in the future? I think unless we have a strong guarantee that can never happen, we'd be better off URL-encoding these values.

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/api/controller/MockOneLoginController.kt
```diff
@@ -0,0 +1,162 @@
...
+        @RequestParam redirect_uri: String,
+        @RequestParam nonce: String,
+    ): ResponseEntity<Unit> {
+        lastReceivedNonce = nonce
+        val locationURI: URI =
+            UriComponentsBuilder
+                .newInstance()
+                .uri(URI.create(redirect_uri))
+                .query("code=SplxlOBeZQQYbYS6WxSbIA")
+                .query("state=$state")
```
rowanhill, line 97: Could we use .queryParam("state", state) instead? Then if state contains anything funky we don't expect (e.g. &s and =s) they'll be encoded as the value of the "state" query parameter, rather than being interpreted as new query parameters.

rowanhill (approved):

PR #10, opened 2024-10-15
PRSD-353: Registration Number Service

isobel-softwire:
### Overarching Changes
- Creates base `ModifiableAuditableEntity` and makes `AuditableEntity` its descendant.
- Creates base `ServiceTest`
- Creates `RegisteredEntity` interface for entities with registration numbers (this allows for generic `RegistrationNumberService` methods)

### Features
- Adds `RegistrationNumber` and a `Landlord` skeleton to the database. 
- Creates `RegistrationNumberService` that can:
  - Create registration numbers
  - Convert between integer and formatted registration numbers
  - Retrieve an entity based on its registration number
- Displays a formatted registration number on the webapp's landing page for PoC.

### Tests
- Tests for `RegistrationNumberService` methods

### Notes
- Decided against displaying a registration number based on seed data as the features required for this are outside this ticket's scope.
- TODOs left in `RegistrationNumberService.retrieveEntity` as entities with registration numbers (apart from landlord) are yet to be implemented.


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/LandlordRepository.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.communities.prsdb.webapp.database.repository
+
+import org.springframework.data.jpa.repository.JpaRepository
+import uk.gov.communities.prsdb.webapp.database.entity.Landlord
+
+interface LandlordRepository : JpaRepository<Landlord?, Long?> {
+    fun findByRegistrationNumberNumber(registrationNumber: Long): Landlord?
```
JasminConterioSW, line 7: Does this function name need an underscore, similar to `findByBaseUser_Id` in the LandlordUserRepository? 
So `findByRegistrationNumber_Number`

isobel-softwire: It does - missed this, thanks!

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberService.kt
```diff
@@ -0,0 +1,62 @@
...
+import uk.gov.communities.prsdb.webapp.database.entity.RegistrationNumber
+import uk.gov.communities.prsdb.webapp.database.repository.LandlordRepository
+import uk.gov.communities.prsdb.webapp.database.repository.RegistrationNumberRepository
+
+@Service
+class RegistrationNumberService(
+    val regNumRepository: RegistrationNumberRepository,
+    val landlordRepository: LandlordRepository,
+) {
+    fun createRegistrationNumber(type: RegistrationNumberType) {
```
Travis-Softwire, line 19: Does this need to be inside a transaction to avoid a read/ write race condition?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberService.kt
```diff
@@ -0,0 +1,62 @@
...
+        regNumRepository.save(RegistrationNumber(type, generateUniqueRegNum()))
+    }
+
+    fun retrieveEntity(formattedRegNum: String): RegisteredEntity? {
+        val charsetRegNum = formattedRegNum.substring(2, 6) + formattedRegNum.substring(7)
```
Travis-Softwire, line 24: Could this be refactored to rely on a constant for the size of the segments in the formatted number?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberService.kt
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    fun formatRegNum(regNum: RegistrationNumber): String {
+        var formattedRegNum = ""
+        var quotient = regNum.number!!
+        while (quotient > 0) {
+            formattedRegNum = REG_NUM_CHARSET[(quotient % REG_NUM_BASE).toInt()] + formattedRegNum
+            quotient /= REG_NUM_BASE
+        }
+        formattedRegNum = formattedRegNum.padStart(8, REG_NUM_CHARSET[0])
```
Travis-Softwire, line 42: Is this 8 the length of the registration number? If so we should probably rely on that constant here instead

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberService.kt
```diff
@@ -0,0 +1,62 @@
...
+
+        return regNum.type.toInitial() + "-" + formattedRegNum.substring(0, 4) + "-" + formattedRegNum.substring(4)
```
Travis-Softwire, line 44: Can we have a constant for the length of the segments?

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberServiceTests.kt
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    @Test
+    fun `retrieveEntity returns null given a non-existent registration number`() {
+        val decRegNum = 1L
+        val formattedRegNum = "L-CCCC-CCCC"
+        val landlord = Landlord()
+
+        `when`(mockLandlordRepository.findByRegistrationNumber_Number(decRegNum)).thenReturn(landlord)
```
Travis-Softwire, line 69: I think probably slightly clearer and simpler to explicitly mock it returning null when receiving the equivalent to the reg number you're querying, as otherwise you're implicitly testing the mock's default behaviour.

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberServiceTests.kt
```diff
@@ -0,0 +1,87 @@
...
+    fun `retrieveEntity retrieves a landlord given their registration number`() {
+        val decRegNums =
+            listOf(
+                MIN_REG_NUM,
+                MAX_REG_NUM,
+            )
+        val formattedRegNums = listOf("L-CCCC-CCCC", "L-9999-9999")
+        val landlord = Landlord()
+
+        for (i in decRegNums.indices) {
```
Travis-Softwire, line 56: Appreciate in this case it's only 2, but it's good practice to use parameterized tests where you want to test multiple values - happy to chat about it if it's not something you've come across before!

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberServiceTests.kt
```diff
@@ -0,0 +1,87 @@
...
+
+        `when`(mockLandlordRepository.findByRegistrationNumber_Number(decRegNum)).thenReturn(landlord)
+
+        assertNull(regNumService.retrieveEntity(formattedRegNum))
+    }
+
+    @Test
+    fun `formatRegNum returns a formatted registration number`() {
+        val regNums =
+            listOf(
```
Travis-Softwire, line 77: As above, this can be parameterized :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/ServiceTest.kt
```diff
@@ -0,0 +1,17 @@
+package uk.gov.communities.prsdb.webapp.services
+
+import org.junit.jupiter.api.BeforeEach
+import org.mockito.Mockito.mock
+import uk.gov.communities.prsdb.webapp.database.repository.LandlordRepository
+import uk.gov.communities.prsdb.webapp.database.repository.RegistrationNumberRepository
+
+abstract class ServiceTest {
+    lateinit var mockRegNumRepository: RegistrationNumberRepository
```
Travis-Softwire, line 9: I think we should only really have a base class like this where the mocks are going to be needed by all or nearly all of the tests that inherit from it. In this case I think it makes more sense for the individual services to mock out the repositories that they rely on, which will vary quite a lot from service to service.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/RegistrationNumber.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    @Column(nullable = false, unique = true)
+    var number: Long? = null
+        private set
+
+    @Column(nullable = false)
+    lateinit var type: RegistrationNumberType
+        private set
+
+    constructor(type: RegistrationNumberType, number: Long) : this() {
```
Travis-Softwire, line 24: I don't _think_ you need this as JPA should generate it behind the scenes?

Travis-Softwire (commented):
Overall looking good I think. I've left a few comments, but would appreciate @rowanhill's view as he designed the registration numbers :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/RegistrationNumberConstants.kt
```diff
@@ -0,0 +1,41 @@
...
+        'Y',
+        'Z',
+        '2',
+        '3',
+        '4',
+        '5',
+        '6',
+        '7',
+        '9',
+    )
```
rowanhill, line 33: The character set I'd proposed excluded L (as it's confusable with i and 1, especially if ever rendered lower case) but included Q (which I think is pretty distinct).

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/enums/RegistrationNumberType.kt
```diff
@@ -0,0 +1,20 @@
+package uk.gov.communities.prsdb.webapp.constants.enums
+
+enum class RegistrationNumberType {
+    PROPERTY,
+    LANDLORD,
+    AGENT,
+    ;
+
+    fun toInitial(): Char = this.toString()[0]
```
rowanhill, line 9: Could we add a test that ensures the initials are unique? Just to guard against someone adding a new enum value in the distant future and not realising that uniqueness is a requirement, or that this calculation exists here.

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/RegistrationNumberConstants.kt
isobel-softwire: Sorry, I must've misread. Fixed it now!

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberService.kt
```diff
@@ -0,0 +1,62 @@
...
+    val regNumRepository: RegistrationNumberRepository,
+    val landlordRepository: LandlordRepository,
+) {
+    fun createRegistrationNumber(type: RegistrationNumberType) {
+        regNumRepository.save(RegistrationNumber(type, generateUniqueRegNum()))
+    }
+
+    fun retrieveEntity(formattedRegNum: String): RegisteredEntity? {
+        val charsetRegNum = formattedRegNum.substring(2, 6) + formattedRegNum.substring(7)
+        val decRegNum = charsetToDec(charsetRegNum)
```
rowanhill, line 25: I think we should pull out a method that parses strings into RegistrationNumber objects, and I think it should be _much_ more permissive about the formats it accepts. The goal should be to parse any string that a user might reasonably type in and mean it to be a registration number - for example spaces in unexpected places, or no separators at all between the alphanumeric bits, etc etc. And we should have a good set of tests covering all the whacky types of input we can think of.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/RegistrationNumber.kt
```diff
@@ -0,0 +1,28 @@
...
+import uk.gov.communities.prsdb.webapp.constants.enums.RegistrationNumberType
+
+@Entity
+class RegistrationNumber() : AuditableEntity() {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private val id: Long? = null
+
+    @Column(nullable = false, unique = true)
+    var number: Long? = null
```
rowanhill, line 17: @Travis-Softwire did we not previously say we'd not use nullable types for not-nullable columns, and use lateinit instead? Is there some reason that's not the best choice, here?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberService.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    private fun generateUniqueRegNum(): Long {
+        var registrationNumber: Long
+        do {
+            registrationNumber = (MIN_REG_NUM..MAX_REG_NUM).random()
+        } while (regNumRepository.existsByNumber(registrationNumber))
+        return registrationNumber
+    }
+
+    private fun charsetToDec(charsetRegNum: String): Long {
```
rowanhill, line 55: I don't think this is quite the right name - we're not converting to decimal, we're converting to an integer (which has no implicit base, because a base only applies to a written representation, i.e. a string). Maybe just "charsetStringToLong" would be better, or perhaps you can suggest something else?

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberServiceTests.kt
isobel-softwire: The mock returns null by default so explicitly returning null would not test that the right argument has been passed to `mockLandlordRepository.findByRegistrationNumber_Number`. I'll change these tests to capture and make assertions about the argument passed instead.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberService.kt
```diff
@@ -0,0 +1,62 @@
...
+@Service
+class RegistrationNumberService(
+    val regNumRepository: RegistrationNumberRepository,
+    val landlordRepository: LandlordRepository,
+) {
+    fun createRegistrationNumber(type: RegistrationNumberType) {
+        regNumRepository.save(RegistrationNumber(type, generateUniqueRegNum()))
+    }
+
+    fun retrieveEntity(formattedRegNum: String): RegisteredEntity? {
```
rowanhill, line 23: I'm not certain whether we want to have a method on RegistrationNumberService that looks up arbitrary entities by regnum, or if we want a method on each entity's service that a) parses the reg num, b) throws/returns null if it's of the wrong type, c) looks up the entity by regnum. I'm just thinking that something somewhere is going to have to do the "check this number is for the type of entity I expect", and that probably belongs in the services somewhere.

Maybe let's leave this for now, but be ready to refactor it into different methods. Agree, @Travis-Softwire?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/RegistrationNumber.kt
isobel-softwire: Kotlin won't let you use lateinit for primitives as they aren't nullable.

rowanhill: Hmm. I had assumed that Long was the boxed type, not the primitive (hence Long? being possible), so I would have thought Kotlin would be smart enough to initialise a lateinit Long to be null (despite the type not being nullable!) and throw a runtime exception if anything accesses the value while it's still null. Apparently not, though!

rowanhill: Yeah, looks like Kotlin uses the JVM primitive long for Long and the JVM boxed type Long for Long?. Feels like it could use the JVM type Long for lateinit Long, too, but seemingly it doesn't :)

https://kotlinlang.org/docs/numbers.html#numbers-representation-on-the-jvm

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/RegistrationNumber.kt
isobel-softwire: Do you mean generate the constructor? Or the RegistrationNumber object? I'm currently using it in `RegistrationNumberService.createRegistrationNumber` to set the attributes that are decided by the user/in the service.

rowanhill: I imagine we'd also want an explicit constructor for the "parse" method I suggested in another comment somewhere.

src/main/resources/db/migrations/V1_0_1__reg_num_and_landlord_skeleton.sql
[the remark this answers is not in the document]
isobel-softwire: I've changed it to v1_0_1 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberService.kt
Travis-Softwire: Ah, yes very good point. I actually wonder if it's worth just doing as part of this ticket - as this function only returns a `RegisteredEntity` which isn't likely to be what we'll want in practice.

src/main/resources/db/migrations/V1_0_1__reg_num_and_landlord_skeleton.sql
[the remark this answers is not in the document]
Travis-Softwire: Yes, good spot 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberService.kt
isobel-softwire: That makes sense. I'll replace this method with one that parses a registration number string into a registration number. And create a landlord service that provides the other functionality Rowan mentioned.

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberServiceTests.kt
Travis-Softwire: [Edit: this might be academic if you're planning to refactor away this method anyway!] 

Ah, ok - I was basing my comment off the name of the test - i.e. testing that if the repository returns null (because the thing it's searching for doesn't exist) then we want to be sure that the outer function also returns null. If the goal is to test the repository searches for the correct Long when given a specific formattedRegNum then you might want to rename the test as well

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/RegistrationNumber.kt
isobel-softwire: I think we should use a data model instead as the registration numbers returned from the parse method will only need `number` and `type` attributes

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberService.kt
isobel-softwire: Ok, I'll filter out non-alphanumeric characters and capitalise the rest. Is there anything else that should be permitted?

rowanhill: I can't think of anything else that wouldn't cover right now.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegistrationNumberService.kt
```diff
@@ -0,0 +1,74 @@
...
+            quotient /= REG_NUM_BASE
+        }
+        regNumString = regNumString.padStart(REG_NUM_LENGTH, REG_NUM_CHARSET[0])
+
+        return regNum.type.toInitial() +
+            "-" +
+            regNumString.substring(0, REG_NUM_SEG_LENGTH) +
+            "-" +
+            regNumString.substring(REG_NUM_SEG_LENGTH)
+    }
```
rowanhill, line 54: My gut instinct is that these methods (and their private friends) would be better placed as static methods on RegistrationNumberDataModel. It probably depends on whether we're aiming for fat models (data + behaviour) or thin models (data only), though.

What do other people think?

Travis-Softwire: To my mind it depends whether we see the 'public' representation of the reg number as part of the business logic or not. If it's business logic then I think it belongs in the service layer. 

If instead we see it as another way of presenting this particular type (equivalent to a toString function) then I agree `stringToRegNum`, `regNumToString` and `validateBaseRegNumString` would live in `RegistrationNumberDataModel`. I think `createRegistrationNumber` and `generateUniqueRegNum` should always be in the service layer though.

rowanhill: I think that's about right: I see these methods as presentation concerns. You could imagine, for example, implementing data binding using stringToRegNum (so our DTOs could hold a RegistrationNumberDataModel directly), and/or a Thymeleaf view calling toString() on a RegistrationNumberDataModel to get the nicely formatted version.

isobel-softwire: I've made this change 👍

rowanhill (approved):
Thanks for all the tweaks - this looks good to me now!

PR #11, opened 2024-10-15
PRSD-NONE: Changes integration test profile to use kebab case

isobel-softwire:
Changes integration test profile to use kebab case


Kiran-R-K (approved):
LGTM

PR #12, opened 2024-10-15
Feat/prsd 422 add form context to db

Kiran-R-K:
# Description

Created a table for storing the in-progress form context for any journeys.
Created `Journey_Type` enum
Made the table auditable
Added a Foreign Key between user id and id on one login database
Create a data model for the form context


Travis-Softwire (approved):
As discussed - I wasn't quite sure about the FormContextDataModel but let's see what it looks like once you start using it in the next PR

src/main/resources/db/migrations/V1_0_2__form_context_table.sql
rowanhill, line 1: I think this should be V1_2_0

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/FormContextDataModel.kt
rowanhill, line 1: I'm probably missing some context (no pun intended), but what will this DataModel class be used for?

Travis-Softwire: Kiran and I spoke and agreed that this would likely be refactored in the next PR as the implementation progresses (e.g. I'm expecting this or something similar to hold the deserialized form context) - so I'm viewing it as a placeholder for now

PR #13, opened 2024-10-16
Feat/prsd 364 add templates to souce control

AEPR:
This primarily updates tests (or testibility) of the email sending.

- Adds "Notify Template Tests" which compare between the resources in the codebase and Notify to ensure they match each other and that every Template with an ID registered in the codebase as a member of the EmailTemplateId enum is available.
- Adds "Email Template Model Tests" which will be the default tests to be added for each new email template model. This checks that the toHashmap function on that model returns a list of keys that corresponds to the "personalisation" tags in the version of the template text in resources.
- Modifies the "Notify Email Notification Service Tests" so that only the service itself is tests (not the Template Model). It does this by creating a Test Template and setting the hashmap returned to compare.

This does mean that we are not explicitly testing the pairing up of keys and values in the "toHashmap" function. However, I don't think this would be a meaningful test to add because it will essentially be restating the list of key/value pairs from the email template.

The "notify" tests should be ignored in the pipeline as the environment variable they depend on will not be set (so they will be ignored). We will need to add an appropriate environment variable to the pipelines!

Two small cleanup changes also included: 
- changing the environment variable name I was using so Spring does not magically overwrite the property value in application.yml
- renaming "TestEmail" to "ExampleEmail" to better match the template and be less confusing in the tests


.gitignore
```diff
@@ -42,6 +42,7 @@ out/
 ### Secrets ###
 *.pem
 *.env
+notify-template-tests.run.xml
```
AEPR, line 45: This is the gradle config for running the notify tests locally which contains the Notify API key. I will add it to keeper so that others can grab to easily run these tests. You can add them to the main test suite by adding the appropriate environment variable to the "prsdb-webapp [test]" configuration in IDEA.

src/test/kotlin/uk/gov/communities/prsdb/webapp/notify/NotifyEmailTemplateTests.kt
```diff
@@ -0,0 +1,98 @@
...
+@EnabledIf("canFetchNotifyTemplates")
+class NotifyEmailTemplateTests {
+    companion object NotifyTestsCompanion {
+        val jsonMetadataList = javaClass.getResource("/emails/emailTemplates.json")?.readText() ?: ""
+        lateinit var notifyTemplates: TemplateList
+
+        fun haveNotifyTemplatesBeenFetched() = ::notifyTemplates.isInitialized
+
+        fun fetchNotifyTemplates() {
+            val apikey = System.getenv("EMAILNOTIFICATIONS_APIKEY")
```
AEPR, line 25: Here I'm using the API key from the environment variable directly. I'm not 100% that it's the correct way to do this, but it means that these tests run faster because we don't need a full Springboot Test set up. I don't expect the tests to be deployed to a different environment before running (which could stop these tests running). We need to know whether to run these tests from a static context for the "@EnabledIf" tag to work, and I think that makes looking at environment variables cleaner (though I might have missed a simple/clean way to access application properties).

src/main/resources/emails/ExampleEmail.md
```diff
@@ -0,0 +1,6 @@
+Hello ((first name)),
```
AEPR, line 1: I'm not sure if this is the correct resources folder for these to live in - they might be better off in test/resources so they don't get included in the build. That said, they are "main" data not "test" data in that they are the actual templates in use (and could be useful at some point for mocking emails or something).

Open to opinions on this!

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationServiceTests.kt
```diff
@@ -3,25 +3,33 @@ package uk.gov.communities.prsdb.webapp.services
...
 import org.mockito.Mockito
-import uk.gov.communities.prsdb.webapp.viewmodel.TestEmail
+import uk.gov.communities.prsdb.webapp.viewmodel.EmailTemplateId
+import uk.gov.communities.prsdb.webapp.viewmodel.EmailTemplateModel
 import uk.gov.service.notify.NotificationClient
 
 class NotifyEmailNotificationServiceTests {
     private lateinit var notifyClient: NotificationClient
-    private lateinit var emailNotificationService: NotifyEmailNotificationService<TestEmail>
+    private lateinit var emailNotificationService: NotifyEmailNotificationService<TestEmailTemplate>
```
AEPR, line 12: This is not a rename (though it did inspire one)! TestEmail has been renamed to "ExampleEmail" and "TestEmailTemplate" is defined below.

AEPR (commented):
Left a couple of comments on files that might be useful for reviewing.

src/main/resources/emails/ExampleEmail.md
Travis-Softwire: For this POC agree it should be in main/resources as it's being used in 'built' code. Once the POC is replaced with a 'real' email sending feature for LA invites then if we still want the template to be used in tests it should move to test/resources.

src/test/kotlin/uk/gov/communities/prsdb/webapp/notify/NotifyEmailTemplateTests.kt
Travis-Softwire: I think strictly speaking it would be to just use the Value annotation on a class property as you would in your application code (possibly needing to be combined with a TestPropertySource annotation on the test class to trigger the configuration scanning). I would hope that the EnabledIf if evaluated before the configuration scan and so this wouldn't cause a problem if the environment variable is missing. If that isn't the case though then happy to stick with accessing the env variable directly in this case.

src/test/kotlin/uk/gov/communities/prsdb/webapp/notify/NotifyEmailTemplateTests.kt
```diff
@@ -0,0 +1,98 @@
...
+            // Act
+            var templateId = metadata.id
+            var notifyTemplate = notifyTemplates.templates.single { template -> template.id.toString() == templateId }
+
+            // Assert
+            assertBodiesMatch(metadata, notifyTemplate)
+            Assertions.assertEquals(metadata.name, notifyTemplate.name)
+            Assertions.assertEquals(metadata.subject, notifyTemplate.subject.orElse(null))
+
+            println("Template \"${metadata.name}\" matches Notify")
```
Travis-Softwire, line 71: Is the idea that this helps us identify which one failed if one doesn't match? If so, I think you can give assertEquals a 'message' parameter that controls what is printed out if the assertion fails which is a little cleaner than using println I think. Or was there a different purpose to these?

src/test/kotlin/uk/gov/communities/prsdb/webapp/notify/NotifyEmailTemplateTests.kt
```diff
@@ -0,0 +1,98 @@
...
+    fun `there is a source controlled copy for each template id`(id: EmailTemplateId) {
+        var metadataList = Json.decodeFromString<List<EmailTemplateMetadata>>(jsonMetadataList)
+
+        metadataList.single { templateMetadata -> templateMetadata.id == id.idValue }
+    }
+
+    @Test
+    fun `all source controlled templates match their notify equivalent`() {
+        // Arrange
+        var metadataList = Json.decodeFromString<List<EmailTemplateMetadata>>(jsonMetadataList)
```
Travis-Softwire, line 58: Drawing inspiration from what you've done below, is there a reason that this test uses a loop vs being parameterized with a method source?

src/test/kotlin/uk/gov/communities/prsdb/webapp/notify/NotifyEmailTemplateTests.kt
AEPR: No - this was to help identify which one failed if it doesn't match. I think the combination of parameterising this test and assert messages would be a lot better!

Travis-Softwire (approved):

src/test/kotlin/uk/gov/communities/prsdb/webapp/notify/NotifyEmailTemplateTests.kt
AEPR: This has now been updated to use Springboot as suggested.

PR #14, opened 2024-10-16
Feat/prsd 364 better error handling

AEPR:
Added a couple of custom exceptions to allow the NotifyService to throw depending on the context. I've made a base exception class intended for the whole of the project so if we want to do a top level catch on that at some point we can.

I'm not entirely sure about the amount of content I've ended up putting into the NotifyService here either:
- All those classes to deserialise the error messages into a useful format plus the helper method to strip out the non-json bit.
- The actual exception message content for the 3 cases I've separated out.
On the other hand, they are private to the service so it makes sense that they are inside it! Let me know what you think.

I've also updated the "Send test email" page to take advantage of this new functionality so that if a temporary or one off error it reloads the page with an updated title. This is going to be hard to test/simulate as you would need to actually cause an internal server error or send enough emails to hit rate limiting.

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/exceptions/CannotSendEmailsException.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/exceptions/CannotSendEmailsException.kt
@@ -1,0 +1,6 @@
+package uk.gov.communities.prsdb.webapp.exceptions
+
+class CannotSendEmailsException(
+    message: String,
+    cause: Throwable,
+) : PrsdbWebException(message, cause)
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/exceptions/EmailWasNotSentException.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/exceptions/EmailWasNotSentException.kt
@@ -1,0 +1,5 @@
+package uk.gov.communities.prsdb.webapp.exceptions
+
+class EmailWasNotSentException(
+    cause: Throwable,
+) : PrsdbWebException("Notify has not sent that email, but retrying may work.", cause)
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/exceptions/EmailWasNotSentException.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/exceptions/EmailWasNotSentException.kt
@@ -1,5 +1,0 @@
-package uk.gov.communities.prsdb.webapp.exceptions
-
-class EmailWasNotSentException(
-    cause: Throwable,
-) : PrsdbWebException("Notify has not sent that email, but retrying may work.", cause)
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationService.kt
@@ -1,5 +1,8 @@
 package uk.gov.communities.prsdb.webapp.services
 
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
 import org.springframework.stereotype.Service
 import uk.gov.communities.prsdb.webapp.viewmodel.EmailTemplateModel
 import uk.gov.service.notify.NotificationClient
@@ -15,4 +18,42 @@ class NotifyEmailNotificationService<EmailModel : EmailTemplateModel>(
         val emailParameters = email.toHashMap()
         notificationClient.sendEmail(email.templateId.idValue, recipientAddress, emailParameters, null)
     }
+
+    fun parseNotifyExceptionErrors(message: String): NotifyErrorType {
+        var nonJsonRegex = Regex("^Status code: \\d\\d\\d")
+        var jsonString = nonJsonRegex.replace(message, "")
+        var parsed = Json.decodeFromString<NotifyErrors>(jsonString)
+
+        return parsed.errors.first().error
+    }
+}
+
+@Serializable
+data class NotifyErrors(
+    val errors: List<NotifyErrorClass>,
+    val status_code: Int,
+)
+
+@Serializable
+data class NotifyErrorClass(
+    val error: NotifyErrorType,
+    val message: String,
+)
+
+@Serializable
+enum class NotifyErrorType {
+    @SerialName("AuthError")
+    AUTH,
+
+    @SerialName("BadRequestError")
+    BAD_REQUEST,
+
+    @SerialName("RateLimitError")
+    RATE_LIMIT,
+
+    @SerialName("TooManyRequestsError")
+    TOO_MANY_REQUESTS,
+
+    @SerialName("Exception")
+    EXCEPTION,
 }
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationService.kt
@@ -4,8 +4,11 @@ import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.Json
 import org.springframework.stereotype.Service
+import uk.gov.communities.prsdb.webapp.exceptions.CannotSendEmailsException
+import uk.gov.communities.prsdb.webapp.exceptions.EmailWasNotSentException
 import uk.gov.communities.prsdb.webapp.viewmodel.EmailTemplateModel
 import uk.gov.service.notify.NotificationClient
+import uk.gov.service.notify.NotificationClientException
 
 @Service
 class NotifyEmailNotificationService<EmailModel : EmailTemplateModel>(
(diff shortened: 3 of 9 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationService.kt
```diff
@@ -13,6 +19,62 @@ class NotifyEmailNotificationService<EmailModel : EmailTemplateModel>(
...
+            }
+        }
+    }
+
+    private fun parseNotifyExceptionError(message: String): NotifyErrorType {
+        var nonJsonRegex = Regex("^Status code: \\d\\d\\d")
+        var jsonString = nonJsonRegex.replace(message, "")
+        var parsed = Json.decodeFromString<NotifyErrors>(jsonString)
+
+        return parsed.errors.first().error
```
Travis-Softwire, line 48: Do we know under what conditions we might get more than one error back (if any)? I.e. how safe is it to only take the first one and throw the rest away?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationService.kt
```diff
@@ -13,6 +19,62 @@ class NotifyEmailNotificationService<EmailModel : EmailTemplateModel>(
...
     ) {
         val emailParameters = email.toHashMap()
-        notificationClient.sendEmail(email.templateId.idValue, recipientAddress, emailParameters, null)
+        try {
+            notificationClient.sendEmail(email.templateId.idValue, recipientAddress, emailParameters, null)
+        } catch (notifyException: NotificationClientException) {
+            val errorType = parseNotifyExceptionError(notifyException.message ?: "")
+            when (errorType) {
+                NotifyErrorType.AUTH, NotifyErrorType.BAD_REQUEST -> throw CannotSendEmailsException(
+                    "prsdb-web does not have the correct credentials to send emails with Notify." +
```
Travis-Softwire, line 28: V. minor, but can we just say that the credentials were not accepted by Notify, and have the call to action be to check that they are correct? The current message is probably correct, but there could be cases where it isn't e.g. Notify's auth has fallen over, or our account has been temporarily suspended etc. and a message saying that the credentials are incorrect could lead a future developer down the wrong path.

Travis-Softwire, line 28: Do we get an exception if we provide an incorrect template ID? If so what type of exception does that come through as?

src/main/kotlin/uk/gov/communities/prsdb/webapp/exceptions/CannotSendEmailsException.kt
```diff
@@ -0,0 +1,6 @@
+package uk.gov.communities.prsdb.webapp.exceptions
+
+class CannotSendEmailsException(
```
Travis-Softwire, line 3: The difference between CannotSend and WasNotSent feels quite subtle. Is there any way we can more clearly bring out the difference via the names?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationService.kt
rowanhill: Without being explicit, the documentation _suggests_ we'll only get one error code back: https://docs.notifications.service.gov.uk/java.html#send-an-email-error-codes

Perhaps we could ask in #govuk-notify in the UK Government Digital slack.

Travis-Softwire:
> Added a couple of custom exceptions to allow the NotifyService to throw depending on the context. I've made a base exception class intended for the whole of the project so if we want to do a top level catch on that at some point we can.
> 
> I'm not entirely sure about the amount of content I've ended up putting into the NotifyService here either:
> 
> * All those classes to deserialise the error messages into a useful format plus the helper method to strip out the non-json bit.
> * The actual exception message content for the 3 cases I've separated out.
>   On the other hand, they are private to the service so it makes sense that they are inside it! Let me know what you think.
> 
> I've also updated the "Send test email" page to take advantage of this new functionality so that if a temporary or one off error it reloads the page with an updated title. This is going to be hard to test/simulate as you would need to actually cause an internal server error or send enough emails to hit rate limiting.

On where the serialization lives - I think that makes sense in the service as a it's specific to Notify. Related to that (but v. minor) we probably want our more generic email exceptions to not reference notify in their error messages as in theory we'd want to reuse them if we had to swap out to a different email service, for example.

On testing the error handling - I think as long as we're testing that everything works with a mocked NotifyClient then I'm comfortable. We'll only have limited e2e testing anyway because it's expensive, and it will mostly focus on the happy paths.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationService.kt
Travis-Softwire: Yes good idea. I think you need a gov.uk email to get on there? If so I can ask on your behalf @AEPR

Travis-Softwire (commented):
A couple of relatively minor comments and Qs, but otherwise LGTM :) I've asked the Q on x gov slack about the error codes so I think we can leave that as-is for now

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationService.kt
AEPR: I don't have access there, so yes please. I have indeed taken it as implicit that there will only ever be one error here - but obviously using first not single means it doesn't error otherwise so definitely worth checking.

src/main/kotlin/uk/gov/communities/prsdb/webapp/exceptions/CannotSendEmailsException.kt
AEPR: Yes, I agree - I struggled a bit to come up with good names for these that express the idea of "That email wasn't sent but no comment on other emails" vs "We are currently unable to send emails and don't expect that to change imminently". I might open it up to the team for suggestions.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationService.kt
Travis-Softwire: From the #govuk-notify channel: "There are situations where the API can return more than one error message in the array.  For example, if there are multiple validation errors.  We are currently in the process of reviewing and updating our API documentation, and the error array is already an item on the list to be updated."

AEPR:
> On where the serialization lives - I think that makes sense in the service as a it's specific to Notify. Related to that (but v. minor) we probably want our more generic email exceptions to not reference notify in their error messages as in theory we'd want to reuse them if we had to swap out to a different email service, for example.

I think it's fine for the error messages to reference notify in the string literals that come from the notify service. If we re-use the exceptions in a different email sending service they can have different error messages when they're thrown from those services.

src/main/kotlin/uk/gov/communities/prsdb/webapp/exceptions/EmailWasNotSentException.kt
```diff
@@ -0,0 +1,5 @@
+package uk.gov.communities.prsdb.webapp.exceptions
+
+class EmailWasNotSentException(
+    cause: Throwable,
+) : PrsdbWebException("Notify has not sent that email, but retrying may work.", cause)
```
Travis-Softwire, line 5: This was the message I was referring to in my other comment!

Travis-Softwire:
> > On where the serialization lives - I think that makes sense in the service as a it's specific to Notify. Related to that (but v. minor) we probably want our more generic email exceptions to not reference notify in their error messages as in theory we'd want to reuse them if we had to swap out to a different email service, for example.
> 
> I think it's fine for the error messages to reference notify in the string literals that come from the notify service. If we re-use the exceptions in a different email sending service they can have different error messages when they're thrown from those services.

Yes agree - but in one case we refer to it in the reusable exception directly - I've added a comment there as well to point it out in case it was a mistake - but let me know if I'm misunderstanding what you mean here!

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationService.kt
AEPR: Status code: 400 {"errors":[{"error":"BadRequestError","message":"Template not found"}],"status_code":400}

So I should change the error message to be less specific in any case.

Travis-Softwire: I'd be tempted to split out Auth from Bad request - I expect most of our errors as we develop will be the latter, due to validation errors on templates etc

AEPR:
> > I think it's fine for the error messages to reference notify in the string literals that come from the notify service. If we re-use the exceptions in a different email sending service they can have different error messages when they're thrown from those services.
> 
> Yes agree - but in one case we refer to it in the reusable exception directly - I've added a comment there as well to point it out in case it was a mistake - but let me know if I'm misunderstanding what you mean here!

Yes, I realised that I'd done this when I went back to start amending the other error messages. I've changed it so all messages are set in the service now.

src/main/kotlin/uk/gov/communities/prsdb/webapp/exceptions/EmailWasNotSentException.kt
AEPR: Moved it to the service.

src/main/kotlin/uk/gov/communities/prsdb/webapp/exceptions/CannotSendEmailsException.kt
AEPR: Taken Rowan's suggestions of "[Permanent/Transient] Email Send Exception"

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationService.kt
AEPR: Yeah, I've actually decided to split out all of the different cases - if we've hit our rate limit it's probably useful to have a different message than if Notify has had an internal server error.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationService.kt
AEPR: I've updated the class to handle multiple errors more gracefully. It will throw the exception matching the "most severe" error (Auth>Bad Request>Daily Limit>Rate Limit>Exception), but if there were multiple errors there's a prefix to the error message saying that that's the case.

I don't think it's necessary to try and extract all the information as the original exception is available as a "Cause".

Travis-Softwire (approved):

PR #16, opened 2024-10-17
PRSD-364: Stub calls to notify for local

AEPR:
Added a stubbed email sending service. I've annotated with the profile "!use-notify" even though "use-notify" doesn't currently exist as it lets you easily connect to notify for real during local development by adding that profile to your build configuration. I've also created some email addressed to simulate the two exception types that can be thrown by the notify email service.


Travis-Softwire (approved):
LGTM :)

PR #19, opened 2024-10-21
PRSD 362: Integrate with OS Places

isobel-softwire:
### Features
- Created `OSPlacesClient` to handle calls to OS Places API
- Created `AddressLookupService` interface with children:
  - `OSPlacesAddressLookupService` calls `OSPlacesClient`
  - `StubAddressLookupService` stubs calls to `OSPlacesClient`
- Created `ExampleAddressLookupController` which allows the user to lookup addresses by postcode
- Configuration:
  - Instructions for adding the API key to the environment and switching between the actual and stubbed services added to 'Optional set up' document
  - API key added to webapp configuration via application properties

### Tests
- Tests for `OSPlacesLookupService`

### Notes
- Different convention used in this ticket's stub service (`StubAddressLookupService`) to an already existing one (`EmailNotificationStubService`). The one used in this ticket fits more with the actual service's name (`OSPlacesAddressLookupService`), but I'm happy to change it for conformity.
- Is there an upcoming ticket that will make use of the address lookup service? If so I can add removing the example controller and template to a TODO for it.
- TODO: research into how to use the API to retrieve addresses by postcode and building number/name.


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ExampleAddressLookupController.kt
```diff
@@ -0,0 +1,43 @@
...
+        model.addAttribute("serviceName", SERVICE_NAME)
+        return "exampleLookupAddress"
+    }
+
+    class Submission(
+        val postcode: String,
+    )
+
+    @PostMapping(consumes = [MediaType.APPLICATION_FORM_URLENCODED_VALUE])
+    fun sendEmail(
```
Travis-Softwire, line 30: Think you probably meant to rename this :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/OSPlacesClient.kt
```diff
@@ -0,0 +1,36 @@
...
+import java.net.http.HttpRequest
+import java.net.http.HttpResponse
+
+class OSPlacesClient(
+    private val apiKey: String,
+) {
+    private val client = HttpClient.newHttpClient()
+
+    fun searchByPostcode(postcode: String): String =
+        getResponse("https://api.os.uk/search/places/v1/postcode?dataset=lpi&postcode=$postcode")
```
Travis-Softwire, line 16: Ideally I think we'd grab this from config rather than hard coding it - that also allows us to mock the endpoint instead of the service, which means that the service code will still be run when developing locally (see related comment below).

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/services/StubAddressLookupService.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import org.springframework.context.annotation.Primary
+import org.springframework.context.annotation.Profile
+import org.springframework.stereotype.Service
+import uk.gov.communities.prsdb.webapp.services.AddressLookupService
+
+@Service
+@Profile("local & !use-os-places")
+@Primary
+class StubAddressLookupService : AddressLookupService {
```
Travis-Softwire, line 11: I think ideally we'd do the stubbing by pointing it back at a localhost endpoint. That way all of our application code is still running when we're developing locally and we can minimise the bugs that only show up when we deploy.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/OSPlacesAddressLookupService.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import org.json.JSONObject
+import org.springframework.stereotype.Service
+import uk.gov.communities.prsdb.webapp.clients.OSPlacesClient
+
+@Service
+class OSPlacesAddressLookupService(
+    val osPlacesClient: OSPlacesClient,
+) : AddressLookupService {
+    override fun searchByPostcode(postcode: String): List<String> {
```
Travis-Softwire, line 11: Are we able to get back a more structured object from the API rather than just the address string? Thinking ahead, when the user selects an address we'll want to be able to then store information about that address in our database, so we'll want to be sure we can look up that specific address once the user clicks on it and sends it back to us.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/OSPlacesAddressLookupService.kt
```diff
@@ -0,0 +1,20 @@
...
+        val response = osPlacesClient.searchByPostcode(postcode.replace(" ", ""))
+        val results = JSONObject(response).getJSONArray("results")
+        val addresses = mutableListOf<String>()
+        for (i in 0 until results.length()) {
+            addresses.add(results.getJSONObject(i).getJSONObject("LPI").getString("ADDRESS"))
```
Travis-Softwire, line 16: If we are getting more than just the single line address (see above) we might want to create a model for our address data inside our application, and a mapper for translating an address from the response into an our address model?

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/OSPlacesAddressLookupServiceTests.kt
```diff
@@ -0,0 +1,42 @@
...
+    @Test
+    fun `searchByPostcode returns a corresponding list of addresses given a valid postcode`() {
+        val addressesJSON = "{'results':[{'LPI':{'ADDRESS':'1 Example Road'}},{'LPI':{'ADDRESS':'2 Example Road'}}]}"
+        val expectedAddresses = listOf("1 Example Road", "2 Example Road")
+
+        `when`(
+            mockOSPlacesClient.searchByPostcode(anyString()),
+        ).thenReturn(addressesJSON)
+
+        assertEquals(expectedAddresses, addressLookupService.searchByPostcode("EG"))
```
Travis-Softwire, line 33: Worth keeping the 'Act' or 'When' part of the test (`addressLookupService.searchByPostcode`) separate from the 'Assert' or 'Then' part - It helps make it clear at a glance what is being test and what the expected behavior is, and can also make it a bit easier to debug without having to refactor first.

Travis-Softwire (commented):
Thanks for this - I've left a few comments, mostly around the mocking where I think ideally we'd take a bit of a different approach vs. what we've done with notify. 

On the question about leaving a TODO for a future ticket, yes please :) PRSD-371 (landlord contact address) is likely the first ticket where this will be used)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ExampleAddressLookupController.kt
isobel-softwire: i did...thanks

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/services/StubAddressLookupService.kt
isobel-softwire: Do we still want to be able to call the actual API from the local configuration (e.g. something like how the `use-os-places` profile is currently being used)?

Travis-Softwire: Yes, we'll want the option to use either the stub or the real service when developing locally, based on the profile we use. @Kiran-R-K was looking into this same issue as part of the One Login mock, so worth comparing notes

src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/OSPlacesClient.kt
isobel-softwire: Would we want to grab the whole URL from config, or just the base address (https://api.os.uk/search/places/v1)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/OSPlacesAddressLookupService.kt
isobel-softwire: (Responding to the last two comments here) The API returns lots more fields - it looks like it matches up with the planned Address entity. I think the model for this ticket should just include the address and postcode (so the address can be looked up later) rather than the whole Address entity.

src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/OSPlacesClient.kt
```diff
@@ -0,0 +1,36 @@
...
+import java.net.http.HttpRequest
+import java.net.http.HttpResponse
+
+class OSPlacesClient(
+    private val baseURL: String,
+    private val apiKey: String,
+) {
+    private val client = HttpClient.newHttpClient()
+
+    fun searchByPostcode(postcode: String): String = getResponse("/postcode?postcode=${postcode.replace(" ", "")}")
```
rowanhill, line 16: We should be URL encoding the postcode, here. 99.9% of the time it'll make no difference, but it'll stop anything funky happening if a user enters something dodgy as a postcode (e.g. "E17&performExpensiveAction=true"). In general: we should never trust user input!

src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/OSPlacesClient.kt
```diff
@@ -0,0 +1,36 @@
...
+                .uri(URI.create(baseURL + endpoint))
+                .build()
+
+        val response = client.send(request, HttpResponse.BodyHandlers.ofString())
+        val responseBody = response.body()
+
+        if (response.statusCode() != 200) {
+            val errorObject = JSONObject(responseBody).getJSONObject("error")
+            throw HttpException("Error ${errorObject.getInt("statuscode")}: ${errorObject.getString("message")}")
+        }
```
rowanhill, line 32: What happens if we get a response with a body that doesn't parse as JSON? Or does, but doesn't have the form we're expecting here? Also, should we have some special handling for rate limiting responses?

isobel-softwire: I'll wrap this in a try-catch in case of that. Rate limiting is a good idea, but I don't have any experience implementing it. Is there a particular library we should use?

rowanhill: Let's make sure we add some tests around unexpected body contents.

In terms of rate limiting, I don't mean we should limit it ourselves, I mean OS will do so - if we make too many requests, they'll start refusing us. We should check their docs to see what kind of response we'll get in that case, and handle it here, somehow: I'd suggest we should throw a special type of exception, so consumers of this client can do something more meaningful (e.g. the registration controllers could show a "we're experiencing a lot of demand, please try again later" type message).

isobel-softwire: Added `OSPlacesClientTests` and created a `RateLimitedExceededException` 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/OSPlacesClient.kt
```diff
@@ -0,0 +1,46 @@
...
+            else -> throw HttpException(getErrorMessage(response))
+        }
+    }
+
+    private fun getErrorMessage(response: HttpResponse<String>): String {
+        var errorMessage = "Error ${response.statusCode()}"
+        try {
+            errorMessage += ": ${JSONObject(response.body()).getJSONObject("error").getString("message")}"
+        } catch (e: JSONException) {
+            println("Warn: Unexpected error response format from OS Places - ${e.message}")
```
rowanhill, line 42: I think we should probably include the raw response body to the errorMessage in this case - what do you think?

isobel-softwire: That makes sense - I've changed it now

rowanhill (approved):

PR #20, opened 2024-10-21
Feat/prsd 358 create frontend page

JasminConterioSW:
First pass at the frontend page.

This version:

- Gets the active LA users for the current users's LA from the database
- Takes hardcoded pending LA users (this will come from the database in a later branch)
- Shows these as a table (this will be paginated in a later branch)
- Does not include the common components of the One Login banner, the "Beta - this is a new service" message or breadcrumbs (these will be in later tickets)

image

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
@@ -37,9 +37,7 @@ class ManageLocalAuthorityUsersController(
         model.addAttribute("title", "Manage Local Authority Users")
         model.addAttribute("serviceName", SERVICE_NAME)
         model.addAttribute("userList", users)
-        model.addAttribute("tableColumn1Heading", "Username")
-        model.addAttribute("tableColumn2Heading", "Access level")
-        model.addAttribute("tableColumn3Heading", "Account status")
+        model.addAttribute("tableColumnHeadings", listOf("Username", "Access level", "Account status", ""))
 
         return "manageLAUsers"
     }
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
@@ -24,6 +24,7 @@ class ManageLocalAuthorityUsersController(
         val currentUserLocalAuthority = localAuthorityDataService.getLocalAuthorityForUser(principal.name)!!
 
         val activeUsers = localAuthorityDataService.getLocalAuthorityUsersForLocalAuthority(currentUserLocalAuthority)
+        // TODO: Get these from LocalAuthorityUserInvitation with userName=email, isManager=false and isPending=true
         val pendingUsers =
             listOf(
                 LocalAuthorityUserDataModel("Invited user 1", isManager = false, isPending = true),
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
@@ -1,5 +1,7 @@
 package uk.gov.communities.prsdb.webapp.controllers
 
+import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.context.MessageSource
 import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.stereotype.Controller
 import org.springframework.ui.Model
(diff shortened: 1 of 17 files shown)


src/main/resources/templates/demoStart.html
```diff
@@ -6,6 +6,6 @@ <h1 class="govuk-heading-xl" th:text="${contentHeader}">Default page template</h
         <span>Example Landlord Registration Number: </span>
         <span th:text="${landlordRegNum}">Example Landlord Registration Number</span>
     </p>
-    <div th:replace="~{fragments/startButton :: button( ${startButtonHref}, ${startButtonText} )}"></div>
+    <div th:replace="~{fragments/startButton :: startButton( href, buttonText )utton( ${startButtonHref}, ${startButtonText} )}"></div>
```
isobel-softwire, line 9: I think this is a mistype?

src/main/resources/templates/fragments/primaryButton.html
```diff
@@ -0,0 +1,6 @@
+<button th:fragment="primaryButton(buttonText)"
+        type="submit"
+        class="govuk-button"
+        data-module="govuk-button"
+        th:text="${buttonText}">
```
isobel-softwire, line 5: Default text could be added here and the other new fragments.

isobel-softwire (commented):
Mostly LGTM 👍 - left a few small comments. 
Also, you could put the page's static text in the `messages.properties` file in case of change to the page's design (not sure how likely change is though).

src/main/resources/templates/demoStart.html
JasminConterioSW: So it is, thanks! I think it autocompleted...

JasminConterioSW:
> Also, you could put the page's static text in the `messages.properties` file in case of change to the page's design (not sure how likely change is though).

General question, how much do we want to put in this file and how much do we just assume isn't going to change? It will get long quickly if we add everything (e.g. column headings, button text etc)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthority.kt
```diff
@@ -10,7 +10,7 @@ import jakarta.persistence.Id
 class LocalAuthority : ModifiableAuditableEntity() {
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
-    private val id: Int? = null
```
Travis-Softwire, line 13: Why did this need to be made public?

src/main/resources/templates/fragments/tables/tableBodyLAUser.html
```diff
@@ -0,0 +1,12 @@
+<tbody th:fragment="tableBody(userList)" class="govuk-table__body">
+    <tr class="govuk_table__row" th:each="user: ${userList}">
+        <td class="govuk-table__cell" th:text="${user.userName}"></td>
+        <td class="govuk-table__cell" th:text="${user.manager ? 'Admin' : 'Basic'}"></td>
```
Travis-Softwire, line 4: As we still don't know if we might need to provide Welsh translations, we should source all of the readable text from message.properties I'm afraid

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthority.kt
JasminConterioSW: Sorry, it doesn't, I think this is a regression because I'm branching off my feature branches before they're merged!

src/main/resources/templates/fragments/tables/tableBodyLAUser.html
JasminConterioSW: Ah ok, that makes sense!

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
```diff
@@ -24,13 +23,21 @@ class ManageLocalAuthorityUsersController(
...
+            )
 
-        model.addAttribute("contentHeader", "Manage Local Authority Users")
+        val users = activeUsers + pendingUsers
+
+        model.addAttribute("contentHeader", "Manage ${currentUserLocalAuthority?.name}'s users")
         model.addAttribute("title", "Manage Local Authority Users")
         model.addAttribute("serviceName", SERVICE_NAME)
+        model.addAttribute("userList", users)
+        model.addAttribute("tableColumnHeadings", listOf("Username", "Access level", "Account status", ""))
```
Travis-Softwire, line 40: Thinking about it, we will also need to switch these (and the page titles) to using the messages.properties file. This would be done by referencing e.g. `#{${title}}` in the template, and then passing in `manageLocalAuthorities.title`, say, in the model. We'll need to update the existing pages (possibly just register as a landlord start page so far) to do the same.

PR #21, opened 2024-10-21
Feat/prsd 404 send invites from example controller

AEPR:
This is the backend logic for storing invitations. They are currently created using the example email page I created for prsd-364, a future PR will give them their own page (and remove the example page).

I'm not 100% sure on value of the "invitationService" (and therefore the "invitationServiceTests") - it's currently basically just a proxy for calling some database methods (and obscures the token format). I think it's worth keeping to make the token opaque and to add more logic to later (at the moment it's the front runner for invitation URL building), but I'm open to other opinions.


src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/LocalAuthorityInvitationEmail.kt
```diff
@@ -1,8 +1,11 @@
 package uk.gov.communities.prsdb.webapp.models.viewModels
 
+import uk.gov.communities.prsdb.webapp.database.entity.LocalAuthority
+import java.net.URI
+
 data class LocalAuthorityInvitationEmail(
-    var localAuthorityName: String,
-    var invitationUrl: String,
+    var localAuthority: LocalAuthority,
+    var invitationUri: URI,
```
AEPR, line 8: I've gone back and forth on whether these should all be strings or whether it's better like this (where you can only build an invite with an authority + URI)

src/main/resources/emails/LocalAuthorityInvitation.md
```diff
@@ -0,0 +1,18 @@
...
+
+## What happens next
+You'll be asked for your work details like your name and work email address.
+
+Once you've registered you can view:
+* property records
+* landlord records
+* reports from the public about the database
+
+You can only view these records that fall within your local authority area ((name of council)).
```
Travis-Softwire, line 13: Not for you to worry about on this ticket, but I believe this copy is incorrect...

Travis-Softwire (approved):
One Q about why we're using nullable instead of lateinit properties for non-primitives in the entity class, but otherwise LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthorityInvitation.kt
```diff
@@ -0,0 +1,37 @@
...
+import java.util.UUID
+
+@Entity
+class LocalAuthorityInvitation() : AuditableEntity() {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private val id: Long? = null
+
+    @Column(nullable = false, unique = true)
+    var token: UUID? = null
```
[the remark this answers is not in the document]
AEPR: I hadn't checked if it worked with "jakarta.persistence" annotations. It does and I've changed it.

src/main/resources/emails/LocalAuthorityInvitation.md
AEPR: Happy to update it - but I've copied it from the image in the ticket you sent. If there's a better souce for copy I can change it (or we can get someone from Madetech to create it directly in Markdown).

PR #22, opened 2024-10-22
Feat/prsd 404 create initial invite pages

AEPR:
Making this a draft pull request because I think it needs at least one more unit test before merging and I don't know how to write it! (I'll leave a comment on the method I think needs testing)

This PR is makes few changes than the files touched:
* Add validation to the emails submitted to the controller (currently non-matching emails just errors)
* Move the existing example functions to the `manage-la-users-controller` (although no UI changes except adding the confirm email)

I've also not added additional controller tests for the manage user controller, even though I've added some functionality in there. Currently all the controller tests just check that they return the correct HTTP code, and nothing else. I could check that it calls all the appropriate service functions or something slightly more heavy-weight (e.g. actually checking logic/validation errors once they're added in). Basically I've decided to follow the status quo but I think we should probably beef up our controller tests.


src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/constraints/ValidConfirmedEmail.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import jakarta.validation.Constraint
+import jakarta.validation.Payload
+import uk.gov.communities.prsdb.webapp.validation.validators.ConfirmedEmailValidator
+import kotlin.reflect.KClass
+
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+@Constraint(validatedBy = [ConfirmedEmailValidator::class])
+annotation class ValidConfirmedEmail(
```
rowanhill, line 11: At the cost of making things _stringly typed_, I wonder if this annotation might be more useful (i.e. usable in more places) if it took two field names, and then the validator used reflection to pull out the values of those fields and compare them?

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/ControllerTest.kt
```diff
@@ -37,4 +40,10 @@ abstract class ControllerTest(
 
     @MockBean
     lateinit var localAuthorityDataService: LocalAuthorityDataService
+
+    @MockBean
+    lateinit var anyEmailNotificationService: EmailNotificationService<EmailTemplateModel>
```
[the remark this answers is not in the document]
AEPR: It creates mock service beans and then is subclassed by each specific controller test class so that e.g. `ManageLocalAuthorityUsersControllerTests` can use the MockMvc created here to test the routes on the `ManageLocalAuthorityUsersController` (and have all these beans added to it).

It could be that we'd be better off not sharing this code between the controller test classes (or at least not sharing the mock beans!)

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/validators/MatchingPropertiesValidator.kt
```diff
@@ -0,0 +1,31 @@
...
+        context: ConstraintValidatorContext?,
+    ): Boolean {
+        if (value == null) return false
+
+        value::class.memberProperties.forEach { property ->
+            if (property.hasAnnotation<Matches>()) {
+                val annotation = property.findAnnotation<Matches>()!!
+                val matchingProperty = value::class.memberProperties.single { it.name == annotation.propertyName }
+
+                if (matchingProperty.call(value) != property.call(value)) {
```
rowanhill, line 23: This might be a when-we-come-to-it thing, but maybe we want to e.g. trim string values before we compare (in cases where we'll trim before storing in the db)?

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/ControllerTest.kt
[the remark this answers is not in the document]
rowanhill: I think I would have expected each of the controller tests to be annotated with @WebMvcTest(ControllerUnderTest::class), and to contain an @Autowired MockMvc plus only the @MockBean dependencies necessary for that controller.

Do you know if that kind of thing was considered / if there's a reason for doing it this way?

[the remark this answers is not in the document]
AEPR: I do not know about that no - this is how these tests were set up when I joined. I feel like refactoring them is slightly beyond the scope of this PR, but maybe one for us to have a discussion tomorrow as an additional task?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityInvitationService.kt
```diff
@@ -26,12 +29,10 @@ class LocalAuthorityInvitationService(
         return invitationRepository.findByToken(tokenUuid).invitingAuthority
     }
 
-// TODO-404: This path should be set to match the necessary route on the new invitation controller once created
+    // TODO-405: This function call should be set to match the controller method on the new invitation controller once created
     fun buildInvitationUri(token: String): URI =
```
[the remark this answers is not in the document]
AEPR: I have now written a test for this - see `test.integration.InvitationUrlTests.kt` - it was more painful that I expected, and worth a thorough review:
* I'm not sure about where I've put it - at the moment all of the tests in "integration" are testing templates, so it should maybe live in another package but it's currently a bit of a one of a kind test
* I've had to add a work around for an issue with using Mockito with Kotlin. I don't know why this issue hasn't emerged previously, but we should maybe consider adding the work around methods to some sort of test helper package in case it comes up again.

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/ControllerTest.kt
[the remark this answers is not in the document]
AEPR: Are you happy to resolve this discussion on this review and maybe the two of us and Travis talk about what we want to do with the controller tests when he's back tomorrow?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityInvitationService.kt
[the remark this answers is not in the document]
rowanhill: On the Kotlin with Mockito stuff, there does seem to be an official library for it: https://github.com/mockito/mockito-kotlin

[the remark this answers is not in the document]
AEPR: I've added it to specifically this file, I've got a commit adding it everywhere but it might be better as a chore after this is merged into main rather than confusing this PR.

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/validators/MatchingPropertiesValidator.kt
```diff
@@ -0,0 +1,57 @@
...
+    ): Boolean {
+        if (value == null) return false
+
+        value::class.memberProperties.forEach { property ->
+            if (property.hasAnnotation<Matches>()) {
+                val annotation = property.findAnnotation<Matches>()!!
+                val matchingProperty = value::class.memberProperties.single { it.name == annotation.propertyName }
+
+                val stringValue = property.call(value) as? String
+                val matchingStringValue = matchingProperty.call(value) as? String
```
Travis-Softwire, line 26: If someone has accidentally mistyped the target field, will we get a runtime error here?

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/constraints/HasMatchingStringProperties.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import jakarta.validation.Constraint
+import jakarta.validation.Payload
+import uk.gov.communities.prsdb.webapp.validation.validators.MatchingPropertiesValidator
+import kotlin.reflect.KClass
+
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+@Constraint(validatedBy = [MatchingPropertiesValidator::class])
+annotation class HasMatchingStringProperties(
```
Travis-Softwire, line 11: Are we expecting to use this more widely than this use case? If not, I'm wondering if we can get away with just an `AssertTrue` annotation on a `isConfirmEmailSameAsEmail` method (I think they have to start with 'is' annoyingly...)?

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/validators/MatchingPropertiesValidator.kt
AEPR: Yes!

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/constraints/HasMatchingStringProperties.kt
AEPR: I'm honestly happy either way. I had initially written a fairly simple validator specifically for the confirmed email data class and Rowan suggested I make it generic. 🤷 

Does 'AssertTrue' count as part of the validation step?

rowanhill: Yeah, I don't mind much either. It doesn't seem impossible we'd ask people to double-enter other strings in future, but perhaps we could say "YAGNI" and go with the simple version for now.

AEPR: OK, I'm signing off for the week now but on Monday I'll return the simpler version and hopefully we can get this PR merged!

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ExampleInvitationTokenController.kt
```diff
@@ -0,0 +1,32 @@
...
+    ): ModelAndView {
+        val authority = invitationService.getAuthorityForToken(token)
+        model.addAttribute("contentHeader", "The local authority issuing that token was: ${authority.name}")
+        model.addAttribute("title", "Magic link")
+        model.addAttribute("serviceName", SERVICE_NAME)
+
+        // TODO PRSD-405: This is using a ModelAndView instead of returning a string to handle an error that otherwise occurs in
+        // the LocalAuthorityInvitationService method that creates the invitation url using MvcUriComponentsBuilder.fromMethodName
+        // see https://stackoverflow.com/questions/28580281/using-mvcuricomponentsbuilderfrommethodcall-with-string-as-the-return-type for details
+        return ModelAndView("index", "index", model)
```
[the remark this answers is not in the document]
AEPR: Switched to using CharSequence as that does work as of the release linked.

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/validators/MatchingPropertiesValidator.kt
AEPR: Discussion made superfluous by later commits/decisions in PR (https://github.com/communitiesuk/prsdb-webapp/pull/22#discussion_r1816787667).

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/validators/MatchingPropertiesValidator.kt
AEPR: Discussion made superfluous by later commits/decisions in PR (https://github.com/communitiesuk/prsdb-webapp/pull/22#discussion_r1816787667).

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/constraints/HasMatchingStringProperties.kt
AEPR: This class has now been removed and I've switched to using a `isConfirmEmailSameAsEmail` method with an `AssertTrue` annotation.

Travis-Softwire (approved):
LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
```diff
@@ -58,4 +69,42 @@ class ManageLocalAuthorityUsersController(
...
+    fun sendEmail(
+        model: Model,
+        @Valid
+        emailModel: ConfirmedEmailDataModel,
+        result: BindingResult,
+        principal: Principal,
+    ): String {
+        model.addAttribute("serviceName", SERVICE_NAME)
+        try {
+            result.allErrors.forEach { println(it.toString()) }
```
rowanhill, line 91: Bit of leftover debug code?

AEPR: Sort of - I hadn't prioritised removing it as I will need to make significant changes to this method to allow the validation errors to be passed back to the frontend.

I'll just raise a micro-PR now to get rid of it.

PR #23, opened 2024-10-23
PRSD-NONE: Remove redundant filter chain config

isobel-softwire:
Removes redundant filter chain config (made redundant by One Login mock)


JasminConterioSW (approved):
LGTM :)

rowanhill:
Can you just explain why /lookup-an-address isn't needed any more?

isobel-softwire:
> Can you just explain why /lookup-an-address isn't needed any more?

I put ` \lookup-an-address` in the filter chain to bypass authentication. The One Login mock now provides this authentication so putting the endpoint in the filter chain isn't needed.

rowanhill:
Ah, okay - so how come we still need to permitAll() on /postcode and /register-as-a-landlord? (Perhaps the latter is the unauthenticated start page for the journey?)

isobel-softwire:
Stubbed calls to OS Places are sent to localhost:8080/postcode - these calls don't require authentication. AFAIK you're right about /register-as-a-landlord

rowanhill:
Ah, okay. I'd thought that /postcode might be the test page for making postcode requests, but that makes sense. Thanks!

PR #28, opened 2024-10-24
PRSD-453: One Login header

isobel-softwire:
### Features
- Creates `authenticatedHeader` fragment based on [One Login Header](https://github.com/govuk-one-login/service-header)
- Creates `header` fragment that shows the `authenticatedHeader` if the user if the user is authenticated and the `anonymousHeader` (the original header) if the user isn't
- Adds a placeholder `/signout` endpoint for 'Sign Out' to point to
- Adds example navigation links to `/registration`

### Screenshot
image


isobel-softwire:
There are line terminator symbol changes in the layout and anonymousHeader fragments so it looks like there are more changes than there actually are.

rollup.config.mjs
```diff
@@ -35,6 +35,14 @@ export default {
                         src: 'node_modules/govuk-frontend/dist/govuk/assets/manifest.json',
                         dest: 'dist',
                     },
+                    {
+                        src: 'node_modules/govuk-one-login-service-header/dist/scripts/service-header.js',
```
Travis-Softwire, line 39: I haven't looked at the library, but have they not managed to set this up as a proper node module that we can import in index.js?

src/main/resources/templates/fragments/header/header.html
```diff
@@ -0,0 +1,10 @@
+<html xmlns:sec="http://www.thymeleaf.org/extras/spring-security">
+<div th:fragment="header(serviceName, navLinks)">
```
Travis-Softwire, line 2: Does this mean we always need to provide a navLinks array even for pages where the user isn't logged in?

isobel-softwire: No, null is being passed when the `header` fragment is called (from `layout`) if the model doesn't provide an array.

Travis-Softwire (approved):
Left a Q about how we import the module from the 1L header library but that aside LGTM :)

rollup.config.mjs
isobel-softwire: There doesn't seem to be

PR #30, opened 2024-10-28
PRSD-NONE: Refactor in case of localisation

isobel-softwire:
- Moves `SERVICE_NAME` constant to `messages.properties`
- Moves static text from controllers to `messages.properties`
- Passes message keys instead of messages to templates
- Adds parameter names to messages with parameters


rowanhill (approved):
Looks good to me, although I must admit I didn't manually check every controller change matches its view change correctly - I assume you've checked each of those in the browser to make sure?

Travis-Softwire (approved):
Thanks for picking this up Isobel, LGTM :) With any refactor like this, please do triple check the pages you've changed for regressions before merging - particularly the non-example ones

Edit: I hadn't noticed Rowan's review, but agree with his comment!

isobel-softwire:
Looked at every controller, the views all LGTM 👍

PR #32, opened 2024-10-28
Feat/prsd 404 create template

AEPR:
This is a draft review, I will continue making changes to it before proposing a merge. The main area that I know still needs improving is the logic and code placement of everything in the controller. That is to say - there are some classes I've left in that file that need their own home and some of the functions and naming are difficult to read.

What I would find useful though would be a review of the Thymeleaf templates and fragments I've written as they're essentially independant of the controller logic (with the obvious exception of property names).


src/main/resources/messages.properties
```diff
@@ -68,4 +68,29 @@ manageLAUsers.contentHeader.part1=Manage
...
+# {0} = the council email address domain (i.e. council.co.uk)
+addLAUser.form.primaryEmailLabel=Enter their email address ending in {0}
+addLAUser.form.confirmEmailLabel=Confirm email address
+addLAUser.form.submitButton=Submit
+
+addLAUser.error.summaryHeading=Something went wrong
+addLAUser.error.missingEmail=You must provide a value
+addLAUser.error.notAnEmail=You must provide a value which is a valid email
+addLAUser.error.noConfirmation=You must re-enter the email to confirm it is correct
+addLAUser.error.confirmationDoesNotMatch=You must enter a confirming email that matches the email you have submitted
```
AEPR, line 94: I've made up all these error messages as I don't believe they were covered on the ticket/in figma. I will bring it up and standup tomorrow, but obviously they're easy to change!

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("ktlint:standard:no-wildcard-imports")
```
AEPR, line 1: I don't know where our klint config comes from or the justifications behind the decisions but I've run into issues with this wildcard import rule a few times because my IDEA keeps helpfully converting imports to a combined wildcard import past a certain number.

I don't mind either way, but if we do want to keep this rule I'll work out how to stop IDEA doing that!

src/main/resources/templates/inviteLAUser.html
```diff
@@ -0,0 +1,48 @@
...
+                </li>
+                <li th:text="#{addLAUser.paragraph.two.bullet.two}">
+                </li>
+                <li th:text="#{addLAUser.paragraph.two.bullet.three}">
+                </li>
+            </ul>
+            <p class="govuk-body">
+                <span th:utext="#{addLAUser.paragraph.four.part1}" th:remove="tag"></span>
+                <strong th:text="${councilName}"></strong>
+                <span th:text="#{addLAUser.paragraph.four.part2}" th:remove="tag"></span>
```
AEPR, line 36: I'm not sure whether I prefer this (having a `<span>` that I then `th:remove="tag"`) or whether it would be better to use `[[#{addLAUser.paragraph.four.part2}]]` (the inline expression syntax).

What do you think/prefer?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
isobel-softwire: In case we end up keeping it, this is what I used to make IntelliJ stop 
image

src/main/resources/templates/inviteLAUser.html
Travis-Softwire: Is another alternative to use th:block combined with th:text? I think I prefer that to span + th:remove if that works. If not then I'd probably go with the inline syntax, which is a bit less pretty but probably the more 'thymeleafy' approach I think.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
```diff
@@ -101,9 +118,63 @@ class ManageLocalAuthorityUsersController(
...
+
+    val hasErrors: Boolean = emailValidation.hasError || confirmEmailValidation.hasError
+}
+
+class FieldValidation(
+    val value: String,
+    val messageKey: String? = null,
+) {
+    val hasError: Boolean = messageKey != null
 }
```
rowanhill, line 180: This is all feeling a little overengineered. Showing validation errors is going to be super super common, so we want it to be as low ceremony as possible.

Could the message param of the validation annotations not just be the messages.properties keys? And plonk them in e.g. a Map on the model keyed by form field name?

src/main/resources/templates/fragments/error-summary.html
```diff
@@ -0,0 +1,11 @@
+<div  th:fragment="summary(message, myListContent)" class="govuk-error-summary" data-module="govuk-error-summary">
```
rowanhill, line 1: Maybe just "listContent"? Or something slightly more descriptive like "errorList"?

src/main/resources/templates/fragments/forms/textInput.html
```diff
@@ -0,0 +1,8 @@
+    <div th:fragment="textInput(label, for, validationError)" th:class="${validationError?.hasError} ? 'govuk-form-group govuk-form-group--error':'govuk-form-group'">
```
rowanhill, line 1: Maybe something like "fieldName" instead of "for"?

And I think using th:classappend to add the error class might be fractionally neater.

src/main/resources/templates/inviteLAUser.html
```diff
@@ -0,0 +1,48 @@
+<!DOCTYPE html>
+<html th:replace="~{fragments/layout :: layout(${(validationErrors?.hasErrors ?: false) ? 'Error: ' : ''} + #{addLAUser.title}, ${serviceName}, ~{::main})}">
```
rowanhill, line 2: Prepending "Error: " is something we'll need to do on practically every page, so I think it's worth moving this logic into the layout (and passing a hasErrors bool).

src/main/resources/templates/inviteLAUser.html
```diff
@@ -0,0 +1,48 @@
...
+<main class="govuk-main-wrapper" id="main-content">
+    <div th:replace="~{fragments/twoThirdsLayout :: twoThirdsLayout(~{::#header})}">
+        <header id="header">
+            <h1 class="govuk-heading-l" th:text="#{addLAUser.heading(${councilName})}"></h1>
+        </header>
+    </div>
+    <div th:replace="${validationErrors?.hasErrors ?: false} ? ~{fragments/error-summary :: summary(#{addLAUser.error.summaryHeading}, ~{:: #errorList})} : ~{}">
```
rowanhill, line 9: Could this be a th:replace plus a th:if?

src/main/resources/templates/inviteLAUser.html
```diff
@@ -0,0 +1,48 @@
...
+        </li>
+        <li th:if="${validationErrors.confirmEmailValidation.hasError}">
+            <a href="#confirmEmail" th:text="#{${validationErrors.confirmEmailValidation.messageKey}}"></a>
+        </li>
+    </ul>
+    </div>
+    <div th:replace="~{fragments/twoThirdsLayout :: twoThirdsLayout(~{::#explanation})}">
+        <section id="explanation">
+            <p th:text="#{addLAUser.paragraph.one}" class="govuk-body">
+            </p>
```
rowanhill, line 22: Here, and in all the tags below, do we want to include some default text? I know we talked about this recently, but I can't remember what we decided we were doing!

src/main/resources/templates/inviteLAUser.html
```diff
@@ -0,0 +1,48 @@
...
+                <span th:utext="#{addLAUser.paragraph.four.part1}" th:remove="tag"></span>
+                <strong th:text="${councilName}"></strong>
+                <span th:text="#{addLAUser.paragraph.four.part2}" th:remove="tag"></span>
+            </p>
+        </section>
+    </div>
+    <div th:replace="~{fragments/twoThirdsLayout :: twoThirdsLayout(~{::#email-form})}">
+    <form class="form" id="email-form" method="post" th:action="@{/manage-users/invite-new-user}">
+        <div th:replace="~{fragments/forms/textInput :: textInput(#{addLAUser.form.primaryEmailLabel(${councilEmail})}, 'email', ${validationErrors?.emailValidation})}"></div>
+        <div th:replace="~{fragments/forms/textInput :: textInput(#{addLAUser.form.confirmEmailLabel}, 'confirmEmail', ${validationErrors?.confirmEmailValidation})}"></div>
```
rowanhill, line 43: Minor thing, but these could probably be `<input>` tags that are being th:replaced?

src/main/resources/templates/inviteLAUser.html
AEPR: I can use th:block and have switched to using that whereever it came up.

src/main/resources/templates/inviteLAUser.html
AEPR: [Turns out th:replace is evaluated before th:if](https://www.thymeleaf.org/doc/tutorials/2.1/usingthymeleaf.html#attribute-precedence), but I think it's better to have an explicit if rather than using the empty fragment. I've acheived it by wrapping this `div` in a `th:block` with the th:if.

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/bindingResultExtensions/BindingResultExtensions.kt
```diff
@@ -0,0 +1,45 @@
...
+        }
+    }
+    return newResult
+}
+
+private fun <T : ValidatableDataModel> BindingResult.rejectFromValidationError(
+    matchingError: ObjectError,
+    dataModel: T,
+) {
+    if (matchingError is FieldError && matchingError.field in dataModel::class.declaredMemberProperties.map { it.name }) {
```
Travis-Softwire, line 25: Should this logic be the other way around? I.e. if we've included it in our map we want to map it to a field, otherwise leave it as-is?

src/main/resources/templates/inviteLAUser.html
```diff
@@ -0,0 +1,39 @@
+<!DOCTYPE html>
+<html th:replace="~{fragments/layout :: layout(#{addLAUser.title}, ${serviceName}, ~{::main}, ${#fields.hasErrors('${confirmedEmailDataModel.*}')})}">
+<main class="govuk-main-wrapper" id="main-content">
+    <div th:replace="~{fragments/twoThirdsLayout :: twoThirdsLayout(~{::#header})}">
```
rowanhill, line 4: Rather than having several divs all with 2/3rds styling applied, would it not make sense to have this div wrap the whole page contents? (You can add an id to this div and then select its children with `~{::#your-id-here/contents()}`)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
```diff
@@ -71,39 +75,61 @@ class ManageLocalAuthorityUsersController(
...
+            val token = invitationService.createInvitationToken(emailModel.email, currentAuthority)
             val invitationLinkAddress = invitationService.buildInvitationUri(token)
             emailSender.sendEmail(
-                emailAddress,
+                emailModel.email,
                 LocalAuthorityInvitationEmail(currentAuthority, invitationLinkAddress),
             )
 
-            model.addAttribute("contentHeader", "You have sent a test email to $emailAddress")
+            model.addAttribute("contentHeader", "You have sent a test email to ${emailModel.email}")
```
rowanhill, line 127: This needs adding to messages.properties

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/bindingResultExtensions/BindingResultExtensions.kt
rowanhill: Isn't this "If this is a FieldError for one of the model's properties, reject that property; otherwise, this error doesn't innately belong to one of the model's properties, so we'd better look this error up in the field map to see if we can figure out what to tie it to"?

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/bindingResultExtensions/BindingResultExtensions.kt
rowanhill, line 1: I think a lot of this code is in service of trying to have only a single error message for a field at a time. Are we sure that's what we should do? I could imagine (but not sure!) it's a GDS anti-pattern, as it could mean that if a user has multiple errors they'll need to resubmit the form multiple times?

Separately... all of this customising BindingResult and having properties to order and map errors feels quite complex and messy. I know we've already u-turned on this once, but would we not be better off with a custom annotation? Looks like we'd need to:

1. Create an annotation with @Constraint(validatedBy =[OurCustomValidator::class]). This annotation would be used on a property, and would specify a validator method.
2. Create OurCustomValidator : ConstraintValidator<OurCustomAnnotation, Any?>. This would invoke the named validator method.
3. Write validator methods for annotated properties. These methods would return an empty list if the field is valid, or one or more message keys if not.

Then I think we could get back to just asking for @Valid data models in our controllers and inspecting BindingResults, and we'd have full control over which messages we want to return, and priorities of validation rules, etc.

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/bindingResultExtensions/BindingResultExtensions.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.validation.BeanPropertyBindingResult
+import org.springframework.validation.BindingResult
+import org.springframework.validation.FieldError
+import org.springframework.validation.ObjectError
+import uk.gov.communities.prsdb.webapp.models.dataModels.ValidatableDataModel
+import kotlin.reflect.full.declaredMemberProperties
+
+fun <T : ValidatableDataModel> BindingResult.toBindingResultWithSingleFieldErrorCodes(dataModel: T): BindingResult {
+    val newResult = BeanPropertyBindingResult(dataModel, dataModel::class.simpleName!!)
+    for (errorMessage in dataModel.errorPrecedenceList) {
```
Travis-Softwire, line 12: If we're having to order all of the error messages anyway, we may as well just include them all in the map? (Maybe renaming it to make clear that the properties are declared in priority order).

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
```diff
@@ -71,39 +75,61 @@ class ManageLocalAuthorityUsersController(
...
+        redirectAttributes: RedirectAttributes,
     ): String {
         model.addAttribute("serviceName", SERVICE_NAME)
+        val currentAuthority = localAuthorityDataService.getLocalAuthorityForUser(principal.name)!!
+        model.addAttribute("councilName", currentAuthority.name)
+
+        val result = validator.validateDataModel(emailModel)
+        if (result.hasErrors()) {
+            redirectAttributes.addFlashAttribute(
+                "serializedDataModel",
```
Travis-Softwire, line 113: Probably a silly question, but can we not bind the binding result instead of the data model to avoid re-running the validation on GET as well?

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/bindingResultExtensions/BindingResultExtensions.kt
Travis-Softwire: There doesn't seem to be anything definitive, but I did stumble across a github thread on the design system where the general consensus seemed to be that you should only show the most helpful message per field. 

I do agree (a bit reluctantly!) that custom validators seem to be the more 'Spring' way to deal with this though. We'll have to figure out how to make that as lightweight as possible.

rowanhill: Yes, I just found the same GitHub thread suggesting one error message at time is recommended: https://github.com/alphagov/govuk-design-system-backlog/issues/47#issuecomment-418786434

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
rowanhill: I know I suggested the PRG pattern here, but I've done a bit more reading and looked up some other MHCLG codebases, and I think the consensus is _not_ to redirect if there's a validation error (i.e. only redirect on success). That means we can sidestep the question of what we need to add to session - I think there are also some Spring conveniences we'd be able to take advantage of (i.e. that the BindingResult is automatically added to the Model if it's put in the right place in the controller params) and perhaps some Thymeleaf ones (`#fields.hasErrors` and `th:errors`)

src/main/resources/templates/fragments/errorSummary.html
Travis-Softwire, line 1: To avoid the need to pass in the list of fields we can use something like:
```
<div class="govuk-error-summary__body">
    <ul class="govuk-list govuk-error-summary__list">
        <li th:each="err : ${#fields.detailedErrors()}">
            <a href="#" th:href="'#'+${err.getFieldName()}" th:text="${err.getMessage()}">
                Error message as link to input with error
            </a>
        </li>
    </ul>
</div>
```
This doesn't guarantee the order but does make sure we catch all of the relevant errors in the summary.

rowanhill: I think I've already made that change (nearly equivalently). Is there something wrong with the version I've written here?

Travis-Softwire: Oh, weird, I'm sure it was still showing me the previous version 🤔. Very mild preference for using the getters since they're in the docs whereas the properties aren't I don't think (and from the [source](https://github.com/thymeleaf/thymeleaf-spring/blob/3.1-master/thymeleaf-spring6/src/main/java/org/thymeleaf/spring6/util/DetailedError.java) it looks like they're private) but I don't feel strongly about it.

rowanhill: I was under the impression that SPeL will automatically call the getter from syntax like this?

Travis-Softwire: Ah, you're probably right - I've not spent any time looking at the details of SPeL so if that's what's happening then that's fair enough 👍

PR #33, opened 2024-10-29
PRSD-NONE: Replace  with whenever

AEPR:
Added `mockito.kotlin` in a previous PR, this just removes the very ugly `` `when` `` from the tests and replaces it with the much easier to read `whenever`.


isobel-softwire (approved):
LGTM 👍

PR #34, opened 2024-10-30
PRSD-452: One Login Logout

isobel-softwire:
### Features
- Creates a `/confirm-sign-out` endpoint with a submit button that logs the user out of OneLogin
- Points the OneLogin header's 'Sign Out' link to `/confirm-sign-out`
- Sets the post OneLogin logout redirect URL to `/signout`

Confirm Sign Out Page:
image

### Notes
- Integration tests are required to check that the logout functionality works, but the configuration required for integration tests has not been done yet

### DIFF
--- a/src/main/resources/templates/exampleConfirmSignOut.html
+++ b/src/main/resources/templates/exampleConfirmSignOut.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html th:replace="~{fragments/layout :: layout('Confirm Log Out', ~{::main})}">
+<main class="govuk-main-wrapper" id="main-content">
+    <form method="post" th:action="@{/logout}">
+        <h1 class="govuk-heading-l">
+            Are you sure you want to log out?
+        </h1>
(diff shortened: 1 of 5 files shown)


Travis-Softwire (approved):
LGTM!

src/main/resources/templates/exampleConfirmSignOut.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html th:replace="~{fragments/layout :: layout('Confirm Log Out', ~{::main})}">
+<main class="govuk-main-wrapper" id="main-content">
+    <form method="post" th:action="@{/logout}">
+        <h1 class="govuk-heading-l">
+            Are you sure you want to log out?
+        </h1>
+        <button type="submit" class="govuk-button" data-module="govuk-button">
+            Log Out
```
rowanhill, line 9: I know we're waiting for updates from the IXD / content team, but FWIW I think we should be using the term "sign out" instead of "log out", to be consistent with the One Login header.

PR #35, opened 2024-10-31
PRSD-362: Refactor OS Places stub for better organisation

isobel-softwire:
- Moves controller stub files into local/api/controllers
- Gives OS Places stub endpoint a clearer name

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/CustomSecurityConfig.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/CustomSecurityConfig.kt
@@ -33,15 +33,13 @@ class CustomSecurityConfig {
                     .permitAll()
                     .requestMatchers("/check/**")
                     .permitAll()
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/config/CustomSecurityConfig.kt
```diff
@@ -35,7 +35,7 @@ class CustomSecurityConfig {
                     .permitAll()
                     .requestMatchers("/one-login-local/**")
                     .permitAll()
-                    .requestMatchers("/postcode")
+                    .requestMatchers("/os-places-local/**")
```
Travis-Softwire, line 38: While we're refactoring, could we bring all of the 3rd party API mocks under a common subfolder in the route? E.g. `/local/**`?

Travis-Softwire (approved):
One thought about bringing our local mocks together route-wise to avoid cluttering the filterchain in the long run, but otherwise happy with this :)

PR #36, opened 2024-10-31
PRSD-499: Add Building Name or Number to OS Places integration

isobel-softwire:
### Features
- Uses the OS Places `/find` endpoint (instead of `/postcode`) to search by building name/number and postcode
- Updates the example address lookup page to take in building name/number and postcode

### Notes
- Requests are parameterised to return up to 10 results of minimum match score 0.4. To get a non-empty result the chosen match score requires any of the following inputs:
  - The first half of a postcode and no building name/number
  - Anything more specific than the first half of a postcode and (optionally) a building name/number

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/OSPlacesClient.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/OSPlacesClient.kt
@@ -15,7 +15,13 @@ class OSPlacesClient(
     private val baseURL: String,
     private val apiKey: String,
 ) {
-    fun searchByPostcode(postcode: String): String = getResponse("/postcode?postcode=${URLEncoder.encode(postcode, "UTF-8")}")
+    fun search(
+        buildingNameOrNumber: String,
+        postcode: String,
+    ): String {
+        val query = URLEncoder.encode("$buildingNameOrNumber $postcode", "UTF-8")
+        return getResponse("/find?minmatch=0.4&maxresults=10&query=$query")
+    }
 
     private fun getResponse(endpoint: String): String {
         val request: HttpRequest =
--- a/adrs/0017-os-places-search-strategy.md
+++ b/adrs/0017-os-places-search-strategy.md
@@ -1,0 +1,47 @@
+# ADR-0017: OS Places Search Strategy
+
+## Status
+
+Accepted
+
+Date of decision: 2024-10-31
+
+## Context and Problem Statement
+
+Users of the Private Rented Sector Database will need to enter addresses: landlord contact addresses and property
+addresses, in particular. In aid of this, users will be able to search for UK addresses through entering a building
+name/number and postcode. We are using OS Places to perform these lookups.
+
+Which OS Places endpoint should we use?
+
+## Considered Options
+
+* `/find`
+* `/postcode`
+
+## Decision Outcome
+
+`/find`, because the data returned will not require any extra processing.
+
+## Pros and Cons of the Options
+
+### `/find`
+
+The `/find` endpoint performs a fuzzy match against a freetext query.
+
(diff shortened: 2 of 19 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/OSPlacesClient.kt
```diff
@@ -15,7 +15,13 @@ class OSPlacesClient(
     private val baseURL: String,
     private val apiKey: String,
 ) {
-    fun searchByPostcode(postcode: String): String = getResponse("/postcode?postcode=${URLEncoder.encode(postcode, "UTF-8")}")
+    fun search(
+        buildingNameOrNumber: String,
+        postcode: String,
+    ): String {
+        val query = URLEncoder.encode("$buildingNameOrNumber $postcode", "UTF-8")
+        return getResponse("/find?minmatch=0.4&maxresults=10&query=$query")
```
Travis-Softwire, line 23: Did we get any results that weren't from the correct postcode in any of your manual tests? 

More generally, it's probably a good moment to parse the response into a data class so that we can consume it more easily in later tickets (although appreciate for now we'll just be turning it straight back into a string!)

Travis-Softwire (commented):
Interested to know what the results we get back are like with the new endpoint and parameters - can you do a mini write-up so we've got a record of why we've gone with this vs the other options? I've also suggested one enhancement - but otherwise looking good :)

rowanhill:
> can you do a mini write-up

In fact, I was going to suggest we write a quick ADR on postcode vs search-with-post-code-and-name/number. It doesn't need to be long at all!

src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/OSPlacesClient.kt
isobel-softwire: The json is being parsed into an `AddressDataModel` in the `OSPlacesAddressLookupService` - should the parsing be done in this layer instead?

isobel-softwire:
> > can you do a mini write-up
> 
> In fact, I was going to suggest we write a quick ADR on postcode vs search-with-post-code-and-name/number. It doesn't need to be long at all!

I haven't used ADRs before - what's the process for writing them?

src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/OSPlacesClient.kt
Travis-Softwire: Ah so it is - I was looking at the wrong `search` method! No doing the mapping in the service is fine :)

rowanhill:
There's an /adrs directory in the root of the repo, which is full of markdown files. You just need to create one of those and fill it in - it's nothing fancier than a short record of a decision we made and why. There's file (new_adr) to help create new, blank ADRs from the template, in both powershell and bash flavour as a tiny convience, but it's not doing anything clever; feel free to just copy and paste the template and rename the file if you prefer / if the script is buggy.

isobel-softwire:
> There's an /adrs directory in the root of the repo, which is full of markdown files. You just need to create one of those and fill it in - it's nothing fancier than a short record of a decision we made and why. There's file (new_adr) to help create new, blank ADRs from the template, in both powershell and bash flavour as a tiny convience, but it's not doing anything clever; feel free to just copy and paste the template and rename the file if you prefer / if the script is buggy.

Thanks, written an ADR for the OS Places search strategy 👍

rowanhill:
That's great, thanks! This all looks good to me, then

isobel-softwire:
@Travis-Softwire if it looks good to you too, can I get an approval please?

Travis-Softwire (approved):

PR #38, opened 2024-11-04
PRSD-422: Adds framework for multipage forms

Travis-Softwire:


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataService.kt
```diff
@@ -0,0 +1,87 @@
...
+    fun getJourneyData(): JourneyData = objectToJourneyData(session.getAttribute("journeyData")) ?: mutableMapOf()
+
+    fun getPageData(
+        journeyData: JourneyData,
+        pageName: String,
+        subPageNumber: Int?,
+    ): PageData? {
+        var pageData = objectToJourneyData(journeyData[pageName])
+        if (subPageNumber != null && pageData != null) {
+            pageData = objectToJourneyData(journeyData[subPageNumber.toString()])
```
rowanhill: I'm a bit confused by this. Should `journeyData` here be `pageData`? Is the top level journey data something like

```
{
    [step name]: {
        [property name]: [property value],
        "1": {
            [property name]: [property value]
        }
    }
}
```

I.e. each step has all the properties from the form, and then there's a bunch of int-as-string keys with their own copies of the same set of properties? If so, why not just make the step data a list, and the default object the one at index 0?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLandlordController.kt
```diff
@@ -1,13 +1,47 @@
...
+    var landlordJourney: LandlordJourney,
+) {
     @GetMapping
     fun index(model: Model): String = "registerAsALandlord"
+
+    @GetMapping("/{stepName}")
+    fun getJourneyStep(
+        @PathVariable("stepName") stepName: String,
+        @RequestParam(value = "subpage", required = false) subpage: Int?,
+        session: HttpSession,
```
rowanhill, line 27: Don't think we need this session

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLandlordController.kt
```diff
@@ -1,13 +1,47 @@
...
+        model: Model,
+    ): String = landlordJourney.populateModelAndGetViewName(landlordJourney.getStepId(stepName), model, subpage)
+
+    @PostMapping("/{stepName}")
+    fun postJourneyData(
+        @PathVariable("stepName") stepName: String,
+        @RequestParam(value = "subpage", required = false) subpage: Int?,
+        @RequestParam formData: PageData,
+        session: HttpSession,
```
rowanhill, line 41: Don't think we need this session either

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -0,0 +1,146 @@
...
+    private val journeyType: JourneyType,
+    val steps: List<Step<T>>,
+    val initialStepId: T,
+    val validator: Validator,
+    val journeyDataService: JourneyDataService,
+) {
+    fun getStepId(stepName: String): StepId {
+        val step = steps.singleOrNull { step -> step.id.urlPathSegment == stepName }
+        if (step == null) {
+            throw IllegalArgumentException("No step with name $stepName found")
```
rowanhill, line 26: As far as I can see this is only called from the controller. Should a misnamed step perhaps be a 404 (as it is below when no step has the given id)?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -0,0 +1,146 @@
...
+        }
+        return step.id
+    }
+
+    fun populateModelAndGetViewName(
+        stepId: StepId,
+        model: Model,
+        subPageNumber: Int?,
+        prevPageData: PageData? = null,
```
rowanhill, line 35: Potentially confusing name, given the talk of previous steps and pages below? Below, "prev" means "preceding in the flow of the journey" but here it means "just submitted". So maybe call this `submittedPageData`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -0,0 +1,146 @@
...
+        var requestedStep =
+            steps.singleOrNull { step -> step.id == stepId } ?: throw ResponseStatusException(
+                HttpStatus.NOT_FOUND,
+                "Step ${stepId.urlPathSegment} not valid for journey ${journeyType.urlPathSegment}",
+            )
+        if (!isReachable(journeyData, requestedStep, subPageNumber)) {
+            return "redirect:/${journeyType.urlPathSegment}/${initialStepId.urlPathSegment}"
+        }
+        val prevStepDetails = getPrevStep(journeyData, requestedStep, subPageNumber)
+        val prevPageUrl = getPrevPageUrl(prevStepDetails?.step, prevStepDetails?.subPageNumber)
```
rowanhill, line 47: I think this maybe better named `prevStepUrl`, and the get function similarly?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordJourney.kt
```diff
@@ -0,0 +1,57 @@
...
+                            formModel = PhoneNumberFormModel(),
+                            templateName = "forms/phoneNumberForm",
+                            contentKeys =
+                                mapOf(
+                                    "fieldSetHeading" to "forms.phoneNumber.fieldSetHeading",
+                                    "fieldSetHint" to "forms.phoneNumber.fieldSetHint",
+                                    "label" to "forms.phoneNumber.label",
+                                    "hint" to "forms.phoneNumber.hint",
+                                    "backUrl" to "/",
+                                ),
```
rowanhill: Does anything actually use these contentKeys? The templates have them hard-coded in.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -0,0 +1,146 @@
...
+        currentStep.updateJourneyData(journeyData, pageData, subPageNumber)
+        journeyDataService.setJourneyData(journeyData)
+
+        if (currentStep.saveAfterSubmit) {
+            val journeyDataContextId = journeyDataService.getContextId()
+            journeyDataService.saveJourneyData(journeyDataContextId, journeyData, journeyType, principal)
+        }
+
+        if (currentStep.handleSubmitAndRedirect != null) {
+            return "redirect:${currentStep.handleSubmitAndRedirect?.let { it(journeyData, subPageNumber) }}"
```
rowanhill, line 82: Do you need this ?.let, given you've just tested handleSubmitAndRedirect isn't null?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/JourneyData.kt
```diff
@@ -0,0 +1,16 @@
+package uk.gov.communities.prsdb.webapp.forms.journeys
+
+typealias JourneyData = MutableMap<String, Any?>
+
+typealias PageData = MutableMap<String, Any?>
+
+fun objectToJourneyData(obj: Any?): JourneyData? {
```
rowanhill, line 7: Given this gets used for journey data, step data, and subpage data, maybe it should just be named `objectToStringKeyedMap`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
```diff
@@ -0,0 +1,57 @@
...
+                            formModel = PhoneNumberFormModel::class,
+                            templateName = "forms/phoneNumberForm",
+                            contentKeys =
+                                mapOf(
+                                    "fieldSetHeading" to "forms.phoneNumber.fieldSetHeading",
+                                    "fieldSetHint" to "forms.phoneNumber.fieldSetHint",
+                                    "label" to "forms.phoneNumber.label",
+                                    "hint" to "forms.phoneNumber.hint",
+                                    "backUrl" to "/",
+                                ),
```
rowanhill, line 37: I think my comment has been swallowed by new commits being added - but I don't think these contentKeys are being used. I think the should be, though, so different journeys can have different text?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/Page.kt
```diff
@@ -0,0 +1,54 @@
...
+    fun populateModelAndGetTemplateName(
+        validator: Validator,
+        model: Model,
+        pageData: Map<String, Any?>?,
+        prevStepUrl: String?,
+    ): String {
+        var bindingResult = bindDataToFormModel(validator, pageData)
+        // TODO: Switch out for generic attribute once more generic template fragments are being used
+        model.addAttribute(BindingResult.MODEL_KEY_PREFIX + "formModel", bindingResult)
+        model.mergeAttributes(contentKeys)
```
rowanhill, line 26: Merging feels needlessly dangerous - why not just add the content keys as their own attribute?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/steps/RegisterLandlordStepId.kt
```diff
@@ -0,0 +1,12 @@
...
+
+enum class RegisterLandlordStepId(
+    override val urlPathSegment: String,
+) : StepId {
+    Start("start"),
+    Email("email"),
+    QuickBreak("quick-break"),
+    BestFriendEmail("best-friend-email"),
+    ReviewPhoneNumbers("check-phone-numbers"),
+    PhoneNumber("phone-number"),
```
rowanhill, line 11: I think you've inherited many of these from my spike, and don't need them any more :)

src/main/resources/ValidationMessages.properties
rowanhill, line 1: I think we're ultimately going to end up not using this file, right?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
Travis-Softwire: The IDE was complaining about it even with the null check. I admit I didn't check whether the compiler was happy though, I'll do that.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
Travis-Softwire: Yep, this was me being hasty when writing the template

src/main/resources/ValidationMessages.properties
Travis-Softwire: Yes, this is a weird one. I found some docs that said Thymeleaf would only pick up validation messages from this file and not messages.properties, but I've just checked and it does indeed find them fine in messages.properties, so I'll just move them there.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataService.kt
Travis-Softwire: On the first point, yes that was a mistake that showed up when manually testing looped pages, hopefully now fixed.

On the keys - if it's not a repeated step then it's key is just the step name, otherwise it's as you've put above. The reason I went with this over a list was that this allows nullable casts whereas a list doesn't, so it should fallback to an empty map if, say, something went wrong with some session data loaded from the database. 

I don't mind switching to a list if we think that possibility is remote enough that we don't need to worry about it though.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
Travis-Softwire: Yes, the IDE is determined that there's an issue with just using the function despite the nullcheck, whereas the compiler itself is fine with it. I've added a not null assertion to the call instead to keep everything happy.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
```diff
@@ -0,0 +1,59 @@
...
+                        Page(
+                            formModel = PhoneNumberFormModel::class,
+                            templateName = "forms/phoneNumberForm",
+                            contentKeys =
+                                mapOf(
+                                    "title" to "registerAsALandlord.title",
+                                    "fieldSetHeading" to "forms.phoneNumber.fieldSetHeading",
+                                    "fieldSetHint" to "forms.phoneNumber.fieldSetHint",
+                                    "label" to "forms.phoneNumber.label",
+                                    "hint" to "forms.phoneNumber.hint",
```
rowanhill, line 36: I am sceptical that these messages will be reusable in other journeys that collect phone numbers (and ditto email below) - but we can switch them to per-journey messages as and when we see that as necessary.

src/main/resources/application-local.yml
```diff
@@ -82,23 +82,23 @@ spring:
...
+#      client:
+#        registration:
+#          one-login:
+#            client-id: l0AE7SbEHrEa8QeQCGdml9KQ4bk
+#            client-authentication-method: private_key_jwt
+#            authorization-grant-type: authorization_code
+#            scope: openid
+#        provider:
+#          one-login:
+#            issuer-uri: https://oidc.integration.account.gov.uk/
```
rowanhill, line 101: Any reason for commenting this out?

src/main/resources/messages.properties
```diff
@@ -128,9 +128,27 @@ invitation.header=The local authority issuing that token was: {0,,token}
...
+forms.email.error.missing=You must provide your email address
+forms.email.error.invalidFormat=Enter an email address in the correct format
+
 pagination.previousText=Previous
 pagination.pageText=page
 pagination.nextText=Next
 
-errorSummary.heading=There is a problem
-errorPrefix=Error\: 
+phoneNumber.missing=You must provide your phone number
```
rowanhill, line 154: Is this key used?

src/main/resources/templates/fragments/forms/crsfToken.html
rowanhill, line 1: Typo: Should be CSRF not CRSF

src/main/resources/templates/fragments/forms/questionPage.html
```diff
@@ -0,0 +1,17 @@
+<th:block th:fragment="questionPage(title, formModel, formContent, backUrl)">
+    <th:block id="main-content"
+              th:replace="~{fragments/layout :: layout(${title}, ~{::#main-content/content()}, ${#fields.hasErrors('${formModel.*}')})}">
+        <a href="#" th:replace="~{fragments/forms/backLink :: backLink(${backUrl})}">Back
+            link</a>
+        <main class="govuk-main-wrapper">
+            <div id="page-content"
+                 th:replace="~{fragments/twoThirdsLayout :: twoThirdsLayout(~{::#page-content/content()})}">
+                <form method="post" novalidate th:object="${formModel}">
+                    <th:block th:replace="~{fragments/forms/crsfToken :: crsfToken()}"></th:block>
```
rowanhill, line 10: I saw Isobel use `th:action="@{''}"` - my assumption is that gives you the current URL, and adds in the CSRF token for free. If so, I think that should be how we tackle this.

rowanhill (commented):
Looking pretty nice!

src/main/resources/application-local.yml
Travis-Softwire: Unintentionally included :) This was part of trying to help diagnose the 1L connection refused issue, to rule out us trying to access the 1L integration site somehow

src/main/resources/templates/fragments/forms/questionPage.html
Travis-Softwire: Nice :)

rowanhill (approved):

PR #39, opened 2024-11-05
Chore: Playwright e2e tests now use mock OneLogin

rowanhill:


Travis-Softwire (approved):

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/IntegrationTest.kt
```diff
@@ -51,9 +38,44 @@ abstract class IntegrationTest {
     @MockBean
     lateinit var osPlacesClient: OSPlacesClient
 
-    @MockBean
-    lateinit var securityFilterChain: SecurityFilterChain
-
     @MockBean
     lateinit var principal: Principal
```
JasminConterioSW, line 42: Do we still need the mock Principal? It seems to be working without this I think?

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/CustomSecurityConfig.kt
```diff
@@ -84,5 +83,4 @@ class CustomSecurityConfig(
 
 @Configuration
 @EnableMethodSecurity
-@Profile("!integration-test")
 class EnableMethodSecurityConfig
```
isobel-softwire, line 86: EnableMethodSecurityConfig can be deleted and @EnableMethodSecurity can be moved to the CustomSecurityConfig

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/IntegrationTest.kt
rowanhill: It's currently referenced from the ManageLAUsers integration test - feel free to remove it as part of your work, though.

PR #40, opened 2024-11-05
Feat/prsd 403 add tests

JasminConterioSW:
I'm expecting this to be the last PR for PRSD-403, it includes

- Pagination: redirect to the first page in the table if the user requests a page that doesn't exist (i.e. page 3 where there are only 2)
- Backend pagination tests
- Frontend playwright tests for Manage LA Users
    - I haven't looked for pending users in the table because they may not show on the first page if many more active users get added!
    - I added a couple more pending users to the database so we have enough to need pagination
    - This checks that the Previous, Next and current page links appear correctly for pagination

### DIFF
--- a/src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/ManageLAUsersTests.kt
+++ b/src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/ManageLAUsersTests.kt
@@ -4,7 +4,6 @@ import com.microsoft.playwright.Page
 import com.microsoft.playwright.assertions.PlaywrightAssertions.assertThat
 import kotlin.test.Test
 
-// @ActiveProfiles("local-no-auth")
 class ManageLAUsersTests : IntegrationTest() {
     val localAuthorityId = 1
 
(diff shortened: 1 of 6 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/ManageLAUsersTests.kt
```diff
@@ -1,20 +1,48 @@
 package uk.gov.communities.prsdb.webapp.integration
 
-import org.junit.jupiter.api.BeforeEach
-import org.mockito.kotlin.whenever
+import com.microsoft.playwright.Page
+import com.microsoft.playwright.assertions.PlaywrightAssertions.assertThat
+import kotlin.test.Test
 
+// @ActiveProfiles("local-no-auth")
```
rowanhill, line 7: Looks like this could be removed?

rowanhill (approved):
Looks reasonable to me :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/ManageLAUsersTests.kt
JasminConterioSW: Oops, yes!

PR #42, opened 2024-11-05
PRSD-404: add success page, link from manage users page, add Page Object style e2e tests

rowanhill:
This PR finishes off the last bits of PRSD-404, namely:

- After inviting a new user, we show a properly styled confirmation page
- The manage LA users page has a button to invite new users, which is now wired up to this page
- There are some controller tests for the invite journey
- There are some e2e tests for the invite journey

Note that as part of the above I've introduced:
- Primary and secondary buttons that look like links
- The beginnings of a Page Object style framework for e2e tests; I'm not that familiar with Playwright, so we might want to evolve a bit as we learn what works well, but here's a starting point

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthority.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthority.kt
@@ -7,12 +7,17 @@ import jakarta.persistence.GenerationType
 import jakarta.persistence.Id
 
 @Entity
-class LocalAuthority : ModifiableAuditableEntity() {
+class LocalAuthority() : ModifiableAuditableEntity() {
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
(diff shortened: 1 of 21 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthority.kt
```diff
@@ -7,12 +7,17 @@ import jakarta.persistence.GenerationType
 import jakarta.persistence.Id
 
 @Entity
-class LocalAuthority : ModifiableAuditableEntity() {
+class LocalAuthority() : ModifiableAuditableEntity() {
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
-    val id: Int? = null
+    var id: Int? = null
```
Travis-Softwire, line 13: If we want to keep this as `val` then I _think_ we can declare it using the primary constructor syntax instead (as we don't need a private setter for it)

PR #47, opened 2024-11-07
feat/PRSD-369: Landlord Registration email page

Kiran-R-K:
When you go to the Register as a Landlord page and click Start Now you are redirected to the Email Input page.
The Email step has been set as the first step in the Landlord Journey (although future tickets will change this).
Updated Error messages to match the Figma designs
Updated Copy on the page and the button name to match the Figma designs.
When an email is successfully added the user is redirected to the next step in the journey (the phone number page)

Added integration tests for the Email step of the Landlord Registration Journey
Updated RegisterLandlordController tests to reflect the changes

{694841D9-9223-4EBA-A009-45679CD0C87E}

{263D20FC-25E4-4CC9-B09C-0DCF0ECDBB11}

{AF993533-E0B7-4843-84E1-BD9B06A297BD}

Note: there are Section heading in the Figma designs that have not been implemented by this ticket - a separate [ticket ](https://mhclgdigital.atlassian.net/browse/PRSD-522)has been raised to implement them.

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
@@ -4,7 +4,7 @@ import org.springframework.stereotype.Component
 import org.springframework.validation.Validator
 import uk.gov.communities.prsdb.webapp.constants.enums.JourneyType
 import uk.gov.communities.prsdb.webapp.forms.pages.Page
-import uk.gov.communities.prsdb.webapp.forms.steps.RegisterLandlordStepId
+import uk.gov.communities.prsdb.webapp.forms.steps.LandlordRegistrationStepId
 import uk.gov.communities.prsdb.webapp.forms.steps.Step
 import uk.gov.communities.prsdb.webapp.models.formModels.EmailFormModel
 import uk.gov.communities.prsdb.webapp.models.formModels.PhoneNumberFormModel
@@ -14,15 +14,32 @@ import uk.gov.communities.prsdb.webapp.services.JourneyDataService
 class LandlordRegistrationJourney(
     validator: Validator,
     journeyDataService: JourneyDataService,
-) : Journey<RegisterLandlordStepId>(
+) : Journey<LandlordRegistrationStepId>(
         journeyType = JourneyType.LANDLORD_REGISTRATION,
-        initialStepId = RegisterLandlordStepId.PhoneNumber,
+        initialStepId = LandlordRegistrationStepId.Email,
         validator = validator,
         journeyDataService = journeyDataService,
         steps =
             listOf(
                 Step(
-                    id = RegisterLandlordStepId.PhoneNumber,
+                    id = LandlordRegistrationStepId.Email,
+                    page =
+                        Page(
+                            formModel = EmailFormModel::class,
+                            templateName = "forms/emailForm",
+                            contentKeys =
+                                mapOf(
+                                    "title" to "registerAsALandlord.title",
+                                    "fieldSetHeading" to "forms.email.fieldSetHeading",
+                                    "fieldSetHint" to "forms.email.fieldSetHint",
+                                    "label" to "forms.email.label",
+                                    "backUrl" to "/${JourneyType.LANDLORD_REGISTRATION.urlPathSegment}",
+                                ),
+                        ),
+                    nextAction = { _, subPageNumber: Int? -> Pair(LandlordRegistrationStepId.PhoneNumber, 0) },
+                ),
+                Step(
+                    id = LandlordRegistrationStepId.PhoneNumber,
                     page =
                         Page(
                             formModel = PhoneNumberFormModel::class,
@@ -34,26 +51,9 @@ class LandlordRegistrationJourney(
                                     "fieldSetHint" to "forms.phoneNumber.fieldSetHint",
                                     "label" to "forms.phoneNumber.label",
                                     "hint" to "forms.phoneNumber.hint",
-                                    "backUrl" to "/",
                                 ),
                         ),
-                    nextAction = { _, subPageNumber: Int? -> Pair(RegisterLandlordStepId.Email, 0) },
-                ),
-                Step(
-                    id = RegisterLandlordStepId.Email,
-                    page =
-                        Page(
-                            formModel = EmailFormModel::class,
-                            templateName = "forms/emailForm",
-                            contentKeys =
-                                mapOf(
-                                    "title" to "registerAsALandlord.title",
-                                    "fieldSetHeading" to "forms.email.fieldSetHeading",
-                                    "fieldSetHint" to "forms.email.fieldSetHint",
-                                    "label" to "forms.email.label",
-                                ),
-                        ),
-                    nextAction = { _, subPageNumber: Int? -> Pair(RegisterLandlordStepId.Email, subPageNumber?.plus(1)) },
+                    nextAction = { _, subPageNumber: Int? -> Pair(LandlordRegistrationStepId.Email, 0) },
                 ),
             ),
     )
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
@@ -36,7 +36,7 @@ class LandlordRegistrationJourney(
                                     "backUrl" to "/${JourneyType.LANDLORD_REGISTRATION.urlPathSegment}",
                                 ),
                         ),
-                    nextAction = { _, subPageNumber: Int? -> Pair(LandlordRegistrationStepId.PhoneNumber, 0) },
+                    nextAction = { _, subPageNumber: Int? -> Pair(LandlordRegistrationStepId.PhoneNumber, null) },
                 ),
                 Step(
                     id = LandlordRegistrationStepId.PhoneNumber,
--- a/src/main/resources/messages.properties
+++ b/src/main/resources/messages.properties
@@ -108,7 +108,7 @@ addLAUser.form.submitButton=Submit
 
 addLAUser.error.missingEmail=You must enter an email address
 addLAUser.error.notAnEmail=Enter an email address in the correct format
-addLAUser.error.noConfirmation=You must enter and confirm your email address
+addLAUser.error.noConfirmation=You must enter and confirm their email address
 addLAUser.error.confirmationDoesNotMatch=Both email address should match
 addLAUser.error.retryable=Please try again
 
--- a/src/main/resources/messages.properties
+++ b/src/main/resources/messages.properties
@@ -163,7 +163,8 @@ search=Search for Private Rented Sector information
 forms.errorSummary.heading=There is a problem
 forms.errorMessage.prefix=Error: 
 
-forms.buttons.continue=Continue
+forms.buttons.startNow=Start Now
+forms.buttons.saveAndContinue=Save and continue
 forms.buttons.back=Back
 
 forms.phoneNumber.fieldSetHeading=What is your phone number?
--- a/src/main/resources/messages.properties
+++ b/src/main/resources/messages.properties
@@ -163,7 +163,7 @@ search=Search for Private Rented Sector information
 forms.errorSummary.heading=There is a problem
 forms.errorMessage.prefix=Error: 
 
-forms.buttons.saveAndContinue=Save and Continue
+forms.buttons.continue=Continue
 forms.buttons.back=Back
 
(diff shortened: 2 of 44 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
```diff
@@ -14,46 +14,46 @@ import uk.gov.communities.prsdb.webapp.services.JourneyDataService
...
-                                    "hint" to "forms.phoneNumber.hint",
-                                    "backUrl" to "/",
+                                    "fieldSetHeading" to "forms.email.fieldSetHeading",
+                                    "fieldSetHint" to "forms.email.fieldSetHint",
+                                    "label" to "forms.email.label",
+                                    "backUrl" to "/${JourneyType.LANDLORD_REGISTRATION.urlPathSegment}",
                                 ),
                         ),
-                    nextAction = { _, subPageNumber: Int? -> Pair(RegisterLandlordStepId.Email, 0) },
+                    nextAction = { _, subPageNumber: Int? -> Pair(LandlordRegistrationStepId.PhoneNumber, 0) },
```
Travis-Softwire, line 39: We don't need the 0 here, it can be null

Kiran-R-K: changed

src/main/resources/templates/forms/emailForm.html
```diff
@@ -12,6 +12,6 @@
         <div th:replace="~{fragments/forms/emailInput :: emailInput(#{${label}}, 'emailAddress', null)}">
         </div>
     </th:block>
-    <button th:replace="~{fragments/primaryButton :: primaryButton(#{forms.buttons.saveAndContinue})}"></button>
+    <button th:replace="~{fragments/primaryButton :: primaryButton(#{forms.buttons.continue})}"></button>
```
Travis-Softwire, line 15: I expect in some journeys we'll want this to be save and continue, and in others just continue. I wonder if it should be passed in a contentKey param instead?

src/main/resources/templates/registerAsALandlord.html
```diff
@@ -38,7 +38,7 @@ <h2 class="govuk-heading-m" th:text="#{registerAsALandlord.access.heading}">
                 <p class="govuk-body" th:text="#{registerAsALandlord.access.paragraph}">
                     registerAsALandlord.access.paragraph
                 </p>
-                <a th:replace="~{fragments/startButton :: startButton( '/registration', 'Start Now' )}"></a>
+                <a th:replace="~{fragments/startButton :: startButton(${registerAsALandlordInitialStep}, 'Start Now' )}"></a>
```
Travis-Softwire, line 41: Worth making 'Start Now' a message

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/BasePage.kt
```diff
@@ -4,7 +4,7 @@ import com.microsoft.playwright.Page
 import uk.gov.communities.prsdb.webapp.integration.pageobjects.components.TextInput
 
 abstract class BasePage(
-    protected val page: Page,
+    val page: Page,
```
Travis-Softwire, line 7: Why does this need to be public?

Travis-Softwire (commented):
A couple of Qs but looking good :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/BasePage.kt
```diff
@@ -16,7 +16,11 @@ abstract class BasePage(
 
     protected val header = page.locator("main header h1")
 
+    protected val fieldSetHeading = page.locator(".govuk-fieldset__heading")
+
     abstract fun validate()
 
     protected fun inputFormGroup(fieldName: String) = TextInput(page.locator(".govuk-form-group:has(input[name=\"$fieldName\"])"))
+
+    protected fun fieldsetInput(fieldName: String) = TextInput(page.locator(".govuk-fieldset:has(input[name=\"$fieldName\"])"))
```
rowanhill, line 25: Why do we have some text inputs in a fieldset and a some in a form group? Should we be doing something consistently that we're not?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/EmailFormPage.kt
```diff
@@ -0,0 +1,32 @@
...
+    override fun validate() {
+        assertThat(fieldSetHeading).containsText("What is your email address?")
+    }
+
+    fun fillEmail(text: String) = emailInputFormGroup.input.fill(text)
+
+    inline fun <reified T : BasePage> submit(): T {
+        submitButton.click()
+        page.waitForLoadState()
+        return BasePage.createValid<T>(page)
```
rowanhill, line 21: I think this is a little behind main: createValid now takes the KClass for the page, and it takes care of the waiting for load state

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/Navigator.kt
```diff
@@ -11,6 +12,16 @@ class Navigator(
         return BasePage.createValid<InviteNewLaUserPage>(page)
     }
 
+    fun goToEmailFormPage(journeyType: JourneyType): EmailFormPage {
```
rowanhill, line 15: This is probably too generic for this name - we'll end up with plenty of form pages that take email addresses.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/Navigator.kt
```diff
@@ -11,6 +12,16 @@ class Navigator(
...
+        navigate("${journeyType.urlPathSegment}/email")
+        return BasePage.createValid<EmailFormPage>(page)
+    }
+
+    fun goToPhoneNumberFormPage(journeyType: JourneyType): PhoneNumberFormPage {
```
rowanhill, line 20: Ditto - we'll have many phone number form pages, I expect

src/main/resources/messages.properties
```diff
@@ -139,7 +139,7 @@ search=Search for Private Rented Sector information
 forms.errorSummary.heading=There is a problem
 forms.errorMessage.prefix=Error: 
 
-forms.buttons.saveAndContinue=Save and Continue
+forms.buttons.continue=Continue
```
rowanhill, line 142: How come we've changed this? Should the landlord registration journey be saved as you go?

Kiran-R-K: I changed it inline with the design for this page but that is also inconsistent so I've asked a designer for clarification on the wording. 

I think we should be saving the journey as we go, I suspect they are moving the designs to all using "Save and continue", let's see what they say

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/BasePage.kt
Kiran-R-K: They are all ultimately inside a form group, this difference is because some are part of the multi-form-framework and some aren't. Inputs created using the multi-form-framework are nested inside a second form group, which is then nested in a fieldset, so when use `inputFormGroup()` it finds two matches but using `fieldsetInput()` gets one.

src/main/resources/templates/forms/emailForm.html
Kiran-R-K: I spoke to Nicola about this and she said on the e-mail page it should be `Save and continue`. I suspect all the ones in this journey are about to be updated to reflect the same (if they're not already) I'm inclined to simply change it back for now and make it a contentKey param if/when we have a requirement to - though I am happy to be suggested otherwise

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/Navigator.kt
Kiran-R-K: Will we? Maybe I've misunderstood something about the multi-page-form but wouldn't all/most form pages that take e- mails be using this form page now?

rowanhill: My concern is that different journeys will have slightly different copy, and that'll mean we need slightly different page objects. Maybe I'm wrong though - let's leave it for now, and we can specialise if needed.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/BasePage.kt
rowanhill: But... is there a good reason that multi-page form journey forms have different markup to other pages?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/EmailFormPage.kt
Kiran-R-K: updated

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/Navigator.kt
Kiran-R-K: My intention was that when we started getting to those cases we would start passing the copy in as parameters but I don't know how viable that is with this set up.

rowanhill: Yeah, something like that might work - worth trying when we get there.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/BasePage.kt
Kiran-R-K: The differences are the the result of parallelising tickets and the page that's not using the same mark up as the multi-page forms will do so in future, a ticket has already been made for this work -[here](https://mhclgdigital.atlassian.net/browse/PRSD-539)

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/BasePage.kt
Kiran-R-K: In `EmailFormPage` I wanted to have the tests be able to specify the type of page that is expected to return, in order to do that I had to make the `submit()` function `inline` which in turn required me to make the `page` value public. I couldn't think of a better way to do this, one option would be to not make it generic at all and have various submit buttons that will expect to be redirected to specific pages? 
```kotlin
  inline fun <reified T : BasePage> submit(): T {
        submitButton.click()
        page.waitForLoadState()
        return createValid(page, T::class)
    }
```

rowanhill: Or pass the class explicitly as a param, rather than reified generic type?

rowanhill: FWIW, though, I'd vote for journey-specific subclasses that document where submit is meant to go, rather than the tests essentially asserting it.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/journeyPageBasePages/EmailFormBasePage.kt
```diff
@@ -1,24 +1,23 @@
-package uk.gov.communities.prsdb.webapp.integration.pageobjects.pages
+package uk.gov.communities.prsdb.webapp.integration.pageobjects.pages.journeyPageBasePages
```
Travis-Softwire, line 1: V. minor, but would probably just call this `basePages` in case we want to use the classes in something other than a journey

Travis-Softwire (approved):
One minor comment but otherwise LGTM :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/journeyPageBasePages/EmailFormBasePage.kt
Kiran-R-K: Good idea, I'll also move `BasePage.kt` into the file for consistency

PR #48, opened 2024-11-07
PRSD-406 / PRSD-407: Disallow LA admins editing or removing themselves

rowanhill:
As discussed on Slack, we think it makes sense to prevent LA admin users from editing their own accounts (so they can't downgrade themselves to basic access, and leave the LA with no admins) or removing themselves from their LA. This PR achieves that by:
- Removing the "Change" link the manage users table for the current user
- Throwing a 403 Forbidden if the user tries to access any pages / make any posts to edit themselves
- Throwing a 403 Forbidden if the user tries to access nay pages / make any posts to remove themselves

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
@@ -2,6 +2,7 @@ package uk.gov.communities.prsdb.webapp.controllers
 
 import jakarta.validation.Valid
 import org.springframework.http.MediaType
+import org.springframework.security.access.AccessDeniedException
 import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.stereotype.Controller
 import org.springframework.ui.Model
@@ -38,8 +39,8 @@ class ManageLocalAuthorityUsersController(
         principal: Principal,
         @RequestParam(value = "page", required = false) page: Int = 1,
     ): String {
-        val currentUserLocalAuthority =
-            localAuthorityDataService.getLocalAuthorityIfAuthorizedUser(localAuthorityId, principal.name)
+        val (currentUser, currentUserLocalAuthority) =
+            localAuthorityDataService.getUserAndLocalAuthorityIfAuthorizedUser(localAuthorityId, principal.name)
 
         val pagedUserList =
             localAuthorityDataService.getPaginatedUsersAndInvitations(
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
```diff
@@ -66,7 +68,10 @@ class ManageLocalAuthorityUsersController(
         principal: Principal,
         model: Model,
     ): String {
-        localAuthorityDataService.getLocalAuthorityIfAuthorizedUser(localAuthorityId, principal.name)
+        val (currentUser, _) = localAuthorityDataService.getUserAndLocalAuthorityIfAuthorizedUser(localAuthorityId, principal.name)
+        if (currentUser.id == localAuthorityUserId) {
```
Travis-Softwire, line 72: Probably not for now, but might be worth us considering injecting the various other types of user id (Landlord, LA user etc) into a custom principal on login to save a DB call every time they want to access something specific to them?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
```diff
@@ -99,7 +104,10 @@ class ManageLocalAuthorityUsersController(
         @ModelAttribute localAuthorityUserAccessLevel: LocalAuthorityUserAccessLevelDataModel,
         principal: Principal,
     ): String {
-        localAuthorityDataService.getLocalAuthorityIfAuthorizedUser(localAuthorityId, principal.name)
+        val (currentUser, _) = localAuthorityDataService.getUserAndLocalAuthorityIfAuthorizedUser(localAuthorityId, principal.name)
+        if (currentUser.id == localAuthorityUserId) {
```
Travis-Softwire, line 108: Is this a candidate for method-level security on the update & delete methods for the `localAuthorityDataService`?  It looks like `preAuthorize` can access the params for the method if you annotate them with `@Param`. We'd lose the custom error message internally, but it would prevent us from accidentally forgetting this check somewhere else?

Travis-Softwire (approved):
This LGTM. I had a couple of high level thoughts but they might well be something we want to consider in the future rather than something for this PR.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
rowanhill: As discussed, I think this suggestion pairs with your other one - ideally the principal would contain the la user (id at least) so we can check on the delete method that we're not deleting the current user's account.

PR #50, opened 2024-11-08
PRSD-NONE: Back link is in nav, uses JS if available

rowanhill:
This is a tiny change to:

1. Enclose the "Back" link in `<nav>`, which I think makes sense
2. Use the browser back functionality via JS, if JS is available

I've tested this works happily in Chrome both with JS enabled (back/forward works as you'd expect) and disabled (clicking the back link adds a new item to history).

### DIFF
--- a/src/main/resources/templates/fragments/forms/backLink.html
+++ b/src/main/resources/templates/fragments/forms/backLink.html
@@ -1,2 +1,9 @@
-<a th:fragment="backLink(backUrl)" th:href="@{${backUrl}}" class="govuk-back-link"
-   th:text="#{forms.buttons.back}">forms.buttons.back</a>
+<nav th:fragment="backLink(backUrl)">
+    <a th:href="@{${backUrl}}"
+       class="govuk-back-link"
+       th:text="#{forms.buttons.back}"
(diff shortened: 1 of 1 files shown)


src/main/resources/templates/fragments/forms/backLink.html
```diff
@@ -1,2 +1,9 @@
-<a th:fragment="backLink(backUrl)" th:href="@{${backUrl}}" class="govuk-back-link"
-   th:text="#{forms.buttons.back}">forms.buttons.back</a>
\ No newline at end of file
+<nav th:fragment="backLink(backUrl)">
+    <a th:href="@{${backUrl}}"
```
Travis-Softwire, line 2: Does this mean that we'll still see the value of backUrl on hover? If so, does that mean it sometimes won't match where the button goes when clicked?

rowanhill: Yes, that's true. As an alternative, we could add a little `<script>` here that changes the href to be `javascript:history.go(-1)`?

Travis-Softwire (approved):

PR #52, opened 2024-11-11
PRSD-NONE: Refactor inviteLAUser template to use new input fragments

Travis-Softwire:


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/BasePage.kt
```diff
@@ -34,5 +34,5 @@ abstract class BasePage(
 
     abstract fun validate()
 
-    protected fun inputFormGroup(fieldName: String) = TextInput(page.locator(".govuk-form-group:has(input[name=\"$fieldName\"])"))
+    protected fun inputFormGroup(fieldName: String) = TextInput(page.locator(".govuk-form-group:has(>input[name=\"$fieldName\"])"))
```
JasminConterioSW, line 37: What does the ">" do?

src/main/resources/templates/inviteLAUser.html
```diff
@@ -24,11 +25,13 @@ <h1 class="govuk-heading-l" th:text="#{addLAUser.heading(${councilName})}"></h1>
...
             </ul>
         </section>
-        <form class="form" id="email-form" method="post" th:object="${confirmedEmailDataModel}" novalidate>
-            <!--/*/ <th:block th:replace="~{fragments/forms/csrfToken :: csrfToken()}"></th:block> /*/-->
-            <input th:replace="~{fragments/forms/basicTextInput :: textInput(#{addLAUser.form.primaryEmailLabel}, 'email')}">
-            <input th:replace="~{fragments/forms/basicTextInput :: textInput(#{addLAUser.form.confirmEmailLabel}, 'confirmEmail')}">
-
+        <form th:action="@{''}" class="form" id="email-form" method="post" th:object="${confirmedEmailDataModel}"
+              novalidate>
+            <!--/*/ <th:block id="fieldset-content"
```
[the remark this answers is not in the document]
Travis-Softwire: They're actually real `th:block`s. Thymeleaf let's you do this to keep your template as valid html overall - which can be useful for templates that define pages, but not so useful for fragments.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/BasePage.kt
Travis-Softwire: This means that the selector will only match form-groups that have an input as their immediate child. The reason for this is that the pattern for question pages in the Gov UK design system have nested form-groups, and so this was failing because it matched both the inner and outer form-groups

src/main/resources/templates/inviteLAUser.html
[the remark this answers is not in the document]
JasminConterioSW: Ah ok, that looks fine then

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/BasePage.kt
JasminConterioSW: Ok, thanks :)

JasminConterioSW (approved):
LGTM :)

PR #53, opened 2024-11-11
Feat/prsd 540 create register la user form

JasminConterioSW:
This is the first part of PRSD-405 (I've broken it into sub-tasks)

This code:

- Sets up the RegisterLAUserController
- Sets up the LAUserRegistrationJourney with the first two steps
    - Name which is a new component
    - Email which modifies the existing component to allow different text on the button
    - Includes validation for both of these steps
    - Journey tests

Later tickets will cover
- The "check answers" page
- Handling the form completion
- Accessing the form from the email link
- Controller tests / page tests (these would cover the entry point and the form completion)

image
image
image
image

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LaUserRegistrationJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LaUserRegistrationJourney.kt
@@ -37,7 +37,7 @@ class LaUserRegistrationJourney(
                                     "backUrl" to "/${JourneyType.LA_USER_REGISTRATION.urlPathSegment}/",
                                 ),
                         ),
-                    nextAction = { _, subPageNumber: Int? -> Pair(RegisterLaUserStepId.Email, subPageNumber?.plus(1)) },
+                    nextAction = { _, subPageNumber: Int? -> Pair(RegisterLaUserStepId.Email, null) },
                 ),
                 Step(
                     id = RegisterLaUserStepId.Email,
@@ -52,17 +52,16 @@ class LaUserRegistrationJourney(
                                     "fieldSetHint" to "registerLAUser.email.fieldSetHint",
                                     "label" to "registerLAUser.email.label",
                                     "submitButtonText" to "forms.buttons.continue",
-                                    "backUrl" to "/${JourneyType.LA_USER_REGISTRATION.urlPathSegment}/${RegisterLaUserStepId.Name}",
                                 ),
                         ),
-                    nextAction = { _, subPageNumber: Int? -> Pair(RegisterLaUserStepId.CheckAnswers, subPageNumber?.plus(1)) },
+                    nextAction = { _, subPageNumber: Int? -> Pair(RegisterLaUserStepId.CheckAnswers, null) },
                 ),
                 /*TODO: PRSD-541 - check answers page
                 Step(
                     id = RegisterLaUserStepId.CheckAnswers,
                     page =
                         Page(),
-                    nextAction = { _, subPageNumber: Int? -> Pair(RegisterLaUserStepId.CheckAnswers, subPageNumber?.plus(1)) },
+                    nextAction = { _, subPageNumber: Int? -> Pair(RegisterLaUserStepId.CheckAnswers, null) },
                 ),*/
             ),
     )
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LaUserRegistrationJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LaUserRegistrationJourney.kt
@@ -1,0 +1,52 @@
+package uk.gov.communities.prsdb.webapp.forms.journeys
+
+import org.springframework.stereotype.Component
+import org.springframework.validation.Validator
+import uk.gov.communities.prsdb.webapp.constants.enums.JourneyType
+import uk.gov.communities.prsdb.webapp.forms.pages.Page
+import uk.gov.communities.prsdb.webapp.forms.steps.RegisterLaUserStepId
+import uk.gov.communities.prsdb.webapp.forms.steps.Step
+import uk.gov.communities.prsdb.webapp.models.formModels.EmailFormModel
+import uk.gov.communities.prsdb.webapp.services.JourneyDataService
+
+@Component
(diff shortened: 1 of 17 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LaUserRegistrationJourney.kt
```diff
@@ -0,0 +1,68 @@
...
+                            formModel = NameFormModel::class,
+                            templateName = "forms/nameForm",
+                            contentKeys =
+                                mapOf(
+                                    "title" to "registerLAUser.title",
+                                    "fieldSetHeading" to "forms.name.fieldSetHeading",
+                                    "fieldSetHint" to "forms.name.fieldSetHint",
+                                    "label" to "forms.name.label",
+                                    "submitButtonText" to "forms.buttons.continue",
+                                    "backUrl" to "/${JourneyType.LA_USER_REGISTRATION.urlPathSegment}/",
```
Travis-Softwire, line 37: Does this point to anything at the moment?

JasminConterioSW: Not yet, I'm planning to put the "Before you Register" page there

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LaUserRegistrationJourney.kt
```diff
@@ -0,0 +1,68 @@
...
+                                ),
+                        ),
+                    nextAction = { _, subPageNumber: Int? -> Pair(RegisterLaUserStepId.Email, subPageNumber?.plus(1)) },
```
Travis-Softwire, line 40: We don't need the subpageNumber here - it's only there to handle looped steps, i.e. when we need to capture multiple copies of the same kind of data, like multiple properties, or multiple interested parties. Where we just have the 1 instance of the page in the journey this can be null.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LaUserRegistrationJourney.kt
```diff
@@ -0,0 +1,68 @@
...
+                                mapOf(
+                                    "title" to "registerLAUser.title",
+                                    "fieldSetHeading" to "registerLAUser.email.fieldSetHeading",
+                                    "fieldSetHint" to "registerLAUser.email.fieldSetHint",
+                                    "label" to "registerLAUser.email.label",
+                                    "submitButtonText" to "forms.buttons.continue",
+                                    "backUrl" to "/${JourneyType.LA_USER_REGISTRATION.urlPathSegment}/${RegisterLaUserStepId.Name}",
+                                ),
+                        ),
+                    nextAction = { _, subPageNumber: Int? -> Pair(RegisterLaUserStepId.CheckAnswers, subPageNumber?.plus(1)) },
```
Travis-Softwire, line 58: Same as the point above - the second pair element can be null.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LaUserRegistrationJourney.kt
```diff
@@ -0,0 +1,68 @@
...
+                            formModel = EmailFormModel::class,
+                            templateName = "forms/emailForm",
+                            contentKeys =
+                                mapOf(
+                                    "title" to "registerLAUser.title",
+                                    "fieldSetHeading" to "registerLAUser.email.fieldSetHeading",
+                                    "fieldSetHint" to "registerLAUser.email.fieldSetHint",
+                                    "label" to "registerLAUser.email.label",
+                                    "submitButtonText" to "forms.buttons.continue",
+                                    "backUrl" to "/${JourneyType.LA_USER_REGISTRATION.urlPathSegment}/${RegisterLaUserStepId.Name}",
```
Travis-Softwire, line 55: Because you're going back to the previous step in the journey you can leave out this key - the multi-page form framework will populate it for you (or it will when we resolve the bug you spotted in the other QA! 😊)

src/main/resources/templates/fragments/forms/nameInput.html
```diff
@@ -0,0 +1,6 @@
+<th:block th:fragment="nameInput(label, fieldName, hint)"
+          th:replace="~{fragments/forms/inputGroup :: inputGroup(${label}, ${fieldName}, ~{::input}, ${hint})}">
+    <input class="govuk-input" th:classappend='${#fields.hasErrors(fieldName) ? "govuk-input--error" : ""}' type="text"
```
Travis-Softwire, line 3: Is the only difference between this and the basicTextInput the hint?

JasminConterioSW: Hmm yes, would it be better to repurpose the basicTextInput component?

Travis-Softwire: I think so - it hopefully shouldn't be too much work to get basicTextInput working with a nullable hint

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LaUserRegistrationJourneyTests.kt
```diff
@@ -0,0 +1,55 @@
+package uk.gov.communities.prsdb.webapp.integration
+
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+
+class LaUserRegistrationJourneyTests : IntegrationTest() {
+    @Nested
```
Travis-Softwire, line 7: Are the nested classes just to break up the tests or are they serving another purpose?

Travis-Softwire (commented):
A few comments and Qs, but overall looking really good :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LaUserRegistrationJourneyTests.kt
JasminConterioSW: I think they are just breaking the tests up, to be honest I cribbed heavily from LandlordRegistrationJourneyTests!

Travis-Softwire (approved):

PR #57, opened 2024-11-13
Feat/prsd 370 landlord registration phone number

Kiran-R-K:
# Implemented the phone number page of the Landlord registration journey

## Description
Added copy, styling and error messages from the Figma designs
Implemented UK and international phone number validation using Google’s `libphonenumber` library as recommended by [GDS](https://design-system.service.gov.uk/patterns/phone-numbers/)
Created integration tests

## Screenshots

{77F8F198-72B0-4235-BA71-131C9E4ABA2B}

{082E6071-A83A-4D78-8E65-521D241FECA4}

{F5430BB1-D8F1-4043-B8A6-E142D5BA3718}

## Risks
The phone number validation flow is:
- Is there a value provided
- Can the value be parsed into a phone number (defaulting to UK formatting if possible)
- Is the value a valid UK number
- If not is the value a valid international number AND does it have a country code

There are tests in `LandlordRegistrationJourneyTests` covering this but some of them are flaky (they have been marked with a `TODO`) this seems to be because some phone numbers that are invalid international phone numbers (or valid international phone numbers without country codes) that should return errors are being accepted by validation because they fit the standard for valid UK numbers.
We can also see the same behaviour when we run that app and input values on the page and his validation pattern does follow the requirements for the page. There is a risk that we could accept a phone number as valid but calling it would lead a phone line different to what the user intended. I'm not sure if there are other ways to mitigate this risk.


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/PhoneNumberFormPage.kt
Kiran-R-K, line 1: This page should be deleted it's been superseded by `PhoneNumberBasePage`

Kiran-R-K: deleted

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordRegistrationJourneyTests.kt
```diff
@@ -29,4 +39,80 @@ class LandlordRegistrationJourneyTests : IntegrationTest() {
...
+                val formResponse: Response? = navigator.navigate("$journeyUrl/phone-number")
+                assertThat(formResponse?.url()).contains("$journeyUrl/phone-number")
+            }
+
+            @Test
+            fun `Submitting correct UK numbers without a country code redirects to the next step`() {
+                val formPage = navigator.goToLandlordRegistrationPhoneNumberFormPage()
+                val number = phoneNumberUtil.getExampleNumber("GB")
+                formPage.fillPhoneNumber("${number.countryCode}${number.nationalNumber}")
+                formPage.submit()
```
Travis-Softwire, line 71: Is there any other assertion we can add here (and at the end of other tests without an assertion) to show explicitly that we've verified the next page loaded?

Travis-Softwire (approved):
Apart from the point about making sure we've got some explicit asserts in the tests (which I think you're picking up separately) this LGTM 👍

PR #61, opened 2024-11-14
Spike/prsd 546 multi oauth poc

AEPR:
Proof of concept for re-validating the user with ID verification only on some routes. For the POC this validates on any "local authority" route. Draft PR as this should never be merged - this is to get feedback/agreement on the approach.

Also, I've fully duplicated the mock one login controller for simplicity (the only difference is in the "vot" field that is set when it returns (used in the custom filter to validate the authentication). 

Don't worry about the package the extra classes are in - they'll find themselves a home in the main ticket.

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/CustomSecurityConfig.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/CustomSecurityConfig.kt
@@ -1,10 +1,5 @@
 package uk.gov.communities.prsdb.webapp.config
 
-import jakarta.servlet.Filter
-import jakarta.servlet.FilterChain
-import jakarta.servlet.ServletRequest
-import jakarta.servlet.ServletResponse
-import jakarta.servlet.http.HttpServletRequest
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 import org.springframework.core.annotation.Order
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/config/CustomSecurityConfig.kt
```diff
@@ -24,6 +27,39 @@ class CustomSecurityConfig(
...
+                config.authorizationEndpoint { customizer ->
+                    customizer.authorizationRequestResolver(
+                        AdditionalParameterAddingRequestResolver(clientRegistrationRepository, "vtr", "Cl.Cm.P2"),
+                    )
+                }
+            }.logout { logout ->
+                logout.logoutSuccessHandler(oidcLogoutSuccessHandler())
+            }.csrf { requests ->
+                requests.ignoringRequestMatchers("/local/**")
+            }.addFilterAfter(OauthTokenSecondaryValidatingFilter(::isOauthTokenValid), SecurityContextHolderFilter::class.java)
```
AEPR, line 54: We want to apply this filter immediately after the `SecurityContextHolderFilter` has been applied so that no other filters see the security context retrieved from the `SecurityContextRepository` if it is not a valid oauth token. In the main implementation it might be worth verifying that it is being applied in the correct place.

PR #62, opened 2024-11-14
Chore/prsd none update page model testing

Kiran-R-K:
# Updating the pattern for testing of pages in the multi-page forms

## Description
Created a `FormBasePage` that inherits from `Base Page` that can create and interact with (get values, input values and has `validate()` function)
The hope is that this can be directly inherited by all pages for all journey types (e.g `EmailFormPageLandlordRegistration`), it may need some additions but it should work for most if not all form pages.
Migrated all assertions to tests pages, and created assertions for valid pages loading if they weren't there.

## Post Merge Notes
These changes will need to be communicated to the dev team so they can update any in progress tickets.

## Notes
I have been strictish about making these changes only for the multi-page form components and not the other pages being tested,  but I think we should consider the benefits of following this same pattern for all testing - though that would be a bigger piece of work to implement and should probably be a separate PR.


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/basePages/FormBasePage.kt
```diff
@@ -0,0 +1,36 @@
...
+abstract class FormBasePage(
+    page: Page,
+    val pageHeading: String,
+    val inputLabel: String,
+) : BasePage(page) {
+    val inputFormGroup = fieldsetInput(inputLabel)
+    val submitButton = page.locator("button[type=\"submit\"]")
+
+    override fun validate() {
+        assertThat(fieldSetHeading).containsText(pageHeading)
```
Travis-Softwire, line 16: I'd be tempted to instead have a base implementation in `BasePage` that checks the url - that way `validate` is checking whether the correct page has been navigated to, whereas our tests can then check whether the correct content has been rendered. That would then be reusable across all pages, and the url _should_ be more stable than the heading. What do you think?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LaUserRegistrationJourneyTests.kt
```diff
@@ -28,32 +34,38 @@ class LaUserRegistrationJourneyTests : IntegrationTest() {
...
 
         @Test
         fun `Submitting a valid email redirects to the next step`() {
             val formPage = navigator.goToLaUserRegistrationEmailFormPage()
-            formPage.fillEmail("test@example.com")
-            val nextStep = formPage.submit()
+            formPage.fillInput("test@example.com")
             // This will need to change when the "check answers" page is implemented
-            nextStep.assertHeadingContains("Page not found")
+            val nextStep = BasePage.createValid(formPage.submit(), PageNotFoundPage::class)
```
Travis-Softwire, line 45: To make it even clearer what is happening here, I wonder whether we could a helper function e.g. something like:
```
// Top level function in BasePage.kt
fun assertIsPage<T : BasePage>(page: Page, targetClass: KClass<T>): T {
    return BasePage.createValid(page, targetClass)
}
```
Then our test flow for moving to a new page would be:
```
val nextPage = form.submit()
val notFoundPage = assertIsPage(nextPage, PageNotFoundPage::class)
assertThat(notFoundPage.heading).containsText("Page not found")
```

Travis-Softwire (approved):
A couple of suggestions but I definitely overall prefer this approach. Please do continue to coordinate with the team RE rebasing when this gets merged

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/basePages/FormBasePage.kt
Kiran-R-K: yeah, I prefer that

PR #63, opened 2024-11-14
Feat/prsd 542 access register la user form

JasminConterioSW:
In this PR:

- Add the "Before you register" page as the first step of the journey 
image

- Create an "acceptInvitation" endpoint that the invitation magic link goes to. This will check that the token is valid (i.e. in the local_authority_invitation table of the database). 
    - If it is valid, the token is stored in the session and the user is directed to the "Before you Register" page. 
    - If the token is invalid, the user is redirected to an "Invalid link" page 
image

- Check for a valid token in the session when a registration journey step is loaded

**I've added tests, but for some reason the RegisterLAUserControllerTests aren't hitting the endpoint, if you can see why that would be really helpful!**


src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserControllerTests.kt
```diff
@@ -0,0 +1,71 @@
...
+        whenever(invitationService.getEmailAddressForToken("token123")).thenReturn("invite@example.com")
+
+        mvc.get("/register-local-authority-user/?token=invalid-token").andExpect {
+            status { is3xxRedirection() }
+        }
+
+        val formData = mutableMapOf<String, Any?>("emailAddress" to "invite@example.com")
+        val journeyData = mutableMapOf<String, Any?>("email" to formData)
+
+        verify(journeyDataService).setJourneyData(journeyData)
```
JasminConterioSW, line 69: I'm not sure why, but these controller tests aren't hitting the endpoint (which is why they are all failing)
They should be getting into the `acceptInvitation` function of the `RegisterLAUserController` but they aren't hitting a debug point in there.

Things that do hit that debug point include:
- Running the app and clicking on a magic link, such as http://localhost:8080/register-local-authority-user/?token=75d3175b-0ce7-469b-826c-cd519f52d093
- The InvitationUrlTests, there is a bit where they get the invitation link `mvc.get(invitationCaptor.firstValue.invitationUri)` which hits the debug point in the acceptInvitation endpoint. `invitationCaptor.firstValue.invitationUri` evaluates to `http://localhost/register-local-authority-user/?token=test%20token`

I'm a bit confused, if you can see why this isn't working it would be great!

Travis-Softwire: As discussed - these will need a `@WithMockUser` annotation to get past auth and hit your endpoint :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserControllerTests.kt
```diff
@@ -0,0 +1,71 @@
...
+
+        verify(invitationService).tokenIsValid("invalid-token")
+        verify(invitationService, never()).storeTokenInSession("invalid-token")
+    }
+
+    @Test
+    fun `acceptInvitation prepopulates the email address in journeyData`() {
+        whenever(invitationService.getEmailAddressForToken("token123")).thenReturn("invite@example.com")
+
+        mvc.get("/register-local-authority-user/?token=invalid-token").andExpect {
```
Travis-Softwire, line 62: Should this be token123 instead of invalid-token?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
```diff
@@ -10,24 +10,51 @@ import org.springframework.web.bind.annotation.RequestParam
...
     @GetMapping("/{stepName}")
     fun getJourneyStep(
         @PathVariable("stepName") stepName: String,
         @RequestParam(value = "subpage", required = false) subpage: Int?,
         model: Model,
-    ): String =
-        laUserRegistrationJourney.populateModelAndGetViewName(
+    ): String {
+        val token = invitationService.getTokenFromSession()
+        if (token == null || !invitationService.tokenIsValid(token)) {
```
Travis-Softwire, line 48: If there's an invalid token in the session should we remove it?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LaUserRegistrationJourney.kt
```diff
@@ -16,11 +17,24 @@ class LaUserRegistrationJourney(
...
+                    page =
+                        Page(
+                            formModel = LandingPageFormModel::class,
+                            templateName = "registerLaUser",
+                            content =
+                                mapOf(
+                                    "title" to "registerLAUser.title",
+                                ),
+                        ),
+                    nextAction = { _, subPageNumber: Int? -> Pair(RegisterLaUserStepId.Name, null) },
```
Travis-Softwire, line 36: subPageNumber can be a `_` here as well since we're not relying on it for the result

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityInvitationService.kt
```diff
@@ -24,15 +25,40 @@ class LocalAuthorityInvitationService(
...
-        return invitationRepository.findByToken(tokenUuid).invitingAuthority
+        val invitation = invitationRepository.findByToken(tokenUuid) ?: throw Exception("Token not found in database")
+
+        return invitation
+    }
+
+    fun tokenIsValid(token: String): Boolean {
+        try {
+            getInvitationFromToken(token)
+        } catch (e: Exception) {
```
Travis-Softwire, line 42: As above RE more specific exceptions

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityInvitationService.kt
```diff
@@ -24,15 +25,40 @@ class LocalAuthorityInvitationService(
...
-    // TODO-405: This function call should be set to match the controller method on the new invitation controller once created
     fun buildInvitationUri(token: String): URI =
         MvcUriComponentsBuilder
-            .fromMethodCall(on(ExampleInvitationTokenController::class.java).acceptInvitation(ExtendedModelMap(), token))
+            .fromMethodCall(on(RegisterLAUserController::class.java).acceptInvitation(token))
             .build()
             .toUri()
+
+    fun storeTokenInSession(token: String) {
+        session.setAttribute("token", token)
```
Travis-Softwire, line 56: Can we make this a more specific key, as we'll have a few different invite flows in the application and we'll want to make sure they don't clobber each other?

Travis-Softwire: Probably also worth making it a constant

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
```diff
@@ -10,24 +10,51 @@ import org.springframework.web.bind.annotation.RequestParam
...
 import java.security.Principal
 
 @Controller
 @RequestMapping("/${REGISTER_LA_USER_JOURNEY_URL}")
 class RegisterLAUserController(
     var laUserRegistrationJourney: LaUserRegistrationJourney,
+    var invitationService: LocalAuthorityInvitationService,
+    var journeyDataService: JourneyDataService,
 ) {
+    @GetMapping("/")
```
isobel-softwire, line 25: I think this endpoint should just be `/register-local-authority-user` (rather than `/register-local-authority-user/`)

src/main/resources/messages.properties
```diff
@@ -145,10 +145,21 @@ deleteLAUserSuccess.whatHappensNext.paragraph.two=You''ll need to invite this us
...
+registerLAUser.requirementsList.workDetails=your work details and contact details
+registerLAUser.beginButtonText=Begin registration
+
 registerLAUser.email.fieldSetHeading=What is your work email address?
 registerLAUser.email.fieldSetHint=This is the email you use for work at the local authority
 registerLAUser.email.label=Email address
 
+invalidLink.title=Invalid link
+invalidLink.heading=This link is not valid
+invalidLink.description=Ask you manager or your admin user for the PRS database to invite you again
```
Travis-Softwire, line 161: Might need to either make this more generic (e.g. Ask the person who sent you the link to invite you again) or to make it an LA user specific key

Travis-Softwire (commented):
Have left a few Qs, but overall looking good :)

src/main/resources/messages.properties
JasminConterioSW: I'm going with making it specifically an invalid LA invitation link page, then we can keep the text in Figma

Travis-Softwire (approved):
LGTM 👍

PR #65, opened 2024-11-20
PRSD-560: Landlord Registration - Country of Residence Form

isobel-softwire:
### Features
- Adds conditionally revealed content functionality to radios component
- Creates select component, which makes use of [accessible-autocomplete](https://github.com/alphagov/accessible-autocomplete)
- Loads in place names from CSV files
- Adds `CountryOfResidence` step to the landlord registration journey
- Sets the last step in the landlord registration journey to `CheckAnswers`
- Updates affected tests

### Screenshots
Country of residence form page:
image

Select component with progressive enhancement:
image

No option selected error:
image

No country selected error:
image

### Notes
- Integration tests for this ticket will be done in a future PR


src/main/js/index.js
```diff
@@ -1,3 +1,16 @@
 import {initAll} from 'govuk-frontend'
+import {enhanceSelectElement} from 'accessible-autocomplete'
 
-initAll()
\ No newline at end of file
+initAll()
+
+window.onload = addAccessibleAutocompleteToSelectElements
```
Travis-Softwire, line 6: Minor, but I'd like to keep index.js as uncluttered as possible - could you pop these into their own file inside a `initAutocomplete` function or similar, and then import and call that function from index.js?

isobel-softwire: Done 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
```diff
@@ -75,8 +79,35 @@ class LandlordRegistrationJourney(
...
+                                    "radioOptions" to
+                                        listOf(
+                                            RadiosViewModel(true, "forms.countryOfResidence.radios.option.yes.label"),
+                                            RadiosViewModel(
+                                                value = false,
+                                                labelMsgKey = "forms.countryOfResidence.radios.option.no.label",
+                                                conditionalFragCall =
+                                                    "fragments/forms/select :: select(" +
+                                                        "#{forms.countryOfResidence.radios.option.no.select.label}, " +
+                                                        "'countryOfResidence', null, \${selectOptions})",
```
Travis-Softwire, line 104: I can't immediately think of a better way to handle this other than specifying a fragment here to replace it with, but can we make this a bit simpler by just dropping a whole, non-parameterized fragment? I.e. creating a `countryOfResidenceSelect` fragment, which is just a `<th:block th:replace=etc></th:block>`. That way our code is only coupled to the existence of the fragment, and not it's internal setup.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/RadiosViewModel.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.communities.prsdb.webapp.models.viewModels
+
+data class RadiosViewModel<T>(
+    val value: T,
+    val labelMsgKey: String,
+    val hintMsgKey: String? = null,
+    val conditionalFragCall: String? = null,
```
Travis-Softwire, line 7: If we're going with the suggestion above of this just being the reference to a fragment instead of a whole call, we could just call it `conditionalFragment`?

Travis-Softwire (approved):
Some refactoring suggestions but otherwise happy with this. If you think it would be easier to refactor in a separate PR then I'm fine for you to merge this one and pick those up separately.

PR #66, opened 2024-11-20
Feat/prsd 541 handle la registration form completion

JasminConterioSW:
This finishes the functionality for registering a new LA User.

In this PR:
- Create a "check answers" page, make this the next step after the "email" step
image
- Create a "success" page 
image
- Add name and email columns to the local_authority_users table
- When the form is completed (by clicking the "Confirm" button on the "check your answers page")
    - Add the new user to the local_authority_users table
    - Remove the invitation from the local_authority_invitation table
    - Check - am I clearing the token from the session?
    - Render the success page

Still to look at (maybe in another PR):
- Tests
- Update the Manage Local Authority Users page so it uses the name from the local_authority_users table rather than the name from One-Login

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
@@ -12,6 +12,7 @@ import uk.gov.communities.prsdb.webapp.forms.journeys.LaUserRegistrationJourney
 import uk.gov.communities.prsdb.webapp.forms.journeys.PageData
 import uk.gov.communities.prsdb.webapp.forms.steps.RegisterLaUserStepId
 import uk.gov.communities.prsdb.webapp.services.JourneyDataService
+import uk.gov.communities.prsdb.webapp.services.LocalAuthorityDataService
 import uk.gov.communities.prsdb.webapp.services.LocalAuthorityInvitationService
 import java.security.Principal
 
@@ -21,6 +22,7 @@ class RegisterLAUserController(
     var laUserRegistrationJourney: LaUserRegistrationJourney,
     var invitationService: LocalAuthorityInvitationService,
     var journeyDataService: JourneyDataService,
+    var localAuthorityDataService: LocalAuthorityDataService,
 ) {
     @GetMapping
     fun acceptInvitation(
@@ -87,10 +89,22 @@ class RegisterLAUserController(
     }
 
     @GetMapping("/success")
-    fun submitRegistration(model: Model): String {
+    fun submitRegistration(
+        model: Model,
+        principal: Principal,
+    ): String {
+        val token = invitationService.getTokenFromSession()
+        if (token == null || !invitationService.tokenIsValid(token)) {
+            invitationService.clearTokenFromSession()
+            return "redirect:invalid-link"
+        }
+        val localAuthority = invitationService.getAuthorityForToken(token)
+
         val journeyData = journeyDataService.getJourneyDataFromSession()
 
-        model.addAttribute("localAuthority", "HARDCODED LA")
+        localAuthorityDataService.registerNewUser(principal.name, localAuthority, "Hardcoded name", "hardcoded@email.com")
+
+        model.addAttribute("localAuthority", localAuthority.name)
 
         return "registerLAUserSuccess"
     }
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
@@ -101,8 +101,13 @@ class RegisterLAUserController(
         val localAuthority = invitationService.getAuthorityForToken(token)
 
         val journeyData = journeyDataService.getJourneyDataFromSession()
+        val name = (journeyData["name"] as PageData)["name"].toString()
+        val email = (journeyData["email"] as PageData)["emailAddress"].toString()
 
-        localAuthorityDataService.registerNewUser(principal.name, localAuthority, "Hardcoded name", "hardcoded@email.com")
+        localAuthorityDataService.registerNewUser(principal.name, localAuthority, name, email)
+
+        val invitation = invitationService.getInvitationFromToken(token)
+        invitationService.deleteInvitation(invitation)
 
         model.addAttribute("localAuthority", localAuthority.name)
 
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
@@ -85,4 +85,13 @@ class RegisterLAUserController(
 
         journeyDataService.setJourneyData(journeyData)
     }
+
+    @GetMapping("/success")
+    fun submitRegistration(model: Model): String {
+        val journeyData = journeyDataService.getJourneyDataFromSession()
+
+        model.addAttribute("localAuthority", "HARDCODED LA")
+
+        return "registerLAUserSuccess"
+    }
 }
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthorityUser.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthorityUser.kt
@@ -23,14 +23,23 @@ class LocalAuthorityUser(
     @Column(nullable = false)
     var isManager: Boolean = false
 
+    @Column(nullable = false)
+    var name: String = ""
+
+    @Column(nullable = false)
+    var email: String = ""
+
     @OneToOne(optional = false)
     @JoinColumn(name = "local_authority_id", nullable = false, foreignKey = ForeignKey(name = "FK_LA_USER_LA"))
     lateinit var localAuthority: LocalAuthority
         private set
 
-    constructor(id: Long, baseUser: OneLoginUser, isManager: Boolean, localAuthority: LocalAuthority) : this(id) {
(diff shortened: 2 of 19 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthorityUser.kt
JasminConterioSW, line 1: We aren't using the name from One-Login any more, but we still need the link to a One-Login user because we use it for assigning user roles / auth.

Do we still want to enforce that the one-login id stored in the local_authority_user table is unique? Or is there a situation where the same one-login user might be added more than once e.g. with a different email address?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityDataService.kt
```diff
@@ -101,4 +104,24 @@ class LocalAuthorityDataService(
...
+
+    fun registerNewUser(
+        baseUserId: String,
+        localAuthority: LocalAuthority,
+        name: String,
+        email: String,
+    ) {
+        val oneLoginUser = oneLoginUserRepository.getReferenceById(baseUserId)
+
+        // TODO: I don't think this is modifying the date create / date modified?
```
JasminConterioSW, line 116: The table has columns for created_date and last_modified_date fields (as do several other tables), do we have to set these when we add/update the entries? Is there a way of automating it?

created_date comes from the AuditableEntity class and last_modified_date comes from the ModifiableAuditableEntity

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
```diff
@@ -85,4 +87,32 @@ class RegisterLAUserController(
...
+        val journeyData = journeyDataService.getJourneyDataFromSession()
+        val name = (journeyData["name"] as PageData)["name"].toString()
+        val email = (journeyData["email"] as PageData)["emailAddress"].toString()
+
+        localAuthorityDataService.registerNewUser(principal.name, localAuthority, name, email)
+
+        val invitation = invitationService.getInvitationFromToken(token)
+        invitationService.deleteInvitation(invitation)
+        invitationService.clearTokenFromSession()
+        journeyDataService.clearJourneyDataFromSession()
```
isobel-softwire, line 112: Is there any chance that there'll be unrelated `journeyData` in the session when this is called? If so, maybe only the fields used in this form should be cleared.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthorityUser.kt
JasminConterioSW: At the moment, if you are logged in with a One-Login account that is already in the local_authority_user table, it lets you go through the registration flow and then fails (rendering the standard "there is a problem with the service" page) because it can't add the new user to the table.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthorityUser.kt
```diff
@@ -23,14 +23,29 @@ class LocalAuthorityUser(
     @Column(nullable = false)
     var isManager: Boolean = false
 
+    @Column(nullable = false)
+    var name: String = ""
```
isobel-softwire, line 27: Do we need this attribute if we already have their OneLogin account?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/LaUserRegistrationSummaryPage.kt
```diff
@@ -0,0 +1,62 @@
...
+                "/${JourneyType.LA_USER_REGISTRATION.urlPathSegment}/${RegisterLaUserStepId.Name.urlPathSegment}",
+            ),
+        )
+        formData.add(
+            FormSummaryDataModel(
+                "registerLaUser.checkAnswers.rowHeading.email",
+                (journeyData["email"] as PageData)["emailAddress"],
+                "/${JourneyType.LA_USER_REGISTRATION.urlPathSegment}/${RegisterLaUserStepId.Email.urlPathSegment}",
+            ),
+        )
```
isobel-softwire, line 58: `formData.addAll()` could be used here

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityDataService.kt
isobel-softwire: I think the annotations we used mean they're supposed to update automatically - is this not happening?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
JasminConterioSW: Good question... I guess if the same user was filling in multiple forms so had multiple journeys this could happen.
If that happens, we might have issues with the same field being used in multiple forms? (E.g. email field in local authority user registration flow, and email field in landlord registration flow, might the same user want different information for the equivalent step in different forms?)

isobel-softwire: I don't think using the same field name will be a problem as each form has its own data map 
(E.g. `journeyData = { landlordJourney : {name : ""}, laJourney : {name : ""} }`)

JasminConterioSW: If I put a debug point in updateJourneyDataAndGetViewNameOrRedirect, my journeyData looks like this:
image

Is this what we expect? There's no laJourney, have I missed a bit?

Travis-Softwire: There can only be one journey in the session at a time. For cases where we want them to be able to swap between different journeys, e.g. with property registration, we have the ability to save and load journey data to the database. _That_ includes info about the journey type as well as the user it belongs to and a unique id, so that we can look up  e.g. all property journeys for user X at a later date.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthorityUser.kt
Travis-Softwire: It should be unique, yes - to avoid people accidentally signing up twice. Ideally, (and this can be a separate ticket in the backlog) if you login and we already have you as an LA user we should redirect you to a sensible page - e.g. the LA User dashboard once that's been built.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/LaUserRegistrationSummaryPage.kt
```diff
@@ -0,0 +1,62 @@
...
+            FormSummaryDataModel(
+                "registerLaUser.checkAnswers.rowHeading.localAuthority",
+                localAuthority?.name,
+                null,
+            ),
+        )
+        formData.add(
+            FormSummaryDataModel(
+                "registerLaUser.checkAnswers.rowHeading.name",
+                (journeyData["name"] as PageData)["name"],
```
Travis-Softwire: To make these casts typesafe you can use the `objectToStringKeyedMap` util function - that will return a nullable `mutableMap<String, Any?>`

src/main/resources/templates/registerLAUserSuccess.html
```diff
@@ -0,0 +1,33 @@
...
+            <li th:text="#{registerLaUser.success.whatHappensNext.bullet.two}">
+                registerLaUser.success.whatHappensNext.bullet.two
+            </li>
+            <li th:text="#{registerLaUser.success.whatHappensNext.bullet.three}">
+                registerLaUser.success.whatHappensNext.bullet.three
+            </li>
+        </ul>
+        <div class="govuk-button-group">
+            <!--TODO: Where should this link go? -->
+            <a th:replace="~{fragments/buttons/primaryButtonLink :: primaryButtonLink(@{/local-authority/{id}/}, #{registerLaUser.success.goToDashboardButtonText})}">
```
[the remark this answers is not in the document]
Travis-Softwire: Not yet - you can point it at `/` for now

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/LaUserRegistrationSummaryPage.kt
```diff
@@ -0,0 +1,62 @@
...
+    ): String {
+        val journeyData = journeyDataService.getJourneyDataFromSession()
+        val formData = mutableListOf<FormSummaryDataModel>()
+        val sessionToken = invitationService.getTokenFromSession()
+
+        val localAuthority =
+            if (sessionToken != null) {
+                invitationService.getAuthorityForToken(sessionToken)
+            } else {
+                null
```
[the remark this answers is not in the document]
JasminConterioSW: Probably should be an exception!

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthorityUser.kt
JasminConterioSW: Added [PRSD-601](https://mhclgdigital.atlassian.net/browse/PRSD-601)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityDataService.kt
JasminConterioSW: Added [PRSD-602](https://mhclgdigital.atlassian.net/browse/PRSD-602)

PR #67, opened 2024-11-20
Feat/prsd 547 verify landlord identity

AEPR:
Adds id verification to the start of the landlord registration journey. Currently doesn't *do* anything with the results of id verification, but that is not one of the ACs for this ticket.

I've added `@Profile("!local | local-auth")` to the IdVerificationConfig so that it is not applied during local development. I'm not attached to this as a solution (and I don't love having a 'local' profile controlling production code) but this means that merging this won't break local until we have a more robust local solution.

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/IdVerificationSecurityConfig.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/IdVerificationSecurityConfig.kt
@@ -43,6 +43,12 @@ class IdVerificationSecurityConfig(
         return http.build()
     }
 
+    /**
+     * Sets the authorization base uri to the specified route and adds claims requests and vector of trust parameters to authorization requests made at that route.
+     *
+     * @param authorizationRequestBaseUri The authorization endpoint used for this filter chain. This **must** be matched by the security matchers for the filter to
+     * ensure that the parameters are added to authorization requests
+     */
     private fun OAuth2LoginConfigurer<HttpSecurity>.AuthorizationEndpointConfig.addIdVerificationParametersToAuthorizationWithBaseUri(
         authorizationRequestBaseUri: String,
     ): OAuth2LoginConfigurer<HttpSecurity>.AuthorizationEndpointConfig =
@@ -56,6 +62,10 @@ class IdVerificationSecurityConfig(
                 ),
             )
 
+    /**
+     * Additional parameters that must be added to authorization requests to one-login for the id verification journey.
+     * [One Login Documentation Reference](https://docs.sign-in.service.gov.uk/integrate-with-integration-environment/authenticate-your-user/#make-a-request-for-authentication)
+     */
     private fun oneLoginIdVerificationParameters(): Map<String, String> {
         val claimsRequest =
             """{"userinfo": {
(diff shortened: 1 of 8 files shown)


Travis-Softwire:
The One Login team are demoing their 'simulator' to Ben & I on Monday which I'm hoping will be able to replace our current local dev solution (and remove the need for the local profile check) - you're welcome to join the demo!

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/IdVerificationSecurityConfig.kt
```diff
@@ -0,0 +1,91 @@
...
+@Configuration
+@EnableMethodSecurity
+class IdVerificationSecurityConfig(
+    val clientRegistrationRepository: ClientRegistrationRepository,
+) {
+    @Bean
+    @Order(1)
+    fun idVerificationFilterChain(http: HttpSecurity): SecurityFilterChain {
+        http
+            .securityMatcher("/register-as-a-landlord/verify-identity", "/id-verification/**")
```
Travis-Softwire, line 26: Minor - I think we have a constant for the `register-as-a-landlord` segment?

AEPR: We do, and I've created one for the `verify-identity` segment as well so I should use that!

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/OauthTokenSecondaryValidatingFilter.kt
```diff
@@ -0,0 +1,40 @@
...
+    ) {
+        val securityContext = SecurityContextHolder.getContext()
+        val auth = securityContext.authentication
+        if (auth != null) {
+            if (!(auth is OAuth2AuthenticationToken && isOauthTokenAcceptable(auth))) {
+                SecurityContextHolder.setContext(SecurityContextHolder.createEmptyContext())
+            }
+        }
+
+        chain!!.doFilter(request, response)
```
Travis-Softwire, line 38: I don't suppose we know in what context chain might be null? Just wondering if it's safer to do `chain?.doFilter(...)` instead of `chain!!.doFilter(..)` unless we know that it should never be null in the normal operation of Spring?

Travis-Softwire (approved):
Couple of minor points, but LGTM 👍

AEPR:
Oh, I almost forgot to ask - I'm not happy with the putting the custom filter and custom auth request resolver in the "config" package, but I'm not sure where they *should* go instead. Maybe filter and request resolver subpackages in config? What do you think?

Travis-Softwire:
> Oh, I almost forgot to ask - I'm not happy with the putting the custom filter and custom auth request resolver in the "config" package, but I'm not sure where they _should_ go instead. Maybe filter and request resolver subpackages in config? What do you think?

I'm fine with that suggestion - I don't know that we'll have many other filters (although probably at least one more) or request resolvers but that's not reason not to do it - and we can always refactor later if a more sensible option emerges.

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/OauthTokenSecondaryValidatingFilter.kt
AEPR: I think that this is a Java-Kotlin conversion thing - in Java all classes are always allowed to be null.

The chain should never be null and in other (Java) filters I've seen call `chain.doFitler()` without null checks. I can change the parameter type from `FilterChain?` to `FilterChain`. I believe this is still effectively a hard null check, but happening implicitly when converting between Java and Kotlin. It does mean we don't have any potentially confusing `!!`.

There being a hard null check is fine as it will make the request fail, but if there's no filter chain then there's nothing to service the request anyway.

PR #69, opened 2024-11-22
Feat/prsd 537 landlord registration date of birth fallback

Kiran-R-K:
# DRAFT: Add date of Birth fallback page for Landlord Registration
- Create a dateForm.html template (according to GDS' date input standard)
- Create a new Journey Step
- Add the step as the second step in the Landlord Registration journey
- Add content keys so the page matchets the designs
- Create a DateOfBirthFormModel  with validation
- Create a validator to ensure the to ensure the date is a valid date
- Create a validator to ensure the the user is between 18 and 120 inclusive (using kotlinx-datetime/kotlinx.datetime/Instant)
- Add integration tests
- Update other tests that may be affected by this change

## Notes
This is a draft PR, there is some testing still to be done but I thought it would be worth putting it up as a draft to start getting feedback


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/steps/LandlordRegistrationStepId.kt
```diff
@@ -8,4 +8,6 @@ enum class LandlordRegistrationStepId(
     PhoneNumber("phone-number"),
     CountryOfResidence("country-of-residence"),
     CheckAnswers("check-answers"),
+    Name("name"),
```
isobel-softwire, line 11: the Name step appears twice here

src/main/resources/templates/fragments/forms/errorSummary.html
```diff
@@ -4,7 +4,7 @@
             <h2 class="govuk-error-summary__title" th:text="#{forms.errorSummary.heading}"></h2>
             <div class="govuk-error-summary__body">
                 <ul class="govuk-list govuk-error-summary__list">
-                    <li th:each="error: ${#fields.detailedErrors()}" th:if="${#fields.hasErrors(error.fieldName)}">
+                    <li th:each="error: ${#fields.detailedErrors()}" th:if="${#fields.hasErrors(error.fieldName) && error.message != '_'}">
```
isobel-softwire, line 7: It would be helpful if you left a comment in `ConstraintDescriptor` about setting `messageKey` to '_' if there's no error message

src/main/resources/templates/forms/dateForm.html
```diff
@@ -0,0 +1,16 @@
+<!--/*@thymesVar id="title" type="java.lang.String"*/-->
+<!--/*@thymesVar id="formModel" type="java.lang.Object"*/-->
+<!--/*@thymesVar id="fieldSetHeading" type="java.lang.String"*/-->
+<!--/*@thymesVar id="fieldSetHint" type="java.lang.String"*/-->
+<!--/*@thymesVar id="submitButtonText" type="java.lang.String"*/-->
+<!--/*@thymesVar id="backUrl" type="java.lang.String"*/-->
+<!DOCTYPE html>
+
+<html th:replace="~{fragments/forms/questionPage :: questionPage(#{${title}}, ${formModel}, ~{::#form-content}, ${backUrl})}">
+<th:block id="form-content">
```
isobel-softwire, line 10: Could the fieldSet fragment be used here instead of duplicating it in basicDateInput?

src/main/resources/templates/fragments/forms/basicDateInput.html
```diff
@@ -0,0 +1,48 @@
+<div th:fragment="dateInput(fieldName, dayField, monthField, yearField, fieldSetHeading, fieldSetHint)" class="govuk-form-group" th:classappend="${#fields.hasAnyErrors()} ? 'govuk-form-group--error'">
+        <fieldset class="govuk-fieldset" role="group" th:attrappend="aria-describedby=${#fields.hasAnyErrors() && fieldSetHint != null} ? ${fieldName} + '-hint ' + ${fieldName} + '-error' : ${fieldName} + '-hint'">
```
isobel-softwire, line 2: The logic for aria-describedby is wrong here - it has been fixed in the fieldSet component

isobel-softwire (commented):
Might want to hold off on merging this until the [integration test refactor](https://github.com/communitiesuk/prsdb-webapp/pull/72) has gone through!

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/formModels/DateOfBirthFormModel.kt
```diff
@@ -0,0 +1,151 @@
...
+
+    fun notDayAndYearBlank(): Boolean = !(dateValidator.isBothBlank(day, year))
+
+    fun notMonthAndYearBlank(): Boolean = !(dateValidator.isBothBlank(month, year))
+
+    fun isValidDay(): Boolean {
+        try {
+            dateValidator.isValidDay(day)
+            return true
+        } catch (e: IllegalArgumentException) {
```
Travis-Softwire, line 123: Why is this one catching an `IllegalArgumentException` but the others aren't? And more generally I wonder if the `DateValidator` shouldn't handle the exception and just return `true` or `false`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/DateValidator.kt
```diff
@@ -0,0 +1,79 @@
...
+        val instant = clock.now().toLocalDateTime(TimeZone.of("UTC"))
+        return LocalDate(instant.year, instant.month.value, instant.dayOfMonth)
+    }
+
+    private fun getFullDateString(
+        day: String,
+        month: String,
+        year: String,
+    ): String {
+        val formattedDay = if (day.length == 2) day else "0$day"
```
Travis-Softwire, line 67: You should just be able to pad the string using https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/pad-start.html

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/DateValidator.kt
```diff
@@ -0,0 +1,79 @@
...
+    fun isValidMonth(month: String): Boolean = month.toInt() in 1..12
+
+    fun isValidYear(year: String): Boolean = year.toInt() in 1900..2099
+
+    fun isValidDateOrIncomplete(
+        day: String,
+        month: String,
+        year: String,
+    ): Boolean {
+        if (isDayOrMonthOrYearNotValid(day, month, year)) return true
```
Travis-Softwire, line 35: What's the scenario where we need this result?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/components/TextInput.kt
```diff
@@ -6,6 +6,7 @@ import com.microsoft.playwright.assertions.PlaywrightAssertions.assertThat
 class TextInput(
     locator: Locator,
 ) : BaseComponent(locator) {
+    // TODO can this just return the basic locator - can it be replaced everywhere with an id locator????
```
Travis-Softwire, line 9: Not sure I understand this ToDo?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordRegistrationJourneyTests.kt
```diff
@@ -60,7 +162,11 @@ class LandlordRegistrationJourneyTests : IntegrationTest() {
             fun setUp() {
                 val formPage = navigator.goToLandlordRegistrationNameFormPage()
                 formPage.fillInput("Arthur Dent")
-                formPage.submit()
+                val datesStep = BasePage.createValid(formPage.submit(), DateOfBirthFormPageLandlordRegistration::class)
```
Travis-Softwire, line 165: I'm not super keen on us doing this in a `BeforeEach` as we'll get a cascading failure if we introduce an error on an early page which could be confusing. I think ideally we either:
- Have fewer tests that test paths through the journey, or
- Use a Spy on the JourneyDataService so that the first time it's called it returns journeyData as though the previous pages had already been traversed, and after that forwards the method call to the real service

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageobjects/pages/InviteNewLaUserPage.kt
```diff
@@ -35,6 +35,6 @@ class InviteNewLaUserPage(
     }
 
     fun assertConfirmEmailErrorContains(text: String) {
-        confirmEmailInputFormGroup.assertErrorMessageContains(text)
+//        confirmEmailInputFormGroup.assertErrorMessageContains(text)
```
Travis-Softwire, line 38: Did you mean to leave this commented out?

src/main/resources/templates/fragments/forms/basicDateInput.html
isobel-softwire, line 1: I think this file should be called dateInput, unless there are other types of date input that we're going to use?

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/DateValidator.kt
```diff
@@ -0,0 +1,79 @@
...
+    fun isBothBlank(
+        firstValue: String,
+        secondValue: String,
+    ): Boolean = isBlank(firstValue) && isBlank(secondValue)
+
+    fun isValidDay(day: String): Boolean = day.toInt() in 1..31
+
+    fun isValidMonth(month: String): Boolean = month.toInt() in 1..12
+
+    fun isValidYear(year: String): Boolean = year.toInt() in 1900..2099
```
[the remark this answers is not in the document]
Kiran-R-K: think so, this isn't checking if the input is a valid year input for any specific date (e.g. date of birth in this case) but it's checking if it is a valid year format, I've given it what feels like a sensible range, although we might have to extend it at some point. We have to do this  (and the `isValidDay`, `isValidMonth`) separate to the isValidDateOfBirth checks to satisfy GDS error handling requirements

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/DateValidator.kt
Kiran-R-K: if any of the values individually are not valid we need this function to return true otherwise the constraint validator (which is checking something else) that calls this method will return a false and an incorrect error will be passed to the user

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordRegistrationJourneyTests.kt
Travis-Softwire: To note: this conversation can be resolved as we're picking this up more widely over the next couple of sprints

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/formModels/DateOfBirthFormModel.kt
Kiran-R-K: Good point, I've taken it off this method, but kept it on the validator for now because that's consistent with what some other validators are doing - but I'm happy to change that?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/formModels/DateOfBirthFormModel.kt
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    fun isValidDateForMaximumAge(): Boolean {
+        if (isDayOrMonthOrYearNotValid()) return true
+        val age = dateValidator.getAgeFromDate(day, month, year)
+        return age <= 120
+    }
+
+    fun notHaveErrorStyling(): Boolean =
+        isDayOrMonthOrYearNotValid() || isValidDate() && isValidDateForMinimumAge() && isValidDateForMaximumAge()
```
Travis-Softwire, line 155: You don't need the OR - because in this case `isValidDate` is always true when `isDayOrMonthOrYearNotValid` is true, which means `isDayOrMonthOrYearNotValid || isValidDate` is equivalent to just `isValidDate`

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/DateValidator.kt
```diff
@@ -0,0 +1,66 @@
...
+        month: String,
+        year: String,
+    ): Boolean = isBlank(day) && isBlank(month) && isBlank(year)
+
+    fun isBothBlank(
+        firstValue: String,
+        secondValue: String,
+    ): Boolean = isBlank(firstValue) && isBlank(secondValue)
+
+    fun isValidDay(day: String): Boolean = day.toInt() in 1..31
```
Travis-Softwire, line 20: After a quick check - I think we need to do the same with these as we have with blanks - otherwise if we enter, say, 32 for day and 13 for month, we get 2 lots of the 'enter a valid date' error message'

Travis-Softwire, line 20: These all need to wrap the cast to Int otherwise they throw an exception that isn't caught

Travis-Softwire (approved):

PR #71, opened 2024-11-25
Feat/prsd 541 add new tests

JasminConterioSW:
EDIT: This is ready but will need tweaking when https://github.com/communitiesuk/prsdb-webapp/pull/72/files#diff-bcdf3d3948c9334c5b5bdafde40fabb0387556cbc6d0d6d0d6eb532de53c9e22 is done - might be worth waiting?

This PR is mostly adding tests:

- Fix a bug where navigating to the "success" page with a valid token in the session but with an incomplete form was letting you get to the page and saving an incomplete user in the database (oops...). Now, it throws an exception and gives you a 500 error.
- Point each integration test class to the sql file that should be used in the tests, rather that always using data-local.sql (I needed MockUser to not be in the local_authority_user database for one of my tests)
- Add tests for:
    - The submitRegistration endpoint in RegisterLAUSerController
    - registerNewUser in the localAuthorityDataService
    - integration tests for checkAnswers and succes in LaUserRegistrationJourney


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LaUserRegistrationJourneyTests.kt
```diff
@@ -105,4 +124,57 @@ class LaUserRegistrationJourneyTests : IntegrationTest() {
             ).containsText("Enter an email address in the right format")
         }
     }
+
```
Travis-Softwire, line 127: I think the thing that is tricky with this approach is that if an early page in the journey has an error we get a cascading set of test failures. I think my preference would either be:
- We have a single test that checks the happy flow through the journey, or
- We use a spy on journeyDataService to inject the correct journey data at the start of each test that will let us load that page directly without navigating through the earlier pages

Travis-Softwire (commented):
One substantive comment about the test structure - there's a couple of options there, but happy to chat through it if that would be helpful

Travis-Softwire (approved):

PR #72, opened 2024-11-25
PRSD-NONE: Refactor Integration Test Framework

isobel-softwire:
- Refactors components to be hierarchical and not include assertions
- Refactors pages to define child components and their interactions
- Refactors integration tests to handle all assertions
- Parameterizes 'redirect to initial step' journey tests
- Allows the navigator to skip to form pages by filling in previous ones


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/EditLAUserTests.kt
```diff
@@ -1,61 +1,58 @@
...
+        assertThat(manageUsersPage.table.getCell(0, ACCESS_LEVEL_COL_INDEX)).containsText("Basic")
 
         // Editing the user shows Arthur Dent's page
-        var editUserPage = row.editUser()
-        assertThat(editUserPage.userName).containsText("Arthur Dent")
-        assertThat(editUserPage.email).containsText("Arthur Dent") // TODO PRSD-405: fix when LA users have email addresses
-        assertEquals(EditLaUserPage.AccessLevelSelection.BASIC, editUserPage.accessLevel())
+        var editUserPage = manageUsersPage.clickChangeLinkAndAssertNextPage(0)
+        assertThat(editUserPage.name).containsText("Arthur Dent")
+        // TODO PRSD-405: fix when LA users have email addresses
```
JasminConterioSW, line 20: We will need to update the Edit User and Delete User pages to use the email address stored in the local_authority_user table (this is on main already), I think this should be a new ticket!

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordRegistrationJourneyTests.kt
```diff
@@ -4,192 +4,128 @@ import com.google.i18n.phonenumbers.PhoneNumberUtil
...
+    @ParameterizedTest
+    @MethodSource("provideNonInitialUrlSegments")
+    fun `Redirects to the first step in the journey if session data is not valid for step`(urlSegment: String) {
+        val journeyResponse: Response? = navigator.navigate("$REGISTER_LANDLORD_JOURNEY_URL/$urlSegment")
+        assertThat(journeyResponse?.url()).contains("/$REGISTER_LANDLORD_JOURNEY_URL/$initialStepUrl")
+    }
+
     @Nested
     inner class LandlordRegistrationStepName {
         @Test
```
JasminConterioSW, line 38: Do we want to add a test at the beginning of each step explicitly testing that the page has loaded? 

It's not reviewed yet so might not be what we want, but for the LaUserRegistrationJourneyTests I've added test that check e.g. the heading text on the page when you navigate to it (https://github.com/communitiesuk/prsdb-webapp/pull/71)

isobel-softwire: The navigator methods call `BasePage.createValidPage()`, which checks that the expected page has loaded by calling its `validate()` method. So I don't think we need to check this again in the tests.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/EditLAUserTests.kt
isobel-softwire: Yeah, that makes sense - will write one up now!

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordRegistrationJourneyTests.kt
JasminConterioSW: I think it may still be worth adding an explicit tests (partly because if something breaks so that the page just doesn't load, it's nice to have a test telling you exactly that).

Do we need to also make sure that all the components have a validate() method that checks the page heading? FormBasePage does check the url, but is that enough?
Pages that aren't inheriting from FormBasePage (e.g. LandingPageUserRegistration) _do_ check the heading.

isobel-softwire: I'll double check that `title` is only being used when its unique

JasminConterioSW: Thanks! Other than that, looks good so far (I've mainly checked the tests which are for pages / components I've worked on)

Travis-Softwire (commented):
Thanks for this Isobel, great effort on a tricky refactor - this should definitely make it easier to make new page models as we implement more of the journeys.

I've left a few cross-cutting comments, some of which are related to the slack thread on integration tests I just kicked off. Happy to chat through any of the comments if that's helpful.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordRegistrationJourneyTests.kt
isobel-softwire: In that case I think it makes sense for validate() to always check URLs

isobel-softwire:
This is ready for another look @Travis-Softwire

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/basePages/BasePage.kt
```diff
@@ -1,47 +1,38 @@
...
 import kotlin.test.assertEquals
 
 abstract class BasePage(
     val page: Page,
-    private val title: String? = null,
+    private val urlSegment: String,
 ) {
     companion object {
-        fun <T : BasePage> createValidPage(
+        fun <T : BasePage> assertPageIs(
```
Travis-Softwire, line 14: Minor, but I wonder if there's still benefit in having both of these (with `assertPageIs` basically wrapping `createValidPage` without returning the result. Semantically they're doing different things even though they basically have a shared implementation - I think it's a bit surprising for an assert method to return a new object, but it's equally a bit surprising for a creation method to do an assertion on the thing it's created, so having both might help keep our tests clear. What do you think?

Travis-Softwire (approved):
One thought/ comment, but otherwise happy with this, good work 👍

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/basePages/BasePage.kt
isobel-softwire: So in the tests we would call `createPage`, then `assertPageIs`? If we go with that I think it makes sense to move the page creation part of `assertPageIs` into `createPage` rather than having two methods that run the same code.

PR #74, opened 2024-11-27
PRSD-562: Landlord Registration Address Lookup

isobel-softwire:
### Features
- Adds `LookUpAddress`, `SelectAddress`, `LookUpContactAddress`, and `SelectContactAddress` to `LandlordRegistrationJourney`
- The select address steps make use of `SelectAddressPage`, which prepopulates the model with the results of the lookup address query
- Updates `OSPlacesAPIStubController` to use the given house number to determine how many addresses to return (this will be used to test the UI)
- Removes example lookup functionality

### Screenshots
- Lookup address page
image

- Lookup address page errors
image

- Lookup contact address
image

- Lookup contact address errors
image

- Select (contact) address page
image

- Select (contact) address page (no addresses found)
image

- Select (contact) address page error
image

- Select (contact) address page with real data
image

### Notes
- Integration tests will be implemented in a future ticket


src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/OSPlacesClient.kt
```diff
@@ -16,10 +16,10 @@ class OSPlacesClient(
     private val apiKey: String,
 ) {
     fun search(
-        buildingNameOrNumber: String,
+        houseNameOrNumber: String,
         postcode: String,
     ): String {
-        val query = URLEncoder.encode("$buildingNameOrNumber $postcode", "UTF-8")
+        val query = URLEncoder.encode("$houseNameOrNumber $postcode", "UTF-8")
```
Travis-Softwire, line 22: See more general comment about comma separating rather than separating by spaces (I think this helps if your building is called, say, "Rose Court" as the API can then more easily guess that for `Rose Court A1 1AA` that `Rose Court` is one search term and `A1 1AA` is also one search term.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/RadiosViewModel.kt
```diff
@@ -2,7 +2,8 @@ package uk.gov.communities.prsdb.webapp.models.viewModels
 
 data class RadiosViewModel<T>(
     val value: T,
-    val labelMsgKey: String,
+    val valueStr: String = value.toString(),
```
Travis-Softwire, line 5: What happens if we want to translate the valueStr? It looks like at the moment we're using hard coded values for this? e.g. "basic", "admin", "yes", "no"?

Travis-Softwire (approved):
A couple of Qs/ suggestions, but overall LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/RadiosViewModel.kt
isobel-softwire: `valueStr` is only being used for setting html tag ids. Before, values were being used but this isn't practical for values that aren't a word

isobel-softwire:
> Not directly related to this PR, but can we order the results we get back from the service by match score to try and get the most likely match as the first result? There's also something in the OS Places docs about comma separating the query elements to improve results which might be worth trying (see end of [this page](https://osdatahub.os.uk/docs/places/technicalSpecification)

Added the comma, but it looks like results already come back ordered by match score. Also, there were some commits I forgot to push before creating the PR, so would you mind having another look please.

PR #75, opened 2024-11-27
Feat/prsd 367 use verified identity

AEPR:
This PR is for pulling the verifiedIdentity creation into a separate service, as well as a quick and dirty local configurable implementation of verified identity to remove the need to connect to One Login.

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/OneLoginConfig.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/OneLoginConfig.kt
@@ -88,16 +88,16 @@ class OneLoginConfig {
 
     private fun extractKeysFromJson(didJson: JSONObject): List<ECKey> {
         val keyArray = didJson.getJSONArray("assertionMethod")
+
         val publicKeysSequence =
-            sequence {
-                for (index in 0 until keyArray.length()) {
-                    yield(keyArray.getJSONObject(index))
+            keyArray
+                .map {
+                    val keyWrapper = it as JSONObject
+                    val jsonKey = keyWrapper.getJSONObject("publicKeyJwk")
+                    val keyId = keyWrapper.getString("id")
+                    jsonKey.put("kid", keyId)
+                    ECKey.parse(jsonKey.toString())
                 }
-            }.map {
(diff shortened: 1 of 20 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/config/OneLoginConfig.kt
```diff
@@ -57,4 +80,36 @@ class OneLoginConfig {
...
+
+    private fun retrieveECKeys(): List<ECKey> {
+        val didJson = retrieveDidJsonObject()
+        return extractKeysFromJson(didJson)
+    }
+
+    private fun extractKeysFromJson(didJson: JSONObject): List<ECKey> {
+        val keyArray = didJson.getJSONArray("assertionMethod")
+        val publicKeysSequence =
+            sequence {
```
AEPR, line 92: I've used the `sequence` + `yield` syntax here to turn the JSONArray into something iterable to map the json keys contained into `ECKey`s. I could do this in a big for loop with a mutable list (and add each to the end in the loop. I personally prefer this but would be happy to switch if preferred.

Travis-Softwire: `JSONArray` should already be iterable - I think you can either do: 

- `keyArray.map { val jsonKey = (it as JsonObject).getJSONObject("publicKeyJwk")... }` 
or, if you want to keep using `getJsonObject`, you can do:
- `keyArray.mappedIndex { index, _ -> val jsonKey = keyArray.getJSONObject(index).getJSONObject("publicKeyJwk")...}`

I'm not against using sequence/ yield, but in this case I wonder if just using a map/ mappedIndex isn't a bit simpler?

Travis-Softwire (approved):
Have left a thought on the sequence/ yield point to consider, but this all looks great to me :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/OneLoginConfig.kt
AEPR: Yeah, that is definitely simpler - updated!

PR #79, opened 2024-12-02
Feat/prsd 494 property registration occupancy

JasminConterioSW:
Occupancy page:
- If this is not filled in, the user sees an error "Select whether the property is occupied"
- If the "yes" option is ticked, the user is taken to the "number-of-households" page
- If the "no" option is ticked, the user is taken to the next step (currently a placeholder page)
image

Households page:
- If this is not filled in, the user should see an error "Enter the number of households living in your property"
- If this is filled in but not with an integer, the use should see an error "Number of households in your property must be a number, like 3"
- If this is filled in with an integer, the user should be taken to the "number-of-people" step
image

People page:
- If this is not filled in, the user should see an error "Enter the number of people living in your property"
- If this is filled in but not with an integer, the use should see an error "Number of people in your property must be a number, like 3"
- If this is filled in with an integer, the user should be taken to the next step (currently a placeholder page - should be the same as the "no" option from the occupancy page)
image

Tests:
- The path for an occupied property is included in the `User can navigate the whole journey if pages are correctly filled in` test
- The path for an unoccupied property is in the OccupancyStep nested class
- Errors are tested in nested classes for each step

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
@@ -123,7 +123,7 @@ class PropertyRegistrationJourney(
                                         ),
                                 ),
                         ),
-                    nextAction = { _, _ -> Pair(RegisterPropertyStepId.NumberOfHouseholds, null) },
+                    nextAction = { journeyData, subpage -> occupancyNextAction(journeyData, subpage) },
                 ),
                 Step(
                     id = RegisterPropertyStepId.NumberOfHouseholds,
@@ -170,4 +170,24 @@ class PropertyRegistrationJourney(
                         ),
                 ),
             ),
-    )
+    ) {
+    companion object {
+        private fun occupancyNextAction(
+            journeyData: JourneyData,
+            subPage: Int?,
+        ): Pair<RegisterPropertyStepId, Int?> =
+            when (
+                val propertyIsOccupied =
+                    objectToStringKeyedMap(journeyData[RegisterPropertyStepId.Occupancy.urlPathSegment])
+                        ?.get("occupied")
+                        .toString()
+            ) {
+                "true" -> Pair(RegisterPropertyStepId.NumberOfHouseholds, subPage)
+                "false" -> Pair(RegisterPropertyStepId.PlaceholderPage, null)
+                else -> throw IllegalArgumentException(
+                    "Invalid value for journeyData[\"${RegisterPropertyStepId.Occupancy.urlPathSegment}\"][\"occupied\"]:" +
+                        propertyIsOccupied,
+                )
+            }
+    }
+}
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
@@ -9,6 +9,7 @@ import uk.gov.communities.prsdb.webapp.forms.pages.Page
 import uk.gov.communities.prsdb.webapp.forms.steps.RegisterPropertyStepId
 import uk.gov.communities.prsdb.webapp.forms.steps.Step
 import uk.gov.communities.prsdb.webapp.models.formModels.LandingPageFormModel
+import uk.gov.communities.prsdb.webapp.models.formModels.NumberOfHouseholdsFormModel
 import uk.gov.communities.prsdb.webapp.models.formModels.OccupancyFormModel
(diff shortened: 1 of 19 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
```diff
@@ -104,8 +165,29 @@ class PropertyRegistrationJourney(
...
                                 ),
                         ),
                 ),
             ),
-    )
+    ) {
+    companion object {
+        private fun occupancyNextAction(
+            journeyData: JourneyData,
+            subPage: Int?,
```
JasminConterioSW, line 177: I've included the subpage here because the landlord will be able to add multiple properties, but I haven't done anything else with it e.g. testing for adding multiple properties because I don't think we have a way to get into that yet.

Is there anything else I should be doing with subpages for this journey?

src/main/resources/templates/forms/numberOfHouseholdsForm.html
```diff
@@ -0,0 +1,26 @@
...
+        <div class="govuk-details__text">
+            <p  th:text="#{'forms.numberOfHouseholds.details.text.paragraph.one'}">
+                forms.numberOfHouseholds.details.text.paragraph.one
+            </p>
+            <p th:text="#{'forms.numberOfHouseholds.details.text.paragraph.two'}">
+                forms.numberOfHouseholds.details.text.paragraph.two
+            </p>
+        </div>
+    </details>
+    <button th:replace="~{fragments/buttons/primaryButton :: primaryButton(#{forms.buttons.continue})}"></button>
```
isobel-softwire, line 24: I think we're using 'save and continue' for the property registration journey

JasminConterioSW: I thought we were using "Continue" everywhere from [this discussion on slack](https://softwire.slack.com/archives/C070WSGTGQY/p1731594933096409)?

isobel-softwire: Oh I thought that was just for the landlord registration journey, as we're setting `saveAfterSubmit` to false in its steps. If it applies across the board then I think the same needs to done in the property registration journey's steps. @Travis-Softwire is that right?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
Travis-Softwire: We have a separate ticket for multiple properties, so for now I'd leave it out and we'll come back and add them as part fo that ticket

src/main/resources/templates/forms/numberOfHouseholdsForm.html
Travis-Softwire: Yes that's right - we decided we weren't going to save/ resume the landlord registration journey, but we will want that for property registration, so it should be Save and Continue here :) Sorry, I didn't communicate that v. well

JasminConterioSW: Ah ok, fair enough - I'll change those to "Save and Continue"

I'm not sure if any of the other journeys need updating?
For LA User registration, it currently says "Continue" on the buttons, but it doesn't have saveAfterSubmit set to false.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationJourneyTests.kt
```diff
@@ -77,4 +101,90 @@ class PropertyRegistrationJourneyTests : IntegrationTest() {
...
+            householdsPage.form.submit()
+            assertThat(householdsPage.form.getErrorMessage()).containsText("Number of households in your property must be a number, like 3")
+        }
+
+        @Test
+        fun `Submitting with a negative integer in the numberOfHouseholds field returns an error`(page: Page) {
+            val householdsPage = navigator.goToPropertyRegistrationHouseholdsPage()
+            householdsPage.householdsInput.fill("-2")
+            householdsPage.form.submit()
+            assertThat(householdsPage.form.getErrorMessage()).containsText("Number of households in your property must be a number, like 3")
```
Travis-Softwire, line 153: If we're using the same error message for negative numbers (which makes sense) should the message say something like `...must be a positive, whole number, like 3`?

Travis-Softwire (approved):
Other than a couple of v. minor points, all looking great :)

src/main/resources/templates/forms/numberOfHouseholdsForm.html
Travis-Softwire: Ah, it probably should have `saveAfterSubmit` set to false as we won't be saving/ resuming that one either. Can you pop a bug ticket in the backlog for that?

PR #80, opened 2024-12-02
PRSD-562: Cache Looked-up Addresses

isobel-softwire:
- Matches `AddressDataModel` to the MVP Address data model
- Creates and tests `AddressDataService`, which handles address caching
- Caches looked-up addresses in after retrieving them in `SelectAddressPage`
- Overrides `isSatisfied` on landlord registration select address steps to check that the submitted value matches a cached address

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
@@ -22,6 +22,7 @@ import uk.gov.communities.prsdb.webapp.models.formModels.PhoneNumberFormModel
 import uk.gov.communities.prsdb.webapp.models.formModels.SelectAddressFormModel
 import uk.gov.communities.prsdb.webapp.models.viewModels.RadiosButtonViewModel
 import uk.gov.communities.prsdb.webapp.models.viewModels.SelectViewModel
+import uk.gov.communities.prsdb.webapp.services.AddressDataService
 import uk.gov.communities.prsdb.webapp.services.AddressLookupService
 import uk.gov.communities.prsdb.webapp.services.JourneyDataService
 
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
```diff
@@ -332,6 +338,14 @@ class LandlordRegistrationJourney(
                 Pair(LandlordRegistrationStepId.CheckAnswers, null)
             }
 
+        private fun isSelectAddressSatisfied(
+            pageData: PageData,
+            addressDataService: AddressDataService,
+        ): Boolean {
+            val selectedAddress = pageData["address"].toString()
```
Travis-Softwire, line 345: What happens if there's no `address` page data in the journey data, e.g. because they've tried to send a GET request for a page that's after this page in the journey without having selected an address?

Travis-Softwire (approved):
On substantive point about how we handle the case where the data is missing from the journey data, which I think we need to handle, but otherwise LGTM 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
isobel-softwire: There's a null check in `Journey.getPrevStep()` that prevents `isSelectAddressSatisfied` from being called in this case. I'll add null-safety in case we end up using the method somewhere else though

isobel-softwire: Actually, Kotlin's indexing operator is already null-safe

Travis-Softwire: Nice :)

PR #83, opened 2024-12-03
Chore/prsd none add filtered journey data to step details

Kiran-R-K:
Added `filteredJourneyData` to `StepDetails` that are returned by `journey.getPrevStep` so we can use the journey data for the journey path a user is currently on separate from all the data they may have input on pages that are no longer part of their journey path

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
@@ -129,9 +129,7 @@ abstract class Journey<T : StepId>(
             // This stores journeyData for only the journey path the user is on
             // and excludes user data for pages in the journey that belong to a different path
             val stepData = journeyDataService.getPageData(journeyData, currentStep.name, null)
-            if (stepData != null && currentStep.isSatisfied(validator, stepData)) {
-                filteredJourneyData[currentStep.name] = stepData
-            }
+            filteredJourneyData[currentStep.name] = stepData
 
             val (nextStepId, nextSubPageNumber) =
                 currentStep.nextAction(journeyData, currentSubPageNumber)
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
@@ -121,10 +121,14 @@ abstract class Journey<T : StepId>(
         var prevStep: Step<T>? = null
         var prevSubPageNumber: Int? = null
         var currentSubPageNumber: Int? = null
-        var filteredJourneyData: JourneyData? = null
+        var filteredJourneyData: JourneyData = mutableMapOf()
         while (!(currentStep.id == targetStep.id && currentSubPageNumber == targetSubPageNumber)) {
             val pageData = journeyDataService.getPageData(journeyData, currentStep.name, currentSubPageNumber)
             if (pageData == null || !currentStep.isSatisfied(validator, pageData)) return null
+            val stepData = journeyDataService.getPageData(journeyData, currentStep.name, null)
+            if (stepData != null && currentStep.isSatisfied(validator, stepData)) {
+                filteredJourneyData[currentStep.name] = stepData
+            }
             val (nextStepId, nextSubPageNumber) =
                 currentStep.nextAction(journeyData, currentSubPageNumber)
             val nextStep = steps.singleOrNull { step -> step.id == nextStepId } ?: return null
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -120,9 +121,18 @@ abstract class Journey<T : StepId>(
         var prevStep: Step<T>? = null
         var prevSubPageNumber: Int? = null
         var currentSubPageNumber: Int? = null
+        var filteredJourneyData: JourneyData = mutableMapOf()
         while (!(currentStep.id == targetStep.id && currentSubPageNumber == targetSubPageNumber)) {
             val pageData = journeyDataService.getPageData(journeyData, currentStep.name, currentSubPageNumber)
             if (pageData == null || !currentStep.isSatisfied(validator, pageData)) return null
+
+            // This stores journeyData for only the journey path the user is on
+            // and excludes user data for pages in the journey that belong to a different path
```
Kiran-R-K, line 130: I'm not sure about the wording of this, is there a clearer way to phrase it?

Travis-Softwire: I think this if fine

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -120,9 +121,18 @@ abstract class Journey<T : StepId>(
...
+            val stepData = journeyDataService.getPageData(journeyData, currentStep.name, null)
+            if (stepData != null && currentStep.isSatisfied(validator, stepData)) {
```
Travis-Softwire, line 132: I don't think we need/ want this check - we know `stepData` is not null from the `pageData` check above, and in the case of there being subpages the `currentStep.isSatisfied` check could fail even though the stepData as a whole is fine

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/JourneyTests.kt
```diff
@@ -358,6 +361,100 @@ class JourneyTests {
             assertEquals("testPropertyValue", propertyValue)
             assertEquals("templateName", result)
         }
+
+        @Test
+        fun `calls populateModelAndGetTemplateName() from page with filteredJourneyData as a parameter when is present`() {
+            val page =
```
Travis-Softwire, line 367: V. minor - I think this lives in the `// Arrange` section :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/JourneyTests.kt
```diff
@@ -358,6 +361,100 @@ class JourneyTests {
...
+                                nextAction = { _, _ -> Pair(TestStepId.StepFour, null) },
+                            ),
+                            Step(
+                                TestStepId.StepFour,
+                                page = spiedOnPage,
+                            ),
+                        ),
+                )
+            val model = BindingAwareModelMap()
+            val pageDataStepOne: PageData = mutableMapOf("testPropertyStepOne" to "testPropertyValueStepOne")
```
Travis-Softwire, line 419: I think something has gone slightly wrong here. The TestFormModel needs a property called `testProperty` in the pageData for `isSatisfied` to return true, otherwise you'll get redirected.

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/JourneyTests.kt
```diff
@@ -358,6 +361,100 @@ class JourneyTests {
...
+                    TestStepId.StepThree.urlPathSegment to pageDataStepThree,
+                )
+            whenever(
+                mockJourneyDataService.getPageData(
+                    anyMap(),
+                    anyString(),
+                    anyOrNull(),
+                ),
+            ).thenReturn(
+                pageDataStepFour,
```
Travis-Softwire, line 442: I think you probably want to spy on the `JourneyDataService` rather than mocking it completely, that way you can tell it to give you your journey data instead of looking it up from the session, but it can still extract the `pageData`/ `stepData` itself when asked. The alternative would be to chain together a series of `thenReturns` here so that it returns the page data for the right sequence of pages (because it will get called twice for each step now I think) but that's quite fiddly.

Travis-Softwire (commented):
I think something has gone a bit wrong with the test, and there's an extra check in the main code that I think needs to be removed, but otherwise this is looking good. If need be you can pull the test out to get the rest merged so that it can be used for the confirmation page

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
Kiran-R-K: ahh ok, I'll remove

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/JourneyTests.kt
Kiran-R-K: ahh I see, thanks

Kiran-R-K:
I'm following your suggestion of pulling out the test so the rest can get merged - so this is ready to go now

Travis-Softwire (approved):

PR #84, opened 2024-12-04
Feat/prsd 490 address lookup

JasminConterioSW:
Add address selection pages to the Property Registration journey

- Lookup address (using the FormModel and template from Landlord Registration)
- Select address (using the FormModel and templatefrom Landlord Registration)
    - Selecting an address (other than "1, Example Road, EG") takes you to the next step (property type)
    - Selecting "I want to manually enter my address" takes you to /register-property/manual-address (doesn't exist yet)
    - If the property is already registered, the user is redirected to an already registered page. For now, this happens if you select "1, Example Road, EG"
We are not properly checking if the property is already registered yet, we'll implement this in https://mhclgdigital.atlassian.net/browse/PRSD-637

image
image
image

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
@@ -8,6 +8,7 @@ import uk.gov.communities.prsdb.webapp.constants.PLACE_NAMES
 import uk.gov.communities.prsdb.webapp.constants.REGISTER_LANDLORD_JOURNEY_URL
 import uk.gov.communities.prsdb.webapp.constants.enums.JourneyType
 import uk.gov.communities.prsdb.webapp.forms.pages.ConfirmIdentityPage
+import uk.gov.communities.prsdb.webapp.forms.pages.LandlordRegistrationSummaryPage
 import uk.gov.communities.prsdb.webapp.forms.pages.Page
 import uk.gov.communities.prsdb.webapp.forms.pages.SelectAddressPage
 import uk.gov.communities.prsdb.webapp.forms.pages.VerifyIdentityPage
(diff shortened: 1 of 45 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
```diff
@@ -402,11 +410,6 @@ class LandlordRegistrationJourney(
             return selectedAddress == MANUAL_ADDRESS_CHOSEN || addressDataService.getAddressData(selectedAddress) != null
         }
 
-        private fun getSelectedAddress(
```
JasminConterioSW, line 405: I moved this / changed it to a generic getFieldValue in the JourneyDataService as I'm using the same function in the PropertyRegistrationJourney

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationJourneyTests.kt
```diff
@@ -65,6 +98,58 @@ class PropertyRegistrationJourneyTests : IntegrationTest() {
...
+            // assertPageIs(page, ManualAddressFormPagePropertyRegistration::class)
+        }
+
+        @Test
+        fun `Selecting and already-registered address navigates to the AlreadyRegistered step`(page: Page) {
+            whenever(
+                osPlacesClient.search("1", "EG1 2AB"),
+            ).thenReturn(
+                "{'results':[{'DPA':{'ADDRESS':'1, Example Road, EG1 2AB'," +
+                    "'LOCAL_CUSTODIAN_CODE':100,'UPRN':'1123456','BUILDING_NUMBER':1,'POSTCODE':'EG1 2AB'}}]}",
```
JasminConterioSW, line 144: For now, the 1123456 uprn is hardcoded to be an already registered property, this will change in https://mhclgdigital.atlassian.net/browse/PRSD-637

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataService.kt
```diff
@@ -88,4 +88,10 @@ class JourneyDataService(
     fun clearJourneyDataFromSession() {
         session.setAttribute("journeyData", null)
     }
+
+    fun getFieldValue(
+        journeyData: JourneyData,
+        urlPathSegment: String,
+        fieldName: String,
+    ): String = objectToStringKeyedMap(journeyData[urlPathSegment])?.get(fieldName).toString()
```
Travis-Softwire, line 96: I'm not against having a helper that extracts individual fields, but I worry that isn't as reusable as we'd need it to be - as I think we can (and probably will) have fields that aren't strings at some point , and we'll need to handle subpages. For now I think it's probably enough to:
- rename to `getFieldStringValue`
- add a subPageNumber parameter which defaults to null
- delegate getting the page data to the `getPageData` function and then here we just need to extract the requested field

Also worth noting this will return the string "null" if the page is missing - I think this needs to return a `String?` and you'll need an extra `?` before the call to `toString()`

Travis-Softwire (approved):
One substantive point, but overall looks great :)

src/main/resources/templates/forms/alreadyRegisteredForm.html
isobel-softwire, line 1: While it is in a journey, I don't think this page counts as a form - maybe it should be something along the lines of `alreadyRegisteredPropertyPage`?

PR #88, opened 2024-12-05
PRSD-496: Search basis adr draft

AEPR:
Proposed ADR for choosing trigram search as the basis of searching through e.g. names, emails, phone numbers etc.

### DIFF
--- a/adrs/0020-exact-trigram-search.md
+++ b/adrs/0020-exact-trigram-search.md
@@ -1,0 +1,162 @@
+# ADR-0020: Exact trigram search
+
+## Status
+
+Draft
+
+Date of decision: Pending
+
+## Context and Problem Statement
+
+Local authority users need to be able to find both landlord and property records in the database using a search
+function so that they can interact with those records. There should be eventually over 4.6 million records in
+the database so search needs to be relatively fast and good at finding intended records over millions of rows.
+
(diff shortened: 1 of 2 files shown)


adrs/0020-exact-trigram-search.md
```diff
@@ -0,0 +1,162 @@
...
+* No indexing
+* Add GIN Index
+* Add GiST Index
+
+## Decision Outcome
+
+No firm decision reached because major factors are unknown - decision to be finalised when there is a realistic test
+dataset to measure performance on.
+
+Interim decision of Concatenate data values, word similarity over threshold & N results with the highest word similarity
```
Travis-Softwire, line 58: Is there an obvious X, Y, Z for "If it's slow we should try X first", "If we're getting lots of false positives we should try Y first", "If we're excluding the desired result a lot we should try Z first"? If so worth capturing that here.

Travis-Softwire (approved):
One thought about capturing some more of what is currently in your brain, but otherwise LGTM. I think we can take 0019 as approved as well as that seems like a slam dunk.

adrs/0020-exact-trigram-search.md
AEPR: I've added a section on this - are you happy for this to get merged in?

Travis-Softwire: LGTM 👍

PR #89, opened 2024-12-05
PRSD-640: Create Landlord and Address Entities

isobel-softwire:
- Adds rest of fields to `Landlord` entity
- Creates `Address` entity
- Creates migration for the created/updated entities
- Updates `AddressDataModel` to match `Address`


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Landlord.kt
```diff
@@ -1,19 +1,45 @@
...
 import jakarta.persistence.GeneratedValue
 import jakarta.persistence.GenerationType
 import jakarta.persistence.Id
 import jakarta.persistence.JoinColumn
 import jakarta.persistence.OneToOne
+import java.util.Date
 
 @Entity
-class Landlord {
+class Landlord : ModifiableAuditableEntity() {
```
Travis-Softwire, line 14: If it's straight forward to do, now is probably also a good moment to pull name and email up into `Landlord` from `OneLoginUser` - this was something that came out of the LA user work, but it might well be that an individual has a different name & email for different roles, e.g. they might have a different surname professionally vs. in their private life.

src/main/resources/db/migrations/V1_6_0__landlord_and_address_tables.sql
```diff
@@ -0,0 +1,66 @@
+CREATE TABLE address
+(
+    id                  BIGINT GENERATED BY DEFAULT AS IDENTITY NOT NULL,
+    last_modified_date  TIMESTAMPTZ(6),
+    created_date        TIMESTAMPTZ(6),
+    uprn                BIGINT,
+    single_line_address VARCHAR(255)                            NOT NULL,
```
Travis-Softwire, line 7: Are 255 characters definitely going to be enough for all of these fields?

src/main/resources/db/migrations/V1_6_0__landlord_and_address_tables.sql
```diff
@@ -0,0 +1,66 @@
...
+);
+
+ALTER TABLE landlord
+    ADD address_id BIGINT;
+
+ALTER TABLE landlord
+    ADD created_date TIMESTAMPTZ(6);
+
+ALTER TABLE landlord
+    ADD date_of_birth TIMESTAMPTZ(6);
```
Travis-Softwire, line 27: This might be one of the few cases where we don't want to store timezone information - as we generally treat someone's DoB as being that date anywhere in the world

src/main/resources/db/migrations/V1_6_0__landlord_and_address_tables.sql
```diff
@@ -0,0 +1,66 @@
...
+    ALTER COLUMN is_active SET NOT NULL;
+
+ALTER TABLE landlord
+    ALTER COLUMN phone_number SET NOT NULL;
+
+ALTER TABLE landlord
+    ALTER COLUMN subject_identifier SET NOT NULL;
+
+ALTER TABLE landlord
+    ADD CONSTRAINT uc_landlord_address UNIQUE (address_id);
```
Travis-Softwire, line 57: Why do we need a unique constraint on the address? What happens if more than one Landlord lives at the same address?

Travis-Softwire (approved):
I had a couple of Qs about some of the fields - but once we've bottomed those out this should be good to go!

src/main/resources/db/migrations/V1_6_0__landlord_and_address_tables.sql
isobel-softwire: I've changed the ones that might be longer, according to OS Places/ our validation

src/main/resources/db/migrations/V1_6_0__landlord_and_address_tables.sql
isobel-softwire: If we used a many-to-one relationship, then when a user changes their address, it would take more careful logic to reflect this in the database without invalidating the records of other users with the same address. The unique constraint enforces a one-to-one relationship (as far as landlords are concerned).

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Landlord.kt
isobel-softwire: Pushed these changes 👍

PR #90, opened 2024-12-06
PRSD-641: Create Landlord

isobel-softwire:
### Features
- Creates minor migration for making `Landlord.uprn` unique
- Creates `AddressDataModel.parseAddressDataModel()` (for converting a set of manual address fields into an `AddressDataModel`)
- Creates `AddressService.findOrCreateAddress()`
- Creates `LandlordService.createLandlordAndReturnRegistrationNumber()`

### Tests
- Tests for created/updated methods in 
  - `RegistrationNumberService`
  - `AddressService`
  - `LandlordService`
  - `AddressDataModel`

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Address.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Address.kt
@@ -5,9 +5,10 @@ import jakarta.persistence.Entity
 import jakarta.persistence.GeneratedValue
 import jakarta.persistence.GenerationType
 import jakarta.persistence.Id
+import uk.gov.communities.prsdb.webapp.models.dataModels.AddressDataModel
 
 @Entity
-class Address : ModifiableAuditableEntity() {
+class Address() : ModifiableAuditableEntity() {
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     private val id: Long? = null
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Address.kt
```diff
@@ -45,4 +47,20 @@ class Address : ModifiableAuditableEntity() {
 
     lateinit var custodianCode: String
         private set
+
+    constructor(addressDataModel: AddressDataModel) : this() {
+        this.uprn = addressDataModel.uprn
+        this.singleLineAddress = addressDataModel.singleLineAddress
+        if (addressDataModel.organisation != null) this.organisation = addressDataModel.organisation
```
Travis-Softwire, line 54: I hadn't clocked that these are all nullable - should the fields also be `String?` rather than `String`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/AddressDataModel.kt
```diff
@@ -15,4 +15,29 @@ data class AddressDataModel(
...
+            county: String? = null,
+        ): AddressDataModel =
+            AddressDataModel(
+                singleLineAddress =
+                    parseSingleLineAddress(addressLineOne, townOrCity, postcode, addressLineTwo, county),
+                townName = townOrCity,
+                postcode = postcode,
+            )
+
+        fun parseSingleLineAddress(
```
Travis-Softwire, line 34: As above, although in this case it's small enough that you couple probably just inline it if we're not using this anywhere else

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/AddressDataModel.kt
```diff
@@ -15,4 +15,29 @@ data class AddressDataModel(
     val locality: String? = null,
     val townName: String? = null,
     val postcode: String? = null,
-)
+) {
+    companion object {
+        fun parseAddressDataModel(
```
Travis-Softwire, line 20: `parse` usually implies it's taking a string. I wonder if this could just be `fromFormData` or something else if this isn't quite form data?

Travis-Softwire (approved):
Couple of minor points but overall looks great!

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/AddressDataModel.kt
isobel-softwire: I think @Kiran-R-K is going to use this method to display manually entered addresses on the check answers page

PR #91, opened 2024-12-09
Feat/prsd 496 landlord search

AEPR:
Adds trigram search behind an example endpoint.


src/main/resources/application.yml
```diff
@@ -7,6 +7,9 @@ spring:
     username: ${RDS_USERNAME}
     password: ${RDS_PASSWORD}
 
+    hikari:
+      connection-init-sql: SET pg_trgm.word_similarity_threshold = 0.2
```
Travis-Softwire, line 11: Nice :)

src/main/resources/db/migrations/V1_7_0__add_trigrams.sql
Travis-Softwire, line 1: Worth checking if anyone else is currently adding a migration

src/main/resources/application.yml
AEPR: 🧙‍♂️

PR #92, opened 2024-12-09
Feat/prsd 491 manual address entry

JasminConterioSW:
- Add a manual address page (similar to the landlord registration journey)
- Add a "Select local authority" page 
- Added tests

image

UPDATED SCREENSHOT
image

### DIFF
--- a/src/main/resources/templates/forms/selectLocalAuthorityForm.html
+++ b/src/main/resources/templates/forms/selectLocalAuthorityForm.html
@@ -6,15 +6,15 @@
 <!--/*@thymesVar id="selectLabel" type="java.lang.Object"*/-->
 <!DOCTYPE html>
 <html th:replace="~{fragments/forms/questionPage :: questionPage(#{${title}}, ${formModel}, ~{::#form-content}, ${backUrl})}">
-<div id="form-content">
-    <div id="fieldset-content"
+<th:block id="form-content">
+    <fieldset id="fieldset-content"
               th:replace="~{fragments/forms/fieldSet :: fieldSet(~{::#fieldset-content/content()}, 'localAuthorityCustodianCode', #{${fieldSetHeading}}, #{${fieldSetHint}})}">
         <p class="govuk-body">
             <span th:text="#{forms.selectLocalAuthority.paragraph.one.beforeLink}">forms.selectLocalAuthority.paragraph.one.beforeLink</span>
             <a class="govuk-link" href="#" th:text="#{forms.selectLocalAuthority.paragraph.one.linkText}">forms.selectLocalAuthority.paragraph.one.linkText</a>
         </p>
-        <th:block th:replace="~{fragments/forms/select :: select(#{${selectLabel}},'localAuthorityCustodianCode',null, ${selectOptions})}"></th:block>
-    </div>
+        <select th:replace="~{fragments/forms/select :: select(#{${selectLabel}},'localAuthorityCustodianCode',null, ${selectOptions})}"></select>
+    </fieldset>
     <button th:replace="~{fragments/buttons/primaryButton :: primaryButton(#{forms.buttons.saveAndContinue})}"></button>
-</div>
+</th:block>
 </html>


src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/LocalAuthorityConstants.kt
```diff
@@ -0,0 +1,24 @@
...
+
+fun readLocalAuthoritiesCsv(fileName: String): List<LocalAuthorityDataModel> {
+    val inputStream: InputStream = ResourceUtils.getFile(fileName).inputStream()
+    val reader = inputStream.bufferedReader()
+    reader.readLine()
+
+    val localAuthorities = mutableListOf<LocalAuthorityDataModel>()
+    reader.forEachLine { line ->
+        val uprn = line.split(",")[0]
+        val nameInQuotationMarks = line.split("\"")
```
JasminConterioSW, line 15: There is one LA which has commas in it's name (Bournemouth, Christchurch and Poole Council), in the CSV file the name is surrounded by "".
This gets that name before splitting up fields by comma!

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/PlaceNameConstants.kt
```diff
@@ -10,7 +10,8 @@ fun readPlaceNamesCsv(fileName: String): List<String> {
 
     val placeNames = mutableListOf<String>()
     reader.forEachLine { line ->
-        val placeName = line.split(",")[1].replace("\"", "")
+        val nameInQuotationMarks = line.split("\"")
```
JasminConterioSW, line 13: Again there are a couple of countries where the full name wasn't showing because it had a comma in the middle (e.g. "Pitcairn, Henderson, Ducie and Oeno Islands")

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/LocalAuthorityConstants.kt
isobel-softwire, line 1: Local authorities aren't capitalized on the check answers page in the Figma - they should probably be formatted here

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/PlaceNameConstants.kt
isobel-softwire: Nice spot 😄

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/steps/RegisterPropertyStepId.kt
```diff
@@ -8,6 +8,7 @@ enum class RegisterPropertyStepId(
     SelectAddress("select-address"),
     ManualAddress("manual-address"),
     AlreadyRegistered("already-registered"),
+    SelectLocalAuthority("select-local-authority"),
```
isobel-softwire, line 11: This can just be `LocalAuthority` and `"local-authority"` if there's no other page that deals with local authorities

src/main/resources/messages.properties
```diff
@@ -289,6 +289,7 @@ forms.selectAddress.error.missing=Select an address
 forms.selectAddress.addAddressManually=I want to manually enter my address
 
 forms.manualAddress.fieldSetHeading=What is your address?
```
isobel-softwire, line 291: This can be changed to `forms.manualAddress.landlordRegistration.fieldSetHeading` to fix the new line

isobel-softwire (approved):
Left a few minor comments but LGTM 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/LocalAuthorityConstants.kt
AEPR: 🤦Of course there is. There's something upsetting about the fact that it's only one exception.

I think that this is non-obvious enough that it's probably worth including a comment in this class to the effect of "commas in local authority names are escaped by containing the name in quotation marks" or similar.

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/LocalAuthorityConstants.kt
```diff
@@ -0,0 +1,24 @@
...
+        val displayName = if (nameInQuotationMarks.size > 1) nameInQuotationMarks[1] else line.split(",")[1]
+
+        localAuthorities.add(LocalAuthorityDataModel(uprn, displayName))
```
AEPR, line 18: I assume that there's been no discussion of treating non-english councils differently? I heard it mentioned on Thursday, but I expect that it's a "maybe someday" feature. Might be worth flagging in standup at some point to make sure there's something in the backlog, on the roadmap or whereever it should belong!

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/LocalAuthorityConstants.kt
```diff
@@ -0,0 +1,24 @@
+package uk.gov.communities.prsdb.webapp.constants
```
AEPR, line 1: I'm made very vaguely uncomfortable that this and placenames are in files called constants when technically they're not `const`s, but being read from a file on startup, but I'm not sure what else we'd call this and they're never changing at run time so I don't think there's any real problem.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/SelectViewModel.kt
```diff
@@ -3,4 +3,5 @@ package uk.gov.communities.prsdb.webapp.models.viewModels
 data class SelectViewModel<T>(
     val value: T,
     val labelMsgKey: String? = null,
```
AEPR, line 5: I don't think `labelMsgKey` is ever actually used, which (if I'm right) would mean you can get rid of it and wouldn't need to have the nested ternary statement in the select form component. 

It also makes sense that when you're using a select there are likely to be enough options that they don't all get their own internationalisable labels. What do you think?

AEPR (changes requested):
Looking good - just a couple of minor thoughts/queries!

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/LocalAuthorityConstants.kt
AEPR: I think we should probably just trust the formatting in the CSV rather than a DIY solution until we have reason to do that and we can make it a separate ticket. It feels like it will probably be a lot of work that would be better served by just storing the nicely formatted names in the first place!

src/main/resources/templates/fragments/forms/select.html
```diff
@@ -4,7 +4,9 @@
             th:attr="aria-describedby=${hint != null} ? ${fieldName} + '-hint'">
         <option value=""></option>
         <option th:each="option: ${options}" th:value="${option.value}"
-                th:text="${option.labelMsgKey != null ? #messages.msg(option.labelMsgKey) : option.value}">
+                th:text="${option.labelMsgKey != null ? #messages.msg(option.labelMsgKey)
+                    : (option.label != null ? option.label
+                    : option.value)}">
```
AEPR, line 9: I think this is a little unclear to read as is - ideally there'd be a way to not do a nested ternary but failing that maybe 
```
${option.labelMsgKey != null 
    ? #messages.msg(option.labelMsgKey)
    : (option.label != null ? option.label : option.value)}
```
?

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/LocalAuthorityConstants.kt
isobel-softwire: Yeah that makes sense - its annoying that the CSV isn't formatted nicely

src/main/resources/data/local_authorities/addressbase-local-custodian-codes.csv
```diff
@@ -0,0 +1,386 @@
+AUTH_CODE,ACCOUNT_NAME,ACCOUNT_TYPE_NAME,ADDRESS_SUBMITTER_FLAG,STREET_SUBMITTER_FLAG,SUBMITTER_FLAG
```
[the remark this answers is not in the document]
JasminConterioSW: Sorry, I should have left a comment telling you not to!

src/main/resources/messages.properties
```diff
@@ -301,6 +302,13 @@ forms.manualAddress.postcode.error.missing=Enter postcode
 
 forms.manualContactAddress.fieldSetHeading=What is your UK contact address?
 
+forms.selectLocalAuthority.custodianCode.error.missing=Select a local authority to continue
+forms.selectLocalAuthority.fieldSetHeading=What local authority area is your property in?
+forms.selectLocalAuthority.fieldSetHint=Select the local authority that your property falls under.
+forms.selectLocalAuthority.paragraph.one.beforeLink=You can
+forms.selectLocalAuthority.paragraph.one.linkText=find your local authority by searching using the property postcode (opens in new tab).
```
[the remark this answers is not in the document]
JasminConterioSW: Oops you're quite right, I missed that bit!

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/SelectViewModel.kt
JasminConterioSW: Hmm I think you're right, when then are only a few options we're more likely to be using radio buttons...
At the moment, this is used in the CountryOfResidence form page but only the value is used.

src/main/resources/templates/fragments/forms/select.html
JasminConterioSW: We don't need this if we are taking out the labelMsgKey (if it turns out we do need it, we can always add it back in!)

src/main/resources/templates/forms/selectLocalAuthorityForm.html
```diff
@@ -0,0 +1,20 @@
...
+    <div id="fieldset-content"
+              th:replace="~{fragments/forms/fieldSet :: fieldSet(~{::#fieldset-content/content()}, 'localAuthorityCustodianCode', #{${fieldSetHeading}}, #{${fieldSetHint}})}">
+        <p class="govuk-body">
+            <span th:text="#{forms.selectLocalAuthority.paragraph.one.beforeLink}">forms.selectLocalAuthority.paragraph.one.beforeLink</span>
+            <a class="govuk-link" href="#" th:text="#{forms.selectLocalAuthority.paragraph.one.linkText}">forms.selectLocalAuthority.paragraph.one.linkText</a>
+        </p>
+        <th:block th:replace="~{fragments/forms/select :: select(#{${selectLabel}},'localAuthorityCustodianCode',null, ${selectOptions})}"></th:block>
+    </div>
+    <button th:replace="~{fragments/buttons/primaryButton :: primaryButton(#{forms.buttons.saveAndContinue})}"></button>
+</div>
```
AEPR, line 19: Sorry - I think I've given a bit of a bum steer on this. I agree with my original point but didn't understand how the `questionPage` fragment worked (it creates an `html` tag and a `form` tag). What I wanted was for the structure of this to approximately match the eventual rendered page i.e.

```
<html...
    <form...
        <fieldset....
        <button....
```

I also think we should avoid using `th:blocks` if they're going to be replaced (but use them when we need to group elements for Thymeleaf but not have a real element!). 

So, I think `#fieldset-content` should be a `fieldset` and the `select` fragment should be a `select`. `#form-content` shouldn't be a real div when thymeleaf has processed so either it should be a `form` with `th:remove="tag"` or it should be a `th:block` (as you had intially).

JasminConterioSW: Ah ok, I think this is right now?

AEPR: Yeah - so do I. I'm probably being a little bit pernickety about this to be honest (and I know you're following the convention from another file). I'll bring it up at Catch Up tomorrow anyway as it would probably good if we standardised on how we're doing this more!

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/LocalAuthorityConstants.kt
AEPR: I've just realised that I've actually also got an outstanding missing external link on another page. I think that both of these issues should probably be raised outside of this PR which otherwise all looks good to me!

AEPR (approved):
Looks good!

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/LocalAuthorityConstants.kt
JasminConterioSW: Moved this discussion to slack :)
https://softwire.slack.com/archives/C07P6L76A1L/p1733837893612729

PR #95, opened 2024-12-10
Prsd 578 select licensing type

AEPR:
Added the lincensing type page and integration tests. I'm not sure if I've added this in exactly the correct place in the journey but updating the `nextAction`s later once the all the pages exist should be relatively easy (and means we're not blocked on previous steps now!)

### DIFF
--- a/src/main/resources/templates/forms/licensingTypeForm.html
+++ b/src/main/resources/templates/forms/licensingTypeForm.html
@@ -1,6 +1,7 @@
 <!--/*@thymesVar id="title" type="java.lang.String"*/-->
 <!--/*@thymesVar id="formModel" type="java.lang.Object"*/-->
 <!--/*@thymesVar id="fieldSetHeading" type="java.lang.String"*/-->
+<!--/*@thymesVar id="fieldSetHint" type="java.lang.String"*/-->
 <!--/*@thymesVar id="backUrl" type="java.lang.String"*/-->
 <!--/*@thymesVar id="radioOptions" type="java.lang.Object"*/-->
 <!DOCTYPE html>
@@ -9,7 +10,7 @@
 using a `form` element with `th:remove="tag" instead of a `th:block` means this page is a form as a prototype -->
 <form id="form-content" th:remove="tag">
     <fieldset id="fieldset-content"
-              th:replace="~{fragments/forms/fieldSet :: fieldSet(~{::#fieldset-content/content()}, 'licensingType', #{${fieldSetHeading}}, null)}">
+              th:replace="~{fragments/forms/fieldSet :: fieldSet(~{::#fieldset-content/content()}, 'licensingType', #{${fieldSetHeading}}, #{${fieldSetHint}})}">
         <input th:replace="~{fragments/forms/radios :: radios(null,'licensingType',null, ${radioOptions})}">
     </fieldset>
     <button th:replace="~{fragments/buttons/primaryButton :: primaryButton(#{forms.buttons.continue})}"></button>
--- a/src/main/resources/templates/forms/licensingTypeForm.html
+++ b/src/main/resources/templates/forms/licensingTypeForm.html
@@ -5,6 +5,8 @@
 <!--/*@thymesVar id="radioOptions" type="java.lang.Object"*/-->
 <!DOCTYPE html>
 <html th:replace="~{fragments/forms/questionPage :: questionPage(#{${title}}, ${formModel}, ~{::#form-content}, ${backUrl})}">
+<!-- The `questionPage` fragment places this content in a `form` element inside an `html` element;
+using a `form` element with `th:remove="tag" instead of a `th:block` means this page is a form as a prototype -->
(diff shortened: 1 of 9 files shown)


src/main/resources/templates/forms/licensingTypeForm.html
```diff
@@ -0,0 +1,15 @@
+<!--/*@thymesVar id="title" type="java.lang.String"*/-->
+<!--/*@thymesVar id="formModel" type="java.lang.Object"*/-->
+<!--/*@thymesVar id="fieldSetHeading" type="java.lang.String"*/-->
+<!--/*@thymesVar id="backUrl" type="java.lang.String"*/-->
+<!--/*@thymesVar id="radioOptions" type="java.lang.Object"*/-->
+<!DOCTYPE html>
+<html th:replace="~{fragments/forms/questionPage :: questionPage(#{${title}}, ${formModel}, ~{::#form-content}, ${backUrl})}">
+<form id="form-content" th:remove="tag">
```
AEPR, line 8: `questionPage` creates a `form` element inside an `html` element (and puts the content inside). The button below is unhappy if it doesn't live in a form element, so I've made this a self-removing `form` so that this works as a form if rendered in prototype mode. 

I don't know 100% if this is the best approach rather than using e.g. a `th:block` instead.

JasminConterioSW:
Could you add a screenshot of the new page please?

src/main/resources/templates/forms/licensingTypeForm.html
benbelow: This doesn't seem unreasonable to me - I wonder if it would be sensible to leave a comment here explaining this though, as I could see it being a bit confusing without this explanation.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationJourneyTests.kt
```diff
@@ -97,6 +98,13 @@ class PropertyRegistrationJourneyTests : IntegrationTest() {
         // fill in and submit
         peoplePage.peopleInput.fill("2")
         peoplePage.form.submit()
+
```
JasminConterioSW, line 101: To follow the pattern of the rest of the test, I think you need 

`val licensingTypePage = assertPageIs(page, LicensingTypeformPagePropertyRegistration`

here.
Then the next few lines should be using licensingTypePage instead of propertyTypePage

If this is working, it does suggest giving new variable names isn't doing anything but it maybe just makes it more readable!

JasminConterioSW: Actually line 79 also has a rogue `propertyTypePage.form.submit()` which should be `ownershipTypePage.form.submit()` (which will have been my fault, sorry!)

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationJourneyTests.kt
```diff
@@ -326,4 +334,14 @@ class PropertyRegistrationJourneyTests : IntegrationTest() {
                 .containsText("Number of people in your property must be a positive, whole number, like 3")
         }
     }
+
+    @Nested
+    inner class LicensingTypeStep {
+        @Test
+        fun `Submitting with no propertyType selected returns an error`(page: Page) {
```
JasminConterioSW, line 341: Again I think the names need updating from propertyType

AEPR: Oops - that's embarrasing. 😳

AEPR:
When first loaded:
image

With an error:
image

JasminConterioSW:
The figma design looks like it has a field set hint (although it's not showing on the error page design... I assume it's meant to be there!)
image

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationJourneyTests.kt
```diff
@@ -338,7 +340,7 @@ class PropertyRegistrationJourneyTests : IntegrationTest() {
     @Nested
     inner class LicensingTypeStep {
         @Test
-        fun `Submitting with no propertyType selected returns an error`(page: Page) {
+        fun `Submitting with no licensingType selected returns an error`(page: Page) {
             val propertyTypePage = navigator.goToPropertyRegistrationLicensingTypePage()
             propertyTypePage.form.submit()
             assertThat(propertyTypePage.form.getErrorMessage()).containsText("Select the type of licensing for the property")
```
JasminConterioSW, line 346: Still a couple of rogue propertyTypePages around I'm afraid!

JasminConterioSW (approved):
A couple of minor points, but generally looks good to me!

PR #98, opened 2024-12-11
Feat/prsd 579 selective licence

AEPR:
Adding the selective licence page if the landlord selects "selective licence" on the licensing page. I've decided to make this the "main line" through the form for journey integration test because I'm doing it first, so I've added the tests for all the other selections as dummy tests (i.e. testing that they reach the placeholder page) so they're easy to populate later. 

Page:
image

No number submitted:
image

Licence number too long:
image

### DIFF
--- a/src/main/resources/templates/forms/selectiveLicenceForm.html
+++ b/src/main/resources/templates/forms/selectiveLicenceForm.html
@@ -1,0 +1,18 @@
+<!--/*@thymesVar id="title" type="java.lang.String"*/-->
+<!--/*@thymesVar id="formModel" type="java.lang.Object"*/-->
+<!--/*@thymesVar id="fieldSetHeading" type="java.lang.String"*/-->
+<!--/*@thymesVar id="label" type="java.lang.String"*/-->
+<!--/*@thymesVar id="backUrl" type="java.lang.String"*/-->
+<!DOCTYPE html>
+<html th:replace="~{fragments/forms/questionPage :: questionPage(#{${title}}, ${formModel}, ~{::#form-content}, ${backUrl})}">
+<form id="form-content" th:remove="tag">
+    <fieldset id="fieldset-content"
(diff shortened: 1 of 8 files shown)


src/main/resources/templates/forms/selectiveLicenceForm.html
```diff
@@ -0,0 +1,18 @@
...
+<html th:replace="~{fragments/forms/questionPage :: questionPage(#{${title}}, ${formModel}, ~{::#form-content}, ${backUrl})}">
+<form id="form-content" th:remove="tag">
+    <fieldset id="fieldset-content"
+              th:replace="~{fragments/forms/fieldSet :: fieldSet(~{::#fieldset-content/content()}, 'licenceNumber', #{${fieldSetHeading}}, null)}">
+        <input th:replace="~{fragments/forms/basicTextInput :: textInput(#{${label}}, 'licenceNumber', null)}">
+    </fieldset>
+    <details id="details-content" th:replace="~{fragments/details :: details(#{'forms.selectiveLicence.detail.summary'},~{::#details-content/content()})}">
+        <div class="govuk-details__text" th:text="#{'forms.selectiveLicence.detail.text'}">detailsText</div>
+    </details>
+    <button th:replace="~{fragments/buttons/primaryButton :: primaryButton(#{'forms.buttons.continue'})}"></button>
```
isobel-softwire, line 16: I think this is supposed to be save and continue

AEPR: So, actually [the designs are ambiguous ](https://www.figma.com/design/ka4URNl45NiQO0LoY8XNe4/MHCLG---Private-Rented-Sector-Database?node-id=9501-157335&node-type=frame&t=ynif7CWALXzvABPT-0) - when there are errors showing the pages say continue but when they aren't they say save and continue.

I'm going to raise this in the design channel, but I'll need to (potentially) update the "licensing type" page as well (as the same thing is true in the deisngs) so I'll apply the fix as a separate PR over both.

PR #100, opened 2024-12-11
Feat/prsd 580 hmo mandatory licence

AEPR:
As the 3 licence number pages are so similar I've commonised them to a single template (though they have different form models to allow for different validation messages). 

Page:
image

Page with errors:
image

Designs: [Figma](https://www.figma.com/design/ka4URNl45NiQO0LoY8XNe4/MHCLG---Private-Rented-Sector-Database?node-id=9548-163434&node-type=frame&t=lZZCYezKoFoFmFj7-0)


src/main/resources/templates/forms/licenceNumberForm.html
```diff
@@ -0,0 +1,31 @@
...
+        <div class="govuk-details__text">
+            <!--/* Iff there is no additional content we should th:remove the div tag to avoid nesting divs, and we don't need the govuk-body class */-->
+            <div th:text="#{${detailMainText}}" class="govuk-body" th:remove="${detailAdditionalContent} ? none : tag">detailsMainText</div>
+            <th:block th:if="${detailAdditionalContent}">
+                <ul class="govuk-list govuk-list--bullet" >
+                    <li th:text="#{${detailAdditionalContent.bulletOne}}">forms.licenceNumber.detail.bullet.one</li>
+                    <li th:text="#{${detailAdditionalContent.bulletTwo}}">forms.licenceNumber.detail.bullet.two</li>
+                </ul>
+                <div class="govuk-body" th:text="#{${detailAdditionalContent.text}}">detailsAdditionalText</div>
+            </th:block>
```
isobel-softwire, line 26: Do you know of any other pages that will make use of this form? If so, it would be good to make it more extensible. E.g. by extracting the additional content into its own fragment then passing the fragment's name via the model (similarly to conditional radio fragments).

AEPR:
I've actually added the 3rd Licence Number form to this PR even though it's on another ticket because the changes required are so minimal (and related!)

Page: 
image

Page with error:
image

src/main/resources/templates/forms/licenceNumberForm.html
AEPR: Currently I think that it's exactly the three licence number pages here - and the third one doesn't have any additional content. I'll have a look and see if "Single Text Input with Detail" is a common repeated pattern.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/propertyRegistrationJourneyPages/HmoMandatoryLicenceFormPagePropertyRegistration.kt
isobel-softwire, line 1: As the license pages use the same form, they should probably inherit from a base `LicenceFormPage` (this should be similar to `NameFormPage` for reference)

AEPR: Good shout! I've made that change now.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationJourneyTests.kt
```diff
@@ -402,4 +404,68 @@ class PropertyRegistrationJourneyTests : IntegrationTest() {
...
+            val hmoMandatoryLicencePage = navigator.goToPropertyRegistrationHmoMandatoryLicencePage()
+            hmoMandatoryLicencePage.form.submit()
+            assertThat(hmoMandatoryLicencePage.form.getErrorMessage()).containsText("Enter the HMO Mandatory licence number")
+        }
+
+        @Test
+        fun `Submitting with a very long licence number returns an error`(page: Page) {
+            val hmoMandatoryLicencePage = navigator.goToPropertyRegistrationHmoMandatoryLicencePage()
+            val aVeryLongString =
+                "This string is very long, so long that it is not feasible that it is a real licence number " +
```
Travis-Softwire, line 429: Enjoyed this :)

PR #101, opened 2024-12-11
PRSD-491: Check field widths in templates

AEPR:
Two main changes here:
* Make separate fragments for a fluid width and fixed width text input to avoid passing lots of nulls.
* Assert that a valid width is being passed as the width parameter to protect against future mistakes.

I dont want to mess up Jasmin's otherwise good PR with this set of changes while she's away but since it's related I'm setting this as a draft merging into that PR.

@Travis-Softwire - if you like this approach I can do the same for the numerical input.

### DIFF
--- a/src/main/resources/templates/fragments/forms/fixedWidthTextInput.html
+++ b/src/main/resources/templates/fragments/forms/fixedWidthTextInput.html
@@ -1,0 +1,5 @@
+<!--The fieldWidth input can be 2, 3, 4, 5, 10, 20, 30-->
+<th:block th:fragment="fixedWidthTextInput(label, fieldName, hint, fieldWidth)"
+          th:assert="${fieldWidth == 2 || fieldWidth == 3 || fieldWidth == 4 || fieldWidth ==  5 || fieldWidth ==  10 || fieldWidth ==  20 || fieldWidth ==  30}">
+    <th:block th:replace="~{fragments/forms/innerTextInput :: innerTextInput(${label}, ${fieldName}, ${hint}, ${fieldWidth})}"></th:block>
+</th:block>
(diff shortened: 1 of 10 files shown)


JasminConterioSW (approved):
I think this looks fine, either approach would work!
This one is maybe more readable as you don't have to work out what nulls are for if you aren't setting the width of your input.

JasminConterioSW:
Just updated the selectiveLicenceForm.html on my branch to make the tests pass, so you might need to change that one back again as well...

src/main/resources/templates/fragments/forms/fixedWidthTextInput.html
```diff
@@ -0,0 +1,5 @@
+<!--The fieldWidth input can be 2, 3, 4, 5, 10, 20, 30-->
+<th:block th:fragment="fixedWidthTextInput(label, fieldName, hint, fieldWidth)"
+          th:assert="${fieldWidth == 2 || fieldWidth == 3 || fieldWidth == 4 || fieldWidth ==  5 || fieldWidth ==  10 || fieldWidth ==  20 || fieldWidth ==  30}">
```
Travis-Softwire, line 3: I forgot to mention that I did manage to find Thymeleaf's `#array.contains()` function yesterday. Is it clever enough to do `#arrays.contains([2, 3, 4, 5, 10, 20, 30], fieldWidth)`?

AEPR: There definitely SHOULD be a way to get this working, but I haven't managed it. I can get it to parse 
```
th:assert="${#arrays.contains({2, 3, 4, 5, 10, 20, 30}, fieldWidth)}
```
but it always evaluates to false even if the fieldWidth is one of those values. I don't think it's worth me spending more time looking into it over the slightly long winded or statement.

Travis-Softwire: Ah, it might be one of cases where the variable needs to be pre-processed with the double underscores, because it's a nested expression, but agree it's not worth spending extra time on

PR #102, opened 2024-12-11
Feat/prsd 637 look for property registration in database

JasminConterioSW:
- Add database entities / repositories for:
    - Property
    - PropertyOwnership
    - Payment (this one is linked to PropertyOwnership)
- Add PropertyRegistrationService with a function to check if the address is registered, along with tests.
- Use the service to check the database in addressAlreadyRegistered in the PropertyRegistrationJourney.

FYI an address is considered registered if 
- The property with that address which is active and
- There is a current and active property ownership for the active property


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyOwnership.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    @OneToOne
+    @JoinColumn(name = "primary_landlord_id", nullable = false, foreignKey = ForeignKey(name = "FK_PROPERTY_OWNERSHIP_PRIMARY_LANDLORD"))
+    lateinit var primaryLandlord: Landlord
+        private set
+
+    @OneToOne(optional = false)
+    @JoinColumn(name = "property_id", nullable = false, foreignKey = ForeignKey(name = "FK_PROPERTY_OWNERSHIP_PROPERTY"))
+    lateinit var property: Property
+        private set
```
AEPR, line 55: Actually, having read the tests - I think this needs to be ManyToOne as well, because we can have several inactive ownerships for a property, but only one active one. I've had a look and I think there's something we can add to the migration to sort this out though. (I'll comment there)

src/main/resources/db/migrations/V1_8_0__property_ownership_tables.sql
```diff
@@ -0,0 +1,62 @@
...
+    payment_date_time TIMESTAMP WITHOUT TIME ZONE,
+    payment_amount    DOUBLE PRECISION                        NOT NULL,
+    CONSTRAINT pk_payment PRIMARY KEY (id)
+);
+
+ALTER TABLE property
+    ADD CONSTRAINT uc_property_address UNIQUE (address_id);
+
+ALTER TABLE property_ownership
+    ADD CONSTRAINT uc_propertyownership_property UNIQUE (property_id);
```
AEPR, line 44: As noted elsewhere, we need to allow multiple `property_ownership`s per `property`, but only want to allow one to be active. We can acheive this by adding a "partial index"([see docs](https://www.postgresql.org/docs/current/indexes-partial.html#INDEXES-PARTIAL-EX3)) like this:
```
CREATE UNIQUE INDEX one_active_ownership_per_property ON property_ownership(property_id)
    WHERE is_active;
```

I'm not sure what performance impacts this will have, but making the database enforce this rule seems likely worth it to me.

AEPR (changes requested):
Some thoughts, mostly, on relationship numbering and constraints. Overall looking good though.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Payment.kt
```diff
@@ -0,0 +1,27 @@
...
+) : AuditableEntity() {
+    @Temporal(TemporalType.TIMESTAMP)
+    lateinit var paymentDateTime: OffsetDateTime
+        private set
+
+    var paymentAmount: Double = 0.00
+
+    @OneToMany(mappedBy = "payment")
+    lateinit var propertyOwnerships: MutableSet<PropertyOwnership>
+        private set
```
isobel-softwire, line 26: Is there a way of enforcing a 'one to at least one' relationship here?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Property.kt
```diff
@@ -0,0 +1,38 @@
...
+) : ModifiableAuditableEntity() {
+    lateinit var status: Status
+        private set
+
+    var isActive: Boolean = false
+
+    lateinit var propertyBuildType: PropertyType
+        private set
+
+    var hasGasSupply: Boolean? = null
```
isobel-softwire, line 27: These should be set to non-nullable as well

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyOwnership.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    lateinit var occupancyType: OccupancyType
+        private set
+
+    lateinit var landlordType: LandlordType
+        private set
+
+    var currentNumHouseholds: Int = 0
+        private set
```
isobel-softwire, line 36: These (except `tenancyStartDate`) need to be non-nullable

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyOwnership.kt
```diff
@@ -0,0 +1,65 @@
...
+    lateinit var primaryLandlord: Landlord
+        private set
+
+    @OneToOne(optional = false)
+    @JoinColumn(name = "property_id", nullable = false, foreignKey = ForeignKey(name = "FK_PROPERTY_OWNERSHIP_PROPERTY"))
+    lateinit var property: Property
+        private set
+
+    @ManyToOne(optional = false)
+    @JoinColumn(name = "payment_id", nullable = false)
```
isobel-softwire, line 58: I think the foreign key name should be set here, to match our convention

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyRegistrationService.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import org.springframework.context.annotation.Scope
+import org.springframework.context.annotation.ScopedProxyMode
+import org.springframework.stereotype.Service
+import org.springframework.web.context.WebApplicationContext
+import uk.gov.communities.prsdb.webapp.database.repository.PropertyOwnershipRepository
+import uk.gov.communities.prsdb.webapp.database.repository.PropertyRepository
+
+@Service
+@Scope(value = WebApplicationContext.SCOPE_REQUEST, proxyMode = ScopedProxyMode.TARGET_CLASS)
```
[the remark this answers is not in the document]
isobel-softwire: I don't think scope is needed here at all, since this service isn't interacting with session data

src/main/resources/data-local.sql
isobel-softwire, line 1: `SELECT setval(pg_get_serial_sequence('table_name', 'id'), (SELECT MAX(id) FROM table_name));` needs to be called after each table's `INSERT` statement to resync the sequence. This has already been done for the tables that existed before this PR, so you only need to add it for property, payment, and property_ownership.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyRegistrationService.kt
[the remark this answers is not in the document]
JasminConterioSW: I managed to crib from the one service that was interacting with the session data didn't I...

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyRegistrationServiceTests.kt
```diff
@@ -0,0 +1,141 @@
...
+
+@ExtendWith(MockitoExtension::class)
+class PropertyRegistrationServiceTests {
+    @Mock
+    private lateinit var propertyRepository: PropertyRepository
+
+    @Mock
+    private lateinit var propertyOwnershipRepository: PropertyOwnershipRepository
+
+    @InjectMocks
```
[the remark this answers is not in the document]
JasminConterioSW: Yep, we're using it for some of the other service tests as well (like LocalAuthorityDataServiceTests)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
```diff
@@ -3,4 +3,8 @@ package uk.gov.communities.prsdb.webapp.database.repository
 import org.springframework.data.jpa.repository.JpaRepository
 import uk.gov.communities.prsdb.webapp.database.entity.PropertyOwnership
 
-interface PropertyOwnershipRepository : JpaRepository<PropertyOwnership?, Int?>
+interface PropertyOwnershipRepository : JpaRepository<PropertyOwnership?, Int?> {
+    // The underscore tells JPA to access fields relating to the referenced table
+    @Suppress("ktlint:standard:function-naming")
+    fun findByProperty_Id(id: Long): List<PropertyOwnership>
```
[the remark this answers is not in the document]
JasminConterioSW: So just to confirm, for multiple landlords / joint ownership, we are expecting a property to have  (up to) 1 active ownership, and that ownership can have multiple landlords (at the moment, it just has one primary landlord, but there might be an "additional_landlords" field later)?

I think that does sound right, in which case your suggestion makes sense!

src/main/resources/db/migrations/V1_8_0__property_ownership_tables.sql
JasminConterioSW: Yes, makes sense thank you!

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyOwnership.kt
```diff
@@ -0,0 +1,65 @@
...
+    @JoinColumn(name = "primary_landlord_id", nullable = false, foreignKey = ForeignKey(name = "FK_PROPERTY_OWNERSHIP_PRIMARY_LANDLORD"))
+    lateinit var primaryLandlord: Landlord
+        private set
+
+    @OneToOne(optional = false)
+    @JoinColumn(name = "property_id", nullable = false, foreignKey = ForeignKey(name = "FK_PROPERTY_OWNERSHIP_PROPERTY"))
+    lateinit var property: Property
+        private set
+
+    @ManyToOne(optional = false)
```
JasminConterioSW, line 57: Do they have to pay before property details are added to the database?
As there is a "draft" registration status I suspect not...

Assuming they don't have to pay before any details are saved, maybe we actually need to make the payment on PropertyOwnership optional (I don't know what we do if they haven't paid, presumably keep the RegistrationStatus as draft, maybe make the property / propertyownership inactive or something like that. Maybe not a question for right now)

Travis-Softwire: The 'draft' registration is a bit of a hangover from before we knew how we were going to implement the journeys. The original idea was that an agent would be able to create a draft property record that a landlord could then take ownership of and pay for. 

Now that we have the concept of saving journey data we will almost certainly use that as the mechanism for draft properties, to avoid lots of hard edge cases that would come up otherwise (e.g. multiple drafts of the same property in the db)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Payment.kt
JasminConterioSW: There should be, we'll remove the Payment table for now as we're not quite sure how we will want it to work!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
```diff
@@ -426,7 +426,7 @@ class PropertyRegistrationJourney(
                 return Pair(RegisterPropertyStepId.ManualAddress, null)
             } else {
                 val addressData = addressDataService.getAddressData(singleLineAddress)
-                if (addressData?.uprn != null && addressAlreadyRegistered(addressData.uprn, propertyRegistrationService)) {
+                if (addressData?.uprn != null && propertyRegistrationService.getIsAddressRegistered(addressData.uprn)) {
```
AEPR, line 429: This function gets set to a `nextAction`, which means that it gets called repeatedly when the user is filling in the form (specificaly once when they submit this page, and twice when they get each subsequent page). That means that we're making a lot of duplicated database calls when we don't expect anything to change and that will probably have a negative impact on performance. 

I think we should probably only check this twice - once when they first select the address and once when we're about to save to the database (N.B. the database is actually enforcing this rule so even if we mess up our data is safe!).

I can think of two ways that we could do that - either we should cache the results of this check in the session so we don't keep hitting the database, OR we write a custom endpoint for `POST select-address` that does this check in the controller before calling `updateJourneyDataAndGetViewNameOrRedirect`.

AEPR (commented):
Looking good! I've noticed a potential performance issue from repeated database calls. I'm not sure whether it's best to fix as part of this PR or merge and raise a new one to handle it? 

Happy to approve if we decide an approach to address this in another PR!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
JasminConterioSW: Ooh I hadn't thought about the nextAction getting called on all the subsequent pages...

I think we can just cache it in the session though, I can add the private function back in and get it to check the cache for a result, then the database if it can't find a result in the cache.

JasminConterioSW: Took a little while but I think it's there...

It's caching the result for each uprn that it looks for (only the one that is selected, but if you keep going back and forth and select different properties it will store all of their results)

src/main/resources/db/migrations/V1_8_0__property_ownership_tables.sql
```diff
@@ -0,0 +1,50 @@
...
+    last_modified_date     TIMESTAMP WITHOUT TIME ZONE,
+    created_date           TIMESTAMP WITHOUT TIME ZONE,
+    is_active              BOOLEAN                                 NOT NULL,
+    tenancy_start_date     TIMESTAMP WITHOUT TIME ZONE,
+    occupancy_type         SMALLINT                                NOT NULL,
+    landlord_type          SMALLINT                                NOT NULL,
+    current_num_households INTEGER                                 NOT NULL,
+    registration_number_id BIGINT                                  NOT NULL,
+    property_id            BIGINT                                  NOT NULL,
+    payment_id             BIGINT,
```
isobel-softwire, line 26: If we're leaving out the payment table, should this line be deleted?

JasminConterioSW: Good catch, yes it should

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/AddressRegisteredDataModel.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.communities.prsdb.webapp.models.dataModels
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class AddressRegisteredDataModel(
```
AEPR, line 6: I don't think you're actually using this class anywhere.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyRegistrationService.kt
```diff
@@ -0,0 +1,27 @@
...
+
+        val property = propertyRepository.findByAddress_Uprn(uprn)
+        if (property == null || !property.isActive || property.id == null) {
+            addressDataService.setCachedAddressRegisteredResult(uprn, false)
+            return false
+        }
+        val propertyOwnership = propertyOwnershipRepository.findByIsActiveTrueAndProperty_Id(property.id)
+        val databaseResult = propertyOwnership != null
+        addressDataService.setCachedAddressRegisteredResult(uprn, databaseResult)
+        return (databaseResult)
```
AEPR, line 25: I don't think `databaseResult` needs to be in brackets here.

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/AddressDataServiceTests.kt
```diff
@@ -83,4 +85,38 @@ class AddressDataServiceTests {
...
+    }
+
+    @Test
+    fun `getCachedAddressRegisteredResult returns the cached result if found`() {
+        val uprn = 1234.toLong()
+        whenever(mockHttpSession.getAttribute("addressRegisteredResults")).thenReturn(mutableMapOf(uprn.toString() to true))
+        assertTrue(addressDataService.getCachedAddressRegisteredResult(uprn) ?: false)
+
+        whenever(mockHttpSession.getAttribute("addressRegisteredResults")).thenReturn(mutableMapOf(uprn.toString() to false))
+        assertFalse(addressDataService.getCachedAddressRegisteredResult(uprn) ?: true)
```
AEPR, line 106: I'd split this up into 4 smaller tests here.

AEPR (approved):
3 really minor comments, but the actual caching looks sweet - I hadn't really thought about it but what you've done seems like a really nice system to make sure we catch someone repeatedly changing the address they're selecting!

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/AddressRegisteredDataModel.kt
JasminConterioSW: Argh, no I'm not... I was in one iteration then realised I didn't need it!

PR #103, opened 2024-12-12
PRSD-642: Landlord Registration Confirmation

isobel-softwire:
### Overarching Changes
- Adds methods to `JourneyDataService` that help with retrieving values from the session
- Moves identity verification from `LandlordRegistrationJourneyTests` to `Navigator` 
- Consolidates `LandlordRegistrationJourneyTests`

### Features
- Creates `/confirmation` endpoint and template for landlord registration journey
- Sets landlord registration journey `Declaration` step's `handleSubmitAndRedirect` to create a landlord, clear the session's journey data, and return the `/confirmation` endpoint

### Screenshots
image

### Tests
- Updates affected unit tests
- Creates tests for confirmation step of landlord registration journey


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
```diff
@@ -430,5 +443,101 @@ class LandlordRegistrationJourney(
...
+                    "address",
+                )!!
+            if (contactAddressChosen == MANUAL_ADDRESS_CHOSEN) {
+                return journeyDataService.getManualAddress(
+                    journeyData,
+                    LandlordRegistrationStepId.ManualContactAddress.urlPathSegment,
+                )!!
+            }
+            return addressDataService.getAddressData(contactAddressChosen)!!
+        }
```
AEPR, line 541: I got a bit lost reading through this bit - it works as it's supposed to AFAICT, but it's a bit difficult to follow. I think the main reason is that there's so many nested multiline blocks and inline conditionals. I think this would be better if this was split out to include some more variables whose names could contain some information to help guide me through what was going on. For example:
```
dateOfBirth =
                    localDateToDate(
                        journeyDataService.getFieldDateValue(
                            journeyData,
                            LandlordRegistrationStepId.VerifyIdentity.urlPathSegment,
                            "birthDate",
                        ) ?: journeyDataService.getDate(
                            journeyData,
                            LandlordRegistrationStepId.DateOfBirth.urlPathSegment,
                        )!!,
                    ),
``` 
is quite difficult to parse, but `dateOfBirth = verifiedDateOfBirth ?: submittedDateOfBirth` communicates more clearly what is going on and requires holding fewer things in your head at once (i.e less working memory).

Very happy to jump on a call if you want to talk it through.

src/main/resources/data-local.sql
```diff
@@ -49,22 +56,35 @@ VALUES ('invited.user@example.com', 1, gen_random_uuid()),
...
+        'mm.smith@importantco.com');
 
-INSERT INTO landlord (id, created_date, last_modified_date, registration_number_id, address_id, date_of_birth, is_active, phone_number, subject_identifier, name, email)
-VALUES (1,  '09/13/24', '09/13/24', 1, 1,'09/13/2000', true, 07111111111, 'urn:fdc:gov.uk:2022:KLMNO', 'Alexander Smith', 'alex.surname@example.com'),
-       (2,  '09/13/24', '09/13/24', 2, 2,'08/13/2001', true, 07111111111, 'urn:fdc:gov.uk:2022:UVWXY', 'Alexandra Davies', 'alexandra.q.davies@example.com'),
-       (3,  '09/13/24', '09/13/24', 3, 3,'07/13/1997', true, 07111111111, 'urn:fdc:gov.uk:2022:PQRST', 'Evan Alexandrescu', 'unrelatedemail@completelydifferentdomain.com'),
-       (4,  '09/13/24', '09/13/24', 4, 4,'06/13/1989', true, 07111111111, 'urn:fdc:gov.uk:2022:07lXHJeQwE0k5PZO7w_PQF425vT8T7e63MrvyPYNSoI', 'Tobias Evans', 'tobyevans@importantco.com'),
-       (5,  '09/13/24', '09/13/24', 5, 5,'05/13/1950', true, 07111111111, 'urn:fdc:gov.uk:2022:mwfvbb5GgiDh0acjz9EDDQ7zwskWZzUSnWfavL70f6s', 'Margaret Mary Smith', 'mm.smith@importantco.com');
\ No newline at end of file
+SELECT setval(pg_get_serial_sequence('landlord', 'id'), (SELECT MAX(id) FROM landlord));
```
AEPR, line 90: I don't think you've made any changes to these values here - just improved the formatting - but I know Jasmin is making some changes in paralell so this reformatting is going to cause some nasty merge conflicts for one of you!

src/main/resources/templates/registerAsALandlordConfirmation.html
```diff
@@ -0,0 +1,51 @@
...
+            <h2 class="govuk-heading-m" th:text="#{registerAsALandlord.confirmation.propertyRegistration.heading}">
+                registerAsALandlord.confirmation.propertyRegistration.heading</h2>
+            <p class="govuk-body" th:text="#{registerAsALandlord.confirmation.propertyRegistration.paragraph.one}">
+                registerAsALandlord.confirmation.propertyRegistration.paragraph.one
+            </p>
+            <p class="govuk-body" th:text="#{registerAsALandlord.confirmation.propertyRegistration.paragraph.two}">
+                registerAsALandlord.confirmation.propertyRegistration.paragraph.two
+            </p>
+        </section>
+        <!-- TODO: Replace with link to dashboard -->
```
AEPR, line 45: Is this TODO for this ticket or a future one? If it's a future one can we add the number to the comment and add it to that ticket? Otherwise TODOs have a habbit of slipping by and living on in the codebase!

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/DateTimeHelper.kt
```diff
@@ -3,13 +3,20 @@ package uk.gov.communities.prsdb.webapp.helpers
...
+    private val clock: Clock = Clock.System
 
     fun getNowAsLocalDate(): LocalDate {
         val instant = clock.now().toLocalDateTime(TimeZone.of("UTC"))
         return LocalDate(instant.year, instant.month.value, instant.dayOfMonth)
     }
+
+    companion object {
+        fun localDateToDate(localDate: LocalDate): Date = Date.from(localDate.atStartOfDayIn(TimeZone.of("UTC")).toJavaInstant())
+    }
```
AEPR, line 21: This class doesn't have any tests! If we are going to have helper classes then we should make sure they're thoroughly tested as we're saying that this code is complicated enough that we don't want to risk people writing it out when it's needed and making mistakes.

For instance - I suspect that `getNowAsLocalDate` may have some bugs during BST!  I appreciate you didn't write that function but could you make sure both of these are thoroughly tested.

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/DateTimeHelper.kt
```diff
@@ -3,13 +3,20 @@ package uk.gov.communities.prsdb.webapp.helpers
...
-    val clock: Clock = Clock.System
+    private val clock: Clock = Clock.System
 
     fun getNowAsLocalDate(): LocalDate {
         val instant = clock.now().toLocalDateTime(TimeZone.of("UTC"))
         return LocalDate(instant.year, instant.month.value, instant.dayOfMonth)
     }
+
+    companion object {
+        fun localDateToDate(localDate: LocalDate): Date = Date.from(localDate.atStartOfDayIn(TimeZone.of("UTC")).toJavaInstant())
```
AEPR, line 20: I'm also a little suspicious of this `localDateToDate` function - what is it doing and why? Local dates are a data structure that represents a day/month/year construct (ideal for date of birth etc.) so converting it into something that represents an instant in time feels wrong. I know that the types don't currently match up for the database but I think we should really be storing these in the database as a Day/Month/Year data type without timezone information - i.e. as a LocalDate.

AEPR (changes requested):

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/DateTimeHelper.kt
isobel-softwire: I agree about storing date of birth as a `LocalDate` - I'll look into it.

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/DateTimeHelper.kt
isobel-softwire: `localDateToDate` will most likely be deleted based on other comments. I'll write some tests for `getNowAsLocalDate` though (I agree about the BST issues)!

src/main/resources/templates/registerAsALandlordConfirmation.html
isobel-softwire: It's for a future one, but I think it's so in the future that it hasn't been written yet! I couldn't find anything in the backlog about creating a dashboard.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
isobel-softwire: That's fair enough - I'll do some refactoring

src/main/resources/templates/registerAsALandlordConfirmation.html
AEPR: I think it's worth raising on slack then and making sure there's a number to track it - even if there's an epic we can note this links to.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordRegistrationJourneyTests.kt
```diff
@@ -51,9 +57,20 @@ class LandlordRegistrationJourneyTests : IntegrationTest() {
...
+        val emailPage = assertPageIs(page, EmailFormPageLandlordRegistration::class)
+        emailPage.emailInput.fill("test@example.com")
+        emailPage.form.submit()
+
+        val phoneNumPage = assertPageIs(page, PhoneNumberFormPageLandlordRegistration::class)
+        phoneNumPage.phoneNumberInput.fill("07123456789")
+        phoneNumPage.form.submit()
+        assertPageIs(page, CountryOfResidenceFormPageLandlordRegistration::class)
+
         val countryOfResidencePage = navigator.goToLandlordRegistrationCountryOfResidencePage()
```
Kiran-R-K, line 74: I think this should be `assertIsPage` not `navigator.goTo....`, this should be changed on all these User can navigate tests (I think)

isobel-softwire: Good catch 😄

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/LandlordJourneyDataHelper.kt
```diff
@@ -129,7 +129,7 @@ class LandlordJourneyDataHelper {
         private fun getSelectedAddress(
             journeyDataService: JourneyDataService,
             journeyData: JourneyData,
-            isContactAddress: Boolean,
+            isContactAddress: Boolean = false,
```
AEPR, line 132: Nice - this will make it much easier to read when it's not relevant (which is obviously the main case)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
```diff
@@ -417,82 +416,103 @@ class LandlordRegistrationJourney(
...
+                    LandlordJourneyDataHelper.getInternationalAddress(
+                        journeyDataService,
+                        journeyData,
+                    ),
+                dateOfBirth = LandlordJourneyDataHelper.getDOB(journeyDataService, journeyData)!!,
+            )
+
+            journeyDataService.clearJourneyDataFromSession()
+
+            return "/$REGISTER_LANDLORD_JOURNEY_URL/$CONFIRMATION_PAGE_PATH_SEGMENT"
```
AEPR, line 515: I know it's been a while, but I absolutely love where this function has got to. IMO it feels so much easier to read what is going into the landlord and it makes it clear that what it does is (a) create a landlord (b) clear the journey data and (c) return a confirmation page URL.

src/main/resources/data-local.sql
```diff
@@ -64,17 +70,23 @@ VALUES (1,  '09/13/24', '09/13/24', 1, '1 Fictional Road'),
        (5,  '09/13/24', '09/13/24', 5, '5 Mythical Place'),
        (6, '12/10/2024', '12/10/2024', 1123456, '1, Example Road, EG');
 
+SELECT setval(pg_get_serial_sequence('address', 'id'), (SELECT MAX(id) FROM address));
+
 INSERT INTO landlord (id, created_date, last_modified_date, registration_number_id, address_id, date_of_birth, is_active, phone_number, subject_identifier, name, email)
 VALUES (1,  '09/13/24', '09/13/24', 1, 1,'09/13/2000', true, 07111111111, 'urn:fdc:gov.uk:2022:KLMNO', 'Alexander Smith', 'alex.surname@example.com'),
-       (2,  '09/13/24', '09/13/24', 2, 2,'08/13/2001', true, 07111111111, 'urn:fdc:gov.uk:2022:UVWXY', 'Alexandra Davies', 'alexandra.q.davies@example.com'),
```
JasminConterioSW, line 69: At the risk of derailing things a bit, we have both a landlord and a landlord_user table.

From the data model it looks like landlord_user is deprecated (so saving to landlord here is correct).

It's probably for another ticket, but I think we might need to get rid of the landlord_user table and use landlord instead (it looks like UserRolesService currently checks the LandlordUserRepository when it's deciding user roles / permissions so that will need updating, I've not looked to see if there is anywhere else that uses the wrong table)

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordRegistrationJourneyTests.kt
```diff
@@ -74,12 +93,94 @@ class LandlordRegistrationJourneyTests : IntegrationTest() {
...
+        emailPage.emailInput.fill("test@example.com")
+        emailPage.form.submit()
+
+        val phoneNumPage = assertPageIs(page, PhoneNumberFormPageLandlordRegistration::class)
+        val number = phoneNumberUtil.getExampleNumber("GB")
+        phoneNumPage.phoneNumberInput.fill("${number.countryCode}${number.nationalNumber}")
+        phoneNumPage.form.submit()
+        assertPageIs(page, CountryOfResidenceFormPageLandlordRegistration::class)
+
+        val countryOfResidencePage = assertPageIs(page, CountryOfResidenceFormPageLandlordRegistration::class)
```
AEPR, line 122: These `assertPageIs` are duplicated.

src/main/resources/data-local.sql
AEPR: Yes, this is a good catch (but I agree it's outside of the scope of this PR)

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordRegistrationJourneyTests.kt
```diff
@@ -53,8 +60,20 @@ class LandlordRegistrationJourneyTests : IntegrationTest() {
 
     // TODO PRSD-622: Add the steps before and after the address section of the journey
     @Test
-    fun `User can navigate the whole journey if pages are correctly filled in`(page: Page) {
-        val countryOfResidencePage = navigator.goToLandlordRegistrationCountryOfResidencePage()
+    fun `User can navigate the whole journey if pages are correctly filled in (verified, UK resident, selected address)`(page: Page) {
```
AEPR, line 63: So, you've added 4 tests that are all very similar and very long here. I thought the idea of the new style of integration tests was that we don't need to run through every combination of whole journey? I also think that having 4 very long tests like this makes it more likely for there to be typos or errors missed in them, as well as potentially making the tests slower to run. I'm not entirely sure whether this is an exception or whether we should be cutting them down to just 1 or 2 runs and testing the rest in smaller tests.

I think my preference would be to have just 1 run through plus smaller tests, and maybe a specialist integration test for the check your answers page which directly manipulates the session somehow? (since it genuinely cares about all the earlier stages and their combinations.)

Failing that, maybe we can cut it down to 2 so we only hit each possible choice once rather than duplicating with combinations (i.e. test 1 is verified, UK, selected; test 2 is unverified, internations, manual) .

AEPR (approved):
Looking really good. I've left a comment on the integration tests but what's there IS thorough AFAICT. I'm happy with everything functional and if Travis is happy with a bit of duplication in the integration tests then this looks ready to go! Alternatively I'm happy to have another chat about where to take them to reduce the number of big tests without reducing coverage.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
isobel-softwire: Yeah, it's definitely much nicer to look at 😄

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordRegistrationJourneyTests.kt
isobel-softwire: I guess 'actually fix rebasing errors' didn't actually fix the rebasing errors 😭

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordRegistrationJourneyTests.kt
isobel-softwire: I definitely think we should talk about what integration testing method we're going to go for - maybe in tomorrow's catch up so the whole team can offer their ideas?

AEPR: That sounds like a good way to handle it. You happy to bring it up?

Happy to just plough on ahead with merging this PR then!

PR #105, opened 2024-12-13
PRSD-495: Add property registration check answers template

AEPR:
Create the template for the register-property check answers page, with placeholder text for the additional landlords and interested parties tabs. The rest of this ticket will involve modifying the "propertyDetails" `List` returned from the `Journey` to have the correct list of details in all circumstances and ensuring the journey flows correctly. This ticket is explictly just for the UI.

image

image

image

image

### DIFF
--- a/src/main/resources/templates/forms/propertyRegistrationCheckAnswersForm.html
+++ b/src/main/resources/templates/forms/propertyRegistrationCheckAnswersForm.html
@@ -15,43 +15,44 @@
             <span th:text="${propertyName}"></span>
         </h1>
     </legend>
-    <div class="govuk-tabs" data-module="govuk-tabs">
-        <ul class="govuk-tabs__list">
-            <li class="govuk-tabs__list-item govuk-tabs__list-item--selected">
-                <a class="govuk-tabs__tab" href="#property-details" th:text="#{forms.checkPropertyAnswers.propertyDetails.heading}">
-                    forms.checkPropertyAnswers.propertyDetails.heading
-                </a>
-            </li>
-            <li class="govuk-tabs__list-item">
-                <a class="govuk-tabs__tab" href="#additional-landlords" th:text="#{forms.checkPropertyAnswers.additionalLandlords.heading}">
-                    forms.checkPropertyAnswers.additionalLandlords.heading
-                </a>
-            </li>
-            <li class="govuk-tabs__list-item">
-                <a class="govuk-tabs__tab" href="#interested-parties" th:text="#{forms.checkPropertyAnswers.interestedParties.heading}">
-                    forms.checkPropertyAnswers.interestedParties.heading
-                </a>
-            </li>
-        </ul>
-        <div class="govuk-tabs__panel" id="property-details">
-            <div id="my-hint" class="govuk-hint" th:text="#{forms.checkPropertyAnswers.propertyDetails.hint}">
-                forms.checkPropertyAnswers.propertyDetails.hint
+    <div th:replace="~{fragments/tabs/tabs :: tabs(~{:: #tabs-content})}" >
+        <th:block id="tabs-content">
+            <ul th:replace="~{fragments/tabs/tabsHeadingList :: tabsHeadingList(~{:: #tab-headings})}">
+                <th:block id="tab-headings">
+                    <li th:replace="~{fragments/tabs/tabsListItem :: tabsListItem('property-details', #{forms.checkPropertyAnswers.propertyDetails.heading})}">
+                    </li>
+                    <li th:replace="~{fragments/tabs/tabsListItem :: tabsListItem('additional-landlords', #{forms.checkPropertyAnswers.additionalLandlords.heading})}">
+                    </li>
+                    <li th:replace="~{fragments/tabs/tabsListItem :: tabsListItem('interested-parties', #{forms.checkPropertyAnswers.interestedParties.heading})}">
+                    </li>
+                </th:block>
+            </ul>
+            <div th:replace="~{fragments/tabs/tabsPanel :: tabsPanel('property-details', ~{:: #property-details-panel-content})}">
+                <th:block id="property-details-panel-content">
+                    <div class="govuk-hint" th:text="#{forms.checkPropertyAnswers.propertyDetails.hint}">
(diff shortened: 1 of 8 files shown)


src/main/resources/templates/forms/propertyRegistrationCheckAnswersForm.html
```diff
@@ -0,0 +1,72 @@
...
+<!--/*@thymesVar id="backUrl" type="java.lang.String"*/-->
+<!DOCTYPE html>
+<html>
+<th:block id="main-content"
+          th:replace="~{fragments/layout :: layout(#{${title}}, ~{::#main-content/content()},false)}">
+    <a href="#" th:replace="~{fragments/forms/backLink :: backLink(${backUrl})}">Back link</a>
+    <main class="govuk-main-wrapper">
+
+        <div id="page-content"
+             th:replace="~{fragments/layouts/twoThirdsLayout :: twoThirdsLayout(~{::#page-content/content()})}">
```
isobel-softwire, line 15: I feel like the `questionPage` fragment should be used in this template - this page seems to follow the same format

src/main/resources/templates/forms/propertyRegistrationCheckAnswersForm.html
```diff
@@ -0,0 +1,72 @@
...
+            <header>
+                <h1 class="govuk-heading-l">
+                    <span th:text="#{forms.checkPropertyAnswers.headingPrefix}"></span>
+                    <br>
+                    <span th:text="${propertyName}"></span>
+                </h1>
```
isobel-softwire, line 21: I think it might be clearer to parameterize `forms.checkPropertyAnswers.heading(Prefix)`. You could put the `<br>` in the message then use `th:utext` here. What do you think?

src/main/resources/templates/forms/propertyRegistrationCheckAnswersForm.html
```diff
@@ -0,0 +1,72 @@
...
+            </header>
+
+            <form th:action="@{''}" method="post" novalidate th:object="${formModel}">
+
+                <div class="govuk-tabs" data-module="govuk-tabs">
```
isobel-softwire, line 26: We're going to use tabs on a few different pages so I think it would be worth creating a `tab.html` fragment.

isobel-softwire (commented):
Looking good 😄 - left a few comments

src/main/resources/templates/forms/propertyRegistrationCheckAnswersForm.html
AEPR: Yep, absolutely, I just missed that here.

Travis-Softwire (approved):
Nothing to add in addition to Isobel's suggestions! But subject to that LGTM

src/main/resources/templates/forms/propertyRegistrationCheckAnswersForm.html
AEPR: I did think about making this a fragment but I think it's actually going to be more trouble than it's worth to turn this into a fragment. My thought was that most of the complexity here is actually in the `<ul>` at the top which is the list of available tabs (and then that they have to correspond to the matching divs below). I can't think of a simple way to turn this into a fragment.

For each tab (which is variable) I need to pass in:
* the display name for the tab
* the id for the tab
* the contents of the tab panel

If I was creating 1 tab I can see how I would do that, but passing it in a variable number of times seems really complicated.

Also, it's not that complicated to use the underlying HTML directly and is likely to be a major component of any page that features it. I also don't think we'll be using it so many times that it's worth spending ages working out how to extract a simple to use fragment which gives the flexibility we'd need.

src/main/resources/templates/forms/propertyRegistrationCheckAnswersForm.html
AEPR: I think we've generally tried to avoid using th:utext feature, and in this case I think seperating these here makes sense. That said, we can maybe come up with a better name for the heading prefix. Maybe it could be `forms.checkPropertyAnswers.heading.one` or `forms.checkPropertyAnswers.heading.firstLine`

AEPR:
I've pushed back on a couple of @isobel-softwire 's suggestions so I don't want to merge without one of you or @Travis-Softwire having another look and resolving those conversations (or continuing them)!

src/main/resources/templates/forms/propertyRegistrationCheckAnswersForm.html
Travis-Softwire: That's fine by me - agree that we should avoid using `th:utext` for anything substantive like anchors. I'm ok with it where it's purely for formatting and makes it easier to understand what the overall copy is saying, but it's a judgement call on a case by case basis - I'm happy to go with whichever you feel makes most sense in this case.

src/main/resources/templates/forms/propertyRegistrationCheckAnswersForm.html
Travis-Softwire: I think we will have quite a few pages with tabs on so I do think it's worth using fragments here rather than having the design system html repeated each time, in case e.g. it changes in a future revision of the design system.

I completely take your point about the difficulties making it a single fragment though - Thymeleaf seems to be missing something equivalent to react's ability to turn the `children` prop into a list which is what we'd want here.

I think it's probably still worth doing something similar to what we've done with forms though, so wrapping the individual elements (tab section(?), tab list, tab panel) in fragments that work together so that we've at least abstracted away the design system classes.

AEPR: OK - I've put the 4 fragments needed in their own directory since they're closely related to each other and nothing else. Probably worth someone check those changes before I merge. @Travis-Softwire @isobel-softwire

Travis-Softwire (approved):
Happy w/ that 👍

src/main/resources/templates/forms/propertyRegistrationCheckAnswersForm.html
isobel-softwire: LGTM!

PR #106, opened 2024-12-13
PRSD-537: Fix getNowAsLocalDate

isobel-softwire:
Sets timezone to London in and adds tests for `DateTimeHelper.getNowAsLocalDate`

### DIFF
--- a/src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/DateTimeHelperTests.kt
+++ b/src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/DateTimeHelperTests.kt
@@ -1,8 +1,10 @@
 package uk.gov.communities.prsdb.webapp.helpers
 
 import kotlinx.datetime.LocalDate
+import kotlinx.datetime.LocalDateTime
+import kotlinx.datetime.LocalTime
 import kotlinx.datetime.TimeZone
-import kotlinx.datetime.atStartOfDayIn
+import kotlinx.datetime.toInstant
 import kotlinx.datetime.toJavaInstant
 import org.junit.jupiter.params.ParameterizedTest
 import org.junit.jupiter.params.provider.CsvSource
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/DateTimeHelperTests.kt
```diff
@@ -0,0 +1,48 @@
...
+        second: Int,
+    ) {
+        val expectedDate = LocalDate(year, month, dayOfMonth)
+        val expectedDateTime = LocalDateTime(expectedDate, LocalTime(hour, minute, second))
+
+        val dateTimeHelper =
+            DateTimeHelper(
+                Clock.fixed(
+                    expectedDateTime.toInstant(TimeZone.of("Europe/London")).toJavaInstant(),
+                    ZoneId.of("UTC"),
```
AEPR, line 42: The functionality of this test is good, but manipulating times and dates is an area that is very easy to either get wrong or later misunderstand the intended bahaviour so I'd lean on the side of over-specifying intentions in the tests where we can. To that end we should also make sure this works regardless of which zone the computer is in by also parameterising the timezone and using more than one.

The best way to do that alongside the different times is to use a [MethodSource](https://www.baeldung.com/parameterized-tests-junit-5#bd-6-method) to enumerate the different options, that way we can have two lists (one of times and one of time zones) and nest for loops to give every combination.

Given that we're using a Method Source, we could also use [named arguments](https://www.baeldung.com/parameterized-tests-junit-5#bd-2-using-the-argumentsnamedname-arg-args-factory) to make sure that all our test cases are easily understandable by the poor soul trying to debug a weird time related bug in five years time (as noted in my previous comment).

AEPR (changes requested):
What a beautiful short PR! I think you're probably right that we should be always evaluating the time as though we're in the UK, but still one that we should check with Travis/MHCLG.

isobel-softwire:
> What a beautiful short PR! I think you're probably right that we should be always evaluating the time as though we're in the UK, but still one that we should check with Travis/MHCLG.
I've already checked this with Travis - sorry, should've mentioned it!

src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/DateTimeHelperTests.kt
isobel-softwire: Made these updates, lmk if they don't line up with what you meant 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/DateValidator.kt
```diff
@@ -66,7 +66,7 @@ class DateValidator {
         // This means spring will allow us to properly mock .getNowAsLocalDate() in our tests
```
AEPR, line 66: Just noticed this comment needs updating now the function name has changed. Maybe worth changing to not reference the function name - 
```
// Using a DateTimeHelper lets us more easily mock times in our tests
```
or something?

AEPR (approved):
This is excellent. I've noticed an outdated comment which should be modified, but then happy for this to go in without another review.

PR #108, opened 2024-12-13
Chore/refactor journey step listings

AEPR:
Suggestion for making the journey easier to read by extracting the steps into separate small functions. This is another review where you might want to look at it both with and without whitespace hidden.

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
@@ -47,336 +47,400 @@ class LandlordRegistrationJourney(
         journeyDataService = journeyDataService,
         steps =
             listOf(
-                Step(
-                    id = LandlordRegistrationStepId.VerifyIdentity,
-                    page = VerifyIdentityPage(),
-                    nextAction = { journeyData, _ ->
-                        if (doesJourneyDataContainVerifiedIdentity(journeyData)) {
-                            Pair(LandlordRegistrationStepId.ConfirmIdentity, null)
-                        } else {
-                            Pair(LandlordRegistrationStepId.Name, null)
-                        }
-                    },
-                    saveAfterSubmit = false,
-                ),
-                Step(
-                    id = LandlordRegistrationStepId.Name,
-                    page =
-                        Page(
-                            formModel = NameFormModel::class,
-                            templateName = "forms/nameForm",
-                            content =
-                                mapOf(
-                                    "title" to "registerAsALandlord.title",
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
```diff
@@ -47,336 +47,400 @@ class LandlordRegistrationJourney(
...
+                manualAddressStep(),
+                internationalAddressStep(),
+                lookupContactAddressStep(),
+                selectContactAddressStep(journeyDataService, addressLookupService, addressDataService),
+                manualContactAddressStep(),
+                checkAnswersStep(),
+                declarationStep(),
+            ),
+    ) {
+    companion object {
```
Travis-Softwire, line 68: Is there a reason that for `LaUserRegistrationJourney` the steps are just private functions whereas here they're on a companion object?

Travis-Softwire:
I'm fine with this - agree that being able to see at a glance what steps are in a journey could be useful. I think there are some trade-offs we'll need to keep an eye on, e.g. making assumptions about the flow of the steps based on their order in the list instead of scrolling down to their definition to find their `nextAction` arguments.

Can you coordinate with the rest of the team on getting this merged, as it will affect all of the other open tickets I think.

Travis-Softwire (approved):

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
AEPR: Not a particularly good one for the difference - I did the `LaUserRegistrationJourney` and having them as naked private functions in the file worked and there was no companion object to add them to. These functions rely on other methods that we already private companion object functions so they needed to be in the class scope to access them. I think it makes more sense for them to be on the companion, given that they're the steps for this specific journey!

Travis-Softwire: Oh, I had not clocked they were outside the class definition for `LaUserRegistrationJourney`! Yes, I think they should definitely be part of the class, and if they can be private static functions then let's just keep them all on the companion objects for consistency.

AEPR: Yes - because they're being called from the constructor the have to exist in a static context, which is also why we need to pass the services into them rather than being able to access them on the class!

Travis-Softwire: Ah yes ofc - we've run into that previously

PR #110, opened 2024-12-16
PRSD-579: Fix fixed input widths

AEPR:
Two minor changes:
* Use a fixed width input for licence number input to better match the designs
* Fix an issue with the numerical input that was causing them to ignore the input value

The width I've selected on the licensing page is a little wider than [the designs](https://www.figma.com/design/ka4URNl45NiQO0LoY8XNe4/MHCLG---Private-Rented-Sector-Database?node-id=8457-364445&t=ly8OKk6KyQV37biw-0) but the designs are between 10 and 20 width and I thought this looked better that the 10 width option (see bottom)

Twenty width:
image

Ten width:
image


JasminConterioSW (approved):
LGTM :)

PR #111, opened 2024-12-16
PRSD-642: Landlord Journey Data Helper

isobel-softwire:
- Adds more field type retrieval methods to `JourneyDataService`
- Creates `LandlordJourneyDataHelper` for retrieving specific fields from the journey data

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataService.kt
@@ -14,6 +14,7 @@ import uk.gov.communities.prsdb.webapp.forms.journeys.JourneyData
 import uk.gov.communities.prsdb.webapp.forms.journeys.PageData
 import uk.gov.communities.prsdb.webapp.forms.journeys.objectToStringKeyedMap
 import java.security.Principal
+import java.time.LocalDate
 
 @Service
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataService.kt
```diff
@@ -98,4 +99,34 @@ class JourneyDataService(
...
         return pageData?.get(fieldName)?.toString()
     }
+
+    fun getFieldIntegerValue(
+        journeyData: JourneyData,
+        urlPathSegment: String,
+        fieldName: String,
+        subPageNumber: Int? = null,
+    ): Int? {
+        val fieldAsString = getFieldStringValue(journeyData, urlPathSegment, fieldName, subPageNumber) ?: return null
```
JasminConterioSW, line 109: It feels odd to be turning into a string then turning it back into an Int, but actually I think you need to!

JasminConterioSW (approved):
LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataService.kt
isobel-softwire: The only other way I could think of is an unsafe cast so I think this is the way to go, weirdly

PR #112, opened 2024-12-17
Feat/prsd 582 property operating type

JasminConterioSW:
Adds the landlord type page, along with some tests.

image
image


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
```diff
@@ -382,6 +385,36 @@ class PropertyRegistrationJourney(
...
+                                        ),
+                                        RadiosButtonViewModel(
+                                            value = LandlordType.JOINT,
+                                            labelMsgKey = "forms.landlordType.radios.option.joint.label",
+                                            hintMsgKey = "forms.landlordType.radios.option.joint.hint",
+                                        ),
+                                    ),
+                            ),
+                    ),
                 nextAction = { _, _ -> Pair(RegisterPropertyStepId.LicensingType, null) },
```
JasminConterioSW, line 418: Page order:

I've added this one in after the Number of People step, or after the Occupancy step if the property is unoccupied as that's where it appears on the "Check Answers" page.

On "Check answers", this is the last step and licensing would go earlier, so we might need to move them around, but that can be checked as part of [PRSD-495](https://mhclgdigital.atlassian.net/browse/PRSD-495)

PR #114, opened 2024-12-18
Feat/prsd 495 data retrieval

JasminConterioSW:
Adds in some helpers to retrieve data from the JourneyData for Property Registration.

These can be used to retrieve data to show on the Check Answers page, and also to get the data to save to the database.

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/PropertyRegistrationJourneyDataHelper.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/PropertyRegistrationJourneyDataHelper.kt
@@ -1,0 +1,52 @@
+package uk.gov.communities.prsdb.webapp.helpers
+
+import uk.gov.communities.prsdb.webapp.constants.MANUAL_ADDRESS_CHOSEN
+import uk.gov.communities.prsdb.webapp.constants.enums.PropertyType
+import uk.gov.communities.prsdb.webapp.forms.journeys.JourneyData
+import uk.gov.communities.prsdb.webapp.forms.steps.RegisterPropertyStepId
+import uk.gov.communities.prsdb.webapp.models.dataModels.AddressDataModel
+import uk.gov.communities.prsdb.webapp.services.AddressDataService
+import uk.gov.communities.prsdb.webapp.services.JourneyDataService
+
+class PropertyRegistrationJourneyDataHelper {
+    companion object {
+        fun getAddress(
+            journeyDataService: JourneyDataService,
+            journeyData: JourneyData,
+            addressDataService: AddressDataService,
+        ): AddressDataModel? {
+            return if (getSelectedAddress(journeyDataService, journeyData) == MANUAL_ADDRESS_CHOSEN) {
+                addressDataService.getManualAddress(journeyDataService, journeyData, RegisterPropertyStepId.ManualAddress.urlPathSegment)
+            } else {
+                val selectedAddress = getSelectedAddress(journeyDataService, journeyData) ?: return null
+                addressDataService.getAddressData(selectedAddress)
+            }
+        }
+
+        fun getPropertyType(
+            journeyDataService: JourneyDataService,
+            journeyData: JourneyData,
+        ): String? = journeyDataService.getFieldStringValue(journeyData, RegisterPropertyStepId.PropertyType.urlPathSegment, "propertyType")
+
+        fun getOwnershipType(
+            journeyDataService: JourneyDataService,
+            journeyData: JourneyData,
+        ): String? =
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/PropertyRegistrationJourneyDataHelper.kt
```diff
@@ -0,0 +1,112 @@
...
+            journeyDataService: JourneyDataService,
+            journeyData: JourneyData,
+            addressDataService: AddressDataService,
+        ): AddressDataModel? {
+            return if (getSelectedAddress(journeyDataService, journeyData) == MANUAL_ADDRESS_CHOSEN) {
+                addressDataService.getManualAddress(journeyDataService, journeyData, RegisterPropertyStepId.ManualAddress.urlPathSegment)
+            } else {
+                val selectedAddress = getSelectedAddress(journeyDataService, journeyData) ?: return null
+                addressDataService.getAddressData(selectedAddress)
+            }
```
isobel-softwire, line 26: I think this would be even easier to read if there was a `isManualAddressChosen()` method, like in the `LandlordJourneyDataHelper`

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/PropertyRegistrationJourneyDataHelper.kt
```diff
@@ -0,0 +1,112 @@
...
+
+        fun getPropertyType(
+            journeyDataService: JourneyDataService,
+            journeyData: JourneyData,
+        ): PropertyType? {
+            val propertyTypeString =
+                journeyDataService.getFieldStringValue(journeyData, RegisterPropertyStepId.PropertyType.urlPathSegment, "propertyType")
+                    ?: return null
+            return PropertyType.valueOf(propertyTypeString)
+        }
```
isobel-softwire, line 37: Do you think it would be worth adding a `getFieldEnumValue` to `JourneyDataService`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/AddressDataService.kt
```diff
@@ -42,4 +43,53 @@ class AddressDataService(
...
+            )
+
+        return AddressDataModel.fromManualAddressData(
+            addressLineOne,
+            townOrCity,
+            postcode,
+            addressLineTwo,
+            county,
+        )
+    }
```
isobel-softwire, line 94: This is definitely more convenient than repeating the method in each journey helper, but I don't think this method belongs here as it doesn't interact with address data stored in session. Maybe instead there should be a base `JourneyDataHelper` for common methods, which the specific helpers can delegate to?

isobel-softwire (changes requested):
This is looking great so far! I left a few thoughts

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/AddressDataService.kt
JasminConterioSW: The other methods aren't talking journeyData but they are still interacting with address-related data in the session, so I think this is ok?

isobel-softwire: I meant more for separation of concerns - the `AddressDataService` should only include methods that deal with cached addresses, but `getManualAddress` requires the service to know about the structure of the journey data.

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/PropertyRegistrationJourneyDataHelper.kt
JasminConterioSW: It gets a little more complicated doing things with a generic enum type.

I think something like this will work though:

```
companion object {
        inline fun <reified E : Enum<E>> getFieldEnumValue(
            journeyData: JourneyData,
            urlPathSegment: String,
            fieldName: String,
            subPageNumber: Int? = null,
        ): E? {
            val fieldAsString = journeyDataService.getFieldStringValue(journeyData, urlPathSegment, fieldName, subPageNumber) ?: return null
            return enumValueOf<E>(fieldAsString)
        }
    }
```

isobel-softwire: LGTM 😄

JasminConterioSW:
@isobel-softwire I think those changes should all be in :)

isobel-softwire (approved):
LGTM 👍

PR #116, opened 2024-12-19
PRSD-641: Tolerate blank manual address fields

isobel-softwire:
Makes `AddressDataModel.manualAddressDataToSingleLineAddress` tolerate blank manual address fields

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/AddressDataModel.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/AddressDataModel.kt
@@ -37,7 +37,8 @@ data class AddressDataModel(
             postcode: String,
             addressLineTwo: String? = null,
             county: String? = null,
-        ) = listOfNotNull(addressLineOne, addressLineTwo, townOrCity, county, postcode)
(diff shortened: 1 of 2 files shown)


JasminConterioSW:
I've not seen this before as a way of writing lots of tests concisely, but actually it does make the test file shorter and maybe more readable.

My only worry is if one test starts failing, does it get more difficult to start debugging it? It's probably fine as you should still be able to add a breakpoint then make it run just the failing test. Normally I'd put a break point and hit the run button, but that will start running all of the tests!

isobel-softwire:
> I've not seen this before as a way of writing lots of tests concisely, but actually it does make the test file shorter and maybe more readable.
> 
> My only worry is if one test starts failing, does it get more difficult to start debugging it? It's probably fine as you should still be able to add a breakpoint then make it run just the failing test. Normally I'd put a break point and hit the run button, but that will start running all of the tests!

The tests are still run sequentially, so I guess you'd just have to step through the cases until you get to the failing one. A little bit inconvenient, but I think it's worth it for the readability.

JasminConterioSW:
Yeah, I can't decide whether I like it or not, but I that's just personal preference.
I do like the increased readability!

JasminConterioSW (approved):

PR #117, opened 2024-12-19
Feat/prsd 729 save property to database

JasminConterioSW:
This is not as complete as I would like it to be, but...

- I think the actual registration should be working.
- I have added tests for the repository specific services
- Some of the PropertyRegistrationJourneyTests are broken
    - The licensing ones - I thinks this is just because they are expecting to go to a placeholder page but now go to a check answers page
    - select address step, the one checking the alreadyRegistered address is broken, I changed how it checks whether it's already registered to make it ignore the cache when it is called during the actual registration, I think this is what's broken it!

Still to do:
- PropertyRegistrationServiceTests (these are in progress but not quite working yet, and what is there was written in a rush...)
- We'll need some journey tests - I've not written anything yet to test the actual handleSubmitAndRedirect function

More notes, potentially related to other tickets:
- The "check answers" step (front end) is just a skeleton, please replace with Alex's version from PRSD-495 when that is done

- We removed the MockUser from the landlord table as part of landlord registration (my bad, didn't twig it would be a problem in review), but actually I think we should add it back in again and say that to do landlord registration you need to be logged in as something like the La User or La Admin (so run the local-auth profile and log in as one of these - details on Keeper. For tests you will need to direct it towards a different data-local.sql - see LaUserRegistrationJourneyTests.kt
- We have a LandlordUser table which I think we don't actually want - it should be replaced by the Landlord table. LandlordUser gets check in the UserRolesService, I haven't looked to see if it is used elsewhere!


src/main/resources/db/migrations/V1_8_0__property_ownership_tables.sql
```diff
@@ -20,10 +19,13 @@ CREATE TABLE property_ownership
     tenancy_start_date     TIMESTAMP WITHOUT TIME ZONE,
     occupancy_type         SMALLINT                                NOT NULL,
     landlord_type          SMALLINT                                NOT NULL,
+    ownership_type         SMALLINT                                NOT NULL,
     current_num_households INTEGER                                 NOT NULL,
+    current_num_tenants    INTEGER                                 NOT NULL,
```
JasminConterioSW, line 24: It's a bit cheeky to add this to an existing migration and we can move it if we prefer, it was making me add a default value if I added a non null column to an existing table and I thought we might get away with changing it here instead...

isobel-softwire:
The comments in the PR's description will be actioned in #119

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Property.kt
```diff
@@ -15,7 +15,7 @@ import uk.gov.communities.prsdb.webapp.constants.enums.RegistrationStatus
 class Property(
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
-    val id: Long? = null,
+    val id: Long = 0,
```
Travis-Softwire, line 18: For generated IDs we've been initialising these to null - is there a rationale for treating this one differently?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyOwnership.kt
```diff
@@ -49,16 +57,76 @@ class PropertyOwnership(
...
+        nullable = false,
+        foreignKey = ForeignKey(name = "FK_PROPERTY_OWNERSHIP_PROPERTY"),
+    )
     lateinit var property: Property
         private set
 
-    constructor(id: Long, isActive: Boolean) : this(id) {
+    @OneToOne
+    @JoinColumn(name = "license_id", nullable = true, foreignKey = ForeignKey(name = "FK_PROPERTY_OWNERSHIP_LICENSE"))
+    var license: License? = null
```
Travis-Softwire, line 79: Seeing this has made me wonder - have we checked that a property can definitely only have 1 type of license? If not, then ask but leave this as-is for this PR and raise a separate ticket if need be.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyRegistrationService.kt
```diff
@@ -1,27 +1,81 @@
...
+
+        val license =
+            if (licenseType != LicensingType.NO_LICENSING) {
+                licenseService.createLicense(licenseType, licenceNumber)
+            } else {
+                null
+            }
+
+        val propertyOwnership =
+            propertyOwnershipService.createPropertyOwnership(
```
Travis-Softwire, line 69: After a user has registered a property, what happens if they try and register the same property in the same session? It doesn't look like we update or invalidate the cached result for `getIsAddressRegistered`?

src/main/resources/db/migrations/V1_8_0__property_ownership_tables.sql
Travis-Softwire: I'm fine with this, BUT there's a bug in our flyway launch config at the moment that means it isn't clearing the database properly in all cases so we might need to fix that in parallel to merging this to make sure the new version of the migration is applied for everyone - should when you come to merge and we can discuss the best way to resolve it.

Travis-Softwire (approved):
A few minor Qs and comments but overall LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Property.kt
isobel-softwire: I changed this one so we don't have to deal with null safety/not-null assertions when accessing the id

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyOwnership.kt
isobel-softwire: I'm not sure, I'll ask 👍

PR #119, opened 2024-12-19
PRSD-597: Added UI for confirmation page

JasminConterioSW:
- Registers a property to the database on submission of the check answers page
- Creates and tests the property registration confirmation page

Screenshot:
image


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
```diff
@@ -518,54 +469,136 @@ class PropertyRegistrationJourney(
...
+                            )!!,
+                        numberOfPeople =
+                            PropertyRegistrationJourneyDataHelper.getNumberOfTenants(
+                                journeyDataService,
+                                journeyData,
+                            )!!,
+                        baseUserId = SecurityContextHolder.getContext().authentication.name,
+                    )
+
+                journeyDataService.clearJourneyDataFromSession()
```
isobel-softwire, line 594: Should we be deleting the journey's form context from the database here as well?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
```diff
@@ -518,54 +469,136 @@ class PropertyRegistrationJourney(
...
+
+                // The propertyOwnershipId will be used on the confirmation page to retrieve the record from the database
+                session.setAttribute("propertyOwnershipId", propertyOwnershipId)
```
isobel-softwire, line 597: I think there should be a `PropertyDataService` (or something similar) for this

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
```diff
@@ -518,54 +469,136 @@ class PropertyRegistrationJourney(
...
+                        numberOfHouseholds =
+                            PropertyRegistrationJourneyDataHelper.getNumberOfHouseholds(
+                                journeyDataService,
+                                journeyData,
+                            )!!,
+                        numberOfPeople =
+                            PropertyRegistrationJourneyDataHelper.getNumberOfTenants(
+                                journeyDataService,
+                                journeyData,
+                            )!!,
```
isobel-softwire, line 590: An error is thrown if the property is unoccupied as these steps are skipped. These parameters should be nullable, or given a default value (0).

isobel-softwire (commented):
Still TODO:
- Action comments
- Get PRN from session in confirmation endpoint
- Determine landlord role in `UserRolesService` from `LandlordRepository` (instead of `LandlordUserRepository`)
- Fix failing `PropertyRegistrationJourneyTests`
- Add tests for the confirmation step of `PropertyRegistrationJourney`

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
```diff
@@ -57,4 +64,27 @@ class RegisterPropertyController(
...
+
+    @GetMapping("/$CONFIRMATION_PAGE_PATH_SEGMENT")
+    fun getConfirmation(
+        model: Model,
+        principal: Principal,
+    ): String {
+        val propertyOwnershipID = session.getAttribute(PROPERTY_OWNERSHIP_ID)?.toString()?.toLong()
+        val propertyOwnership =
+            propertyOwnershipID?.let { propertyOwnershipService.retrievePropertyOwnership(it) }
+                ?: throw PrsdbWebException("User ${principal.name} has not recently registered a property")
```
JasminConterioSW, line 76: I'm not sure this error message is right, am I right in thinking that this will be thrown if the PROPERTY_OWNERSHIP_ID is not found in the session (so it's null) or if there is no entry with propertyOwnershipID in the database? I think these should be two separate errors and I don't think we need to know the user's one-login id (which is what is stored in principal.name).

Maybe this bit can be replaced with something like:
```
        val propertyOwnershipID =
            session.getAttribute(PROPERTY_OWNERSHIP_ID)?.toString()?.toLong()
                ?: throw PrsdbWebException("$PROPERTY_OWNERSHIP_ID was not found in the session")
        val propertyOwnership =
            propertyOwnershipService.retrievePropertyOwnership(propertyOwnershipID)
                ?: throw PrsdbWebException("No property ownership with id $propertyOwnershipID was found in the database")
```
What do you think?

src/test/resources/data-mockuser-not-landlord.sql
```diff
@@ -0,0 +1,5 @@
+INSERT INTO one_login_user (id, name, email, created_date, last_modified_date)
+VALUES ('urn:fdc:gov.uk:2022:UVWXY', 'Mock User', 'test@example.com', '10/14/24', '10/14/24');
+
+INSERT INTO landlord_user (subject_identifier, phone_number, date_of_birth, created_date, last_modified_date)
```
JasminConterioSW, line 4: Do you need the landlord_user table for this?
I think this is just used to assign the landlord role and let the user do things like register properties, but they shouldn't need to have the landlord role when registering as a landlord.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
isobel-softwire: You're right - that's definitely clearer. I've implemented this change and added some tests for it 👍

src/test/resources/data-mockuser-not-landlord.sql
isobel-softwire: No I don't - removed it 👍

JasminConterioSW:
LGTM but I think someone else should review and approve as I wrote half of it!

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
```diff
@@ -57,4 +64,29 @@ class RegisterPropertyController(
...
         )
+
+    @GetMapping("/$CONFIRMATION_PAGE_PATH_SEGMENT")
+    fun getConfirmation(
+        model: Model,
+        principal: Principal,
+    ): String {
+        val propertyOwnershipID =
+            session.getAttribute(PROPERTY_OWNERSHIP_ID)?.toString()?.toLong()
+                ?: throw AccessDeniedException("$PROPERTY_OWNERSHIP_ID was not found in the session")
```
Travis-Softwire, line 75: Feels slightly odd to throw an access denied exception here? Most likely they have bookmarked the page or similar, so this is more of a bad request than an auth issue.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataService.kt
```diff
@@ -138,7 +147,9 @@ class JourneyDataService(
             fieldName: String,
             subPageNumber: Int? = null,
         ): E? {
-            val fieldAsString = journeyDataService.getFieldStringValue(journeyData, urlPathSegment, fieldName, subPageNumber) ?: return null
+            val fieldAsString =
```
Travis-Softwire, line 150: Sorry, this is the first time I'm looking at most of the helpers in the service, but is there a reason they're not ALL on the companion object? They mostly seem to be pure functions

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyRegistrationService.kt
```diff
@@ -23,18 +23,23 @@ class PropertyRegistrationService(
...
+    ): Boolean {
+        if (!ignoreCache) {
+            val cachedResult = addressDataService.getCachedAddressRegisteredResult(uprn)
+            if (cachedResult != null) return cachedResult
+        }
 
         val property = propertyRepository.findByAddress_Uprn(uprn)
         if (property == null || !property.isActive) {
-            addressDataService.setCachedAddressRegisteredResult(uprn, false)
+            if (!ignoreCache) addressDataService.setCachedAddressRegisteredResult(uprn, false)
```
Travis-Softwire, line 37: Why do we want to not set the cache here?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyRegistrationService.kt
```diff
@@ -23,18 +23,23 @@ class PropertyRegistrationService(
...
             return false
         }
 
         val databaseResult = propertyOwnershipRepository.existsByIsActiveTrueAndProperty_Id(property.id)
-        addressDataService.setCachedAddressRegisteredResult(uprn, databaseResult)
+        if (!ignoreCache) addressDataService.setCachedAddressRegisteredResult(uprn, databaseResult)
```
Travis-Softwire, line 42: And especially here - if we've established that the property has been registered in the meantime I think we'd want the cache to reflect that?

src/main/resources/templates/registerPropertyConfirmation.html
```diff
@@ -0,0 +1,32 @@
...
+                <tbody class="govuk-table__body">
+                    <tr class="govuk_table__row">
+                        <td class="govuk-table__cell govuk-!-width-one-half" th:text="${singleLineAddress}"></td>
+                        <td class="govuk-table__cell govuk-!-width-one-half" th:text="${prn}"></td>
+                    </tr>
+                </tbody>
+            </table>
+            <h1 class="govuk-heading-l" th:text="#{common.confirmationPage.whatHappensNext}">common.whatHappensNext</h1>
+            <p class="govuk-body" th:utext="#{registerProperty.confirmation.whatHappensNext.paragraph.one}">
+                rregisterProperty.confirmation.whatHappensNext.paragraph.one
```
Travis-Softwire, line 20: Typo - extra r on rregisterProperty

Travis-Softwire (commented):
I haven't reviewed the tests yet (will do in the morning) but I had a few Qs & comments in the meantime

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
```diff
@@ -480,60 +474,132 @@ class PropertyRegistrationJourney(
...
+                        ownershipType =
+                            PropertyRegistrationJourneyDataHelper.getOwnershipType(
+                                journeyDataService,
+                                journeyData,
+                            )!!,
+                        numberOfHouseholds =
+                            PropertyRegistrationJourneyDataHelper.getNumberOfHouseholds(
```
Travis-Softwire, line 583: Missed this before, but if a null in the journey data means zero here, should the helper do that conversion for us?

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyControllerTests.kt
```diff
@@ -43,11 +53,57 @@ class RegisterPropertyControllerTests(
...
+            .perform(
+                MockMvcRequestBuilders
+                    .get("/$REGISTER_PROPERTY_JOURNEY_URL/$CONFIRMATION_PAGE_PATH_SEGMENT")
+                    .sessionAttr(PROPERTY_OWNERSHIP_ID, propertyOwnershipID),
+            ).andExpect(MockMvcResultMatchers.status().isOk())
+    }
+
+    @Test
+    @WithMockUser(roles = ["LANDLORD"])
+    fun `getConfirmation returns 403 if there's no property ownership ID in session`() {
```
Travis-Softwire, line 83: As above, if we decide to change this to a 400 rather than a 403 then we'll need to change these tests as well

src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/PropertyRegistrationJourneyDataHelperTests.kt
```diff
@@ -184,7 +188,9 @@ class PropertyRegistrationJourneyDataHelperTests {
             ),
         ).thenReturn(true)
 
-        assertTrue(PropertyRegistrationJourneyDataHelper.getIsOccupied(mockJourneyDataService, mockJourneyData) ?: false)
+        assertTrue(
```
Travis-Softwire, line 191: Won't assertTrue still fail if the result of getIsOccupied is null?

Travis-Softwire (approved):
I've reviewed the tests as well now. Happy with this overall, and don't want to hold it up given it's been open a little while and is already quite a large PR. If any of the comments/ suggestions are non-trivial to address let me know as we might be better merging this and addressing them in a new PR

src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/PropertyRegistrationJourneyDataHelperTests.kt
isobel-softwire: I agree, but in this case `null` implies that the form hasn't been completed as this field is compulsory. I've changed this to a not-null assertion instead

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyRegistrationService.kt
isobel-softwire: That makes sense - changed this 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataService.kt
isobel-softwire: Will address this in a new PR 👍

PR #124, opened 2025-01-07
PRSD-748: Adds build and deploy pipeline

Travis-Softwire:


.github/workflows/build-and-test.yml
```diff
@@ -1,8 +1,5 @@
 name: Build and Test
 on:
-  push:
```
Travis-Softwire, line 3: Removed as we will run Build & Deploy on merge instead now

.github/workflows/build-and-test.yml
```diff
@@ -35,18 +32,4 @@ jobs:
           ONE_LOGIN_PUBLIC_KEY: |
             ${{ secrets.ONE_LOGIN_PUBLIC_KEY }}
           EMAILNOTIFICATIONS_APIKEY: ${{ secrets.NOTIFY_KEY }}
-        run: ./gradlew clean check --no-daemon
```
Travis-Softwire, line 38: As above

.run/flywayClean-local.run.xml
Travis-Softwire, line 1: As part of adding a new environment variable to ensure that runs correctly Intellij has decided to reformat it. THe only change is the addition of the FLYWAY_LOCATIONS variable

src/main/resources/application.yml
```diff
@@ -10,15 +10,24 @@ spring:
...
-        dialect: org.hibernate.dialect.PostgreSQLDialect
+        ddl-auto: validate
+      properties:
+        hibernate:
+          dialect: org.hibernate.dialect.PostgreSQLDialect
+      defer-datasource-initialization: true
 
   flyway:
-    enabled: false
+    enabled: true
```
Travis-Softwire, line 22: We will want to consider how/ when we want to run our migrations in higher envs, but for now I'm leaving these as running on application start-up

src/main/resources/application.yml
```diff
@@ -10,15 +10,24 @@ spring:
...
+    locations: classpath:db/migrations
+    validate-on-migrate: true
+
+  sql:
```
Travis-Softwire, line 26: Similarly, we'll want to decide how we want to add seed data to the database in each env, but for now we can use data-<env>.sql files and let hibernate populate the database.

.github/workflows/build-and-test.yml
```diff
@@ -35,18 +32,4 @@ jobs:
...
-
-      - name: Report pipeline status
```
benbelow, line 40: this feels like a sensible move, though I do think it's worth checking in with the team whether they find this useful, vs. relying on github notifications when their PR build breaks. If so we could consider setting up another slack channel for these to separate them from the main build status. 

(My gut says that this won't be useful, as Github notifications are more personalised and a slack channel will come with a lot of false negatives for other peoples' branches - but I don't want to make anyone's life harder because of an assumption!)

src/main/resources/application.yml
benbelow: do we have a card for that in the backlog? 

I'm a bit wary of this as something that could get accidentally forgotten, and then run into some issues at an unfortunate time once we're live.

benbelow (approved):

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,80 @@
+name: Build and Deploy
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build-and-deploy:
```
benbelow, line 8: we don't run the tests again here - is that definitely correct?

the tests should have passed on a PR, but I think there are cases where the final state on main can be different to that in the PR. (specifically, if something is approved, but there have been non-conflicting changes on main. GitHub will happily merge your change as there's no conflicts, and each of the changes individually has passed the tests, but the combination breaks one. Unless there's a setting in GitHub PRs that enforces it's been run on up to date main that I've missed, I think it might be safer to run them again here...

benbelow (changes requested):

.github/workflows/build-and-deploy.yml
benbelow: Oh ignore this, I'd missed that we do actually run the tests here, wrapped up in the gradle check. Ignore me!

benbelow (approved):

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,80 @@
...
+      - name: Login to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v2
+
+      - name: Build, tag, and push image to Amazon ECR
+        id: build-image
+        env:
+          ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
+          ECR_REPOSITORY: "integration-webapp"
+          IMAGE_TAG: ${{ github.sha }}
```
benbelow, line 59: Probably not one to block this PR, but could we consider using SemVer tags instead (or as well as?) the SHA? I feel like SHA only will be a bit less readable, and could be more likely to enable human error in the case of an emergency rollback.

benbelow (approved):

.github/workflows/build-and-deploy.yml
Travis-Softwire: Yes that's a good shout. Looks like EIP use a couple of 3rd party actions to manage this so I'll add those in a separate PR. Looks like it needs to be bootstrapped with an initial git tag but that's easy to do.

src/main/resources/application.yml
Travis-Softwire: We do now - PRSD-767

.github/workflows/build-and-test.yml
Travis-Softwire: At the moment we were only using this on merge/ push, not on PR. Because we're not running this pipeline on merge any more this block wasn't going to be doing anything. I've started a straw poll on slack to see what the team would prefer though.

benbelow (approved):

.github/workflows/build-and-test.yml
Travis-Softwire: For completeness - the consensus was to continue with only getting slack notifications on merge rather than on PR builds

PR #125, opened 2025-01-08
Feat/prsd 668 cancel la user invite

JasminConterioSW:
- Creates a "Cancel invitation" page. This includes the invitation id in the url
    - Only LA Admins in the correct local authority should be able to reach this page as the LA_ADMIN role is checked when manage users is loaded and only invitations for the user's local authority are shown
    - In case this changes, we check here that the logged in user is an LA admin for the current local authority (the id is specified in url), and that the inviting authority for the invitation being cancelled is the current local authority
    
- Creates a "Cancel invitation success" page
- Links to `cancel-invitation` from the "change" button on a pending user on the manage-users page
- Clicking "Cancel invite" removes the invitation from the database then redirects to the success page

image
image


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalAuthorityUsersController.kt
```diff
@@ -253,4 +253,51 @@ class ManageLocalAuthorityUsersController(
...
+    ): String {
+        val invitation = invitationService.getInvitationById(invitationId)
+
+        val (_, authority) =
+            localAuthorityDataService.getUserAndLocalAuthorityIfAuthorizedUser(
+                localAuthorityId,
+                principal.name,
+            )
+
+        if (authority.id != invitation.invitingAuthority.id) {
```
JasminConterioSW, line 272: This checks that the invitation belongs to the local authority with id localAuthorityId

src/main/resources/templates/manageLAUsers.html
```diff
@@ -28,10 +28,10 @@ <h1 class="govuk-heading-l" th:text="#{manageLAUsers.contentHeader(${localAuthor
                                 </span>
                     </td>
                     <td class="govuk-table__cell">
-                        <a th:href="@{${user.pending ? '#' : 'edit-user/' + user.id}}"
+                        <a th:href="@{${user.pending ? 'cancel-invitation/' + user.id : 'edit-user/' + user.id}}"
                            class="govuk-link"
                            th:text="#{manageLAUsers.table.change}"
-                           th:if="${currentUser.id} != ${user.id}"
+                           th:if="(${currentUser.id} != ${user.id}) or ${user.pending}"
```
JasminConterioSW, line 34: user.id is not unique in this case as it's using a view that combines two tables (local_authority_user and local_authority_invitation).
So if the logged in user had id 1 (for the local_authority_user table), the invitation with id 1 (for the local_authority_invitation table) wasn't showing the change link.

src/test/resources/data-la-users-and-invitations.sql
```diff
@@ -0,0 +1,14 @@
+INSERT INTO one_login_user (id, name, email, created_date, last_modified_date)
```
JasminConterioSW, line 1: Using a separate data file here as I need the invitation I'm looking for to stay in the same position in the table, so I didn't want to use data-local.sql as that is frequently updated

isobel-softwire (approved):
LGTM 😄 - left a few minor thoughts

PR #127, opened 2025-01-09
PRSD-748: Use semver for tagging images

Travis-Softwire:


.github/workflows/build-and-deploy.yml
```diff
@@ -41,6 +41,28 @@ jobs:
...
+        with:
+          repository: communitiesuk/prsdb-webapp
+          prefix: 'v0.'
+
+      - name: Bump version tag
+        uses: cabinetoffice/action-bump-semver@v1
+        id: bump-semver
+        with:
+          current_version: ${{ steps.get-latest-tag.outputs.tag }}
+          level: minor
```
benbelow, line 56: do we want a minor version bump every time we push to int? 

It feels like we'll get flooded with version numbers this way (which would probably be fine for true continuous deployment - but given we're not following that I wonder if it would be easier to keep track of minor versions being bumped when we deploy to test - that way we can use the minor versions to keep a suitable changelog for e.g. UAT)

.github/workflows/build-and-deploy.yml
```diff
@@ -41,6 +41,28 @@ jobs:
           EMAILNOTIFICATIONS_APIKEY: ${{ secrets.NOTIFY_KEY }}
         run: ./gradlew clean check --no-daemon
 
+      - name: Get latest version tag for minor version
+        id: get-latest-tag
+        uses: oprypin/find-latest-tag@v1
+        with:
+          repository: communitiesuk/prsdb-webapp
+          prefix: 'v0.'
```
benbelow, line 49: this hard-coded v0 will work fine until v1, but I'm worried it could get forgotten by then. Is it worth a ticket to track either making the automation more robust, or documenting the process for major version bumps?

.github/workflows/build-and-deploy.yml
```diff
@@ -41,6 +41,28 @@ jobs:
...
+
+      - name: Bump version tag
+        uses: cabinetoffice/action-bump-semver@v1
+        id: bump-semver
+        with:
+          current_version: ${{ steps.get-latest-tag.outputs.tag }}
+          level: minor
+
+      - name: Push version tag
```
benbelow, line 58: should we move this to be post-ECR upload, in case the build fails and we end up with a git tag that was never released?

benbelow (approved):

benbelow (changes requested):
accidentally selected approve last time with outstanding discussions...

.github/workflows/build-and-deploy.yml
Travis-Softwire: The other approach would be to think of it more as build promotion, so bumping the minor version _after_ each release to test, and ditto with the major version after each release to production. E.g. we might have 1.1.1, 1.1.2, and then release 1.1.2 to test, after which the next version on int would be 1.2.0. I think that would resolve the issue above, but would be a trickier to automate.

.github/workflows/build-and-deploy.yml
Travis-Softwire: That's a fair point (this was another copy/ paste from EIP). I think it depends heavily on the answer to our versioning above though - some options would be pretty hard to automate

benbelow:
Results of an offline conversation: 

* We'll use `maj.min.path-SHA` for ECR tags for the continuously built images from main (this removes churn in version numbers, and allows us to keep semVer semantic, while still having the value of versions in our ECR tags for e.g. hotfixes)
* We'll update the minor version when we "spin off" releases to test, to allow us to use them to communicate feature sets with other teams e.g. testers and business teams. 
  * We can automate that now if it's trivial, but otherwise we can push this down the line until we get to a point where we're deploying anything to test
* Updating the major version will be very infrequent, so it's fine for this to be manual.

.github/workflows/build-and-deploy.yml
```diff
@@ -78,13 +63,21 @@ jobs:
         env:
           ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
           ECR_REPOSITORY: "integration-webapp"
-          IMAGE_TAG: ${{ steps.bump-semver.outputs.new_version }}
+          IMAGE_TAG: "${{ steps.bump-semver.outputs.new_version }}-${{ github.sha }}"
```
benbelow, line 66: sadly we've removed `bump-semver` so I don't think this knows the version any more 😓

benbelow (changes requested):

.github/workflows/build-and-deploy.yml
Travis-Softwire: Git-fail on my part, only some of my changes ended up in the commit (I also accidentally left the git tagging in)

Travis-Softwire: Git-fail on my part, only some of my changes ended up in the commit (I also accidentally left the git tagging in)

benbelow (approved):

PR #128, opened 2025-01-09
Feat/prsd 669 send invite cancellation email

JasminConterioSW:
In this PR:

- Create an email template for LA invitation cancellation emails (this has also been added on notify)
- Send the email when the invitation is removed from the database
- (Also a couple of minor UI changes from QA of PRSD-668)

This is the email which appears in my inbox:
image

And the template on Notify
image


Travis-Softwire (approved):
This all LGTM :)

PR #134, opened 2025-01-13
PRSD-NONE: Gives non-nullable entity fields default values

isobel-softwire:


JasminConterioSW (approved):
This looks sensible to me, I'm assuming we haven't used those null values anywhere (checking that the field has been set or anything), if we had I think there'd be a failing test somewhere!

PR #137, opened 2025-01-14
PRSD-650: Landlord Search - No Results

isobel-softwire:
Overarching changes:
- Adds optional caption to `table.html`

Features:
- Updates search landlord page to show error when no results are found
- Creates placeholder property search endpoint

Tests:
- Updates integration tests for landlord search page

Screenshot:
image

### DIFF
--- a/src/main/resources/templates/searchLandlord.html
+++ b/src/main/resources/templates/searchLandlord.html
@@ -34,11 +34,13 @@
             </table>
             <div th:if="${searchResults.size() == 0}" class="govuk-!-text-align-centre" id="no-results">
                 <h1 class="govuk-heading-l" th:text="#{landlordSearch.error.heading}">
-                    landlordSearch.error.heading</h1>
+                    landlordSearch.error.heading
+                </h1>
                 <div class="govuk-hint">
                     <span th:remove="tag" th:text="#{landlordSearch.error.hint.beforeLink}">landlordSearch.error.hint.beforeLink</span>
                     <a class="govuk-link" th:href="@{${propertySearchURL}}"
-                       th:text="#{landlordSearch.error.hint.link}">landlordSearch.error.hint.link</a>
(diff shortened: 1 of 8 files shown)


src/main/resources/templates/searchLandlord.html
```diff
@@ -32,6 +32,15 @@ <h2 class="govuk-heading-m" th:text="#{landlordSearch.table.heading}">landlordSe
                 </tr>
                 </tbody>
             </table>
+            <div th:if="${searchResults.size() == 0}" class="govuk-!-text-align-centre" id="no-results">
+                <h1 class="govuk-heading-l" th:text="#{landlordSearch.error.heading}">
+                    landlordSearch.error.heading</h1>
```
AEPR, line 37: I think if we're breaking the html element over two lines, it's clearer to read if you put the closing tag on its own line as well. So this would be:
```
<h1 class="govuk-heading-l" th:text="#{landlordSearch.error.heading}">
    landlordSearch.error.heading
</h1>
```

src/main/resources/templates/searchLandlord.html
```diff
@@ -32,6 +32,15 @@ <h2 class="govuk-heading-m" th:text="#{landlordSearch.table.heading}">landlordSe
...
+                <div class="govuk-hint">
+                    <span th:remove="tag" th:text="#{landlordSearch.error.hint.beforeLink}">landlordSearch.error.hint.beforeLink</span>
+                    <a class="govuk-link" th:href="@{${propertySearchURL}}"
+                       th:text="#{landlordSearch.error.hint.link}">landlordSearch.error.hint.link</a>
```
AEPR, line 41: Same thing here - I think this should either all be on one line or we should put the opening tag, content and closing tag on each their own line.

AEPR (approved):
Nitpick on formatting of the HTML template. Happy for this to merge.

PR #140, opened 2025-01-15
Feat/prsd 652 landlord search pagination

JasminConterioSW:
- Update LandlordService `searchForLandlords` and landlordRepositority `searchMatching` to return a page of results
- Update SearchRegisterController `searchForLandlords` to
    - Take a `page `parameter from the url
    - Request page `page` of results 
    - Redirect if the user requests a non-existent page
- Update the pagination components to include url parameters (these are just passed in from the controller)
- Update manageLAUsers to work with the updated pagination components
- Fix a bug in manageLAUsers where the pagination component was showing even if there was only one page of results

image


src/main/resources/templates/fragments/pagination/next.html
```diff
@@ -1,5 +1,5 @@
-<div th:fragment="next(nextPage)" class="govuk-pagination__next">
-    <a class="govuk-link govuk-pagination__link" th:href="@{${#ctx.#requestdatavalues.url('')}(page=${nextPage})}" rel="next">
```
JasminConterioSW, line 2: I was trying to do something using thymeleaf's inbuilt function to add parameters, but it ignored url parameters that were already there unless it was explicitly told to add them (they needed to be named). I couldn't see a way of making it just add all the existing parameters and then just update the `page` parameter

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/ManageLaUsersPage.kt
```diff
@@ -12,7 +12,6 @@ class ManageLaUsersPage(
     page: Page,
 ) : BasePage(page, "/manage-users") {
     val table = Table(page)
-    val pagination = Pagination(page)
```
JasminConterioSW, line 15: This component shouldn't show up if there is only one page of results

src/main/resources/templates/fragments/pagination/next.html
isobel-softwire: I'm going to run into this issue when I implement the search filter, as there'll be a few more request parameters to keep track of. I think we should make a URL builder helper/service to make this more manageable. Happy to pair on this as it'll affect us both!

src/main/resources/templates/manageLAUsers.html
```diff
@@ -39,9 +39,9 @@ <h1 class="govuk-heading-l" th:text="#{manageLAUsers.contentHeader(${localAuthor
                 </tr>
                 </tbody>
             </table>
-            <nav th:if="${totalPages > 1}"
```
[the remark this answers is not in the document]
isobel-softwire: Yeah I've run into this before - the `th:replace` takes priority over the `th:if`. Good catch 😄

src/main/resources/templates/searchLandlord.html
```diff
@@ -10,6 +10,11 @@ <h1 class="govuk-heading-l" th:text="#{landlordSearch.heading}">landlordSearch.h
             #{landlordSearch.searchBar.label}, #{landlordSearch.searchBar.hint}, ${searchWrapperModel})}">
         </th:block>
         <th:block th:if="${searchResults != null}">
+            <hr class="govuk-section-break govuk-section-break--m">
+            <th:block th:if="${totalPages > 1}">
+                <nav th:replace="~{fragments/pagination/pagination :: pagination(${totalPages}, ${currentPage}, ${parameterStringNoPage})}"></nav>
+            </th:block>
+            <h2 class="govuk-heading-m" th:text="#{landlordSearch.table.heading}">landlordSearch.table.heading</h2>
```
isobel-softwire, line 17: The results table title has been turned into a table caption and a margin has been added to the bottom of the search bar so I don't think lines 13 and 17 are needed anymore. Could you please rebase on main and check? It would be good to get rid of my commits as well when you rebase!

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordServiceTests.kt
```diff
@@ -195,21 +197,60 @@ class LandlordServiceTests {
...
         val searchResults = landlordService.searchForLandlords(registrationNumber)
 
-        assertEquals(expectedSearchResults, searchResults)
+        assertEquals(expectedSearchResults, searchResults.content)
     }
 
     @Test
     fun `searchForLandlords returns a corresponding list of LandlordSearchResultDataModels`() {
         val searchQuery = "query"
+        val maxEntriesOnPage = 25
```
isobel-softwire, line 211: Couldn't the pagination constant be used here instead of setting a new variable?

isobel-softwire (commented):
Mostly LGTM. Left a few minor comments, and a thought about how to deal with updating request parameters - lmk what you think!

src/main/resources/templates/fragments/pagination/next.html
JasminConterioSW: A bit nicer than what I have currently, we can do something like in the controller (or in a helper):

```
val urlWithoutPage =
                UriComponentsBuilder
                    .newInstance()
                    .path("/search/landlord")
                    .queryParam("query", query)
                    .build()
                    .toUriString()

            model.addAttribute("unpagedUrl", urlWithoutPage)
```

Pass that through to the pagination link components, then do `th:href="${unpagedUrl} == null ? @{''(page=${nextPage})} : @{${unpagedUrl}(page=${nextPage})}"` (and similar for the previous and pageLink components).
If there are no query parameters (e.g. the Manage LA Users Page), the pagination component can be called with unpageUrl set to null and it still works.

JasminConterioSW: I'm pushing the above version if you want to take a look!

isobel-softwire: Yeah, I was thinking along these lines, but more general so we could use it in other situations if needed. 
Either a helper method like `updateQueryParam(currentURL,paramToUpdate,newValue)` or a service that retrieves the current URL from the request context so only needs the `paramToUpdate` and `newValue` arguments.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/SearchRegisterController.kt
```diff
@@ -19,10 +19,25 @@ class SearchRegisterController(
...
         model: Model,
         @RequestParam(required = false) query: String?,
+        @RequestParam(value = "page", required = false) page: Int = 1,
     ): String {
         if (!query.isNullOrBlank()) {
-            val results = landlordService.searchForLandlords(query)
-            model.addAttribute("searchResults", results)
+            if (page < 1) {
+                return "redirect:/search/landlord?query=$query"
+            }
```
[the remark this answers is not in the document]
isobel-softwire: ah ok - probably fine to just leave it then

src/main/resources/data-local.sql
```diff
@@ -62,7 +88,33 @@ VALUES (1, '09/13/24', 2001001001, 1),
...
+       (24,'01/15/25',1001001017,1),
+       (25,'01/15/25',1001001018,1),
+       (26,'01/15/25',1001001019,1),
+       (27,'01/15/25',1001001020,1),
+       (28,'01/15/25',1001001021,1),
+       (29,'01/15/25',1001001022,1),
+       (30,'01/15/25',1001001023,1),
+       (31,'01/15/25',1001001024,1),
+       (32,'01/15/25',1001001025,1),
+       (33,'01/15/25',100100106,1);
```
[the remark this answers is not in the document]
JasminConterioSW: I think we still need these as we need unique registration numbers for the landlord entries?

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordServiceTests.kt
JasminConterioSW: Mainly I was thinking if we use the pagination constant and someone changes it to 50 it'll make the test fail - it depends whether we care about that here!

If we just care about it giving a paged result, setting a value here (that will work with the data...sql file) is fine, if we want to make sure it works with whatever the pagination constant is set to we can use the pagination constant (but if someone changes it to 50 for some reason, the test data file will need more than 50 entries to make the test pass again).

I think checking it's giving some paged result is enough, so I gave a value that should always work with that test data file (I could have made my life easier and set it to 5, but figured we'd want the 25 entries for UAT anyway...)

src/main/resources/data-local.sql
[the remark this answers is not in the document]
isobel-softwire: You're right - I wasn't looking at the expanded view so I thought these were entries into the landlord user table. Sorry!

src/main/resources/templates/searchLandlord.html
JasminConterioSW: I tend to merge in main rather than rebase so it doesn't make me resolve all the conflicts for every commit, but it will squash all the commits when this merges to main anyway. It's messier because I branched from yours before it was on main!

src/main/resources/templates/fragments/pagination/next.html
isobel-softwire: Oh, I meant making a helper bean, so we'll be able to access its methods in templates (e.g. `th:href=${@helper.method()}`). We could use something like a `addQueryParam(url,name,value)` method in place of setting parameters in the controller or with thymeleaf's in-built method.

src/main/resources/templates/fragments/pagination/next.html
```diff
@@ -1,5 +1,5 @@
...
+<div th:fragment="next(nextPage,extraParametersString)" class="govuk-pagination__next">
+    <a class="govuk-link govuk-pagination__link" th:href="${unpagedUrl} == null ? @{''(page=${nextPage})} : @{${unpagedUrl}(page=${nextPage})}" rel="next">
```
isobel-softwire, line 2: should `extraParametersString` be `unpagedUrl` (or vice versa)?

src/main/resources/templates/searchLandlord.html
```diff
@@ -34,6 +37,9 @@ <h1 class="govuk-heading-l" th:text="#{landlordSearch.heading}">landlordSearch.h
                 </tr>
                 </tbody>
             </table>
+            <th:block th:if="${totalPages > 1}">
+                <nav th:replace="~{fragments/pagination/pagination :: pagination(${totalPages}, ${currentPage},${parameterStringNoPage})}"></nav>
```
isobel-softwire, line 41: Should `parameterStringNoPage` be `unpagedUrl`?

isobel-softwire (approved):
Two naming qs, but LGTM 👍

src/main/resources/templates/fragments/pagination/next.html
JasminConterioSW: We're going with leaving this for now, then adding helpers as needed in later tickets

PR #143, opened 2025-01-16
PRSD-649: Landlord Search Back Link

isobel-softwire:
- Bundles JS for MoJ design system
- Adds back link to landlord search page

### DIFF
--- a/src/main/resources/templates/searchLandlord.html
+++ b/src/main/resources/templates/searchLandlord.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
-<html th:replace="~{fragments/layout :: layout(#{landlordSearch.title}, ~{::html/content()}, false)}">
-<!--TODO PRSD-647: Set backlink to LA landing page-->
-<a th:replace="~{fragments/forms/backLink :: backLink('')}">backLink</a>
+<html>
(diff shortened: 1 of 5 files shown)


src/main/resources/templates/searchLandlord.html
```diff
@@ -1,5 +1,7 @@
 <!DOCTYPE html>
-<html th:replace="~{fragments/layout :: layout(#{landlordSearch.title}, ~{::main}, false)}">
+<html th:replace="~{fragments/layout :: layout(#{landlordSearch.title}, ~{::html/content()}, false)}">
```
JasminConterioSW, line 2: This is probably fine, but on other pages with a back link we've put the back link and main block into a <body> tag (see for example deleteLAUser.html) or a <th:block> (see for example questionPage.html), so we might want to do that for consistency? I can't see anywhere else where we do it in the html tag.

We are also normally passing a backUrl into the template from the controller rather than including it in the template, so it might be better to do that too :)

JasminConterioSW (changes requested):

src/main/resources/templates/searchLandlord.html
isobel-softwire: Made these changes 👍

JasminConterioSW (approved):

PR #145, opened 2025-01-16
Prsd 701 use page header actions

AEPR:
This also adds tests and moves the logic to a ViewModel class.

image


src/main/resources/templates/landlordDetailsView.html
```diff
@@ -3,7 +3,22 @@
...
+                <h1 class="govuk-heading-l" th:text="${name}">landlord.name</h1>
+            </div>
+
+            <div class="moj-page-header-actions__actions">
+                <div class="moj-button-group moj-button-group--inline">
+                    <a th:replace="~{fragments/buttons/primaryButtonLink :: primaryButtonLink('delete-record', #{landlordDetails.removeRecord})}"></a>
+                </div>
+            </div>
+
+        </div>
```
isobel-softwire, line 18: I think it would be worth making a page header actions fragment, similar to the tabs one

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
```diff
@@ -32,10 +30,14 @@ class LandlordDetailsController(
             landlordService.retrieveLandlordByBaseUserId(principal.name)
                 ?: throw PrsdbWebException("User ${principal.name} is not registered as a landlord")
 
+        val landlordViewModel = LandlordViewModel(landlord)
+
+        model.addAttribute("name", landlordViewModel.name)
+        model.addAttribute("personalDetails", landlordViewModel.personalDetails)
+        model.addAttribute("consentInformation", landlordViewModel.consentInformation)
```
isobel-softwire, line 37: You could just pass the whole `landlordViewModel` into the model for brevity

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/LandlordViewModel.kt
isobel-softwire, line 1: This is a great idea! Probably a bit nit-picky, but I feel like a data class with a static `fromLandlord()` method should be used here, to fit with our other models.

src/test/kotlin/uk/gov/communities/prsdb/webapp/mockObjects/MockLandlordData.kt
```diff
@@ -100,5 +103,37 @@ class MockLandlordData {
...
+            whenever(landlord.phoneNumber).thenReturn("07123456789")
+            whenever(landlord.address).thenReturn(createAddress())
+            whenever(landlord.registrationNumber).thenReturn(RegistrationNumber(RegistrationNumberType.LANDLORD, 0L))
+            whenever(landlord.internationalAddress).thenReturn(null)
+            whenever(landlord.dateOfBirth).thenReturn(LocalDate.now())
+            whenever(landlord.lastModifiedDate).thenReturn(OffsetDateTime.now())
+            whenever(landlord.createdDate).thenReturn(OffsetDateTime.now())
+
+            return landlord
+        }
```
isobel-softwire, line 137: To reduce duplication, `createLandlord()` could return a spied on `Landlord` instead of having separate mock landlord methods

src/test/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/LandlordViewModelTest.kt
```diff
@@ -0,0 +1,291 @@
...
+        val headerList = viewModel.personalDetails.map { it.fieldHeading }
+
+        assertEquals(headerList[0], "landlordDetails.personalDetails.registrationDate")
+        assertEquals(headerList[1], "landlordDetails.personalDetails.lrn")
+        assertEquals(headerList[2], "landlordDetails.personalDetails.name")
+        assertEquals(headerList[3], "landlordDetails.personalDetails.dateOfBirth")
+        assertEquals(headerList[4], "landlordDetails.personalDetails.emailAddress")
+        assertEquals(headerList[5], "landlordDetails.personalDetails.telephoneNumber")
+        assertEquals(headerList[6], "landlordDetails.personalDetails.ukResident")
+        assertEquals(headerList[7], "landlordDetails.personalDetails.contactAddress")
```
isobel-softwire, line 55: You could create an `expectedHeaderList` and make a (deep) equality assertion here

isobel-softwire (changes requested):
LGTM - left a few suggestions 👍

src/main/resources/templates/landlordDetailsView.html
AEPR: Yeah, 100% agree.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
AEPR: Sure, and that makes sense from a page structure page anyway.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/LandlordViewModel.kt
AEPR: I feel like the data I pass to front end IS a ViewModel, since its two properties are both lists of view models, and it directly backs a page being viewed. Happy to have a discussion about it, though.

isobel-softwire: I agree that it's a view model - I meant that it should be a Kotlin data class (i.e. declared as `data class LandlordViewModel()`). Sorry, should've made that clearer.

isobel-softwire (approved):
LGTM!

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
```diff
@@ -32,10 +30,13 @@ class LandlordDetailsController(
             landlordService.retrieveLandlordByBaseUserId(principal.name)
                 ?: throw PrsdbWebException("User ${principal.name} is not registered as a landlord")
 
+        val landlordViewModel = LandlordViewModel(landlord)
+
+        model.addAttribute("name", landlordViewModel.name)
```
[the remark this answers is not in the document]
AEPR: No it doesn't strictly need to be, but I think it should be because in the template it's used as the title of the page. When you're on the template there's the "landlord" attribute (for the personal details tab) and the "name" attribute (for the page heading). I think it's worth keeping them separate.

PR #146, opened 2025-01-16
bug/PRSD-NONE fising the isTenantedMessasgemKey property name on land…

Kiran-R-K:
…lordDetailsView

fixing the property name `tenantedMessageKey` -> `isTenantedMessageKey` on `lanldordDetailsView.html` so that the page loads and populates correctly


isobel-softwire (approved):
LGTM 👍

PR #149, opened 2025-01-21
Feat/prsd 654 listed property count

JasminConterioSW:
The outcome of this is that for each landlord, we count the number of active property ownerships where they are the primary landlord and display this in the search results table.

- Create a query and a view that returns the landlords with their listed property counts
    - The view returns LandlordWithListedPropertyCount entities, and the associated repository searches the view for records matching the landlord_id
- Include the listedPropertyCount in LandlordSearchResultDataModel
- Add listed properties to the displayed table
- Add new service and data model tests, update existing integration test to include the listed properties column in the table

image


src/main/resources/data-local.sql
```diff
@@ -174,13 +184,23 @@ VALUES (1, '09/13/24', '09/13/24', 1, 1, '09/13/2000', true, 07111111111, 'urn:f
...
+       (5, 1, true, 1, 10),
+       (6, 1, false, 1, 11);
 
 SELECT setval(pg_get_serial_sequence('property', 'id'), (SELECT MAX(id) FROM property));
 
 INSERT INTO property_ownership (id, is_active, occupancy_type, landlord_type, ownership_type, current_num_households,
                                 current_num_tenants,
                                 registration_number_id, primary_landlord_id, property_id)
-VALUES (1, true, 0, 0, 1, 1, 2, 6, 1, 1);
+VALUES (1, true, 0, 0, 1, 1, 2, 6, 1, 1),
```
JasminConterioSW, line 199: In this table, I am adding:

- 4 property ownership for Alexander Smith, but one is inactive (listedPropertyCount should be 3)
- 1 property ownership for Alexandra Davies, but it is inactive (listedPropertyCount should be 0)
- 1 property ownership for Tobias Evans (listedPropertyCount should be 1)

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/LandlordSearchResultDataModel.kt
```diff
@@ -23,5 +25,13 @@ data class LandlordSearchResultDataModel(
...
                 phoneNumber = landlord.phoneNumber,
             )
+
+        fun fromLandlordWithListedPropertyCount(
+            landlordWithListedPropertyCount: LandlordWithListedPropertyCount,
+        ): LandlordSearchResultDataModel {
+            val dataModel = fromLandlord(landlordWithListedPropertyCount.landlord)
+            dataModel.listedPropertyCount = landlordWithListedPropertyCount.listedPropertyCount
+            return dataModel
+        }
```
isobel-softwire, line 35: I don't think `fromLandlord` needs to be a separate method - are we using it anywhere?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordService.kt
```diff
@@ -67,7 +69,13 @@ class LandlordService(
...
                 retrieveLandlordByRegNum(registrationNumber)?.let { landlord ->
-                    return PageImpl(listOf(LandlordSearchResultDataModel.fromLandlord(landlord)))
+                    return PageImpl(
+                        listOf(
+                            LandlordSearchResultDataModel.fromLandlordWithListedPropertyCount(
+                                landlordWithListedPropertyCountRepository.findByLandlordId(landlord.id),
+                            ),
+                        ),
+                    )
                 }
```
isobel-softwire, line 79: You could use `landlordWithListedPropertyCountRepository.findByLandlord_RegistrationNumber_Number()` here and move `landlordRepository.searchMatching()` into `landlordWithListedPropertyCountRepository` instead of using multiple db queries. If you go with this then I think this method should be moved into `LandlordWithListedPropertyCountService`.

JasminConterioSW: I think we'll keep it as two separate queries because we don't need the database to find listedPropertyCounts for every landlord in the database, only those on our page of results (up to 25).

Also we might not be interested in listedPropertyCount in every search (it might not be a column on every table), there might be other columns to add... we could get a complicated query!

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/LandlordWithListedPropertyCountRepository.kt
```diff
@@ -0,0 +1,10 @@
+package uk.gov.communities.prsdb.webapp.database.repository
+
+import org.springframework.data.jpa.repository.JpaRepository
+import uk.gov.communities.prsdb.webapp.database.entity.LandlordWithListedPropertyCount
+
+interface LandlordWithListedPropertyCountRepository : JpaRepository<LandlordWithListedPropertyCount?, Long?> {
+    fun findByLandlordIdIn(primaryLandlordId: List<Long>): List<LandlordWithListedPropertyCount>
+
+    fun findByLandlordId(primaryLandlordId: Long): LandlordWithListedPropertyCount
```
JasminConterioSW, line 9: With the changes @isobel-softwire is making on [PRSD-651](https://github.com/communitiesuk/prsdb-webapp/tree/feat/PRSD-651-landlord-search-la-filter), we might not actually need this version that takes a single Id (as the landlord search functions are changing so they both return a page of Landlord)

JasminConterioSW: Let's pause this PR until PRSD-651 is in so we can make sure adding the listed property count is working with the updated version of landlord search.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/LandlordSearchResultDataModel.kt
JasminConterioSW: I don't think we are using it at the moment (apart from in a couple of tests, but they can be updated), but I'm still tempted to keep it because I'm wondering if we will always want the listed property count! 

I'll have another look when I come back to it though, if we're not using it after the 651 changes I might take it out, we can always add it back in if needed.

JasminConterioSW: I've had another look and I think you're right, removed this one now :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordService.kt
```diff
@@ -71,12 +74,20 @@ class LandlordService(
...
 
-        return (
+        val landlordPage = (
             if (lrn == null) {
                 landlordRepository.searchMatching(searchTerm, laUserId, useLAFilter, pageRequest)
             } else {
                 landlordRepository.searchMatchingLRN(lrn.number, laUserId, useLAFilter, pageRequest)
             }
-        ).map { LandlordSearchResultDataModel.fromLandlord(it) }
+        )
```
isobel-softwire, line 83: I don't think the brackets around the if-statement are needed anymore

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordServiceTests.kt
```diff
@@ -247,33 +266,80 @@ class LandlordServiceTests {
...
+                laUserBaseId,
+                currentPageNumber = currentPageNumber,
+                pageSize = pageSize,
+            )
+
+        // Actually we should probably check this with > 1 page of results and check that the total pages on the output is correct. Maybe as a separate test?
+
+        assertEquals(expectedSearchResults, searchResults.content)
+    }
+
```
isobel-softwire, line 306: Is this test not the same as the one on line 147?

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordServiceTests.kt
```diff
@@ -247,33 +266,80 @@ class LandlordServiceTests {
...
+        val searchResults =
+            landlordService.searchForLandlords(
+                searchQuery,
+                laUserBaseId,
+                currentPageNumber = currentPageNumber,
+                pageSize = pageSize,
+            )
+
+        // Actually we should probably check this with > 1 page of results and check that the total pages on the output is correct. Maybe as a separate test?
```
isobel-softwire, line 302: I think the next test does this - was this comment supposed to be removed?

isobel-softwire (approved):
Left a few minor questions - LGTM!

src/main/resources/db/migrations/V1_11_0__landlord_with_listed_property_view.sql
```diff
@@ -0,0 +1,5 @@
+CREATE VIEW landlord_with_listed_property_count AS
+SELECT l.id as landlord_id, COUNT (*) FILTER (WHERE po.is_active = true) AS listed_property_count
```
Travis-Softwire, line 2: I think I'm missing something, but why are we using FILTER on the aggregated rows instead of using a WHERE clause pre-aggregation?

Travis-Softwire (approved):
I had a Q about the query we're using to create the view, and I think Isobel has a few outstanding Qs, but LGTM once those are all resolved 👍

src/main/resources/db/migrations/V1_11_0__landlord_with_listed_property_view.sql
JasminConterioSW: I wanted the "where" to apply to the count, but not the whole table because I still wanted to get rows for landlords with no listed properties.

Travis-Softwire: Hmm, but that will only be true for landlords with inactive records, not those who've never added any properties? 

I _think_ if we want to include all Landlords we need to include the `is_active` check in the `ON` clause, but happy to chat about it

JasminConterioSW: Both of these give similar results:

```
SELECT l.id as landlord_id, COUNT (*) FILTER (WHERE po.is_active = true) AS listed_property_count
    FROM property_ownership po
    RIGHT JOIN landlord l ON po.primary_landlord_id = l.id
    GROUP BY l.id
```
or 
```
SELECT l.id as landlord_id, COUNT (po.primary_landlord_id) AS listed_property_count
    FROM property_ownership po
    RIGHT JOIN landlord l ON po.primary_landlord_id = l.id AND po.is_active = true
    GROUP BY l.id
```
We can go with the second version for efficiency, thanks!

PR #150, opened 2025-01-21
PRSD-560: loads PLACE_NAMES using a CSV mapper instead of manually

isobel-softwire:


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/DataLoaderService.kt
```diff
@@ -0,0 +1,57 @@
...
+import com.fasterxml.jackson.dataformat.csv.CsvSchema
+import com.fasterxml.jackson.module.kotlin.registerKotlinModule
+import org.springframework.stereotype.Service
+import org.springframework.util.ResourceUtils
+import uk.gov.communities.prsdb.webapp.models.dataModels.PlaceNameDataModel
+
+@Service
+class DataLoaderService {
+    companion object {
+        val PLACE_NAMES = loadPlaceNames().sortedBy { it.name }
```
AEPR, line 13: Actually, having looked here - I think that this shouldn't live on the `DataLoaderService`, and in fact, I'm not convinced that this should be a `@Service` at all - everything it does is on the companion object, so there's no reason for us to have a Bean instantiatable for the class. 

I think that having a data loader seems very sensible. I think that maybe we should think of this as configuration, but I'm not completely sure. 

I don't think I've been very clear hear and I'm happy to have a conversation about it but I think that the code looks good but we need to make sure it's in the right place. EITHER it should be in a Static context (i.e. on companion objects or outside classes), in which case the logic shouldn't be on a service and `PLACE_NAMES` should be on a class with a sensible name, OR we move it all to instance context (i.e. not on companion objects) and we use dependency injection everywhere.

AEPR (changes requested):
The actual switch to a library looks good, just a thought about where it all lives.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/DataLoaderService.kt
isobel-softwire: I've turned PLACE_NAMES into a constant and DataLoader into a helper 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/DataLoader.kt
```diff
@@ -0,0 +1,54 @@
+package uk.gov.communities.prsdb.webapp.helpers
+
+import com.fasterxml.jackson.dataformat.csv.CsvMapper
+import com.fasterxml.jackson.dataformat.csv.CsvSchema
+import com.fasterxml.jackson.module.kotlin.registerKotlinModule
+import org.springframework.util.ResourceUtils
+import uk.gov.communities.prsdb.webapp.models.dataModels.PlaceNameDataModel
+
+class DataLoader {
```
AEPR, line 9: This should probably have a test or two if it doesn't already. It might be that to make this testable we want to move the classpaths out of this function and make _them_ configuration, that way we can pass in/set a test file.

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/PlaceNamesConstants.kt
```diff
@@ -0,0 +1,5 @@
+package uk.gov.communities.prsdb.webapp.constants
+
+import uk.gov.communities.prsdb.webapp.helpers.DataLoader
+
+val PLACE_NAMES = DataLoader.loadPlaceNames()
```
AEPR, line 5: I think this being "PLACE_NAMES" is a a little confusing - I certainly misunderstood for some of our conversation yesterday - maybe "Country Names" or "All Countries"? Something that communicates that it's a list containing all of the names of countries in the world anyway.

AEPR (changes requested):

PR #151, opened 2025-01-23
PRSD-651: Landlord Search LA Filter Service

isobel-softwire:
- Creates `LandlordRepository` methods for searching by a LRN or query with an optional LA filter
- Updates `LandlordService.searchForLandlords()` to use these repository methods
- Updates the affected tests

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/LandlordRepository.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/LandlordRepository.kt
@@ -24,7 +24,8 @@ interface LandlordRepository : JpaRepository<Landlord?, Long?> {
     )
     fun searchMatching(
         @Param("searchQuery") searchQuery: String,
-        @Param("laUserBaseId") laUserBaseId: String? = null,
+        @Param("laUserBaseId") laUserBaseId: String,
+        @Param("useLAFilter") useLAFilter: Boolean = false,
         pageable: Pageable,
     ): Page<Landlord>
 
@@ -37,7 +38,8 @@ interface LandlordRepository : JpaRepository<Landlord?, Long?> {
     )
     fun searchMatchingLRN(
         @Param("searchLRN") searchLRN: Long,
-        @Param("laUserBaseId") laUserBaseId: String? = null,
+        @Param("laUserBaseId") laUserBaseId: String,
+        @Param("useLAFilter") useLAFilter: Boolean = false,
         pageable: Pageable,
     ): Page<Landlord>
 
@@ -55,7 +57,7 @@ interface LandlordRepository : JpaRepository<Landlord?, Long?> {
                           WHERE l.id = po.primary_landlord_id 
                           AND po.is_active 
                           AND ol.id = :laUserBaseId)
-                  OR :laUserBaseId IS NULL) 
+                  OR NOT :useLAFilter) 
             """
     }
 }
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/LandlordRepository.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/LandlordRepository.kt
@@ -18,11 +18,44 @@ interface LandlordRepository : JpaRepository<Landlord?, Long?> {
         "SELECT l.* " +
             "FROM landlord l  " +
             "WHERE (l.phone_number || ' ' || l.email || ' ' || l.name) %> :searchQuery " +
+            LA_FILTER +
             "ORDER BY (l.phone_number || ' ' || l.email || ' ' || l.name) <->> :searchQuery",
         nativeQuery = true,
     )
     fun searchMatching(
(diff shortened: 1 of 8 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordServiceTests.kt
```diff
@@ -53,48 +47,6 @@ class LandlordServiceTests {
     @InjectMocks
     private lateinit var landlordService: LandlordService
 
-    companion object {
-        private val landlord = createLandlord()
-
-        @JvmStatic
-        fun provideRegNumsAndExpectedSearchResults() =
```
JasminConterioSW, line 60: I think we will need to add in tests for searchMatchingLRN covering similar cases to these.

For searchMatching and searchMatchingLRN, I think we also need to test whether get unfiltered results if the filter is not applied (so if the user hasn't selected "landlords in my la", they should see landlords even if they don't have listed properties in the la), then test that the filtering works.

I was thinking the tests could go in the next PR but actually it's safer to put them in here so we can be sure that changes I make won't break the changes you make!

JasminConterioSW (changes requested):
It looks like the filter is being applied whenever the user searches for landlords, but I think it needs to be optional (so the la user can see all landlords, even with no listed properties in their la, unless they select the "landlords in my authority" filter).

There'll also be some more tests to add, it might actually be better to add them to this PR rather than the next one so it's clear whether my changes break your changes!

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/LandlordRepository.kt
```diff
@@ -18,11 +18,44 @@ interface LandlordRepository : JpaRepository<Landlord?, Long?> {
...
+        @Param("searchQuery") searchQuery: String,
+        @Param("laUserBaseId") laUserBaseId: String? = null,
         pageable: Pageable,
     ): Page<Landlord>
+
+    @Query(
+        "SELECT l.* " +
+            "FROM landlord l JOIN registration_number r on l.registration_number_id = r.id " +
+            "WHERE r.number = :searchLRN " +
+            LA_FILTER,
```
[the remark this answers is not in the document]
isobel-softwire: The filter is only used if `laUserBaseId` is not null, so it's optional.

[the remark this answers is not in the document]
JasminConterioSW: I would probably assume that laUserBaseId will be passed to the service / repository function as long as the user is logged in / the controller has found the id.

We might have the case though where the user is logged in, so the laUseBaseId is not null, but the user has chosen to view all landlords even if they don't have active ownerships in their LA.

Adding a parameter like "laFilterApplied : Boolean" and making LA_FILTER only apply when that is true would I think be clearer - it will make it more obvious what the controller needs to do in order to apply / not apply the filter.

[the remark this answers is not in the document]
isobel-softwire: Ah ok, that makes sense

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordServiceTests.kt
isobel-softwire: Testing searchMatching and searchMatchingLRN means setting up repository testing, which as far as I know will involve a lot of config as we haven't written tests for any other repository. I'm not sure that it's worth it considering we can indirectly test these methods through integration tests on the UI PR.

JasminConterioSW: You're right, we don't need to test the repository layer.

JasminConterioSW: We might be ok with what you've got actually, I was thinking we might need to test with and without the flag set, but the service doesn't actually do anything different if that flag is set or not (the repo does), so we don't need anything there.

And you have added one testing the LRN query.

You've taken out some tests for what happens when the LRN is the wrong type etc, is it worth adding a test back in to make sure it's returning an empty page if there are no matching results (covering the LRN being the wrong type etc)?

JasminConterioSW: Mainly I worry that we've taken out some tests and not replaced them, unless there is a particular reason you think we don't need them anymore?

isobel-softwire: I agree - I've put in tests for if the wrong registration number type or a query with no matches is given.

JasminConterioSW (approved):

JasminConterioSW:
LGTM, thanks :)

PR #152, opened 2025-01-23
Feat/prsd 658 local authority view landlord property details

Kiran-R-K:
Added the registered properties table to the local authority view of the landlord's details
Added methods to retrieve a landlord's registered properties from the database using the landlord's id

## Notes
Only unit tests have been added so far but there will be integration tests added in a separate PR


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
```diff
@@ -9,10 +9,17 @@ interface PropertyOwnershipRepository : JpaRepository<PropertyOwnership, Long> {
...
     fun findAllByPrimaryLandlord_BaseUser_IdAndIsActiveTrueAndProperty_Status(
         userId: String,
         status: RegistrationStatus,
     ): List<PropertyOwnership>
+
+    // This returns all active PropertyOwnerships for a given landlord from their landlord_Id with a particular RegistrationStatus
+    @Suppress("ktlint:standard:function-naming")
+    fun findAllByPrimaryLandlord_IdAndIsActiveTrueAndProperty_Status(
+        landlordId: Long,
+        status: RegistrationStatus,
```
isobel-softwire, line 23: You could give this a default value - it looks like you're always calling it with `REGISTERED`

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
```diff
@@ -66,6 +66,10 @@ class LandlordDetailsController(
         model.addAttribute("lastModifiedDate", lastModifiedDate)
         model.addAttribute("landlord", landlordViewModel)
 
+        val registeredPropertiesList = propertyOwnershipService.getRegisteredPropertiesForLandlord(id)
+
+        model.addAttribute("registeredPropertiesList", registeredPropertiesList)
+
```
isobel-softwire, line 72: `RegisteredPropertyDataModel` seems more like a view model to me, as it's only being used to set template content. What do you think?

isobel-softwire (approved):
LGTM - left some minor suggestions 😄

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
Kiran-R-K: yeah, agree - I'm gonna change it

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
Kiran-R-K: I did think about that, but I think it's fairly likely we'll be using this soon to call with `DRAFT` so didn't think it was worth it tbh

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
Kiran-R-K: I'm actually gonna raise a separate PR for this change

PR #155, opened 2025-01-24
chore/PRSD-NONE-add-css-stylesheet

Kiran-R-K:
Added:
- a custom css stylesheet
- the govuk red as a css global variable
- a css class for setting the border-color to this red
- updated `warningInsetText.html` to use the new css class


src/main/resources/css/custom.css
```diff
@@ -0,0 +1,7 @@
+:root {
+    --govuk-colour-red: #d4351c;
+
+    .border-color-red {
```
Travis-Softwire, line 4: Does the class need to be inside the root block?

Travis-Softwire (approved):
One Q about the CSS but otherwise LGTM

src/main/resources/css/custom.css
Kiran-R-K: no it doesn't, I'll pull it out

PR #157, opened 2025-01-27
Feat/prsd 598 property registration confirmation email

JasminConterioSW:
- When saving the new property to the database, return the registration number (instead of the property ownership id)
- Instead of saving the property ownership id to the session and using that to retrieve the ownership from the database on the confirmation page, do this with the registration number
    - (my thinking here was that it doesn't matter which we use on the confirmation page, but we do now need the prn when sending the email, so if we just return this when saving the property to the database we don't have to retrieve it in another database call)

- Add a new email template and send a confirmation email once the new property has been saved to the database
- Update email and notify tests to include the new email template
- Update the journey test to check an email is sent

image


Travis-Softwire (approved):
Looks great :)

PR #160, opened 2025-01-27
Feat/prsd 651 landlord search la filter

isobel-softwire:
### Overarching changes:
- Creates `filter.html` and `scrollablePane.html` fragments
- Creates `FilterViewModel` and (abstract) `SearchModel` to support search pages (this replaces `SearchWrapperModel` and makes passing `page` as a search endpoint request parameter redundant)
- Creates `URLQueryBuilder` service to update URL query parameters 
  - Refactors pagination to use this

### Features:
- Adds filter component to landlord search page
- Creates `LandlordSearchModel`
- Adds filter functionality to landlord search endpoint
- Wraps landlord search results table in scrollable pane

### Tests:
- `URLQueryBuilder`
- Integration tests for the filter component

### Screenshots:
Filter panel closed:
image

No filters applied:
image

LA filter applied:
image


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/URLQueryBuilder.kt
```diff
@@ -0,0 +1,91 @@
...
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Scope
+import org.springframework.context.annotation.ScopedProxyMode
+import org.springframework.stereotype.Service
+import org.springframework.web.context.WebApplicationContext
+import org.springframework.web.context.request.RequestContextListener
+
+@Service
+@Scope(value = WebApplicationContext.SCOPE_REQUEST, proxyMode = ScopedProxyMode.TARGET_CLASS)
+class URLQueryBuilder(
```
rowanhill, line 13: Instinctively, this feels like quite a low-level utility for us to need to build. Did you take a look at e.g. UriComponentsBuilder and decide it wasn't fit for our purposes?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/URLQueryBuilder.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    private fun getHiddenParam(param: String) = "_$param"
+
+    private fun resetParamMap() {
+        paramMap = httpServletRequest.parameterMap.toMutableMap()
+    }
+}
+
+@Bean
+fun requestContextListener(): RequestContextListener = RequestContextListener()
```
rowanhill, line 91: What's this for?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/URLQueryBuilder.kt
```diff
@@ -0,0 +1,91 @@
...
+                    .map { arrayEntry ->
+                        arrayEntry.value.joinToString("&") { singleValue ->
+                            arrayEntry.key + if (singleValue.isNotBlank()) "=$singleValue" else ""
+                        }
+                    }.joinToString(separator = "&", prefix = "?")
+            } else {
+                ""
+            }
+
+        resetParamMap()
```
rowanhill, line 31: I'm a bit nervous about having a per-request singleton where you can mess with the state and leave it in an unexpected state if you don't call build(). What do you think of having a factory that produces a builder, instead, and every time you create a builder it's starting fresh from the servlet request params?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/SearchRegisterTests.kt
```diff
@@ -111,4 +113,42 @@ class SearchRegisterTests : IntegrationTest() {
         assertContains(page.url(), "page=2")
         assertPageIs(page, SearchLandlordRegisterPage::class)
     }
+
+    @Test
+    fun `filter component functionality works`(page: Page) {
```
rowanhill, line 118: "X works" test names are pet peeve of mine: can we be more descriptive? E.g `filter panel can be used to refine search results` or something?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/URLQueryBuilder.kt
```diff
@@ -0,0 +1,91 @@
...
+            .map { arrayEntry ->
+                val hiddenIgnoreList = ignoreList.map { getHiddenParam(it) }
+                if (arrayEntry.key in ignoreList + hiddenIgnoreList) {
+                    emptyList()
+                } else {
+                    arrayEntry.value.map { singleValue -> arrayEntry.key to singleValue }
+                }
+            }.flatten()
+
+    private fun getHiddenParam(param: String) = "_$param"
```
rowanhill, line 83: I'm not sure I'm following what these "hidden" params are all about. Can you explain?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/URLQueryBuilder.kt
isobel-softwire: It gives the thread access to the current request, which allows for dependency injection of `HttpServletRequest`

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/URLQueryBuilder.kt
isobel-softwire: [Spring + Thymeleaf](https://www.thymeleaf.org/doc/tutorials/3.1/thymeleafspring.html#checkbox-fields) adds a hidden field to checkboxes so that browsers still submit a value if none are checked.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/URLQueryBuilder.kt
rowanhill: Okay - maybe worth a quick comment?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/URLQueryBuilder.kt
rowanhill: Ah, right, that makes sense, thanks.

src/main/resources/templates/fragments/pagination/pageLink.html
```diff
@@ -1,7 +1,7 @@
-<a th:fragment="pageLink(pageNumber, isCurrentPage, unpagedUrl)" class="govuk-link govuk-pagination__link"
-   th:href="${unpagedUrl} == null ? @{''(page=${pageNumber})} : @{${unpagedUrl}(page=${pageNumber})}"
+<a th:fragment="pageLink(pageNumber, isCurrentPage)" class="govuk-link govuk-pagination__link"
+   th:href="${@URLQueryBuilder.updateParam('page',pageNumber).build()}"
```
Travis-Softwire, line 2: In this case I could probably be persuaded that it makes sense to call some sort of utility function here to keep this flexible rather than repeating pagination link computation code in every controller that uses it

src/main/resources/templates/searchLandlord.html
```diff
@@ -3,62 +3,84 @@
...
+                     th:replace="~{fragments/layouts/scrollablePane :: scrollablePane(~{::#scrollable-content/content()})}">
+
+                    <table th:replace="~{fragments/tables/table :: table(
+                            ${ {'landlordSearch.table.heading.landlord','landlordSearch.table.heading.address','landlordSearch.table.heading.info', 'landlordSearch.table.heading.listedProperties'} },
+                            ~{::tbody},
+                            #{landlordSearch.table.heading}
+                        )}">
+                        <tbody class="govuk-table__body">
+                        <tr class="govuk_table__row" th:each="landlord: ${searchResults}">
+                            <td class="govuk-table__cell govuk-!-width-one-quarter">
```
Travis-Softwire, line 36: I think these cells being fixed @ 1/4 width guarantees that we get horizontal scrolling regardless of the filter panel width? We might need to think about whether we can do anything about that, e.g. putting the filter panel in a fixed width and adjusting the column widths when it's open?

Travis-Softwire (changes requested):
Sorry, I've had to come back to this a few times and I think my comments have overlapped with Rowan's but hopefully we're not contradicting each other! Happy to chat about any of these.

(This was probably also a good candidate to be broken down into multiple PRs as it ended up being quite complex, but appreciate that's easier to spot with hindsight!)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/SearchRegisterController.kt
```diff
@@ -5,61 +5,62 @@ import org.springframework.stereotype.Controller
...
         model: Model,
-        @RequestParam(required = false) query: String?,
-        @RequestParam(value = "page", required = false) page: Int = 1,
+        landlordSearchModel: LandlordSearchModel,
         principal: Principal,
     ): String {
-        if (!query.isNullOrBlank()) {
-            if (page < 1) {
-                return "redirect:/search/landlord?query=$query"
+        if (landlordSearchModel.query?.isBlank() == true) {
```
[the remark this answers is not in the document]
isobel-softwire: Yes as `landlordSearchModel.query?.isBlank()` can be null

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/searchModels/LandlordSearchModel.kt
```diff
@@ -0,0 +1,28 @@
...
+
+class LandlordSearchModel : SearchModel() {
+    var useLAFilter: Boolean = false
+
+    override val filters =
+        listOf(
+            FilterViewModel(
+                headingMsgKey = "landlordSearch.filter.la.heading",
+                field = "useLAFilter",
+                options = listOf(CheckboxViewModel(value = true, labelMsgKey = "landlordSearch.filter.la.label")),
```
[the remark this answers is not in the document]
isobel-softwire: The `value` property is constant, its purpose is to tell us to set `useLAFilter` to `true` if it's checked. We could make set `value` to `true/false` based on its previous value, but I don't think that would work for checkboxes with multiple options.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/URLQueryBuilder.kt
isobel-softwire: I'm not sure if the updated version of `URLQueryBuilder` will require this anymore, but I'll leave a comment if it does 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/URLQueryBuilder.kt
isobel-softwire: Sounds good to me 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/SearchRegisterController.kt
```diff
@@ -5,61 +5,62 @@ import org.springframework.stereotype.Controller
...
+            return "redirect:landlord"
+        }
+
+        if (landlordSearchModel.query != null) {
+            if (landlordSearchModel.page < 1) {
```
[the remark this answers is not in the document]
isobel-softwire: It is, I'll make this change and raise a small PR for doing the same wherever else the pagination component is used (I think this is just on the manage LA user's page)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/SearchRegisterController.kt
```diff
@@ -5,61 +5,62 @@ import org.springframework.stereotype.Controller
...
+                landlordService.searchForLandlords(
+                    landlordSearchModel.query!!,
+                    principal.name,
+                    landlordSearchModel.useLAFilter,
+                    currentPageIndex,
+                )
 
-            if (pagedLandlordList.totalPages in 1..<page) {
-                return "redirect:/search/landlord?query=$query"
+            if (pagedLandlordList.totalPages < currentPageIndex) {
```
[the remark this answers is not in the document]
isobel-softwire: `pagedLandlordList.totalPages` would equal `currentPageIndex` so this doesn't work... I'll change it to be  `pagedLandlordList.totalPages < page` instead

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/searchModels/LandlordSearchModel.kt
```diff
@@ -0,0 +1,28 @@
+package uk.gov.communities.prsdb.webapp.models.searchModels
+
+import uk.gov.communities.prsdb.webapp.models.viewModels.CheckboxViewModel
+import uk.gov.communities.prsdb.webapp.models.viewModels.FilterViewModel
+
+class LandlordSearchModel : SearchModel() {
```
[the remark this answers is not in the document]
isobel-softwire: At the moment it represents both. I'll separate them out into a `FilterViewModel` and a model for the query params (should this be a data model?)

src/main/resources/templates/fragments/filter.html
```diff
@@ -0,0 +1,103 @@
...
+                        <div class="moj-filter__heading-title">
+                            <h2 class="govuk-heading-m" th:text="#{filter.selected.heading}">
+                                filter.selected.heading
+                            </h2>
+                        </div>
+                        <div class="moj-filter__heading-action">
+                            <p>
+                                <a class="govuk-link govuk-link--no-visited-state"
+                                   th:text="#{filter.selected.clearLink}"
+                                   th:href="${@URLQueryBuilder.removeParams(searchModel.filterFields).build()}">
```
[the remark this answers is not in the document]
isobel-softwire: I did it this way so we don't have to generate the links in every controller that returns a view with a filter component, but I agree that this doesn't really fit with what we've done so far. To get around this I could make generating the links part of the view model?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/searchModels/LandlordSearchModel.kt
```diff
@@ -0,0 +1,28 @@
...
+    var useLAFilter: Boolean = false
+
+    override val filters =
+        listOf(
+            FilterViewModel(
```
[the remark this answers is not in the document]
isobel-softwire: Yes - I guess it isn't super clear. I'll do some renaming...

src/main/resources/templates/fragments/pagination/pageLink.html
isobel-softwire: These links could be generated in a view model as well, as we're already doing some pagination computation on controllers that use it

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/URLQueryBuilder.kt
rowanhill: I would hope you can use UriComponentsBuilder in all the methods, not just build(), so this class ends up a fairly thin wrapper to make the methods chainable. I.e. your UriComponentsBuilder would hold your state, rather than paramMap.

src/main/resources/templates/fragments/filter.html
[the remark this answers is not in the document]
Travis-Softwire: If we can compute it in the view model sensibly then I think that could work 👍

src/main/resources/templates/searchLandlord.html
Travis-Softwire: As discussed on this one - for this PR let's leave it as per the designs and then explain the issue to others and see what they say

isobel-softwire:
Closed as Travis requested changes but is away

PR #161, opened 2025-01-28
Feat/prsd 600 property registration task list

AEPR:
Pull request for the UI components and page template for the task list component. There's a difference to the designs, specifically in the numbered list of headings - this is an open question on Slack as what I've implemented is what's supported by [the gov design system](https://design-system.service.gov.uk/patterns/complete-multiple-tasks/). The content on the page is currently a placeholder to show each possible state. 

Still to go:
* Reading the actual statuses of the tasks out of session data
* Restoring session data from database data
* Routing logic to ensure users land on this page whenever they return to this journey

image


src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/TaskListItem.kt
isobel-softwire, line 1: Minor, but I think this should be called `TaskListItemViewModel` for clarity

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/TaskStatusViewModel.kt
```diff
@@ -0,0 +1,21 @@
...
+    val tag: Boolean = false,
+    val tagClass: String? = null,
+    val cannotStart: Boolean = false,
+) {
+    companion object {
+        fun fromStatus(status: TaskStatus): TaskStatusViewModel {
+            when (status) {
+                TaskStatus.CANNOT_START_YET -> return TaskStatusViewModel("taskList.status.cannotStartYet", cannotStart = true)
+                TaskStatus.NOT_YET_STARTED -> return TaskStatusViewModel("taskList.status.notYetStarted", true, "govuk-tag--blue")
+                TaskStatus.IN_PROGRESS -> return TaskStatusViewModel("taskList.status.inProgress", true, "govuk-tag--light-blue")
```
isobel-softwire, line 16: It would be helpful to use named parameters for the Booleans here, like you've done with `cannotStart = true`

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/TaskStatusViewModel.kt
```diff
@@ -0,0 +1,21 @@
+package uk.gov.communities.prsdb.webapp.models.viewModels
+
+import uk.gov.communities.prsdb.webapp.constants.enums.TaskStatus
+
+data class TaskStatusViewModel(
+    val textKey: String,
+    val tag: Boolean = false,
+    val tagClass: String? = null,
```
isobel-softwire, line 8: Is `tag` true iff `tagClass` isn't null?

isobel-softwire (approved):
Left some minor comments - LGTM!

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/TaskStatusViewModel.kt
AEPR: Yes, that's right - I'll make it a function!

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/TaskListItem.kt
AEPR: I'm not sure I agree, since it's already in the view model package - but for consistency I'm going to change it now and bring it up at catch up tomorrow!

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
```diff
@@ -50,6 +53,44 @@ class RegisterPropertyController(
...
         )
 
+    @GetMapping("/task-list")
+    fun getTaskList(model: Model): String {
+        val registerTaskList =
+            listOf(
+                TaskListItemViewModel(
+                    "registerProperty.taskList.register.addAddress",
+                    TaskStatusViewModel.fromStatus(TaskStatus.COMPLETED),
+                    "https:www.google.com",
```
Travis-Softwire, line 63: Extremely minor, but wonder if we should get into the habit of using Gov.uk or similar as placeholder links just in case something goes horribly wrong and we release code with a placeholder link still in place

Travis-Softwire (approved):

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
AEPR: OK - happy to switch to that in future.

PR #162, opened 2025-01-28
PRSD-719: bare bones of property details page

JasminConterioSW:
Bare bones version of PRSD-719, which should unblock PRSD-723 
- Creates the controller and endpoint, including the property ownership id in the path
- Retrieve the property ownership from the database (should be able to get the information needed for both tabs from this, I think)
- Create a template with tabs for Property details and Landlord details

image


isobel-softwire (approved):
LGTM!

PR #163, opened 2025-01-28
PRSD-814: Converts 400 errors

Travis-Softwire:
Small tweak to make sure that errors caused by the user changing the url always get a not found page instead of the generic error page. For now 400s in response to POST requests should still get the generic error page.


rowanhill (approved):

Kiran-R-K (approved):
I left one suggestion but it all LGTM

PR #166, opened 2025-01-28
Feat/prsd 594 property registration declaration

Kiran-R-K:
Added the declaration step to the `PropertyRegistrationJourney` re-using the existing `declarationForm`
Changed the check answers step to have the declaration step as it's next action.
Added `handleSubmitAndRedirect` to the declaration step to as it's now the last step before confirmation.
Created separate message properties for the fine amounts for declaration steps in `PropertyRegistrationJourney` and `LandlordRegistrationJourney`, and updated them to pass the message keys to the form
Updated affected tests
Added tests for the declaration step

the new declaration page for the `PropertyRegistrationJourney`:
image

the error message on the declaration page for the `PropertyRegistrationJourney`:
image

the declaration page on the `LandlordRegistrationJourney` (this should look unchaanged):
image


Travis-Softwire (approved):
Looks great :)

PR #167, opened 2025-01-30
PRSD-600: Don't diplay absent links

AEPR:
The names of "Cannot start yet" tasks should not display as links.

image

### DIFF
--- a/src/main/resources/templates/fragments/taskList/taskListItem.html
+++ b/src/main/resources/templates/fragments/taskList/taskListItem.html
@@ -1,7 +1,7 @@
 
 <li th:fragment="taskListItem(task, id)"  class="govuk-task-list__item" th:classappend="${task.url != null} ? 'govuk-task-list__item--with-link' : ''">
     <div class="govuk-task-list__name-and-hint">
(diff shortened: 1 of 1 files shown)


src/main/resources/templates/fragments/taskList/taskListItem.html
```diff
@@ -1,7 +1,7 @@
 
 <li th:fragment="taskListItem(task, id)"  class="govuk-task-list__item" th:classappend="${task.url != null} ? 'govuk-task-list__item--with-link' : ''">
     <div class="govuk-task-list__name-and-hint">
-        <a class="govuk-link" th:classappend="${task.url != null} ? 'govuk-task-list__link' : ''"
+        <a class="govuk-link 'govuk-task-list__link" th:remove="${task.url == null} ? tag : none"
```
isobel-softwire, line 4: I don't think the quotation mark in front of `govuk-task-list__link` is supposed to be there...

isobel-softwire (approved):
LGTM!

src/main/resources/templates/fragments/taskList/taskListItem.html
AEPR: Quite right.

PR #168, opened 2025-01-30
Feat/prsd 600 property details from session

AEPR:
This is missing tests for the TaskList class and similar, as well as integration tests. (Which is why this is a draft review)

Still outstanding (for a subsequent PR) will be redirecting to the task-list page on starting or revisiting the journey (and loading state from the database when doing so).

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/tasks/CheckAndSubmitPropertiesTaskList.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/tasks/CheckAndSubmitPropertiesTaskList.kt
@@ -1,0 +1,42 @@
+package uk.gov.communities.prsdb.webapp.forms.tasks
+
+import org.springframework.stereotype.Component
+import org.springframework.validation.Validator
+import uk.gov.communities.prsdb.webapp.constants.enums.TaskStatus
+import uk.gov.communities.prsdb.webapp.forms.journeys.PropertyRegistrationJourney
+import uk.gov.communities.prsdb.webapp.forms.steps.RegisterPropertyStepId
+import uk.gov.communities.prsdb.webapp.models.viewModels.TaskListItemViewModel
+import uk.gov.communities.prsdb.webapp.models.viewModels.TaskStatusViewModel
+import uk.gov.communities.prsdb.webapp.services.JourneyDataService
+
+@Component
+class CheckAndSubmitPropertiesTaskList(
+    propertyRegistrationJourney: PropertyRegistrationJourney,
+    journeyDataService: JourneyDataService,
+    validator: Validator,
+) : TaskList<RegisterPropertyStepId>(propertyRegistrationJourney, journeyDataService, validator) {
+    override val taskList: List<TaskListItemDataModel<RegisterPropertyStepId>>
+        get() =
+            listOf(
+                TaskListItemDataModel(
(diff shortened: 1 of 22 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/tasks/TaskList.kt
```diff
@@ -0,0 +1,88 @@
...
+    private val journey: Journey<T>,
+    private val journeyDataService: JourneyDataService,
+    private val validator: Validator,
+) {
+    private val steps = journey.steps
+    protected abstract val taskList: List<TaskListItemDataModel<T>>
+
+    open fun getTaskListViewModels(): List<TaskListItemViewModel> {
+        val journeyData = journeyDataService.getJourneyDataFromSession()
+        val registerPropertyList =
```
rowanhill, line 23: I think this is probably misnamed after a refactor?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/tasks/TaskList.kt
```diff
@@ -0,0 +1,88 @@
...
+                null
+            } else {
+                taskListItem.startId.urlPathSegment
+            },
+        )
+    }
+
+    private fun getStatusForTask(
+        journeyData: JourneyData,
+        firstStepId: T,
```
rowanhill, line 48: This is the first step _of a task_ right? Rather than the first step of the whole journey?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/tasks/CheckAndSubmitPropertiesTaskList.kt
```diff
@@ -0,0 +1,42 @@
...
+    override val taskList: List<TaskListItemDataModel<RegisterPropertyStepId>>
+        get() =
+            listOf(
+                TaskListItemDataModel(
+                    "registerProperty.taskList.checkAndSubmit.checkAnswers",
+                    RegisterPropertyStepId.CheckAnswers,
+                    null,
+                ),
+            )
+    // TODO PRSD-589: Add another property
```
rowanhill, line 27: Not really your problem in this ticket, but: is adding another property really a task list item? Feels like an action you might decide to take at some point in the journey, but not a task list item to be completed.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/tasks/TaskList.kt
AEPR: Indeed - and it's actually unnecessary anyway since I can return the expression directly without putting it in a variable first.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/tasks/TaskList.kt
AEPR: Yes, that's right - it will probably be clearer if I pass in the `TaskListItemDataModel` here.

AEPR (commented):
Updated two points you've picked up.

Travis-Softwire:
This all LGTM :)

isobel-softwire (approved):
LGTM - left a couple of refactoring suggestions

PR #170, opened 2025-01-30
Feat/prsd 719 property details

JasminConterioSW:
-  Adds a "Key information" panel at the top of the page
    - This is a non-GDS component
    - It uses a details list (chosen for accessibility, taking advice from cop-accessibility and govuk-design-system slack channels)
    - It's using custom CSS with a flexbox display (I spent some time trying to make it work with GDS classes but didn't get it looking right and didn't want to spend too long on it)
 - Adds the property record to the Property details tab
 - Adds controller tests
 - Does not add integration tests - these will be in a separate PR after [#171 ](https://github.com/communitiesuk/prsdb-webapp/pull/171/files#diff-0f537b71f3fbf70bb22b7cd8bf9ff77f3a6eb80d759322fe6603f352bb636f62) is merged

image


src/main/resources/templates/propertyDetailsView.html
```diff
@@ -4,7 +4,13 @@
 <main class="govuk-main-wrapper" id="main-content">
   <div id="page-contents">
     <h2 class="govuk-heading-m" th:text="#{propertyDetails.heading}">propertyDetails.heading</h2>
-
+    <div th:replace="~{fragments/buttons/pageHeaderActions :: pageHeaderActions(${propertyDetails.address}, ~{:: #delete-record-button})}">
+      <th:block id="delete-record-button">
+        <a th:replace="~{fragments/buttons/primaryButtonLink :: primaryButtonLink('delete-record', #{propertyDetails.deleteRecordButton.text})}"></a>
```
rowanhill, line 9: Is deleting the record actually implemented?

rowanhill (approved):

src/main/resources/templates/propertyDetailsView.html
rowanhill: And, actually, shouldn't this be a red button rather than the primary green?

JasminConterioSW: Ah, the design has updated since the screenshots were added to the ticket, so the button is now green.
We've been going with Figma as the source of truth rather than necessarily the screenshots on the ticket! https://www.figma.com/design/ka4URNl45NiQO0LoY8XNe4/MHCLG---Private-Rented-Sector-Database?node-id=10593-94706&m=dev

JasminConterioSW: Deleting the record isn't implemented yet though, no - that might be part of deregistration when we get to that!

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/PropertyDetailsViewModel.kt
```diff
@@ -0,0 +1,140 @@
...
+            ),
+            SummaryListRowViewModel(
+                "propertyDetails.propertyRecord.propertyType",
+                MessageKeyConverter.convert(propertyOwnership.property.propertyBuildType),
+                null,
+            ),
+            SummaryListRowViewModel(
+                "propertyDetails.propertyRecord.ownershipType",
+                MessageKeyConverter.convert(propertyOwnership.ownershipType),
+                "#",
```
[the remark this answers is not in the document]
JasminConterioSW: They are placeholders, looks like there are a load of tickets for these in this sprint, I'll add some TODOs!

PR #171, opened 2025-01-31
Feat/prsd 723 property details view landlord details tab

Kiran-R-K:
Landlord details tab on the property record page

Get's the details of the primary landlord for the property, formats and returns them to the view  

UK landlord:
image

International landlord:
image


src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/PropertyDetailsLandlordViewModel.kt
```diff
@@ -0,0 +1,79 @@
...
+                toggleChangeLink("$UPDATE_ROUTE/address"),
+            ),
+        )
+
+    private fun toggleChangeLink(link: String?): String? =
+        if (withChangeLinks) {
+            link
+        } else {
+            null
+        }
```
Kiran-R-K, line 78: This duplicates the method in `LandlordViewModel` I think it should be pulled out somewhere but I'm not sure where would be sensible?

JasminConterioSW: Maybe we can add a ViewModelHelper or similar?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
```diff
@@ -1,25 +1,54 @@
...
+
+        val propertyOwnershipLandlordId = propertyOwnership.primaryLandlord.id
+        val principalLandlordIdFrom =
+            landlordService.retrieveLandlordIdByBaseUserId(principal.name)
+                ?: throw PrsdbWebException("User ${principal.name} is not registered as a landlord")
+
+        if (propertyOwnershipLandlordId != principalLandlordIdFrom) {
+            throw ResponseStatusException(HttpStatus.FORBIDDEN)
+        }
+        val landlordsDetails =
```
JasminConterioSW, line 43: You shouldn't need to call the database here because this landlordDetails is the same as `propertyOwnership.primaryLandlord` 

(If the landlord isn't in the database, it's going to throw an error when you try to access this on line 35.  We could try to catch the error there, but I don't think the database will let us get to the position where we have a property ownership with no primary landlord as it's not a nullable field)

JasminConterioSW (changes requested):

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
```diff
@@ -1,25 +1,54 @@
...
+            landlordService.retrieveLandlordById(propertyOwnershipLandlordId)
+                ?: throw ResponseStatusException(HttpStatus.NOT_FOUND, "Landlord $propertyOwnershipLandlordId not found")
```
isobel-softwire, line 45: I think this should be abstracted into a `PropertyOwnershipService` method that calls `PropertyOwnershipRepository.findByIdAndPrimaryLandlord_Id` and returns the primary landlord of the retrieved property ownership. If this change is made I'm don't think you'll need the part above, so maybe talk to Jasmin about whether she needs it.

src/main/resources/templates/fragments/summaryRowWithValueLink.html
isobel-softwire, line 1: I think the summary list fragments should be put in a folder. Also, the `summaryRow` fragment should delegate to this so we only have to make changes to it in one place.

isobel-softwire: Actually, I'm wondering if it would be worth adding a `valueUrl` field to `SummaryListRowViewModel`. Then there wouldn't need to be separate `summaryRow` fragments. What do you think?

isobel-softwire (changes requested):
mostly LGTM - left a few comments

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/PropertyDetailsLandlordViewModel.kt
rowanhill: I'd suggest we just leave it for now. I don't think it belongs on a helper (it depends on a member property), and it's not worth setting up any inheritance / strategy pattern / etc for such a tiny method.

src/main/resources/templates/fragments/summaryRowWithValueLink.html
rowanhill: That sounds good to me - it would mean we wouldn't need to treat the landlord name so specially in the view model, too.

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsControllerTests.kt
```diff
@@ -17,53 +19,81 @@ class PropertyDetailsControllerTests(
...
+        mvc
+            .get("$propertyDetailsUrl/$propertyOwnershipId")
+            .andExpect {
+                status { isForbidden() }
+            }
     }
 
     @Test
-    @WithMockUser(roles = ["LANDLORD"])
-    fun `getPropertyDetails returns 404 if the requested property ownership is not found`() {
```
JasminConterioSW, line 39: I think this test might have disappeared when you merged by the way!

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsControllerTests.kt
```diff
@@ -17,53 +19,81 @@ class PropertyDetailsControllerTests(
...
+        mvc
+            .get("$propertyDetailsUrl/$propertyOwnershipId")
+            .andExpect {
+                status { isForbidden() }
+            }
     }
 
     @Test
     @WithMockUser(roles = ["LANDLORD"])
-    fun `getPropertyDetails returns error if the requested property ownership is inactive`() {
```
JasminConterioSW, line 47: Also this one

Kiran-R-K: as above - I've made sure to add it back

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
Kiran-R-K: @JasminConterioSW do you have any opinions on this - I could write a query that checks against the `principal.name` and `propertyOwnershipId` at once (and then when we get to implementing joint ownerships it should be easily expandable) I can also add `isActive` into the query so all requirements are checked at db level when getting the data. 
It feels sensible to me but I also only have one side of the story - what do you think?

JasminConterioSW: Either way I don't think you need this call to the database to get the landlord because it's already available in the returned property ownership

Kiran-R-K: yeah let's hop on a call

Kiran-R-K: We discussed this and came to the conclusion that:
- there should be a method in `PropertyOwnershipService` that calls `retrievePropertyOwnershipById` and checks that the landlord of the property matches `principal.name` and returns appropriate errors including changing the error if the landlord ids don't match from `FORBIDDEN` to `NOT_FOUND`
- `retrievePropertyOwnershipById` should call a method on `PropertyOwnershipRepository` that returns a PropertyOwnership by it's Id only if it is also active

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/PropertyDetailsLandlordViewModel.kt
```diff
@@ -0,0 +1,82 @@
...
+    private fun formatUkAddressDetails(landlord: Landlord) =
+        listOf(
+            SummaryListRowViewModel(
+                "landlordDetails.personalDetails.contactAddress",
+                landlord.address.singleLineAddress,
+                toggleChangeLink("$UPDATE_ROUTE/address"),
+            ),
+        )
+
+    private fun toggleChangeLink(link: String?): String? =
```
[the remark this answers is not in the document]
Kiran-R-K: I'm hesitant to do that because on the PR where I created this method on the `LandlordViewModel ` it was called something like `setChangeLink` and was changed to `toggleChangeLink` in response to a comment because it's deciding whether or not the link will be visible.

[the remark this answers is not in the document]
rowanhill: Okay, alternate suggestion, then: in Jasmin's recent PR I suggested that we [use a mutable list with .apply](https://github.com/communitiesuk/prsdb-webapp/blob/e3f5fbb8191e5ba3f51f15c6e0f9b10258fba6e7/src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/PropertyDetailsViewModel.kt#L47) to create the list of `SummaryListRowViewModel`s in the view model, and also to use a [little extension method](https://github.com/communitiesuk/prsdb-webapp/blob/e3f5fbb8191e5ba3f51f15c6e0f9b10258fba6e7/src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/PropertyDetailsViewModel.kt#L107) to make it prettier.

How about we do the same here but the extension method can ignore the supplied `link` if `withChangeLinks == false`. I think that'll a) be a bit neater, b) be more consistent with Jasmin's work, c) avoid needing this method at all, so we don't have to agree on a name :)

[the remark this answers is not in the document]
Kiran-R-K: That sounds good, there's quite a lot changing in this PR now so I think I'll have this change be in a separate PR to make things clearer

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsControllerTests.kt
Kiran-R-K, line 1: The deleted tests in this file will require equivalent tests to be added to PropertyOwnershipServiceTests - they will be added in a separate Pr

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/PropertyDetailsPage.kt
```diff
@@ -0,0 +1,21 @@
+package uk.gov.communities.prsdb.webapp.integration.pageObjects.pages
```
JasminConterioSW, line 1: Sorry, I've just changed these in my integration test PR - there's now a base page and the landlord and la version inherit from that. It follows a similar pattern to the landlord version so hopefully won't be too bad!

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDetailsTests.kt
```diff
@@ -0,0 +1,47 @@
...
+            val detailsPage = navigator.goToPropertyDetails(1)
+            detailsPage.goToLandlordDetails()
+
+            assertEquals(detailsPage.getActiveTabPanelId(), "landlord-details")
+        }
+
+        @Test
+        fun `when the landlord details tab is active clicking the property details tab shows property details tab`(page: Page) {
+            val detailsPage = navigator.goToPropertyDetails(1)
+            println(detailsPage.page.content())
```
JasminConterioSW, line 32: This line isn't needed, I think

JasminConterioSW (approved):
LGTM - hopefully I haven't caused issues with your integration tests, you'll get some merge conflicts when you merge in main.

The functionality should be the same for the tests you added, the page objects are slightly different now as I have a landlord version and an la version that both inherit from a base version. Give me a shout if it's being a pain!

isobel-softwire (approved):
LGTM - left a minor suggestion

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -49,6 +52,26 @@ class PropertyOwnershipService(
...
+                    HttpStatus.NOT_FOUND,
+                    "Property ownership $propertyOwnershipId not found",
+                )
+
+        val landlordIdFromPrincipal =
+            landlordService.retrieveLandlordIdByBaseUserId(baseUserId)
+
+        if (propertyOwnership.primaryLandlord.id != landlordIdFromPrincipal) {
+            throw ResponseStatusException(HttpStatus.NOT_FOUND)
+        }
```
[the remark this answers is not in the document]
Kiran-R-K: `propertyOwnership.primaryLandlord.baseUser.id` is a private value so I'll leave it for now

PR #172, opened 2025-02-03
PRSD-NONE: Store address data within the journey data

AEPR:
When we search for addresses using postcode and house name/number we get a list of addresses which we are currently separately caching in the session. This change instead stores it in journey data (still in the session) so that when we save the journey data to the database it will be persisted.

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/AddressDataService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/AddressDataService.kt
@@ -3,28 +3,30 @@ package uk.gov.communities.prsdb.webapp.services
 import jakarta.servlet.http.HttpSession
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
-import org.springframework.context.annotation.Scope
-import org.springframework.context.annotation.ScopedProxyMode
 import org.springframework.stereotype.Service
-import org.springframework.web.context.WebApplicationContext
 import uk.gov.communities.prsdb.webapp.forms.journeys.objectToStringKeyedMap
 import uk.gov.communities.prsdb.webapp.models.dataModels.AddressDataModel
 
 @Service
-@Scope(value = WebApplicationContext.SCOPE_REQUEST, proxyMode = ScopedProxyMode.TARGET_CLASS)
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/AddressDataService.kt
```diff
@@ -3,28 +3,30 @@ package uk.gov.communities.prsdb.webapp.services
...
 import kotlinx.serialization.json.Json
-import org.springframework.context.annotation.Scope
-import org.springframework.context.annotation.ScopedProxyMode
 import org.springframework.stereotype.Service
-import org.springframework.web.context.WebApplicationContext
 import uk.gov.communities.prsdb.webapp.forms.journeys.objectToStringKeyedMap
 import uk.gov.communities.prsdb.webapp.models.dataModels.AddressDataModel
 
 @Service
-@Scope(value = WebApplicationContext.SCOPE_REQUEST, proxyMode = ScopedProxyMode.TARGET_CLASS)
```
isobel-softwire, line 14: Is this not needed anymore?

rowanhill:
I think I'm missing the context, here: why do we want address search results to be persisted in the DB with the journey data?

If I were resuming a registration journey and hadn't previously selected an address, personally I would expect to have to search again; ditto if I _had_ selected an address but wanted to change it.

AEPR:
These search results are actually used directly to store the address that was selected - the Radio button selection only stores the first line of the address in the journey and then whenever we look up the full address value we take the first line and use it to retrieve the full address from the cached results.

We could change how this works more significantly - e.g. storing the full selected address value in the journey and as you mention forcing the user to re-search if they want to change their selection if the sessions expired, but I was just making a smallest change that allowed the journey to work after restoring data from the database.

rowanhill:
Right, okay. So, with the changes in this PR, if someone wants to change the address after resuming a registration journey, will the UI / journey logic cause them to re-search?

AEPR:
If the user navigates directly to the /register-property/select-address (or equivilent) url then they'll see the cached results with their previously selected address. If they use the task list they'll be taken to the lookup-address screen and be prompted to search again (with their details already filled in).

PR #175, opened 2025-02-04
PRSD-403: Repeatable LA Table Population Migration 

isobel-softwire:
- Makes LA table population migration repeatable and adds instructions as to how to make the migration repeat
- Copies `local_authorities` folder to docker container (instead of file)


ReadMe.md
```diff
@@ -86,11 +86,10 @@ populate the database with seed data.
...
 
-The project uses a migration (`V1_10_0__populate_la_table.sql`) to populate the `local_authority` table with data from
-`src/main/resources/db/migrations/data/local_authorities/local_authorities.csv`. If the CSV file is updated, create a
-copy of
-it and call it `local_authorities_V<version number>.csv`, where `version number` is one more than the latest version in
-`src/main/resources/db/migrations/data/local_authorities`.
+The project uses a repeatable migration (`R__populate_la_table.sql`) to populate the `local_authority` table with data 
+from `src/main/resources/db/migrations/data/local_authorities/local_authorities_V<version number>.csv`. To update this 
+data, create a new `local_authorities_V<version number>.csv` file (where `version number` is one more than the latest 
+version) and update `R__populate_la_table.sql` to use it.
```
rowanhill, line 92: Why do we want the csv file name to be versioned? Is it purely so the checksum of the repeatable migration updates when the data changes?

src/main/resources/db/migrations/R__populate_la_table.sql
```diff
@@ -9,11 +9,22 @@ CREATE TABLE tmp
...
+    FROM '/data/local_authorities/local_authorities_V1.csv'
     WITH (FORMAT CSV, HEADER);
 
 INSERT INTO local_authority(custodian_code, name)
 SELECT AUTH_CODE, ACCOUNT_NAME
 FROM tmp;
 
-DROP TABLE tmp;
\ No newline at end of file
+DROP TABLE tmp;
```
rowanhill, line 19: Instead of explicitly dropping tmp, why don't we create it as a `TEMPORARY TABLE`?

src/main/resources/db/migrations/R__populate_la_table.sql
```diff
@@ -9,11 +9,22 @@ CREATE TABLE tmp
 );
 
 COPY tmp
-    FROM '/data/local_authorities.csv'
+    FROM '/data/local_authorities/local_authorities_V1.csv'
     WITH (FORMAT CSV, HEADER);
 
 INSERT INTO local_authority(custodian_code, name)
 SELECT AUTH_CODE, ACCOUNT_NAME
 FROM tmp;
```
rowanhill, line 17: If this is re-run every time it changes, I guess we're relying on the unique constraint on the custodian_code to prevent duplicates? What about deleting stale data (e.g. if an LA is removed)? And what happens if/when this migration encounters a unique constraint violation?

src/main/resources/db/migrations/R__populate_la_table.sql
```diff
@@ -9,11 +9,22 @@ CREATE TABLE tmp
...
+
+CREATE TABLE tmp
+(
+    AUTH_CODE              VARCHAR(255),
+    ACCOUNT_NAME           VARCHAR(255),
+    ACCOUNT_TYPE_NAME      VARCHAR(255),
+    ADDRESS_SUBMITTER_FLAG CHAR,
+    STREET_SUBMITTER_FLAG  CHAR,
+    SUBMITTER_FLAG         CHAR
+);
```
rowanhill, line 29: Why do we recreate this table immediately after dropping it? And in general why do we leave it in place at the end of the migration, now?

isobel-softwire: This isn't supposed to be here - I'll remove it

src/main/resources/db/migrations/data/local_authorities/local_authorities.csv
rowanhill, line 1: How come we're no longer keeping this in source control?

src/main/resources/db/migrations/R__populate_la_table.sql
isobel-softwire: As there's uncertainty over how local authorities can change, we decided to leave the migration like this for now, and update it if/when we need to

src/main/resources/db/migrations/data/local_authorities/local_authorities.csv
isobel-softwire: This is a copy of the most up-to-date versioned file. The migration uses the versioned file now so this is no longer required.

rowanhill: Okay, but shouldn't the versioned file be added to source control?

src/main/resources/db/migrations/R__populate_la_table.sql
rowanhill: I think we can be pretty confident that LAs will be created, updated, and deleted, so our process should handled all of those cases.

Writing a migration that doesn't work for those cases just feels like a foot-gun, and we're going to confuse ourselves in future. What's the benefit of making the migration repeatable like this, rather than just leaving it as-is for now, if we really think we don't know enough about how LA data is going to change?

src/main/resources/db/migrations/data/local_authorities/local_authorities.csv
isobel-softwire: It was added to source control in a previous ticket 👍

src/main/resources/db/migrations/R__populate_la_table.sql
isobel-softwire: I meant more in terms of the cascading effects of changes, which will most likely require manual intervention. For example, if an LA subsumes another, the addresses/LA users associated with the subsumed LA will need updating.

rowanhill: Yes, I agree with all of that - changes to LAs might have wider impacts than just changing the local_authority table's data, and we'll need to think that through carefully at some point.

But even just narrowly thinking about the local_authority table, I don't think this PR is going to do what we want (because of errors on duplicates, leaving stale data, etc). Am I missing the benefit this change gives us?

isobel-softwire: This PR is more about deciding how we'll deal with changes to LAs in the future (i.e. by updating the repeatable migration instead of creating a new migration)

PR #176, opened 2025-02-04
Feat/prsd 600 property details from session

AEPR:
Full review for PRSD-600. Includes the changes from https://github.com/communitiesuk/prsdb-webapp/pull/168, but with redirection to the task page and tests.


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegisterPropertyTaskListService.kt
isobel-softwire, line 1: I'm not sure that this needs to be its own service. In the `/task-list` endpoint you could call `journeyDataService.initializeJourneyData()` to load journey data into session and add the task list data if needed, then call `journey.populateModelAndGetViewName()`. Happy to have a call about this!

isobel-softwire (changes requested):
Mostly LGTM - left a few thoughts that I'm happy to talk about if you'd like :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/FormContextRepository.kt
```diff
@@ -1,6 +1,14 @@
...
 import uk.gov.communities.prsdb.webapp.database.entity.FormContext
 
-interface FormContextRepository : JpaRepository<FormContext, Long?>
+// The underscore tells JPA to access fields relating to the referenced table
+@Suppress("ktlint:standard:function-naming")
+interface FormContextRepository : JpaRepository<FormContext, Long?> {
+    fun findByUser_IdAndJourneyType(
+        principalName: String,
+        journeyType: JourneyType,
+    ): FormContext?
```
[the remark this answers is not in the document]
rowanhill: I thought we were meant to allow multiple incomplete property registration journeys per user, so I wouldn't expect (user ID, journey type) to be uniquely identifying.

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/TaskListTests.kt
```diff
@@ -0,0 +1,197 @@
...
+
+            // Assert
+            assertIterableEquals(
+                listOf("taskList.status.completed", "taskList.status.inProgress", "taskList.status.cannotStartYet"),
+                viewModel.map { it.status.textKey },
+            )
+        }
+
+        @Test
+        fun `when a task with a null completion step is completed, it is not marked as `() {
```
rowanhill, line 171: The test name kinda trails off, here... :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/Navigator.kt
```diff
@@ -264,8 +265,14 @@ class Navigator(
...
 
+    fun goToPropertyRegistrationTaskList(): TaskListPagePropertyRegistration {
+        navigate("register-property/task-list")
+        return createValidPage(page, TaskListPagePropertyRegistration::class)
+    }
+
     fun goToPropertyRegistrationLookupAddressPage(): LookupAddressFormPagePropertyRegistration {
-        navigate("register-property/lookup-address")
+        val taskListPage = goToPropertyRegistrationTaskList()
+        taskListPage.clickRegisterTaskWithName("Add the property address")
```
rowanhill, line 275: This feels very surprising - wouldn't we expect the Navigator to just go directly to the url? And if a test needs to go via the task list, it's free to do so?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationJourneyTests.kt
```diff
@@ -206,6 +211,37 @@ class PropertyRegistrationJourneyTests : IntegrationTest() {
         assertEquals("/", URI(page.url()).path)
     }
 
+    @Nested
+    inner class TaskListStep {
+        @Test
+        fun `Completing preceding steps will show a task as not yet started and completed steps as complete`(page: Page) {
+            navigator.goToPropertyRegistrationOccupancyPage()
```
rowanhill, line 218: Can you explain how vising this page means a step/task is now complete?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/FormContextRepository.kt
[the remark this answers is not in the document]
AEPR: OK, in that case what *are* we using to allow a user to resume this journey once it isn't in session state any more?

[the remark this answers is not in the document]
isobel-softwire: `FormContext` can store multiple property journeys at once so I think (user ID, journey type) is supposed to be uniquely identifying but this constraint isn't enforced

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationJourneyTests.kt
AEPR: Reaching a particular page means completing all of the pages preceding it.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/FormContextRepository.kt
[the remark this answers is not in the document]
rowanhill: Good question - I guess we either only allow one journey of each type per user (and `(user ID, journey type)` is unique), or we need some kind of UI to allow users to pick the journey to resume. I'm asking what the latest thinking is in this area.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationJourneyTests.kt
AEPR: That is actually exactly what `navigator.goToPropertyRegistrationOccupancyPage()` does - all of the navigator functions for forms complete the form journey up to that point. The thought here is that for all of the integration tests targetting specific pages (e.g. the occupancy page), we just want to arrive happily on that page so we can start testing it.

rowanhill: Huh, okay. That's _very_ surprising to me! IMO the Navigator object should just do simple navigation, and if we want to have some "fill in the whole journey" logic, that belongs in a different class. But that's a discussion that's out of scope of this PR :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/TaskListTests.kt
AEPR: And the test itself has the wrong expectation and is failing...

I fixed it first time out but forgot to `git add .` 🤦

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegisterPropertyTaskListService.kt
AEPR: I think the logic in `loadJourneyDataIntoSessionIfNotLoaded` is enough to justify not being in a controller, especially the direct journey data manipulation in `addTaskListStepDataToJourneyData`. Happy to have a call about it this morning though.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/tasks/CheckAndSubmitPropertiesTaskList.kt
```diff
@@ -0,0 +1,35 @@
...
+import uk.gov.communities.prsdb.webapp.services.JourneyDataService
+
+@Component
+class CheckAndSubmitPropertiesTaskList(
+    propertyRegistrationJourney: PropertyRegistrationJourney,
+    journeyDataService: JourneyDataService,
+    validator: Validator,
+) : TaskList<RegisterPropertyStepId>(propertyRegistrationJourney, journeyDataService, validator) {
+    override val taskList: List<Task<RegisterPropertyStepId>>
+        get() =
```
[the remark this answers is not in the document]
AEPR: It can't change, no - but that's already specified by the fact it's a `val` not a `var`. I should just keep it simple and use an assignment here and in `RegisterPropertiesTaskList` (though fundementally I don't think it massively changes the functionality).

eta: The actual difference is that the assignment creates one list that lives in memory in the variable, whereas the getter creates a new list every time this value is called. I don't think that's in any way significant here.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
```diff
@@ -95,9 +96,17 @@ class PropertyRegistrationJourney(
                     confirmationEmailSender,
                     session,
                 ),
+                taskListStep(),
```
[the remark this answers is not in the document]
AEPR: There's a slightly contrary part of me that thinks it would be better if the tasks here were in a completely jumbled order so you it's harder to have the misapprehension that the order of them matters. 

That said, it's already a `Set` which doesn't have order so I think you're right that I should prioritise readability.

AEPR (commented):
Accepted a couple of suggestions: 
* @isobel-softwire keen for a chat on the `RegisterPropertyTaskListService`
* There's still an open discussion on how we're restoring journey state from the database

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegisterPropertyTaskListService.kt
isobel-softwire: We've had a conversation about this, and decided that there should be an abstract class that stores a journey's task lists and performs the functionality currently in `RegisterPropertyTaskListService`

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegisterPropertyTaskListService.kt
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    private fun loadJourneyDataIntoSessionIfNotLoaded(principalName: String) {
+        val data = journeyDataService.getJourneyDataFromSession()
+        if (data.isEmpty()) {
+            val contextId = journeyDataService.getContextId(principalName, JourneyType.PROPERTY_REGISTRATION)
+            if (contextId == null) {
+                addTaskListStepDataToJourneyData(data)
+            } else {
+                journeyDataService.loadJourneyDataIntoSession(contextId)
```
rowanhill, line 33: It's a bit of a shame to look the context up from the database, take the id, then pass the id to another method which retrieves it from the database. I think ideally we'd just pass the whole context (as an overload to journeyDataService.loadJourneyDataIntoSession) - but, given we expect this stuff to change when we allow multiple concurrent incomplete journeys, probably not worth fixing now.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/FormContextRepository.kt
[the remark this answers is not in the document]
AEPR: 👍 Sounds good.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegisterPropertyTaskListService.kt
AEPR: Done! Probably needs a re review. I've called it a MultiTaskTransaction based on the naming here: https://design-system.service.gov.uk/patterns/complete-multiple-tasks/

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/RegisterPropertyTaskListService.kt
AEPR: I've paraphrased this review comment into a code comment with a TODO. Have you raised a ticket to address this, and if so what is the PRSD number? Otherwise I'll update the comment to not have a number (or raise a ticket myself).

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
```diff
@@ -54,39 +53,14 @@ class RegisterPropertyController(
...
+    fun getTaskList(
+        model: Model,
+        principal: Principal,
+    ): String {
+        val viewModelLists = registerPropertyTransaction.getTaskListPageViewModels(principal.name)
 
-        model.addAttribute("registerTasks", registerTaskList)
-        model.addAttribute("checkAndSubmitTasks", checkAndSubmitTaskList)
+        model.addAttribute("registerTasks", viewModelLists[0])
+        model.addAttribute("checkAndSubmitTasks", viewModelLists[1])
```
isobel-softwire, line 63: Not a big fan of relying on the order of the lists here. Maybe `getTaskListPageViewModels` should return key-value pairs where the key is the `TaskList` type (or something along those lines)

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/MultiTaskTransactionTests.kt
```diff
@@ -0,0 +1,123 @@
...
+
+            // Act
+            transaction.getTaskListPageViewModels(principalName)
+
+            // Assert
+            val captor = argumentCaptor<JourneyData>()
+            verify(journeyDataService).setJourneyData(captor.capture())
+            assertIs<MutableMap<*, *>>(
+                captor.firstValue[testUrlSegment],
+            )
```
isobel-softwire, line 85: I think this should check that `captor.firstValue[testUrlSegment]` equals `mutableMapOf()` instead of making a type assertion. Also, would `captor.value` work instead of  `captor.firstValue`?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/components/TaskList.kt
```diff
@@ -0,0 +1,25 @@
...
+            .getByText("In progress")
+            .or(
+                locator
+                    .locator("li")
+                    .getByText("Not yet started"),
+            ).nth(0)
+            .click()
+
+    fun getTaskStatus(name: String): String =
+        getChildComponent("li", Locator.LocatorOptions().setHasText(name)).locator(".govuk-task-list__status").textContent()
```
isobel-softwire, line 24: You could abstract locating task elements to a private `getTask()` method. Also it would be good to locate the status using `getChildComponent()` so we can assert that it exists.

isobel-softwire (changes requested):
`MultiTaskTransaction` looks really good! Left a comment about relying on list order in the controller and a few suggestions in the tests, but apart from that LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
AEPR: It's an interesting one, because my thought is that the MultiTaskTransaction has responsibility for the order of the taskList segments in the list it owns but I now realise that this doesn't really work because you could easily switch the order of them attributes here. As I see it we need to go all the way one way or all the way the other:
* The attribute we add here is a list of taskList view models each of which has a heading value as well as the list of taskListItems. We pass that list to the fragment as is and there's a MultiTaskTransaction fragment that creates a "Task list with heading" for each item in the list. Order IS captured by the order in the list, but that's meaningful because it's corresponding to order on the page.
* Do what you suggest, change the return a `Map<,>` from `getTaskListPageViewModels()`, probably update `taskLists: List<TaskList<T>>` to be a `Set<>` instead and have no ordering happening outside of the page fragments.

I think I can see the appeal of both and probably lean towards the latter. What do you think?

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/MultiTaskTransactionTests.kt
AEPR: Yeah, that's fair enough and an easy change. My thought here was that we don't ever care about the contents of the map, only that it exists. But equally, we don't care that it's a map either - we really only care that it's not null. Either way, just testing that it matches the value we just set it to seems reasonable.

No, I'm using `argumentCaptor()` here (from `mockito.kotlin`) instead of `captor()` (from base `mockito`). It plays more nicely with kotlin and we should prefer it, but it means the captor values are in a list. I could do `.allValues.single()` to ensure there's only one though.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
isobel-softwire: Probably outside the scope of this PR, but I think `registerPropertyTaskList.html` should be generalised to a `taskList.html` template (and maybe put in the forms folder). The other content could be added in the controller (or `MultiTaskTransaction` could have a `populateModel` method). On that basis, I'd lean towards the first option.

AEPR: OK, I'm adding a comment with a TODO for this ticket - I'll do that update after merging this PR. Leaving the code as is for now.

PR #179, opened 2025-02-05
PRSD-659: Property Search

isobel-softwire:
- Creates methods for querying the property ownerships table by PRN, UPRN, or (fuzzy) address
- Creates and tests `PropertySearchResultViewModel`
- Creates and tests `AddressHelper.parseUprnOrNull()` (checks if a given string could be a UPRN)
- Creates and tests `PropertyOwnershipService.searchForProperties()`
- Creates skeleton property search page
- Adds more property ownerships to data-local.sql
- Note: UI and integration tests will be implemented in a future PR


src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/PropertySearchResultViewModel.kt
```diff
@@ -0,0 +1,37 @@
...
+                landlord =
+                    PropertySearchResultLandlordViewModel(
+                        id = propertyOwnership.primaryLandlord.id,
+                        name = propertyOwnership.primaryLandlord.name,
+                    ),
+            )
+    }
+}
+
+data class PropertySearchResultLandlordViewModel(
```
JasminConterioSW, line 34: I know it's tiny, but I think we've generally been going with a separate file for each class

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -66,6 +69,23 @@ class PropertyOwnershipService(
...
+
+        val matchingProperties =
+            if (prn != null) {
+                propertyOwnershipRepository.searchMatchingPRN(prn.number)
+            } else {
+                val uprnMatch =
+                    AddressHelper
+                        .parseUprnOrNull(searchTerm)
+                        ?.let { uprn -> propertyOwnershipRepository.searchMatchingUPRN(uprn) } ?: emptyList()
+                uprnMatch + propertyOwnershipRepository.searchMatching(searchTerm)
```
JasminConterioSW, line 83: I'm not sure about this, if the UPRN matches, this still searches the database for any addresses matching the search term. Do we still need it to do that or can we just return the match to UPRN?

Actually I just checked the ticket and yes it does expect you to return other records that might match the uprn string.

So you're right but I'll leave the comment here anyway in case anyone has the same idea!

JasminConterioSW: What it does here is return if there is an exact match for the uprn, then if there is a fuzzy match on the address.

Should it be returning an exact match on the uprn, then a fuzzy match on the uprn (and maybe then a fuzzy match on the address)?

JasminConterioSW: This is what it says on the ticket:

```
When I search for a string that might be a UPRN (1-12 numeric digits, no spaces or other characters)
Then I am shown the property (if any) with an exact match for the UPRN first, followed by the other matching properties
```

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
```diff
@@ -25,4 +27,40 @@ interface PropertyOwnershipRepository : JpaRepository<PropertyOwnership, Long> {
...
+
+    @Query(
+        "SELECT po.* " +
+            "FROM property_ownership po " +
+            "JOIN property p ON po.property_id = p.id " +
+            "JOIN address a ON p.address_id = a.id " +
+            "WHERE po.is_active AND a.uprn = :searchUPRN",
+        nativeQuery = true,
+    )
+    fun searchMatchingUPRN(
```
JasminConterioSW, line 50: I'm not sure if you are going to need a search for an exact match, then a fuzzy search on the uprn - it's not clear to me whether the requirement on the ticket for showing "other matching properties" means matching bits of uprn or address or both!

Or maybe just a fuzzy search will be enough, it should give you the exact match first if you are ordering it...

AEPR, line 50: For both of these functions you don't need to do a query - you can just use the baked in JPA repo syntax. You just need to add functions like:

```findAllByIsActiveTrueAndProperty_Address_Uprn(searchUprn: Long)```
and 
```findAllByIsActiveTrueAndRegistrationNumber_Number(searchPRN: Long)```

Happy to talk this through if you like.

Comment (AEPR) [reply src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt:50]: My reading matches Isobel's here - that the UPRN needs to be an exact match but that then you do a normal search.

That said, that strikes me as kind of silly, especially for small numbers - (a) in what world would searching for "211" be a sensible address search? (b) how low do real UPRNs go?

We might want to have some sort of minimum length before we allow fuzzy matching to happen on a string or something. I think that we should be careful doing too much as part of this ticket though - once search exists and is integrated into the website tweaking the behaviour later will be easier and doing it with more realistic data will give better outcomes.

**Review by JasminConterioSW — CHANGES_REQUESTED** [review]
This mostly looks good, what I'm not clear in is quite what they want you to return if someone enters a UPRN!

At the moment you return anything where URPN exactly matches the `searchTerm`, then anything where the address is similar to the `searchTerm`. Should we also be returning properties where the uprn is similar to the `searchTerm`? 

Maybe that gets complicated though, I'm thinking more that they might input a 12 digit number and put one digit wrong, but maybe it's fine that they then don't see anything that matches too closely! You may well have it completely right, I just think it's ambiguous.

File: src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/AddressHelperTests.kt [hunk 4]
```diff
@@ -0,0 +1,47 @@
...
+                    Named.of("a non-numeric string", "not-a-uprn"),
+                    null,
+                ),
+                Arguments.of(
+                    Named.of("an invalid length string (short)", ""),
+                    null,
+                ),
+                Arguments.of(
+                    Named.of("an invalid length string (long)", "1234567890123"),
+                    null,
```
AEPR, line 32: Does a `null` `NamedArgument?` have name "null"?

src/main/resources/data-local.sql
```diff
@@ -202,6 +277,31 @@ VALUES (1, true, 0, 0, 1, 1, 2, 6, 1, 1, '01/15/25'),
        (4, true, 0, 0, 1, 1, 2, 36, 1, 4, '01/15/25'),
        (5, true, 0, 0, 1, 1, 2, 37, 1, 5, '01/15/25'),
        (6, false, 0, 0, 1, 1, 2, 38, 1, 6, '01/15/25'),
-       (7, true, 0, 0, 1, 0, 0, 39, 1, 7, '02/02/25');
+       (7, true, 0, 0, 1, 0, 0, 39, 1, 7, '02/02/25'),
```
JasminConterioSW, line 280: By the way, main now has a property ownership (id 7) with a license...

If you don't want to add null licenses to all your other properties you could miss out po 7 in this insert statement then have another one at the bottom that adds the license id as well!

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/SearchRegisterTests.kt
```diff
@@ -90,9 +90,7 @@ class SearchRegisterTests : IntegrationTest() {
         searchLandlordRegisterPage.searchBar.search("non-matching searchTerm")
         searchLandlordRegisterPage.getPropertySearchLink().click()
 
-        // TODO PRSD-659: Replace with landlord details page assertion
-        assertPageIs(page, ErrorPage::class)
-        assertContains(page.url(), "/search/property")
+        assertPageIs(page, SearchPropertyRegisterPage::class)
```
AEPR, line 93: We should probably write some real integration tests for searching given that it will hit the database in an interesting way. I feel like it might need a litttle careful thought though so I'd be happy with them as a separate PR/ part of the UI PR.

src/test/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/PropertySearchResultViewModelTests.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import org.junit.jupiter.api.Test
+import uk.gov.communities.prsdb.webapp.mockObjects.MockLandlordData.Companion.createPropertyOwnership
+import uk.gov.communities.prsdb.webapp.models.dataModels.RegistrationNumberDataModel
+import kotlin.test.assertEquals
+
+class PropertySearchResultViewModelTests {
+    @Test
+    fun `fromPropertyOwnership returns a corresponding PropertySearchResultViewModel`() {
+        val propertyOwnership = createPropertyOwnership()
```
AEPR, line 11: I'd probably manually invoke `MockLandlordData.createPropertyOwnership()`, to make it clearer where this data is coming from.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/PropertySearchResultViewModel.kt
isobel-softwire: In some instances, like `VerifiedCredentialModel` and `RadiosViewModel`, we've been grouping closely related classes together. `PropertySearchResultLandlordViewModel` only exists to be a part of `PropertySearchResultViewModel` so I think it makes sense here.

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipServiceTests.kt
```diff
@@ -222,4 +223,61 @@ class PropertyOwnershipServiceTests {
             assertEquals(expectedResults, result)
         }
     }
+
+    @Test
+    fun `searchForProperties returns a corresponding list of PropertySearchResultViewModels (PRN search)`() {
```
AEPR, line 228: I think all of these test names should be a little more directly descriptive rather than relying on the implication in brackets. For example:

"searchForProperties returns a single matching property when the search term is a PRN"
"searchForProperties returns a UPRN match ahead of other results when the search term is a UPRN" etc.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
isobel-softwire: Hmmm, I'm not sure what the intended behaviour is. I'll ask around.

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipServiceTests.kt
```diff
@@ -222,4 +223,61 @@ class PropertyOwnershipServiceTests {
...
+        val searchPRN = RegistrationNumberDataModel(RegistrationNumberType.PROPERTY, 123)
+        val expectedPropertyOwnerships = listOf(createPropertyOwnership())
```
AEPR, line 230: I'd rename this second variable `propertyOwnershipsMatchingPRN` or something to make it clearer that you're going to use the mock to return this value - this isn't really the "expected" list since we're just passing it back to the service from the database.

(When I first read this I for a moment assumed that the PRN was going to have to match the MockLandlordData property ownership PRN).

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipServiceTests.kt
```diff
@@ -222,4 +223,61 @@ class PropertyOwnershipServiceTests {
...
+
+    @Test
+    fun `searchForProperties returns a corresponding list of PropertySearchResultViewModels (UPRN search)`() {
+        val searchUPRN = "123"
+
+        val uprnMatchingPropertyOwnership =
+            listOf(createPropertyOwnership(property = createProperty(address = createAddress(uprn = searchUPRN.toLong()))))
+        val fuzzyMatchingPropertyOwnerships = listOf(createPropertyOwnership(), createPropertyOwnership())
+
+        val expectedPropertyOwnerships = uprnMatchingPropertyOwnership + fuzzyMatchingPropertyOwnerships
```
AEPR, line 251: In reference to the comment above, `uprnMatchingPropertyOwnership` and `fuzzyMatchingPropertyOwnerships` are great names and now the `expectedPropertyOwnerships` is fine because it's we're never passing back to the service!

AEPR (changes requested):
Couple of things - very happy to have a call to discuss any if you like!

src/main/resources/data-local.sql
isobel-softwire: Thanks for the heads up!

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
isobel-softwire: The [current lowest UPRN](https://www.ordnancesurvey.co.uk/governance/public-information/search-information-responses/lowest-uprn-not-marked-historic-foi241231#:~:text=The%20lowest%20UPRN%20number%20that%20is%20not%20marked,FindMyAddress%20for%20the%20information%20associated%20to%20UPRN%201.) is 1, believe it or not - I agree that searching '1' is weird though. Probably something to raise in the next dev catch-up

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/SearchRegisterController.kt
```diff
@@ -73,5 +77,18 @@ class SearchRegisterController(
 
     // TODO PRSD-659: implement property search endpoint
```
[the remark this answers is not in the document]
isobel-softwire: I'm going to address this in the next PR for this ticket 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/SearchRegisterController.kt
```diff
@@ -73,5 +77,18 @@ class SearchRegisterController(
...
-    fun searchForProperties() = "error/404"
+    fun searchForProperties(
+        model: Model,
+        searchRequest: PropertySearchRequestModel,
+    ): String {
+        var searchResults: List<PropertySearchResultViewModel>? = null
+
+        if (searchRequest.searchTerm != null) {
+            searchResults = propertyOwnershipService.searchForProperties(searchRequest.searchTerm!!)
+        }
```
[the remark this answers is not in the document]
isobel-softwire: More logic is going to be added to this if-statement in future PRs (see the landlord search endpoint for an idea) so I don't think either of those options would work here. The alternative would be to do 
`if (searchTerm != null) model.add(result) else model.add(null)`
but I don't think that's as readable as having all the model attributes assembled at the end of the method.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
AEPR: Yeah, agree it's worth a discussion - searching for '1' should probably _just_ return that property rather than returning that property AND every property that has 1 in its road number or postcode. That said it will also depend on the exact behviour of the trigram search e.g. what we're setting the `pg_trgm.word_similarity_threshold` to. 

But yeah, the complexity of exactly how this behaves can be kind of cut out of this PR IMO. Let's get the rest of the logic right and we can come back to any limits on fuzzy matching.

isobel-softwire: You're right, I've done them as queries to make adding filtering to them easier later on

src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/AddressHelperTests.kt
isobel-softwire: Yes

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/SearchRegisterController.kt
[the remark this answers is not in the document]
AEPR: > I don't think that's as readable as having all the model attributes assembled at the end of the method.
I agree with this, FWIW (and had considered that as a suggestion). That starts adding complexity all over again as attributes could be overwritten on different paths!

I think that in that case that you should pull this logic out into a separate function and return the search results from that. It's maybe beyond the scope of this PR, but I'd do the same for the landlord search endpoint as well - I'm going to raise a small PR to that effect to show what I mean.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
AEPR: OK, good thinking - can you add a numbered TODO comment or something to indicate that this isn't the expected final state and that the queries are here to make a future ticket easier. :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/AddressHelperTests.kt
AEPR: Nice.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
```diff
@@ -25,4 +27,40 @@ interface PropertyOwnershipRepository : JpaRepository<PropertyOwnership, Long> {
...
+        @Param("searchUPRN") searchUPRN: Long,
+    ): List<PropertyOwnership>
```
rowanhill, line 52: Do these need to be native queries? Could they be derived methods?
```kotlin
fun findByIsActiveTrueAndRegistrationNumber_Number(searchPRN: Long): List<PropertyOwnership>

fun findByIsActiveTrueAndProperty_Address_Uprn(searchUPRN: Long): List<PropertyOwnership>
```

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
isobel-softwire: Update: UPRN searches will result in an exact match or no result (i.e. no fuzzy matching)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
isobel-softwire: [Alex commented something similar](https://github.com/communitiesuk/prsdb-webapp/pull/179#discussion_r1944538619) - I'm going to action his suggestion of adding an explanatory TODO comment here

rowanhill: Aha, sorry for missing that.

JasminConterioSW (approved):

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -89,6 +92,25 @@ class PropertyOwnershipService(
...
+                propertyOwnershipRepository.searchMatchingPRN(prn.number)
+            } else {
+                val uprn = AddressHelper.parseUprnOrNull(searchTerm)
+
+                if (uprn != null) {
+                    propertyOwnershipRepository.searchMatchingUPRN(uprn)
+                } else {
+                    propertyOwnershipRepository.searchMatching(searchTerm)
+                }
+            }
```
AEPR, line 109: I think this would be a little easier to read as a single `if` `else if` `else`, rather than nesting `if`s. You can pull the `uprn` assignment out of the statement altogether.

As a general rule nested if statements can get confusing to follow so IMO it's generally best to either try and avoid them (by combining into a single statement or completely seperating the logic into non-nesting statements) or to abstract the inner if statement away (by creating a function with responsibility for that logic).

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipServiceTests.kt
```diff
@@ -225,4 +226,55 @@ class PropertyOwnershipServiceTests {
...
+        val searchResults = propertyOwnershipService.searchForProperties(searchUPRN)
+
+        assertEquals(expectedSearchResults, searchResults)
+    }
+
+    @Test
+    fun `searchForProperties returns a collection of fuzzy matches when the search term is not a PRN or UPRN`() {
+        val searchTerm = "EG1 2AB"
+
+        val expectedPropertyOwnerships = listOf(createPropertyOwnership(), createPropertyOwnership())
```
AEPR, line 268: Can we rename this value as wel?

AEPR (approved):
Two minor notes - but happy with this now.

PR #184, opened 2025-02-06
Feat/prsd 655 add last updated to property details

JasminConterioSW:
- Make a `lastModifed `fragment from what was in `localAuthorityLandlordDetailsView`
- Use the `lastModified `fragment to add a "Last updated by..." section to the local authority view of the property details page (but not the landlord view)

image

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
@@ -1,5 +1,6 @@
 package uk.gov.communities.prsdb.webapp.controllers
 
+import kotlinx.datetime.toKotlinInstant
 import org.springframework.http.HttpStatus
 import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.stereotype.Controller
@@ -8,6 +9,7 @@ import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.server.ResponseStatusException
+import uk.gov.communities.prsdb.webapp.helpers.DateTimeHelper
 import uk.gov.communities.prsdb.webapp.models.viewModels.PropertyDetailsLandlordViewModel
 import uk.gov.communities.prsdb.webapp.models.viewModels.PropertyDetailsViewModel
 import uk.gov.communities.prsdb.webapp.services.LandlordService
@@ -62,6 +64,12 @@ class PropertyDetailsController(
             throw ResponseStatusException(HttpStatus.BAD_REQUEST, "Property ownership $propertyOwnershipId is inactive")
         }
 
+        val lastModifiedDate =
+            DateTimeHelper.getDateInUK(
+                propertyOwnership.lastModifiedDate?.toKotlinInstant() ?: propertyOwnership.createdDate.toKotlinInstant(),
+            )
+        val lastModifiedBy = propertyOwnership.primaryLandlord.name
+
         val propertyDetails =
             PropertyDetailsViewModel(
                 propertyOwnership = propertyOwnership,
@@ -71,6 +79,8 @@ class PropertyDetailsController(
             )
 
         model.addAttribute("propertyDetails", propertyDetails)
+        model.addAttribute("lastModifiedDate", lastModifiedDate)
+        model.addAttribute("lastModifiedBy", lastModifiedBy)
 
         // TODO PRSD-647: Replace with link to dashboard
         model.addAttribute("backUrl", "/")
--- a/src/main/resources/templates/fragments/lastModified.html
+++ b/src/main/resources/templates/fragments/lastModified.html
@@ -1,0 +1,11 @@
+<th:block th:fragment="lastModified(lastModifiedBy, lastModifiedDate)"
+     th:if="${lastModifiedDate != null}" >
+    <div th:replace="~{fragments/warningInsetText :: warningInsetText(~{:: #last-updated-text})}">
+        <th:block id="last-updated-text">
+            <strong th:text="${lastModifiedBy}">name</strong>
+            <span th:text="' ' + #{lastUpdated.afterName} + ' '">landlordDetails.lastUpdated.afterName</span>
+            <strong th:text="${lastModifiedDate}">lastModifiedDate</strong>
+            <span th:text="#{lastUpdated.afterDate}">landlordDetails.lastUpdated.afterDate</span>
+        </th:block>
+    </div>
+</th:block>
--- a/src/main/resources/messages.properties
+++ b/src/main/resources/messages.properties
@@ -293,11 +293,12 @@ landlordSearch.error.hint.link=search for a property
 landlordSearch.details.heading=Help find a landlord record
 landlordSearch.details.body=If you can't find a landlord record, they may not be registered. Local authorities can request the landlord registers with the database.
 
+lastUpdated.afterName=updated these details on
+lastUpdated.afterDate=.
+
 landlordDetails.title=Landlord record
 landlordDetails.heading=Landlord record
 landlordDetails.removeRecord=Delete record
-landlordDetails.lastUpdated.afterName=updated these details on
-landlordDetails.lastUpdated.afterDate=.
 landlordDetails.personalDetails.heading=Personal details
 landlordDetails.personalDetails.lrn=Landlord registration number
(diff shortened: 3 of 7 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
```diff
@@ -62,6 +64,12 @@ class PropertyDetailsController(
             throw ResponseStatusException(HttpStatus.BAD_REQUEST, "Property ownership $propertyOwnershipId is inactive")
         }
 
+        val lastModifiedDate =
+            DateTimeHelper.getDateInUK(
+                propertyOwnership.lastModifiedDate?.toKotlinInstant() ?: propertyOwnership.createdDate.toKotlinInstant(),
```
isobel-softwire, line 69: We're using this logic in multiple places so I'm wondering if it's worth adding a `getLastModifiedDate(ModifiableAuditableEntity)` method to `DateTimeHelper`

src/main/resources/templates/fragments/lastModified.html
isobel-softwire, line 1: I think the purpose of this fragment would be clearer if it were called `lastModifiedInsetText`

isobel-softwire (approved):
LGTM! Left a few suggestions

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/DateTimeHelper.kt
```diff
@@ -20,5 +21,8 @@ class DateTimeHelper(
             val dateTimeInUK = instant.toLocalDateTime(TimeZone.of("Europe/London"))
             return LocalDate(dateTimeInUK.year, dateTimeInUK.month.value, dateTimeInUK.dayOfMonth)
         }
+
+        fun getLastModifiedDate(entity: ModifiableAuditableEntity): LocalDate =
+            getDateInUK(entity.lastModifiedDate?.toKotlinInstant() ?: entity.createdDate.toKotlinInstant())
```
AEPR, line 26: I don't think this belongs on the DateTime helper - this is about `ModifiableAuditableEntity`s (or perhaps just `AuditableEntity`s). I would suggest that this should be an open function on `AuditableEntity` with an override on `ModifiableAuditableEntity`.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
```diff
@@ -62,6 +63,9 @@ class PropertyDetailsController(
             throw ResponseStatusException(HttpStatus.BAD_REQUEST, "Property ownership $propertyOwnershipId is inactive")
         }
 
+        val lastModifiedDate = DateTimeHelper.getLastModifiedDate(propertyOwnership)
+        val lastModifiedBy = propertyOwnership.primaryLandlord.name
```
AEPR, line 67: Obviously at the moment only the primary landlord can modify properties but I'm a little nervous of us using this value here rather than something that actually comes audit data - it feels like this will be a likely candidate to be missed later when we allow more than just the primary landlord to modify properties.

AEPR (changes requested):

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/DateTimeHelper.kt
AEPR: I wouldn't mess with the getter - I'm not 100% sure what jakarta persistence/the JPA repositories are doing with entities in the background so I think it's probably best to make a new function for this. `getMostRecentlyUpdated()` perhaps?

Or alternatively, we could just set the `lastModifiedDate` to the `createdDate` on initialisation:
```
var lastModifiedDate: Instant? = createdDate
```

isobel-softwire: Setting `lastModifiedDate` to `createdDate` on initialisation sounds good to me!

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
JasminConterioSW: I originally left this out for this reason!
But in [QA](https://mhclgdigital.atlassian.net/browse/PRSD-665?focusedCommentId=646158), Kiran said this came up on the equivalent landlord details ticket (LA view of landlord details) and the decision on that ticket (after discussion with Sam and Travis) was to add the feature and just set it to be the primary landlord.

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/DateTimeHelper.kt
JasminConterioSW: Hmm, it doesn't seem to be setting lastModifiedDate when it saves a new entity, any idea why that would be?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/ModifiableAuditableEntity.kt
```diff
@@ -14,7 +14,7 @@ abstract class ModifiableAuditableEntity :
     Serializable {
     @LastModifiedDate
     @Temporal(TemporalType.TIMESTAMP)
-    @Column(insertable = false)
-    var lastModifiedDate: Instant? = null
+    @Column
```
JasminConterioSW, line 17: Is this change ok - was the `(insertable = false)` there for a particular reason?
We would need `(insertable = true)` to be able to initialise lastModifiedDate to createdDate

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/AuditableEntity.kt
```diff
@@ -16,6 +16,6 @@ abstract class AuditableEntity : Serializable {
     @CreatedDate
     @Temporal(TemporalType.TIMESTAMP)
     @Column(updatable = false)
-    lateinit var createdDate: Instant
```
JasminConterioSW, line 19: Actually I'm a bit worried about initialising this here, it's making tests fail...

In the tests, the value of createdDate is what gets created here, it isn't getting recreated when the property is saved

JasminConterioSW: The real one seems to work though

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
AEPR: OK - let's leave it for now, but I'll bring it up at catch up.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/ModifiableAuditableEntity.kt
AEPR: I think there is a reason for this to be `(insertable = false)` - it states that this service is not responsible for updating this field on any `ModifiableAuditableEntity` (or inheriting entity). We're instead relying on the database to keep it up to date.

I think I've lead us all a little astray on this one. I think it would be better to leave these (as in `created` and `lastModified`) alone and fall back to a method `ModifiableAuditableEntity` that `= lastModifiedDate ?: createdDate`.

AEPR (approved):
I think it isn't worth us messing with the built in auditing in the database - we should just have a combining function on the `ModifiableAuditableEntity`. Once that's back I'm happy for this to merge.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/ModifiableAuditableEntity.kt
Travis-Softwire: Agree with Alex here ^ ^ - I think we're at risk of running into undocumented/ unpredictable behaviour if we're asking Spring to manage these fields in most cases but occasionally trying to set them ourselves

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
Travis-Softwire: Agree that this is going to be a tricky one to resolve in the general case, and that this won't be the end state for how we set the modifiedBy, but fine with this as a stopgap.

PR #185, opened 2025-02-07
PRSD-659: Links property search results to LA view of property record

isobel-softwire:
Links property search results to LA view of property record (instead of landlord view)


Kiran-R-K (approved):
LGTM

PR #188, opened 2025-02-07
PRSD-661: Property Search Pagination

isobel-softwire:
Features:
- Updates `PropertyOwnershipService.searchForProperties` to return a (requested) page
- Adds pagination to property search page

Tests:
- Updates affected `PropertyOwnershipService.searchForProperties` tests
- Tests pagination in 
  - PropertyOwnershipService.searchProperties
  - Property search endpoint
  - Property search page

Screenshot:
image

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/SearchRegisterController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/SearchRegisterController.kt
@@ -51,7 +51,7 @@ class SearchRegisterController(
                 searchRequest.searchTerm!!,
                 principal.name,
                 searchRequest.restrictToLA,
-                currentPageNumber = page - 1,
+                requestedPageIndex = page - 1,
             )
 
         if (isPageOutOfBounds(pagedLandlordList, page)) {
@@ -90,7 +90,7 @@ class SearchRegisterController(
         }
 
         val pagedSearchResults =
-            propertyOwnershipService.searchForProperties(searchRequest.searchTerm!!, currentPageNumber = page - 1)
+            propertyOwnershipService.searchForProperties(searchRequest.searchTerm!!, requestedPageIndex = page - 1)
 
         if (isPageOutOfBounds(pagedSearchResults, page)) {
             return getRedirectForPageOutOfBounds(httpServletRequest)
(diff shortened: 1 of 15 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/SearchRegisterController.kt
```diff
@@ -55,11 +54,7 @@ class SearchRegisterController(
                 currentPageNumber = page - 1,
             )
 
-        if (isPageOutOfBounds(pagedLandlordList, page)) {
-            return "redirect:${
-                URIQueryBuilder.fromHTTPServletRequest(httpServletRequest).removeParam("page").build().toUriString()
-            }"
-        }
+        getRedirectIfPageOutOfBounds(pagedLandlordList, page, httpServletRequest)?.let { redirect -> return redirect }
```
rowanhill, line 57: I might just be set in my ways, but this feels less easy to quickly understand than `if (condition) { return value }`.

What do you think of the below, instead?
```kotlin
if (isPageOutOfBounds(pagedLandlordList, page)) {
    return uriWithoutPageParam(httpServletRequest)
}
```

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -92,17 +95,22 @@ class PropertyOwnershipService(
     fun retrievePropertyOwnershipById(propertyOwnershipId: Long): PropertyOwnership? =
         propertyOwnershipRepository.findByIdAndIsActiveTrue(propertyOwnershipId)
 
-    fun searchForProperties(searchTerm: String): List<PropertySearchResultViewModel> {
+    fun searchForProperties(
+        searchTerm: String,
+        currentPageNumber: Int = 0,
```
rowanhill, line 100: It's conventional for `<something>Number` to be 1-indexed and `<something>Index` to be 0-indexed. So I would suggest this method should take `currentPageIndex`?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/SearchRegisterTests.kt
```diff
@@ -264,5 +264,34 @@ class SearchRegisterTests : IntegrationTest() {
...
             assertPageIs(page, SearchLandlordRegisterPage::class)
         }
+
+        @Test
+        fun `pagination component does not show if there is only one page of results`(page: Page) {
+            val searchPropertyRegisterPage = navigator.goToPropertySearchPage()
+            searchPropertyRegisterPage.searchBar.search("Way")
+
+            val exception = assertThrows<AssertionFailedError> { searchPropertyRegisterPage.getPaginationComponent() }
+            assertContains(exception.message!!, "Expected 1 instance of Locator@.govuk-pagination >> nth=0, found 0")
```
rowanhill, line 274: Hmm, I don't love tying our tests to the exception messages of some third party library. Is there a cleaner way of detecting the pagination component's presence?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/SearchRegisterController.kt
AEPR: Great minds think alike https://github.com/communitiesuk/prsdb-webapp/pull/188#discussion_r1946858978

isobel-softwire: Made this change 👍

PR #189, opened 2025-02-07
Feat/prsd 724 la view landlord detfails on property record

Kiran-R-K:
this PR returns the landlord's details to the property record for the local authority view
I also reformat the PropertyDetailsLandlordViewModel following the PropertyDetailsViewModel pattern and added tests

Uk landlord:
image

International landlord:
image


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
```diff
@@ -55,12 +58,7 @@ class PropertyDetailsController(
         @PathVariable propertyOwnershipId: Long,
         model: Model,
     ): String {
-        val propertyOwnership =
-            propertyOwnershipService.retrievePropertyOwnershipById(propertyOwnershipId)
-                ?: throw ResponseStatusException(HttpStatus.NOT_FOUND, "Property ownership $propertyOwnershipId not found")
-        if (!propertyOwnership.isActive) {
-            throw ResponseStatusException(HttpStatus.BAD_REQUEST, "Property ownership $propertyOwnershipId is inactive")
-        }
+        val propertyOwnership = propertyOwnershipService.getPropertyOwnership(propertyOwnershipId)
```
rowanhill, line 61: As per comments in Slack, I wonder if we couldn't use `getPropertyOwnershipIfAuthorizedUser` here, and expand the concept of 'authorised user' to mean 'LA or owning landlord' rather than just 'owning landlord'.

I think there's a benefit to enforcing access policy at the service level rather than at the controller. If you imagine multiple controllers using the authless `getPropertyOwnership` for LAs, plus some for super users, but then we decide that actually LA users should only be permitted to see info about properties in their LA then we'd have to implement some new service method and then make sure we use it in all the controllers. Whereas if we were using `getPropertyOwnershipIfAuthorizedUser` we could just update the logic of what it means to be an authorized user.

Kiran-R-K: ah, I understand, thanks for explaining.

JasminConterioSW:
Rowan's suggestion sounds good, other than that this all LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -59,12 +59,7 @@ class PropertyOwnershipService(
...
-        val propertyOwnership =
-            retrievePropertyOwnershipById(propertyOwnershipId)
-                ?: throw ResponseStatusException(
-                    HttpStatus.NOT_FOUND,
-                    "Property ownership $propertyOwnershipId not found",
-                )
+        val propertyOwnership = getPropertyOwnership(propertyOwnershipId)
 
         val landlordIdFromPrincipal =
             landlordService.retrieveLandlordIdByBaseUserId(baseUserId)
```
isobel-softwire, line 65: Instead of doing an extra db lookup here, couldn't we compare `propertyOwnership.primaryLandlord.baseUser.id` to `baseUserId`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
Travis-Softwire: Still working through the PRs from last week, but just to add I agree that a service level access policy check is useful for the reasons Rowan has said, but as the controllers stand it'll need to be in addition to controller checks, as landlords shouldn't be able to access `/local-authority/property-details/*` and LA users shouldn't be able to access `/property-details/*`.

We _could_ avoid the duplicate check by refactoring so that we have only one `property-details` endpoint and then select the view based on the user-type, but I'm not keen on that since we have more user-types to come in the future with potentially more divergent sets of permissions on what they can/ can't do on a property-details page.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
JasminConterioSW: I think that should be fine actually, yes

Kiran-R-K: `baseUser.id` is currently private so that would require changing on the `OneLoginUser` entity (unless I've missed something) which I feel a bit unsure about

isobel-softwire: We've make ids public on other entities in order to access them, so I don't think it'll have any side-effects. Maybe ask Travis/Rowan if you want to double check?

Kiran-R-K: Oh, I hadn't realised that, well if we're already doing it I'm happy to change it here too

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/PropertyDetailsLandlordViewModel.kt
```diff
@@ -6,75 +6,62 @@ import uk.gov.communities.prsdb.webapp.database.entity.Landlord
...
-        return residencyIndependentPersonalDetails + residencyPersonalDetails
+    private fun MutableList<SummaryListRowViewModel>.addRow(
+        key: String,
+        value: Any?,
+        changeLink: String? = null,
+        valueUrl: String? = null,
+    ) {
+        val changeLinkOrNull = if (withChangeLinks) changeLink else null
+        add(SummaryListRowViewModel(key, value, changeLinkOrNull, valueUrl))
     }
```
isobel-softwire, line 66: I think this method should be accessible outside of this class, so we don't end up repeating it in other view models. Maybe it should be in a `SummaryListViewModelExtensions` helper?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityUserService.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.communities.prsdb.webapp.services
+
+import org.springframework.stereotype.Service
+import uk.gov.communities.prsdb.webapp.database.repository.LocalAuthorityUserRepository
+
+@Service
+class LocalAuthorityUserService(
+    private val localAuthorityUserRepository: LocalAuthorityUserRepository,
+) {
+    fun getIsLocalAuthorityUser(baseUserId: String): Boolean = localAuthorityUserRepository.findByBaseUser_Id(baseUserId) != null
```
isobel-softwire, line 10: I think this belongs in the `LocalAuthorityDataService`

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -62,22 +62,31 @@ class PropertyOwnershipService(
...
 
-        val landlordIdFromPrincipal =
-            landlordService.retrieveLandlordIdByBaseUserId(baseUserId)
+        val isLocalAuthority = localAuthorityUserService.getIsLocalAuthorityUser(baseUserId)
 
-        if (propertyOwnership.primaryLandlord.id != landlordIdFromPrincipal) {
+        val isPrimaryLandlord = getIsPrimaryLandlord(propertyOwnership, baseUserId)
+
+        if (!isLocalAuthority && !isPrimaryLandlord) {
             throw ResponseStatusException(HttpStatus.NOT_FOUND)
```
isobel-softwire, line 72: It would be worth adding an error message here

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsControllerTests.kt
isobel-softwire, line 1: You could add tests for invalid landlord/LA requests as well

isobel-softwire (approved):
left a few suggestions - apart from that LGTM!

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsControllerTests.kt
Kiran-R-K: I'm inclined not to do that because the validity check of the requests is not happening in the controller anymore it's in the `PropertyOwnershipService` and ~I think we want the tests to fail there and not on the controller if there's a problem - let me know if you think otherwise though?

PR #191, opened 2025-02-10
Feat/prsd 793 update landlord email template

AEPR:
This is missing tests and some tidying up of function names, but I wanted to get a review on the approach of modifying Journey to allow alternate `getPreviousStep` method and alternate `isStepReachable` method.


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordService.kt
```diff
@@ -66,6 +67,18 @@ class LandlordService(
...
+    fun updateLandlordEmailForBaseUserId(
+        baseUserId: String,
+        landlordUpdate: LandlordUpdateModel,
+    ): Landlord {
+        val landlordEntity = retrieveLandlordByBaseUserId(baseUserId)!!
+
+        landlordUpdate.email.ifPresent { landlordEntity.email = it }
+
+        return landlordEntity
+    }
```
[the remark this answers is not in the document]
AEPR: I have tested it and it doe work - just updating the value in a `@Transactional` function is enough. Save is needed to get the data framework to start tracking an object, but if we retrieve it from the database it is already tracked.

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/UpdateDetailsJourneyDataExtensions.kt
```diff
@@ -0,0 +1,18 @@
...
+        Optional.of(
+            JourneyDataHelper.getFieldStringValue(
+                this,
+                UpdateDetailsStepId.UpdateEmail.urlPathSegment,
+                "emailAddress",
+            )!!,
+        )
+    } else {
+        Optional.empty()
+    }
```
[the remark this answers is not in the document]
AEPR: I agree that we should have a discussion of extensions vs helpers. I prefer extensions but did slightly include this here as a conversational prompt 😅 .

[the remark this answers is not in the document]
AEPR: I think in that case I would be better off just using a nullable type since it's more idiomatic Kotlin. And actually that's probably what I should do for now.

The potential problem that can come up is - what do we do in the future to distinguish between "no update" and "update to `null`". The idea here is that `Optional.empty()` = no update and that an `Optional<String?>` could update to null.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateDetailsJourney.kt
```diff
@@ -0,0 +1,93 @@
...
+        targetStep: Step<UpdateDetailsStepId>,
+        targetSubPageNumber: Int?,
+    ): StepDetails<UpdateDetailsStepId>? {
+        // This stores journeyData for only the journey path the user is on
+        // and excludes user data for pages in the journey that belong to a different path
+        val stepData = JourneyDataHelper.getPageData(journeyData, targetStep.name)
+        val filteredJourneyData = mutableMapOf<String, Any?>(targetStep.name to stepData)
+
+        return StepDetails(viewDetailsStep, null, filteredJourneyData)
+    }
```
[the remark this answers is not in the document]
AEPR: Now defunct (see update), but yes this 100% does not work for any multi-step updates. My intention was to implement it in a way that it would OBVIOUSLY not work for such, so whoever is doing that work would have to update it.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateDetailsJourney.kt
```diff
@@ -0,0 +1,93 @@
...
+        validator = validator,
+        journeyDataService = journeyDataService,
+    ) {
+    override val initialStepId = UpdateDetailsStepId.ReturnToDetails
+
+    private val viewDetailsStep =
+        Step(
+            id = UpdateDetailsStepId.ReturnToDetails,
+            page = Page(NoInputFormModel::class, "redirect:/${UpdateDetailsStepId.ReturnToDetails}", mapOf()),
+        )
```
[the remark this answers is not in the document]
AEPR: Also defunct, but I've decided to just step pages that should never be reached to internal error pages. The difference with the verify page (IIRC) is that the verify is different because when you `GET` the verify endpoint, we treat that as a post to the same, meaning that success looks like redirecting to the next step and a validation failure could in theory return the page.

In this case there should be no controller routing that can return this page so I think just returning an error is fine. 

Returning a self-redirect is certainly a terrible idea because if it ever did get reached it results in an infinite redirect loop.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordService.kt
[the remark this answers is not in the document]
isobel-softwire: Ah ok - good to know!

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/UpdateDetailsJourneyDataExtensions.kt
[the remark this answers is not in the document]
isobel-softwire: Actually I think using `Optional<String?>` makes sense then. Maybe there should be a `JourneyDataHelper.getOptionalFieldStringValue()` method so the logic won't have to be repeated in all the update methods.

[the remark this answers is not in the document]
AEPR: Having had a look I've seen some guidance that suggests avoiding `Optional` in Kotlin as `?` should be replacing it where possible. This is a use case that *might* call for it at some point (but I think only if updating from `nonEnglandOrWales` resident for which see below) but if it doesn't I'd prefer to avoid adding more complexity than needed.

For the `nonEnglandOrWalesAddress` updating to `null`, that is fully determined by the `countryOfResidence` field. I think we'd be better off updating based on such. That is to say if we update `countryOfResidence` to "England or Wales" then we can set `nonEnglandOrWalesAddress` to `null` and in any other case we treat `nonEnglandOrWalesAddress` as though it is a non-nullable value in the database.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
```diff
@@ -23,18 +27,35 @@ class LandlordDetailsController(
     val landlordService: LandlordService,
     val addressDataService: AddressDataService,
     val propertyOwnershipService: PropertyOwnershipService,
+    val updateDetailsJourney: UpdateLandlordDetailsJourney,
 ) {
+    @PreAuthorize("hasRole('LANDLORD')")
+    @GetMapping("update/details")
```
isobel-softwire, line 33: `landlord-details/update/details` feels a bit clunky. Could this just be `landlord-details/update/`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
```diff
@@ -49,6 +70,35 @@ class LandlordDetailsController(
         return "landlordDetailsView"
     }
 
+    @PreAuthorize("hasRole('LANDLORD')")
+    @GetMapping("update/{stepName}")
+    fun getUpdateEmail(
```
isobel-softwire, line 75: Reminder to update this function name (and the post request one)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateLandlordDetailsJourney.kt
```diff
@@ -0,0 +1,130 @@
...
+class UpdateLandlordDetailsJourney(
+    validator: Validator,
+    journeyDataService: JourneyDataService,
+    val landlordService: LandlordService,
+) : Journey<UpdateDetailsStepId>(
+        journeyType = JourneyType.UPDATE_DETAILS,
+        validator = validator,
+        journeyDataService = journeyDataService,
+    ) {
+    override val initialStepId = UpdateDetailsStepId.UpdateEmail
```
isobel-softwire, line 30: How will this change as we add more update pages?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateLandlordDetailsJourney.kt
```diff
@@ -0,0 +1,130 @@
...
+
+    private val updateSessionStep =
+        Step(
+            id = UpdateDetailsStepId.ChangeDetailsSession,
+            page = Page(NoInputFormModel::class, "error/500", mapOf()),
+            handleSubmitAndRedirect = { journeyData, _ -> submitAllChanges(journeyData) },
```
isobel-softwire, line 36: Will this `handleSubmitAndRedirect` ever get called? This page isn't supposed to be reached right?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateLandlordDetailsJourney.kt
```diff
@@ -0,0 +1,130 @@
...
+                        mapOf(
+                            "title" to "forms.update.title",
+                            "fieldSetHeading" to "forms.update.email.fieldSetHeading",
+                            "fieldSetHint" to "forms.email.fieldSetHint",
+                            "label" to "forms.email.label",
+                            "submitButtonText" to "forms.buttons.continue",
+                        ),
+                ),
+            handleSubmitAndRedirect = { _, _ -> getRedirectToUpdateSessionPage() },
+            nextAction = { _, _ -> Pair(UpdateDetailsStepId.ChangeDetailsSession, null) },
```
isobel-softwire, line 56: Is there a point to setting a `nextAction` here?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateLandlordDetailsJourney.kt
```diff
@@ -0,0 +1,130 @@
...
+            )
+
+        landlordService.updateLandlordForBaseUserId(
+            SecurityContextHolder.getContext().authentication.name,
+            landlordUpdate,
+        )
+
+        journeyDataService.clearJourneyDataFromSession()
+
+        return "/landlord-details"
```
isobel-softwire, line 107: `/landlord-details` should probably be a constant

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/steps/UpdateDetailsStepId.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.communities.prsdb.webapp.forms.steps
+
+enum class UpdateDetailsStepId(
+    override val urlPathSegment: String,
+) : StepId {
+    UpdateEmail("email"),
+    ChangeDetailsSession("details"),
```
isobel-softwire, line 7: I'm not sure about `ChangeDetailsSession` as a name. I think something like `UpdateDetails`/`UpdateAll`/`Update` would work better

src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/UpdateDetailsJourneyDataExtensionsTests.kt
```diff
@@ -0,0 +1,36 @@
+package uk.gov.communities.prsdb.webapp.helpers
+
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.mockito.Mockito.mock
+import uk.gov.communities.prsdb.webapp.helpers.UpdateLandlordDetailsJourneyDataHelper.Companion.getEmailUpdateIfPresent
+import uk.gov.communities.prsdb.webapp.mockObjects.JourneyDataBuilder
+import kotlin.test.assertEquals
+
+class UpdateDetailsJourneyDataExtensionsTests {
```
isobel-softwire, line 10: I know you're changing the helpers to extensions anyway, but this name doesn't match the current implementation

src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/UpdateDetailsJourneyDataExtensionsTests.kt
```diff
@@ -0,0 +1,36 @@
...
+    fun setup() {
+        journeyDataBuilder = JourneyDataBuilder(mock(), mock())
+    }
+
+    @Test
+    fun `getEmailUpdateIfPresent returns an optional email if the email page is in journeyData`() {
+        val newEmail = "new email address value"
+        val testJourneyData = journeyDataBuilder.withEmailAddressUpdate(newEmail).build()
+
+        val emailUpdate = getEmailUpdateIfPresent(testJourneyData)
```
isobel-softwire, line 23: I think the helper methods should be qualified

src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/UpdateDetailsJourneyDataExtensionsTests.kt
```diff
@@ -0,0 +1,36 @@
...
+class UpdateDetailsJourneyDataExtensionsTests {
+    private lateinit var journeyDataBuilder: JourneyDataBuilder
+
+    @BeforeEach
+    fun setup() {
+        journeyDataBuilder = JourneyDataBuilder(mock(), mock())
+    }
+
+    @Test
+    fun `getEmailUpdateIfPresent returns an optional email if the email page is in journeyData`() {
```
isobel-softwire, line 19: These test names haven't been updated :)

isobel-softwire (changes requested):
From what I've understood, this is looking good so far. I've left a lot of questions, mostly because I'm not completely sure what the intended functionality is! It would be really helpful if you could document what this PR aims to do (+ add screenshots of the pages you've created).

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -151,6 +150,31 @@ abstract class Journey<T : StepId>(
...
+
+            val (nextStepId, nextSubPageNumber) = currentStepDetails.step.nextAction(journeyData, currentStepDetails.subPageNumber)
+            currentStepDetails =
+                nextStepId?.let {
+                    val nextStep = steps.single { step -> step.id == nextStepId }
+                    StepDetails(nextStep, nextSubPageNumber, currentStepDetails!!.filteredJourneyData)
+                }
+        }
+
+        return null
```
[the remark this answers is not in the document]
AEPR: Yes, that's the idea.

src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/UpdateDetailsJourneyDataExtensionsTests.kt
AEPR: Yep, that's just me being lazy - I'll make it consistent on this one and make the change a single PR.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateLandlordDetailsJourney.kt
AEPR: This is going to stay as the initial step in the journey - I've pushed the name change to this PR to show how it will work with more options.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateLandlordDetailsJourney.kt
AEPR: Yes - I've added a submit button on the `/update/details` page (with a TODO to make it clear it's not permanent) so that we can use and test the database update.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateLandlordDetailsJourney.kt
AEPR: Yes - it's necessary for the holistic validation that happens - it means that jumping around in the journey won't allow inconsistent data.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/steps/UpdateDetailsStepId.kt
AEPR: I think there is value in calling this something session, because this IS meant to be the session page. Maybe `UpdateSession`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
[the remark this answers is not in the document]
isobel-softwire: I don't think the method's title suggests that, but I can't think of a more intuitive name

AEPR:
This creates a new page that represents a session of updating landlord details. From that page you can access the individual details and update them. Currently the only available pages are the name and email updates (also added by this PR). The actual pages used are just recycled from the landlord registration journey with new headings! 

We are using the Journey class is a mildly unauthodox way because it closely corresponds to the desired behaviour. I'll explain the desired flow and then how it's implemented here.
Flow:
* On the "Session" page you see your landlord details, and click a link to update a particular detail
* When you successfully submit on that page (if it is independent) you are directed back to the "Session" page
* You can submit on the session page (currently this updates the database, but there will be a declaration)
* If you update a value that has other dependent data (i.e. update to or from England or Wales Landlord), you are required to complete the other details attached to you choice (e.g. adding an international address and UK contact address)
* If you navigate to another page in the journey while ignoring these additional steps, you are redirected to the page you should be filling in

This is implemented:
* This is a journey that (when evaluating validity of the journey data overall) starts with the journey data pre-populated from the database. This is done by saving a copy of the landlord data in the journey data when first arriving on the session page.
* The `nextAction`s of the steps take you through the journey as normal, ensuring that `getPrevStep` is able to validate all the data in the journey up to your current page
* The `handleSubmitAndRedirect` on each update page that is independant (or has no dependants) redirects back to the "Session" page
* There is a new `getUnreachableStepRedirect()` method that calculates where the journey should redirect based on the journeyData, replacing the previous property. This allows the Update Journey to redirect to the last reachable step of the journey (which if there is a problem with the data due a user manually navigating will allow them to fix it). If the landlord data has not been loaded from the database the redirect is instead to the session page to start a session!
* I have added a temporary submit button on the landlord-details page (hidden unless an attribute is set), with a TODO to remove when the session and declaration pages are added.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
```diff
@@ -18,23 +23,42 @@ import uk.gov.communities.prsdb.webapp.services.PropertyOwnershipService
...
-@RequestMapping("/landlord-details")
+@RequestMapping(LandlordDetailsController.LANDLORD_DETAILS_ROUTE)
 class LandlordDetailsController(
     val landlordService: LandlordService,
     val addressDataService: AddressDataService,
     val propertyOwnershipService: PropertyOwnershipService,
+    val updateDetailsJourney: UpdateLandlordDetailsJourney,
 ) {
+    @PreAuthorize("hasRole('LANDLORD')")
+    @GetMapping("$UPDATE_PATH_SEGMENT/details")
```
Travis-Softwire, line 34: Minor, but I think we're relying on the string `details` here matching the one in the urlPathSegment for `UpdateDetailsStepId.ChangeDetailsSession` - should we be using a constant?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -137,14 +136,15 @@ abstract class Journey<T : StepId>(
...
+    protected open fun getPrevStep(
         journeyData: JourneyData,
         targetStep: Step<T>,
         targetSubPageNumber: Int?,
     ): StepDetails<T>? {
+        if (targetStep.id == initialStepId && targetSubPageNumber == null) return null
         val initialStep = steps.singleOrNull { step -> step.id == initialStepId } ?: return null
         var currentStep = initialStep
-        var prevStep: Step<T>? = null
+        lateinit var prevStep: Step<T>
```
Travis-Softwire, line 147: Will this not cause an error if we never enter the for loop?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -168,6 +167,31 @@ abstract class Journey<T : StepId>(
         return StepDetails(prevStep, prevSubPageNumber, filteredJourneyData)
     }
 
+    protected fun getLastReachableStep(journeyData: JourneyData): StepDetails<T>? {
```
Travis-Softwire, line 170: Intuitively feels like there must be a way to refactor into a single function between `getPrevStep` and this function, or at least commonise the internal logic? Happy for that to be a separate refactor ticket though

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateLandlordDetailsJourney.kt
Travis-Softwire: Ah nice - never really thought about how a page would look if we added a specific GET endpoint for it but not a POST, this makes sense.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateLandlordDetailsJourney.kt
```diff
@@ -0,0 +1,162 @@
...
+            saveAfterSubmit = false,
+        )
+
+    override val sections =
+        createSingleSectionWithSingleTaskFromSteps(
+            initialStepId,
+            setOf(
+                emailStep,
+                fullNameStep,
+                updateSessionStep,
```
Travis-Softwire, line 92: V. minor, but wondering if we need an explanatory comment here to warn future devs that this needs to stay as the last step in the journey, as I'm not sure it's as intuitive as for the journeys where the user progresses through them linearly

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
AEPR: Yeah, I agree but couldn't think of a great way to do it immediately. Happy to make a sub ticket for it.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
AEPR: It would do, if line 144 didn't make sure that we're going to enter the loop!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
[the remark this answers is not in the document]
AEPR: Do you think it should return the last step in that case instead?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/steps/UpdateDetailsStepId.kt
AEPR: Sure, I can just dodge calling it a session - we can come up with separate terminology later.

AEPR (commented):
Making a couple if small fixes as noted.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
Travis-Softwire: Ah - a very good point!

isobel-softwire (approved):
LGTM!

PR #193, opened 2025-02-10
Feat/prsd 858 implement section headers on property registration journey

JasminConterioSW:
This starts the front end changes for prsd-858

- Add a (nullable) section header fragment 
- Add section header to question pages (set to null for almost all of them)
- Add non-null section header to the lookup address page for the property registration journey with hardcoded values

image
image


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
```diff
@@ -125,6 +126,8 @@ class PropertyRegistrationJourney(
                             "houseNameOrNumberLabel" to "forms.lookupAddress.houseNameOrNumber.label",
                             "houseNameOrNumberHint" to "forms.lookupAddress.houseNameOrNumber.hint",
                             "submitButtonText" to "forms.buttons.saveAndContinue",
+                            "sectionHeaderInfo" to
+                                SectionHeaderViewModel("registerProperty.taskList.register.heading", 1, 3),
```
JasminConterioSW, line 130: Hardcoded for now - getting this value (here and for all the other steps) will come in a later PR

src/main/resources/templates/fragments/sectionHeader.html
```diff
@@ -0,0 +1,6 @@
+<p th:fragment="sectionHeader(sectionHeaderInfo)"
+   th:if="${sectionHeaderInfo} != null"
+    class="govuk-hint govuk-!-margin-top-0">
+    <span th:text="#{sectionHeader.sectionXofY(${sectionHeaderInfo.sectionNumber}, ${sectionHeaderInfo.totalSections})}">Section X of Y</span>
+    <span th:text="#{${sectionHeaderInfo.sectionNameKey}}"></span>
```
isobel-softwire, line 5: Do these need to be separate? I.e., could you have 
```
#{sectionHeader(
    ${sectionHeaderInfo.sectionNumber}, 
    ${sectionHeaderInfo.totalSections}, 
    #{${sectionHeaderInfo.sectionNameKey}}
)}
```

isobel-softwire (approved):
LGTM - left one small Q

src/main/resources/templates/fragments/sectionHeader.html
JasminConterioSW: Actually yes, the text can all go into the <p> and we don't need the spans...

PR #194, opened 2025-02-10
PRSD-600: Give tasks a list of constituent steps

AEPR:
Changes how the task list tracks which steps are in a task. The reason for this change is (a) it's neater to have task have a set of steps that count as their members and (b) it will help with determining with section a page is in so we can display it.


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/tasks/CheckAndSubmitPropertiesTaskList.kt
```diff
@@ -20,7 +20,6 @@ class CheckAndSubmitPropertiesTaskList(
             Task(
                 "registerProperty.taskList.checkAndSubmit.checkAnswers",
                 RegisterPropertyStepId.CheckAnswers,
```
JasminConterioSW, line 22: I think when there is only one step, we should still include the list of stepIds (albeit with only one entry) for consistency.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/tasks/TaskList.kt
```diff
@@ -71,6 +67,6 @@ abstract class TaskList<T : StepId>(
     data class Task<T : StepId>(
         val nameKey: String,
         val startId: T,
```
JasminConterioSW, line 69: Calling this `startingStepId `feels clearer to me, but that one might be a matter of opinion!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/tasks/CheckAndSubmitPropertiesTaskList.kt
isobel-softwire: Maybe the last argument should be `otherStepIds` with the full set instantiated in the class? Then not passing a third argument when there's only one step would feel less inconsistent.

isobel-softwire (approved):
LGTM - left a thought on Jasmin's comment

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/tasks/TaskList.kt
AEPR: No - I agree especially now that the list is `stepIds`.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/tasks/CheckAndSubmitPropertiesTaskList.kt
AEPR: I'm going to make a companion function `Task.withOneStep()` - that way it's clearer what's going on in the listing but without having to repeat the list, and without adding extra logic to combine them.

PR #196, opened 2025-02-11
PRSD-660: Fixes Search Page Style

isobel-softwire:
- Removes unnecessary margins

Screenshots:
- With filter panel closed
image

- With filter panel open
image


rowanhill:
Can I ask what's driving the scrollbar change? I'm surprised we'd need to (or want to) tweak the behaviour of a component that's been through thorough accessibility testing.

isobel-softwire:
> Can I ask what's driving the scrollbar change? I'm surprised we'd need to (or want to) tweak the behaviour of a component that's been through thorough accessibility testing.

In QA, a comment was made about hiding the scrollbar if there's nothing to scroll.

AEPR:
I agree that if possible we should avoid messing with css.

Having tried it, I don't believe we need the side scrolling panel at all - I think the basic gov.uk table already does side scrolling on a small enough width.

Travis-Softwire:
> I agree that if possible we should avoid messing with css.
> 
> Having tried it, I don't believe we need the side scrolling panel at all - I think the basic gov.uk table already does side scrolling on a small enough width.

I think the MoJ scrollable pane class does a few extra things for us unfortunately, like adding a shadow to hint that there is content to see by scrolling horizontally. My view would be that we leave it as-is since it's just the behaviour of this component, and if Product/ Service owners are concerned about it we can raise an issue in the github for the design system

isobel-softwire:
Ok, I've reverted the scrollbar change

JasminConterioSW:
Could you add a quick screenshot of what it looks like with the updated margins please? Thanks!

isobel-softwire:
> Could you add a quick screenshot of what it looks like with the updated margins please? Thanks!

Added 👍

JasminConterioSW:
Do we still want the margin change if we are keeping the scrollbar, or does it look better with the extra margin?
This is with the margin:
image

isobel-softwire:
I don't think the change is needed anymore - I'll close this PR

PR #197, opened 2025-02-11
PRSD-600: Fix incorrect task completion logic and tests

AEPR:
(a) Fixes the bug that came up in internal demo earlier today
(b) Fixes the rather egregious fact that both my integration tests that should have caught that issue were non functional (such that they always passed)


Travis-Softwire (approved):
LGTM :) (Presumably the cause of the bug was that for tasks with branching steps the previous check was failing since the non-chosen branch would not be complete?)

AEPR:
> Presumably the cause of the bug was that for tasks with branching steps the previous check was failing since the non-chosen branch would not be complete?

Yeah, that's exactly right - I previously changed this last minute (in a separate PR) to allow the task lists to be used in determining which section a step is in.

PR #201, opened 2025-02-12
Feat/prsd 632 ls user registration set save after submit to false

Kiran-R-K:
- For all the steps in `LaUserRegistrationJourney` set `saveAfterSubmit` to `false`
- Added a test to `JourneyTests` to ensure that when `saveAfterSubmit` is set to `false` the `JourneyDataService.saveJourneyData()` method isn't called

### DIFF
--- a/src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/JourneyTests.kt
+++ b/src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/JourneyTests.kt
@@ -17,6 +17,7 @@ import org.mockito.Mockito.mock
 import org.mockito.junit.jupiter.MockitoExtension
 import org.mockito.kotlin.argumentCaptor
 import org.mockito.kotlin.eq
+import org.mockito.kotlin.never
 import org.mockito.kotlin.spy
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/JourneyTests.kt
```diff
@@ -674,6 +675,67 @@ class JourneyTests {
             assertEquals("testPropertyValue", resultPageData?.get("testProperty"))
         }
 
+        @Test
+        fun `saves the journey data when saveAfterSubmit is set to false on the step`() {
```
Travis-Softwire, line 679: Should this be 'does not save the journey data when...'?

Travis-Softwire (approved):

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/JourneyTests.kt
Kiran-R-K: yup, thanks

PR #202, opened 2025-02-13
Fix/prsd 600 unreachable in progress tasks

AEPR:
Under the current schema it is possible for all the steps in a task to be complete, but for it to still be in status "Cannot start yet", if a previous task has gone from being complete to not being complete. This is possible if you alter an earlier answer to change the route through that task (for example if you change the answer to "Do you have tennants?" from "No" to "Yes", but haven't yet given the tennancy details).

Both this and the previous "In progress" bug were caused by the behaviour of multi-path tasks, so I've added a section of tests for those. I've tried cherry-picking these tests onto earlier commits to prove they fail when the bugs are present.

### DIFF
--- a/src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/TaskListTests.kt
+++ b/src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/TaskListTests.kt
@@ -5,10 +5,9 @@ import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Nested
 import org.mockito.Mock
 import org.mockito.Mockito.mock
-import org.mockito.kotlin.any
-import org.mockito.kotlin.isNull
 import org.mockito.kotlin.whenever
 import org.springframework.validation.Validator
+import uk.gov.communities.prsdb.webapp.constants.enums.JourneyType
 import uk.gov.communities.prsdb.webapp.forms.journeys.Journey
 import uk.gov.communities.prsdb.webapp.forms.steps.Step
 import uk.gov.communities.prsdb.webapp.forms.steps.StepId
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/TaskListTests.kt
```diff
@@ -20,114 +19,273 @@ class TaskListTests {
...
+                mutableMapOf(
+                    "step1" to mutableMapOf<String, String>(),
+                    "step2a" to mutableMapOf<String, String>(),
+                    "step2b" to mutableMapOf<String, String>(),
+                    "step3" to mutableMapOf<String, String>(),
+                    "step4start" to mutableMapOf<String, String>(),
+                    "step4path1" to mutableMapOf<String, String>(),
+                    "step4path2a" to mutableMapOf<String, String>(),
+                    "step4path2b" to mutableMapOf<String, String>(),
                 ),
```
rowanhill, line 73: Just a stylistic thing, but I think you could rewrite this a little terser as:

```kotlin
listOf(
    "step1", "step2a", "step2b", "step3",
    "step4start", "step4path1", "step4path2a", "step4path2b"
).associateWith { mutableMapOf<String, String>() }.toMutableMap()
```

Or even:

```kotlin
TestStepId.entries.associate { it.urlPathSegment to mutableMapOf<String, String>() }.toMutableMap()
```

rowanhill (approved):

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/TaskListTests.kt
```diff
@@ -20,114 +19,273 @@ class TaskListTests {
...
-                Step(
-                    TestStepId.StepTwo,
-                    mock(),
-                    isSatisfied = { _, _ -> stepTwoCompleted },
-                    nextAction = { _, _ -> Pair(TestStepId.StepThree, null) },
+            )
+        }
+
+        fun getMultiPathTestTaskListWithMainlineCompleted(useMainline: Boolean): TestTaskList {
```
isobel-softwire, line 77: Is the mainline completed in the returned tasklist because of the mocking in `setup()`?

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/TaskListTests.kt
```diff
@@ -20,114 +19,273 @@ class TaskListTests {
...
-                    TestStepId.StepThree,
-                    mock(),
-                    isSatisfied = { _, _ -> stepThreeCompleted },
-                    nextAction = { _, _ -> Pair(TestStepId.StepFour, null) },
+            )
+        }
+
+        @Suppress("ktlint:standard:max-line-length")
+        @Test
+        fun `when the mainline of a journey is followed and the journey is using the mainline, all filled in tasks tasks show as completed`() {
```
JasminConterioSW, line 170: To me "when the mainline of a journey is followed" and "the journey is using the mainline" sound like they are saying the same thing.

Maybe just `when the journey is using the mainline route, all filled in tasks show as completed`?

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/TaskListTests.kt
JasminConterioSW: I think it's the "isSatisfied" on the steps in the journey

AEPR: Yeah, that's right.

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/TaskListTests.kt
AEPR: `when a multi route task is completed along a one route, that task and subsequent filled in tasks show as completed`

PR #203, opened 2025-02-13
feat/PRSD-785 Align licensing terminology across screens

Kiran-R-K:
On the Landlord Record page in the property details tab - changed some of the displayed names for `LicencingType`, by using the `MessageKeyConverter` in `RegisteredPropertyViewModel` so that the names are unified across all screens.

Changes:
HMO licence -> HMO mandatory licence
Additional licence -> HMO additional licence
Not Licenced -> None

image


JasminConterioSW (approved):
Maybe rename the variable to make it clearer that it's a message key but other that that LGTM :)

PR #204, opened 2025-02-14
PRSD-663: Property Search License Filter

isobel-softwire:
Features:
- Adds license filter to `PropertyOwnershipRepository` search methods
- Adds license filter to `PropertySearchRequestModel` and `PropertyFilterPanelViewModel`
- Adds license filter functionality to property search endpoint

Tests:
- Adds license filter to property search integration tests
- Updates seed data (local and search)

Screenshots:
- No licences selected
image

- Some licenses selected
image


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
```diff
@@ -37,14 +38,16 @@ interface PropertyOwnershipRepository : JpaRepository<PropertyOwnership, Long> {
...
+            FILTERS,
         nativeQuery = true,
     )
     fun searchMatchingPRN(
         @Param("searchPRN") searchPRN: Long,
         @Param("laUserBaseId") laUserBaseId: String,
         @Param("restrictToLA") restrictToLA: Boolean = false,
+        @Param("restrictToLicenses") restrictToLicenses: Collection<LicensingType> = LicensingType.entries,
         pageable: Pageable,
+        @Param("noLicenseType") noLicenseType: LicensingType = LicensingType.NO_LICENSING,
```
rowanhill, line 50: If I'm understanding correctly, `noLicenseType` is only being passed here so you can annotate it as a `@Param` and then use it in the query string - and in fact, it'd break things if any value other than `NO_LICENSING` was passed. Is that right? If so, it's a bit of a shame.

Are you sure it's not possible to just directly reference the value in `LICENSE_FILTER` somehow? This SO question has a few answers with suggestions, though I've not tested any: https://stackoverflow.com/questions/18402988/is-there-a-way-to-use-constants-inside-spring-data-query-annotation-value

Failing that, I at least think we should add a comment to each of these methods explaining what's going on.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/SearchRegisterTests.kt
```diff
@@ -319,24 +329,34 @@ class SearchRegisterTests : IntegrationTest() {
             laFilter.checkCheckbox("true")
             filter.clickApplyFiltersButton()
 
-            val laFilterSelectedHeadingText = filter.getSelectedHeadings(expectedCount = 1).first().innerText()
+            val laFilterSelectedHeadingText = filter.getSelectedHeadings(expectedCount = 1)[0].innerText()
```
rowanhill, line 332: Out of curiosity, what was wrong with `first()`?

rowanhill (changes requested):
Broadly looks good to me, just one question about how noLicenseType

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/SearchRegisterTests.kt
isobel-softwire: Nothing - I changed it to match `filter.getSelectedHeadings(expectedCount = 2)[1]` later in the test for consistency

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
isobel-softwire: We can reference it using its fully qualified name - i.e. 
`#{T(uk.gov.communities.prsdb.webapp.constants.enums.LicensingType).NO_LICENSING}`
Made this change👍

rowanhill: Nice!

PR #205, opened 2025-02-14
PRSD-663: Manually Bind Filter Request to View

isobel-softwire:
- Manually determines the state of each filter checkbox rather than letting Spring do it. This is done to avoid the hidden values [Spring+Thymeleaf](https://www.thymeleaf.org/doc/tutorials/3.1/thymeleafspring.html#checkbox-fields) adds to checkboxes.
- Makes `SearchRequestModel` filter properties nullable so filters only appear in URLs if they're active.

### DIFF
--- a/src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/URIQueryBuilderTests.kt
+++ b/src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/URIQueryBuilderTests.kt
@@ -153,7 +153,7 @@ class URIQueryBuilderTests {
 
     @Test
     fun `removeParamValue returns the updated URI (multi-value param)`() {
-        mockHTTPServletRequest.queryString = "name1=value1a&name1=value1b&&name2=value2"
+        mockHTTPServletRequest.queryString = "name1=value1a&name1=value1b&name2=value2"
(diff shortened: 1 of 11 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/URIQueryBuilderTests.kt
```diff
@@ -153,8 +153,8 @@ class URIQueryBuilderTests {
 
     @Test
     fun `removeParamValue returns the updated URI (multi-value param)`() {
-        mockHTTPServletRequest.queryString = "name1=value1a&name1=value1b&_name1=hiddenValue1&name2=value2"
-        val expectedUpdatedURI = "${mockHTTPServletRequest.requestURI}?_name1=hiddenValue1&name2=value2&name1=value1b"
+        mockHTTPServletRequest.queryString = "name1=value1a&name1=value1b&&name2=value2"
```
rowanhill, line 156: There's an extra `&` in here.

rowanhill (approved):

src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/URIQueryBuilderTests.kt
isobel-softwire: Fixed 👍

PR #207, opened 2025-02-17
Chore/prsd none generate la table migration

Travis-Softwire:
Changing the LA data migration to no longer use the 'COPY' command as RDS doesn't support it. Also adding a quick script to generate an UPSERT migration for any future changes to the csv


isobel-softwire (approved):
LGTM :) - the 'Updating Local Authority Data' section of the ReadMe should be updated to match this

PR #208, opened 2025-02-17
PRSD-878: Remove Details from 1L User

isobel-softwire:
- Removes name and email from `OneLoginUser` table
- Updates affected tests


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OneLoginUser.kt
```diff
@@ -1,23 +1,9 @@
...
-    @Column(nullable = false)
-    lateinit var email: String
-        private set
-
-    constructor(id: String, name: String, email: String) : this(id) {
-        this.name = name
-        this.email = email
-    }
-}
+) : ModifiableAuditableEntity()
```
rowanhill, line 9: Is this still modifiable if it's only column is an ID? 🤔

rowanhill (commented):
So, we're doing this because we want each user type (landlord, LA, etc) to have their own independent name and email, even if owned by the same 1L user? And we're happy with the implication that someone with multiple accounts would have to change their name in multiple places?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OneLoginUser.kt
isobel-softwire: I guess not - changed this 😄

AEPR:
I guess with name (and date of birth, actually) that's something that might be coming from one login if the user has a verified identity. I appreciate that the point here is that people might have different legal and proffessional names (e.g. someone has changed their name legally but keeps using their former name at work), it does feel like allowing these to be different should be to support functionality that's come from service design.

Also, just looking at PRSD-878 there's nothing on there that mentions that this is removing names from OneLogin users.

I do agree that we shouldn't be duplicating this data though - if it live on e.g. Landlord and Local Authority User then it shouldn't also live on One Login User (unless those mean different things!).

isobel-softwire:
> I guess with name (and date of birth, actually) that's something that might be coming from one login if the user has a verified identity. I appreciate that the point here is that people might have different legal and proffessional names (e.g. someone has changed their name legally but keeps using their former name at work), it does feel like allowing these to be different should be to support functionality that's come from service design.
> 
> Also, just looking at PRSD-878 there's nothing on there that mentions that this is removing names from OneLogin users.
> 
> I do agree that we shouldn't be duplicating this data though - if it live on e.g. Landlord and Local Authority User then it shouldn't also live on One Login User (unless those mean different things!).

It's not specifically part of PRSD-878 - I don't think it has its own ticket either. @Travis-Softwire and I agreed that it should be done under this ticket as it's related.

Travis-Softwire:
> And we're happy with the implication that someone with multiple accounts would have to change their name in multiple places?

I think this is fine, and consistent with the expectation that users will view the app as multiple services depending on which front door they've used. 

> I appreciate that the point here is that people might have different legal and professional names (e.g. someone has changed their name legally but keeps using their former name at work), it does feel like allowing these to be different should be to support functionality that's come from service design.

There is also a more fundamental point that if you have a role that requires ID verification and one that doesn't, then we'd need to manage the interaction between those if we just keep them in the One Login user table - e.g. you enter your professional name when signing up as a LA user, and then later you sign up as a Landlord then your legal name would clobber your professional name, which would be very surprising for the user. Similarly we'd need logic to avoid allowing a landlord overwrite their verified name because they've also started working as a Housing Officer.

Travis-Softwire (approved):
👍

PR #209, opened 2025-02-17
PRSD-884 Part 1: replace getMyThing functions with MyThing component classes

rowanhill:
This is a fairly mechanical set of changes. Although there's quite a lot in total, each change is quite simple - you might like to review each commit in turn.

The overview is:
* We no longer assert number of matching nodes when creating locators (in component constructors and in the various `getChildComponent`-like methods)
* In turn, this means we no longer need the various `isHidden` and `expectedNumber` type parameters that were plumbed through various methods
* The various `getMyThing` type methods on `BaseComponent.Companion` are now represented by equivalent `MyThing` component classes
* And where `MyThing` is used, it's now a `val` property. Relatedly, there's a preference for `page.component.performAction()` rather than `page.performActionOnComponent()`
* In particular `clickAndWait` (which waits for the page to load) has been pulled out into an interface with a default implementation, for use by those components that can be clicked to navigate.
* Because components are mostly a convenience wrapper around Locators, I've provided an `assertThat(component)` method that just passes through to `assertThat(locator)`

This is just part 1 of the tidy-up. Still to come will be tackling some of the more complex existing components, where changes might be a bit more involved, and then finally getting rid of the various `getChildCompnent` style methods.

### DIFF
--- a/src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/components/BaseComponent.kt
+++ b/src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/components/BaseComponent.kt
@@ -7,7 +7,7 @@ import com.microsoft.playwright.assertions.LocatorAssertions
 import com.microsoft.playwright.assertions.PlaywrightAssertions.assertThat
 
 abstract class BaseComponent(
-    protected val locator: Locator,
+    protected open val locator: Locator,
 ) {
     companion object {
         fun assertThat(component: BaseComponent): LocatorAssertions = assertThat(component.locator)
@@ -54,12 +54,6 @@ abstract class BaseComponent(
         fun getSubHeading(page: Page) = getComponent(page, "main header p")
 
         fun getConfirmationPageBanner(page: Page) = getComponent(page, ".govuk-panel--confirmation")
-
-        fun getLink(
-            page: Page,
-            text: String,
-            index: Int = 0,
-        ) = getComponent(page, ".govuk-link", LocatorOptions().setHasText(text), index)
     }
 
     protected fun getChildrenComponents(
--- a/src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/components/BaseComponent.kt
+++ b/src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/components/BaseComponent.kt
@@ -44,11 +44,6 @@ abstract class BaseComponent(
             return component.all()
         }
 
-        fun getSection(
-            page: Page,
-            index: Int = 0,
-        ) = getComponent(page, "section", index = index)
-
         fun getHeading(page: Page) = getComponent(page, "main header h1")
 
         fun getSubHeading(page: Page) = getComponent(page, "main header p")
--- a/src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/components/BaseComponent.kt
+++ b/src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/components/BaseComponent.kt
@@ -44,10 +44,6 @@ abstract class BaseComponent(
             return component.all()
         }
 
-        fun getHeading(page: Page) = getComponent(page, "main header h1")
-
-        fun getSubHeading(page: Page) = getComponent(page, "main header p")
-
         fun getConfirmationPageBanner(page: Page) = getComponent(page, ".govuk-panel--confirmation")
     }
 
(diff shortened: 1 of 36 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/components/BaseComponent.kt
```diff
@@ -3,96 +3,56 @@ package uk.gov.communities.prsdb.webapp.integration.pageObjects.components
...
-    ): List<Locator> = Companion.getChildrenComponents(locator, locatorStr, expectedChildren, locatorOptions)
+    ): List<Locator> = Companion.getChildrenComponents(locator, locatorStr, locatorOptions)
 
     protected fun getChildComponent(
         locatorStr: String,
         locatorOptions: Locator.LocatorOptions? = null,
         index: Int = 0,
-        isVisible: Boolean = true,
-    ): Locator = Companion.getChildComponent(locator, locatorStr, locatorOptions, index, isVisible)
+    ): Locator = Companion.getChildComponent(locator, locatorStr, locatorOptions, index)
```
isobel-softwire, line 57: I assume these are to-be-deleted as well?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/components/Button.kt
```diff
@@ -0,0 +1,17 @@
...
+
+class Button(
+    override val locator: Locator,
+) : BaseComponent(locator),
+    ClickAndWaitable {
+    companion object {
+        fun byText(
+            page: Page,
+            text: String? = null,
+        ): Button = Button(page.locator(".govuk-button", if (text == null) null else LocatorOptions().setHasText(text)))
```
isobel-softwire, line 15: I don't think `text` should be nullable if the method's called `byText`

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/components/Link.kt
```diff
@@ -0,0 +1,18 @@
...
+class Link(
+    override val locator: Locator,
+) : BaseComponent(locator),
+    ClickAndWaitable {
+    companion object {
+        fun byText(
+            page: Page,
+            text: String,
+            index: Int = 0,
+        ): Button = Button(page.locator(".govuk-link", LocatorOptions().setHasText(text)).nth(index))
```
isobel-softwire, line 16: Is this supposed to return a `Link`?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/landlordRegistrationJourneyPages/ConfirmationPageLandlordRegistration.kt
```diff
@@ -3,15 +3,20 @@ package uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.landlordRe
...
-    val registrationNumberText: String = getChildComponent(confirmationBanner, "strong").innerText()
+    val confirmationBanner = LandlordRegistrationConfirmationBanner(page)
+    val goToDashboardButton = Button.byText(page, "Go to Dashboard")
 
-    fun clickGoToDashboard() = clickComponent(BaseComponent.getButton(page, "Go to Dashboard"))
+    class LandlordRegistrationConfirmationBanner(
+        page: Page,
+    ) : ConfirmationBanner(page) {
+        val registrationNumberText: String
+            get() = locator.locator("strong").innerText()
```
isobel-softwire, line 20: Is a custom getter needed here?

isobel-softwire (approved):
Left a few minor thoughts. LGTM!

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/components/BaseComponent.kt
rowanhill: Yep

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/landlordRegistrationJourneyPages/ConfirmationPageLandlordRegistration.kt
rowanhill: Yes, but only to be purist about the "reactive" nature of components. If we didn't have a getter, we'd get the text at the point the constructor was called and never look it up again. In practice, I think that would be fine (as this text isn't dynamic), but this would be the only property on our components that behaves this way. By adding a custom getter, it's evaluated again each time.

Alternatively, I could have only exposed val registrationNumber: Locator and called innerText() in the test, but that seemed more repetitive.

PR #210, opened 2025-02-17
Feat/prsd 742 save landlords country of residence during landlord registration

Kiran-R-K:
Save Landlord's country of residence during Landlord registration and renamed Landlord's `internationalAddress` to `nonEnglandOrWalesAddress`

On the `landlord` table - renamed `internationalAddress` to `nonEnglandOrWalesAddress` and added  `country_of_residence` column
Ensured `country_of_residence` is saved when a Landlord is created, adding the value "England or Wales" for appropriate landlords.
Created `isEnglandOrWalesResident()` Landlord entity to use when checking the residency of the landlord

⚠️ Note: I will create a separate PR to update message keys and terminology across the code base that reference "UK resident" and "international address"


src/main/resources/db/migrations/V1_12_0__landlord_table_add_country_of_residence_and_rename_international_address.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE landlord
+    ADD country_of_residence VARCHAR(255);
+
+ALTER TABLE landlord
+    ALTER COLUMN country_of_residence SET NOT NULL;
```
rowanhill, line 5: Two things: a) can't we do this in one line, but more importantly b) won't this throw an exception if there is some existing landlord data (because the country_of_residence doesn't have a value, cannot be null, and there's no default set)? Are we happy we'll never be in that situation at the moment (i.e. this migration will only ever be applied to empty databases)?

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/CountryOfResidenceConstants.kt
```diff
@@ -0,0 +1,3 @@
+package uk.gov.communities.prsdb.webapp.constants
+
+const val ENGLAND_OR_WALES = "England or Wales"
```
isobel-softwire, line 3: I feel like this could go in `PlaceNamesConstants`

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/PropertyDetailsLandlordViewModel.kt
```diff
@@ -50,7 +50,7 @@ class PropertyDetailsLandlordViewModel(
                 } else {
                     addRow(
                         "propertyDetails.landlordDetails.addressOutsideEnglandOrWales",
-                        landlord.internationalAddress,
+                        landlord.nonEnglandOrWalesAddress,
                         "$UPDATE_ROUTE/address",
```
isobel-softwire, line 54: Here as well?

src/main/resources/db/migrations/V1_12_0__landlord_table_add_country_of_residence_and_rename_international_address.sql
isobel-softwire, line 1: #208 also has a `V1_12_0__` migration 😟

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordServiceTests.kt
```diff
@@ -140,6 +141,7 @@ class LandlordServiceTests {
                 "example@email.com",
                 "07123456789",
                 addressDataModel,
+                "England or Wales",
```
isobel-softwire, line 144: You could use the constant here and on line 125

isobel-softwire (approved):
LGTM 😄 - left a few thoughts

src/main/resources/db/migrations/V1_12_0__landlord_table_add_country_of_residence_and_rename_international_address.sql
Kiran-R-K: 🐎 whichever one merges in first will get to keep the name 🕐

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/PropertyDetailsLandlordViewModel.kt
Kiran-R-K: As above, this can be ignored for now

src/main/resources/db/migrations/V1_12_0__landlord_table_add_country_of_residence_and_rename_international_address.sql
Kiran-R-K: a) yes, although after I respond to point b we won't be able to do it anymore.
b) I thought we were going to be only applying it to an empty database but I was wrong so I'll update it to populate the column for existing rows before setting it to not null now

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/LandlordViewModel.kt
```diff
@@ -69,32 +69,31 @@ class LandlordViewModel(
                 ),
                 SummaryListRowViewModel(
                     "landlordDetails.personalDetails.ukResident",
```
AEPR, line 71: I assume that this will get changed in the "content sweep"?

PR #214, opened 2025-02-18
Feat/prsd 858 section headings property registration

JasminConterioSW:
- Add a boolean to Step to decide whether to display a section heading on that step (default false)
- Add a function to Journey which can get the section header info and call it in populateModelAndGetViewName 
- Pass sectionHeaderInfo into Page's populateModelAndGetTemplateName and add it to the model
EDIT:
- Also adding sectionHeaders to the Landlord registration journey

Tests:
- add tests for new getSectionHeaderInfo
- check that the section header appears on pages in the PropertyRegistrationJourney and LandlordRegistratinoJourney integration tests


src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PropertyRegistrationCheckAnswersPageTests.kt
```diff
@@ -46,7 +46,7 @@ class PropertyRegistrationCheckAnswersPageTests {
     }
 
     private fun getPropertyDetails(journeyData: MutableMap<String, Any?>): List<SummaryListRowViewModel> {
-        page.populateModelAndGetTemplateName(validator, model, pageData, prevStepUrl, journeyData)
+        page.populateModelAndGetTemplateName(validator, model, pageData, prevStepUrl, journeyData, null)
```
Travis-Softwire, line 49: Minor, but do you need the explicit null here?

Travis-Softwire (commented):
One tiny comment but otherwise LGTM :) (Although probably worth @AEPR casting an eye over it as well)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/Page.kt
```diff
@@ -41,6 +42,20 @@ open class Page(
         journeyData: JourneyData?,
     ): String = populateModelAndGetTemplateName(validator, model, pageData, prevStepUrl)
 
+    open fun populateModelAndGetTemplateName(
```
Travis-Softwire, line 45: On reflection - this overload isn't quite right. I don't think the default `null` for sectionHeaderInfo will ever be used, as I think any call with that argument missing will resolve to the version of the function above instead.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -59,15 +61,34 @@ abstract class Journey<T : StepId>(
         val prevStepUrl = getPrevStepUrl(prevStepDetails?.step, prevStepDetails?.subPageNumber)
         val pageData =
             submittedPageData ?: JourneyDataHelper.getPageData(journeyData, requestedStep.name, subPageNumber)
+
+        val sectionHeaderInfo = getSectionHeaderInfo(requestedStep)
```
isobel-softwire, line 65: Can a journey without a task list have sections/section headers?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -59,15 +61,34 @@ abstract class Journey<T : StepId>(
...
             prevStepDetails?.filteredJourneyData,
+            sectionHeaderInfo,
         )
     }
 
+    fun getSectionHeaderInfo(step: Step<T>): SectionHeaderViewModel? {
+        if (step.displaySectionHeader) {
+            val sectionContainingStep = sections.first { it.isStepInTaskList(step.id) }
+            if (sectionContainingStep.headingKey == null) {
+                throw PrsdbWebException("Section heading requested but heading message key not found")
```
isobel-softwire, line 81: Not sure that this should be a web exception - maybe a `NullPointerException` instead?

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PropertyRegistrationCheckAnswersPageTests.kt
JasminConterioSW: The tests are failing without it, it keeps trying to call the overridden populateModelAndGetTemplateName from PropertyRegistrationCheckAnswersPage.

Actually, probably what I should be doing is overriding both versions of populateModelAndGetTemplateName on that page so it works whether the section header is included or not...

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/Page.kt
JasminConterioSW: Yes I think you're right!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
```diff
@@ -55,29 +55,69 @@ class LandlordRegistrationJourney(
...
+                registerDetailsTasks(addressLookupService, addressDataService),
+                "registerAsALandlord.section.yourDetails.heading",
+            ),
+            JourneySection(
+                checkAndSubmitDetailsTasks(addressDataService, landlordService, emailNotificationService),
+                "registerAsALandlord.section.checkAndSubmit.heading",
+            ),
+        )
+
+    private fun privacyNoticeTasks(): List<JourneyTask<LandlordRegistrationStepId>> = emptyList()
```
isobel-softwire, line 70: Should we be adding this if we haven't implemented the privacy notice pages yet?

src/main/resources/messages.properties
```diff
@@ -117,6 +117,10 @@ registerAsALandlord.confirmation.propertyRegistration.heading=You must register
 registerAsALandlord.confirmation.propertyRegistration.paragraph.one=You can be fined or prosecuted for not registering all your properties on the database.
 registerAsALandlord.confirmation.propertyRegistration.paragraph.two=Make sure to complete your property registration as soon as possible.
 
+registerAsALandlord.section.privacyNotice.heading=Privacy notice
```
isobel-softwire, line 120: Same here

src/main/resources/templates/forms/confirmIdentityForm.html
isobel-softwire, line 1: This isn't part of your ticket, but `confirmIdentityForm` should probably use the `questionPage` (and maybe `fieldSet`) fragments

src/main/resources/templates/forms/confirmIdentityForm.html
```diff
@@ -11,11 +11,13 @@
...
-            <header>
-                <h1 class="govuk-heading-l" th:text="#{forms.confirmDetails.heading}">forms.checkAnswers.heading</h1>
-                <p  class="govuk-body" th:text="#{forms.confirmDetails.summary}"></p>
-            </header>
+
             <form th:action="@{''}" method="post" novalidate th:object="${formModel}">
+                <p th:replace="~{fragments/sectionHeader :: sectionHeader(${sectionHeaderInfo})}" ></p>
+                <header>
+                    <h1 class="govuk-heading-l" th:text="#{forms.confirmDetails.heading}">forms.checkAnswers.heading</h1>
+                    <p  class="govuk-body" th:text="#{forms.confirmDetails.summary}"></p>
```
isobel-softwire, line 19: The fallback text doesn't match the `th:text` here

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PropertyRegistrationCheckAnswersPage.kt
```diff
@@ -30,15 +31,34 @@ class PropertyRegistrationCheckAnswersPage(
...
+    override fun populateModelAndGetTemplateName(
+        validator: Validator,
+        model: Model,
+        pageData: Map<String, Any?>?,
+        prevStepUrl: String?,
+        journeyData: JourneyData?,
+    ): String {
+        addPropertyDetailsToModel(model, journeyData)
+        return super.populateModelAndGetTemplateName(validator, model, pageData, prevStepUrl, journeyData)
+    }
```
isobel-softwire, line 49: Should this ever be called?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/components/Form.kt
```diff
@@ -11,6 +11,8 @@ class Form(
     fun getErrorMessage(fieldName: String? = null) =
         getChildComponent(if (fieldName == null) ".govuk-error-message" else "p[id='$fieldName-error']")
 
+    fun getSectionHeader() = getChildComponent("#section-header")
```
isobel-softwire, line 14: `getChildComponent()` is being removed in #218 - you could replace it with `locator.locator("#section-header")`

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
AEPR: We should ideally be throwing a `PrsdbWebException` for pretty much anything we explicitly need exceptions for - that's al it means as an exception, one thrown by `PrsdbWeb`. If we need something more specific we should make a custom subclass. In this case, I think it makes sense to just use the default class, unless we create an exception to represent "This should have been unreachable - two things that should match up don't!"

I don't think there's a situation where a `NullPointerException` should be manually thrown by service code - it should be reserved for Kotlin being unhappy that a non-nullable value is actually `null`.

Edit: see my later comment that would potentially make this moot!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
JasminConterioSW: Yes, there's been a bit of back on forth on this but they do want section headers (matching what's on the journey pages on Figma) for the landlord registration journey even though they don't know if and when they want a task list. We can move tasks around as needed if they move between sections

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/JourneySection.kt
```diff
@@ -7,6 +7,8 @@ class JourneySection<T : StepId>(
     val tasks: List<JourneyTask<T>>,
     val headingKey: String? = null,
 ) {
+    fun isStepInTaskList(stepId: T): Boolean = stepId in tasks.flatMap { task -> task.steps.map { step -> step.id } }
```
AEPR, line 10: This should be `isStepInSection`, we shouldn't refer to task lists except as the page or component (it will get confusing!).

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
isobel-softwire: Ah ok!

AEPR: It makes a functional difference, because it means the sections currently present are "2/3" and "3/3". I don't think it does any harm.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PropertyRegistrationCheckAnswersPage.kt
JasminConterioSW: Hmm.

So it _does_ get called in PropertyRegistrationCheckAnswersPageTests (updated after one of Travis' comments).

I'll think about whether it should go back, maybe it should just call the other version passing null for sectionHeaderInfo as that is how it gets called from Journey

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -59,15 +61,34 @@ abstract class Journey<T : StepId>(
...
+            }
+            return SectionHeaderViewModel(
+                sectionContainingStep.headingKey,
+                sections.indexOf(sectionContainingStep) + 1,
+                sections.size,
+            )
+        }
+        return null
+    }
```
AEPR, line 90: I think the logic of "should a page show section information" should live on the `Page` class itself, rather than in the Journey. This feels like a violation of encapsulation to me.

I think that this should return `null` instead of throwing an exception if there's no section name and shouldn't check whether the step wants to display the data or not!

Then, the `Page` itself should be responsible for determining whether the `SectionHeaderViewModel` is needed and whether it should be displayed. (I'll leave another comment there!)

(Comment part 1 of 3)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/Page.kt
```diff
@@ -41,6 +42,20 @@ open class Page(
...
+        validator: Validator,
+        model: Model,
+        pageData: Map<String, Any?>?,
+        prevStepUrl: String?,
+        journeyData: JourneyData?,
+        sectionHeaderInfo: SectionHeaderViewModel?,
+    ): String {
+        if (sectionHeaderInfo != null) {
+            model.addAttribute("sectionHeaderInfo", sectionHeaderInfo)
+        }
```
AEPR, line 55: I think this is where we should be checking whether this `Page` wants to show a section header or not, and we should base it on the page itself, not the data passed in. That is to say, this should read something like:
```
if (shouldShouldSectionHeader) {
    addSectionHeaderToModel(model, sectionHeaderInfo)
}
```
where `addSectionHeaderToModel()` replicates the logic from the current function in journey (including that exception if the `SectionHeaderInfo` is null when it shouldn't be!).

(Comment part 2 of 3)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/steps/Step.kt
```diff
@@ -22,6 +22,7 @@ class Step<T : StepId>(
             null,
         )
     },
+    val displaySectionHeader: Boolean = false,
```
AEPR, line 25: And the upshot of all of that is that this boolean doesn't live here - it should be a parameter passed into `Page` instead!

(Comment part 3 of 3)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
JasminConterioSW: The problem with putting it on the Page class is the Page class doesn't know about all of `sections`, is it worth having a quick chat?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PropertyRegistrationCheckAnswersPage.kt
```diff
@@ -33,32 +35,14 @@ class PropertyRegistrationCheckAnswersPage(
...
-        journeyData: JourneyData?,
-    ): String {
-        addPropertyDetailsToModel(model, journeyData)
-        return super.populateModelAndGetTemplateName(validator, model, pageData, prevStepUrl, journeyData)
-    }
-
-    private fun addPropertyDetailsToModel(
-        model: Model,
-        journeyData: JourneyData?,
-    ) {
```
AEPR: Nitpick: I actually quite liked this being extracted to a little function, even if we're only calling it once. You could make it take `JourneyData` (instead of nullable) and then call it with `!!` in the populate function.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/Page.kt
```diff
@@ -50,9 +52,13 @@ open class Page(
...
         sectionHeaderInfo: SectionHeaderViewModel?,
     ): String {
-        if (sectionHeaderInfo != null) {
+        if (displaySectionHeader) {
+            if (sectionHeaderInfo == null) {
+                throw PrsdbWebException("Section heading requested but heading message key not found")
+            }
             model.addAttribute("sectionHeaderInfo", sectionHeaderInfo)
         }
+
```
AEPR, line 61: Nitpick: Similarly, I think this is enough complexity to rate a little function if you wanted.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/Page.kt
```diff
@@ -6,15 +6,18 @@ import org.springframework.validation.BindingResult
...
 import uk.gov.communities.prsdb.webapp.models.formModels.FormModel
+import uk.gov.communities.prsdb.webapp.models.viewModels.SectionHeaderViewModel
 import kotlin.reflect.KClass
 import kotlin.reflect.full.createInstance
 
 open class Page(
     private val formModel: KClass<out FormModel>,
     private val templateName: String,
     private val content: Map<String, Any>,
+    val displaySectionHeader: Boolean = false,
```
AEPR, line 20: Nitpick: I think we should prefer naming booleans things like `shouldDisplaySectionHeader`, or `isTheVeryBestThatNoOneEverWas` because this could in theory be a header for a display section.

AEPR (approved):
Couple of nitpicks, feel free to change and merge or ignore.

AEPR (approved):
LGTM!

PR #217, opened 2025-02-18
PRSD-751: Adds ECS service update step to deploy pipeline

Travis-Softwire:


rowanhill (approved):
Looks good. How are we going to do deploys to the test environment?

Travis-Softwire:
> Looks good. How are we going to do deploys to the test environment?

There are a few options and I thought we'd probably want a discussion and likely an ADR to decide and record our process. Options include:

- A manually triggered pipeline
- An automatic pipeline that is launched on successful merges to main but requires an approval gate
- A separate long-lived branch that tracks `test` in the same way `main` tracks `integration` and an accompanying pipeline that automatically deploys to it

PR #225, opened 2025-02-20
PRSD-648: Interact with Session via Services

isobel-softwire:
- Makes LA and property registration journeys interact with session via `LocalAuthorityDataService` and `PropertyRegistrationService` respectively
- Updates the affected tests


Kiran-R-K (commented):
I'm curious about the method name pattern being `setThingThisSession` and `getThingThisSession`. The JourneyDataService uses `setThing` and `getThingFromSession`.

At first I thought it meant the thing was being registered in that session, but someone could register multiple properties per session right?

Kiran-R-K (approved):
LGTM :)

PR #228, opened 2025-02-21
feat/PRSD-747 store and return id verification status for landlord

Kiran-R-K:
Created a migration to add an `is_verified`  column to the landlord table
Add `isVerified` to the Landlord entity
Ensured that when a new Landlord is created their `isVerified` status is also saved.
Added a verification status row to `LandlordViewModel` and `PropertyDetailsLandlordViewModel`
Updated tests
Updated data sql files.

Landlord record:
image

Property record:
image


src/test/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/PropertyDetailsLandlordViewModelTests.kt
```diff
@@ -197,8 +199,12 @@ class PropertyDetailsLandlordViewModelTests {
         val viewModel = PropertyDetailsLandlordViewModel(testLandlord)
 
         // Assert
-        for (i in viewModel.landlordsDetails) {
-            assertNotNull(i.changeUrl)
+        for (row in viewModel.landlordsDetails) {
+            if (row.fieldHeading == "landlordDetails.personalDetails.oneLoginVerified") {
```
AEPR, line 203: I think it's worth pulling this out explicitly into a list of "rowHeadingsWithoutChangeLinks", just so if someone reads this later it makes more sense.

Obviously it's the same thing, but to me it's easier to understand if there's an explicit assignment to say "There are some rows that shouldn't have change links".

isobel-softwire (approved):
LGTM :)

AEPR (approved):
Looks good to me.

One nitpick on a test. Happy for a merge with or without the change.

PR #229, opened 2025-02-21
Chore/prsd 901 refactor step details iteration

AEPR:
Turns `Journey` into an `Iterable<StepDetails>` by creating an `Iterator`. This massively simplifies functions that need to iterate through the journey, and lets us leverage kotlins built in iterable functionality.

 Draft because `StepDetailsIterator`...
* Should be in its own file
* Have more sensible exceptions
* Have explicit tests

To make `Journey` iterable, however, we need to get journey data from the `JourneyDataService` for each `Iterator` created. That in itself should be no problem, but I was worried about the potential performance hit of accessing data from the session multiple times so I've added an optional additional change to make `JourneyDataService` have a manual in-memory cache of the journey data.

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
@@ -7,6 +7,7 @@ import org.springframework.web.server.ResponseStatusException
 import org.springframework.web.util.UriComponentsBuilder
 import uk.gov.communities.prsdb.webapp.constants.enums.JourneyType
 import uk.gov.communities.prsdb.webapp.constants.enums.TaskStatus
+import uk.gov.communities.prsdb.webapp.exceptions.PrsdbWebException
 import uk.gov.communities.prsdb.webapp.forms.steps.Step
 import uk.gov.communities.prsdb.webapp.forms.steps.StepDetails
 import uk.gov.communities.prsdb.webapp.forms.steps.StepId
@@ -219,4 +220,51 @@ abstract class Journey<T : StepId>(
         initialStepId: T,
         steps: Set<Step<T>>,
     ): List<JourneySection<T>> = listOf(JourneySection.withOneTask(JourneyTask(initialStepId, steps)))
+}
+
+class JourneyWithJourneyData<T : StepId>(
+    val journeyData: JourneyData,
+    val steps: Set<Step<T>>,
+    val initialStepId: T,
+    val validator: Validator,
+) : Iterable<StepDetails<T>> {
+    override fun iterator(): Iterator<StepDetails<T>> = JourneyIterator(journeyData, steps, initialStepId, validator)
+}
+
+class JourneyIterator<T : StepId>(
+    val journeyData: JourneyData,
+    val steps: Set<Step<T>>,
+    val initialStepId: T,
+    val validator: Validator,
+) : Iterator<StepDetails<T>> {
+    lateinit var currentStepDetails: StepDetails<T>
+
+    override fun hasNext(): Boolean {
+        if (!this::currentStepDetails.isInitialized) return steps.count { step -> step.id == initialStepId } == 1
+        val pageData = JourneyDataHelper.getPageData(journeyData, currentStepDetails.step.name, currentStepDetails.subPageNumber)
+        if (pageData == null || !currentStepDetails.step.isSatisfied(validator, pageData)) {
+            return false
+        }
+
+        return currentStepDetails.step.nextAction(journeyData, currentStepDetails.subPageNumber).first != null
+    }
+
+    override fun next(): StepDetails<T> {
+        if (!this::currentStepDetails.isInitialized) {
+            currentStepDetails = StepDetails(steps.single { step -> step.id == initialStepId }, null, mutableMapOf())
+        } else {
+            val pageData = JourneyDataHelper.getPageData(journeyData, currentStepDetails.step.name, currentStepDetails.subPageNumber)
+            if (pageData == null || !currentStepDetails.step.isSatisfied(validator, pageData)) {
+                throw PrsdbWebException("Does not have next")
+            }
+            val stepData = JourneyDataHelper.getPageData(journeyData, currentStepDetails.step.name, null)
+            currentStepDetails.filteredJourneyData[currentStepDetails.step.name] = stepData
+
+            val (nextStepId, nextSubPageNumber) = currentStepDetails.step.nextAction(journeyData, currentStepDetails.subPageNumber)
+            if (nextStepId == null) throw PrsdbWebException("Does not have next")
+            val nextStep = steps.single { step -> step.id == nextStepId }
+            currentStepDetails = StepDetails(nextStep, nextSubPageNumber, currentStepDetails.filteredJourneyData)
+        }
+        return currentStepDetails
+    }
 }
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
@@ -21,7 +21,7 @@ abstract class Journey<T : StepId>(
     private val journeyType: JourneyType,
     protected val validator: Validator,
     protected val journeyDataService: JourneyDataService,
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -211,12 +162,53 @@ abstract class Journey<T : StepId>(
...
     protected fun <T : StepId> createSingleSectionWithSingleTaskFromSteps(
         initialStepId: T,
         steps: Set<Step<T>>,
     ): List<JourneySection<T>> = listOf(JourneySection.withOneTask(JourneyTask(initialStepId, steps)))
+
+    override fun iterator(): Iterator<StepDetails<T>> =
+        StepDetailsIterator(journeyDataService.getJourneyDataFromSession(), steps, initialStepId, validator)
+}
+
+class StepDetailsIterator<T : StepId>(
```
AEPR, line 178: This should maybe be renamed something like `ReachableStepDetailsIterator`

Travis-Softwire (commented):
This is really smart, and we should definitely pursue this.

As mentioned, I don't think we need the optimisation of holding a copy of the cache in memory as the implementation of HttpSession should be doing that for us. 

I also paused over the semantics of the journey itself being the iterable as opposed to the steps being held in a custom iterable, but happy to chat about it in more detail on Monday.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/ReachableStepDetailsIterator.kt
```diff
@@ -0,0 +1,103 @@
...
+import uk.gov.communities.prsdb.webapp.helpers.JourneyDataHelper
+
+class ReachableStepDetailsIterator<T : StepId>(
+    journeyData: JourneyData,
+    private val steps: Iterable<Step<T>>,
+    private val initialStepId: T,
+    private val validator: Validator,
+) : Iterator<StepDetails<T>> {
+    private lateinit var currentStepDetails: StepDetails<T>
+    private var currentFilteredJourneyData: Map<String, Any?> = mapOf()
```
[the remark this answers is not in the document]
Travis-Softwire: Alternatively worth looking at where we currently manipulate the journey data (just in `Step.updateJourneyData`?) and considering whether that could be refactored to use an immutable version of JourneyData

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/ReachableStepDetailsIteratorTest.kt
```diff
@@ -0,0 +1,289 @@
...
+
+        @Test
+        fun `hasNext returns false if it hasn't been initialised and there is no initial step`() {
+            // Arrange
+            val testIterator =
+                TestIteratorBuilder()
+                    .withMissingFirstStep()
+                    .addStepToEnd(TestStepModel("1", isSatisfied = true))
+                    .addStepToEnd(TestStepModel("2", isSatisfied = false))
+                    .withMissingFirstStep()
```
Travis-Softwire, line 81: Why do we need the `withMissingFirstStep` call twice?

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/TestIteratorBuilder.kt
```diff
@@ -0,0 +1,92 @@
...
+            steps.zipWithNext {
+                    stepModel,
+                    nextStepModel,
+                ->
+                testStep(stepModel.urlPathSegment, nextStepModel.urlPathSegment, stepModel.isSatisfied, stepModel.customNextActionAddition)
+            } + testStep(steps.last().urlPathSegment, isSatisfied = steps.last().isSatisfied)
+
+        val firstStepId =
+            if (missingFirstStep) {
+                // A concatenation of all strings in a list cannot be contained by the list
```
Travis-Softwire, line 55: True, but not sure it's immediately obvious why it matters here (after a min I got it, but would probably make it clearer either with a more explicit explanatory comment or an intermediate variable e.g. `val impossibleUrlPathSegment = `

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/TestIteratorBuilder.kt
```diff
@@ -0,0 +1,92 @@
...
+    fun build(): ReachableStepDetailsIterator<TestStepId> {
+        val linkedSteps =
+            steps.zipWithNext {
+                    stepModel,
+                    nextStepModel,
+                ->
+                testStep(stepModel.urlPathSegment, nextStepModel.urlPathSegment, stepModel.isSatisfied, stepModel.customNextActionAddition)
+            } + testStep(steps.last().urlPathSegment, isSatisfied = steps.last().isSatisfied)
+
+        val firstStepId =
```
Travis-Softwire, line 53: Wonder if it would be clearer (esp given your test for multiple first steps) to have an explicit `withFirstStep(urlPathSegment: String)` builder function?

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/TestIteratorBuilder.kt
```diff
@@ -0,0 +1,92 @@
...
+        missingFirstStep = true
+        return this
+    }
+
+    fun addStepToEnd(stepModel: TestStepModel): TestIteratorBuilder {
+        journeyData[stepModel.urlPathSegment] = mapOf("urlPathSegment" to stepModel.urlPathSegment, "isSatisfied" to stepModel.isSatisfied)
+        return addStepToEndWithoutJourneyData(stepModel)
+    }
+
+    fun addStepToEndWithoutJourneyData(stepModel: TestStepModel): TestIteratorBuilder {
```
Travis-Softwire, line 39: Minor, but would call this `addStepToEndWithoutPageData` as it's the data for that step that's missing, rather than the journey data as a whole

Travis-Softwire (commented):
Overall looking good to me, a few thoughts in the comments

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/ReachableStepDetailsIterator.kt
```diff
@@ -0,0 +1,98 @@
...
+    private fun initialStepDetails() =
+        steps.singleOrNull { step -> step.id == initialStepId }.let {
+            if (it == null) {
+                throw NoSuchElementException("Journey does not have initial step")
+            } else {
+                StepDetails(it, null, currentFilteredJourneyData.toMutableMap())
+            }
+        }
+
+    private fun getInitialStepDetailsProcedural(): StepDetails<T> {
```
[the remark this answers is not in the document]
AEPR: Oops! Yes, I've deliberately included two different ways of doing `getInitialStepDetails` and included them with the plan that I leave a comment to spark discussion.

Basically, `getInitialStepDetails` and `getInitialStepDetailsProcedural` are two different implementations of the same function. I prefer the more functional version that uses `?.let{  ...  }`, but I thought I'd include both and get an opinion.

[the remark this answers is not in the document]
Travis-Softwire: Yeah I'm fine with that, although should it actually use `?.let{ ...}` rather than just `.let{ .... }`? In which case I think your exception call just moves to the end of the statement after an `?:` operator?

AEPR (commented):
Responded. Mostly accepted your suggestions (and will make the changes presently).

Couple that bear discussion.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/ReachableStepDetailsIterator.kt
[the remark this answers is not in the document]
AEPR: Discussed separately and will refactor to immutable maps in a separate PR.

Travis-Softwire (approved):
Happy with those changes :)

PR #230, opened 2025-02-21
Feat/prsd 900 landlord details journey tests

AEPR:
Adds two simple sets of tests, one a unit test for saving data to the database and an integration test for the journey.

I think that this is sufficient for now, but there will need to be more interesting tests once we have branching journeys (i.e. updating residency)


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/Navigator.kt
```diff
@@ -409,5 +410,10 @@ class Navigator(
         return createValidPage(page, PropertyDetailsPageLocalAuthorityView::class)
     }
 
+    fun goToUpdateLandlordDetailsPage(): LandlordUpdateDetailsPage {
+        navigate("landlord-details/update/details")
```
isobel-softwire, line 414: You could use the constant here

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/basePages/LandlordDetailsBasePage.kt
```diff
@@ -9,6 +10,7 @@ abstract class LandlordDetailsBasePage(
     urlSegment: String,
 ) : BasePage(page, urlSegment) {
     val tabs = LandlordDetailsTabs(page)
+    val personalDetails = LandlordPersonalDetailsSummaryList(page)
     val table = Table(page)
```
isobel-softwire, line 14: I think these should be renamed to `personalDetailsSummaryList` and `registeredPropertiesTable`

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/UpdateLandlordDetailsJourneyTests.kt
```diff
@@ -0,0 +1,97 @@
...
+        val landlordEmail = "new@email.test"
+        landlordDetailsUpdatePage = updateLandlordEmailAndReturn(landlordDetailsUpdatePage, landlordEmail)
+
+        // Submit changes TODO PRSD-355 add proper submit button and declaration page
+        landlordDetailsUpdatePage.submitButton.clickAndWait()
+        val landlordDetailsPage = assertPageIs(page, LandlordDetailsPage::class)
+
+        // Check changes have occurred
+        assertThat(landlordDetailsPage.personalDetails.emailRow.value).containsText(landlordEmail)
+    }
```
isobel-softwire, line 96: Do we need to check each step's happy path? I think the combined test is sufficient. Probably worth writing tests for each step's unhappy path though.

isobel-softwire (approved):
LGTM - left a suggestion about tests to edit/add

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/basePages/LandlordDetailsBasePage.kt
AEPR: Yeah, that's better isn't it.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/UpdateLandlordDetailsJourneyTests.kt
AEPR: This isn't really a test of the step in question, but of all the other steps - I wanted to capture that any given step can either be changing or not and it will work as expected. Having a single happy path that changes every value doesn't test partial changes. Having a "just update email" and "just update name" guarrantees that every possible field has NOT been updated once.

Regarding unhappy paths, I'm not sure there are any at the moment really. There's validation failures, but they should be tested on other tests already as there are no new pages.

isobel-softwire: Ah ok, that's fair enough

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/Navigator.kt
AEPR: Awkwardly, `navigate` prepends with a "/", so I cant use the route constant directly. I think I'm just going to update `navigate` to leave it off and add it to the start of all the other paths.

PR #235, opened 2025-02-25
Feat/prsd 647 la dashboard

JasminConterioSW:
- Adds the local authority dashboard, along with some dashboard fragments
- Adds links from the dashboard to CTAs which are implemented
- Adds links to the dashboard from other LA pages 
    - Property details, property search and landlord search pages - back links go to the LA dashboard
    - "Go to dasbhoard" buttons on the Invite LA User success page, Register LA User success page, Manage LA User page

Tests:
- Controller tests (authorized / unauthorised users)
- Service tests (getting the LA user)
- Integration tests
    - Dashboard - display users name, local authority,
    - Dashboard - CTA links work
    - Links from other pages to the dashboard work

(I might still adjust the css a bit, some of the padding doesn't quite match the designs but it is close) 

image
image


src/main/resources/css/custom.css
```diff
@@ -39,3 +41,24 @@
...
+
+.prsd-dashboard-panel__title {
+    color: var(--govuk-colour-white);
+}
+
+.prsd-dashboard-panel__body {
+    color: var(--govuk-colour-white);
+}
+
+.prsd-section {
```
rowanhill, line 58: This is quite a generic name, but if it was really that generic we'd just apply this to `section` rather than a class. I don't know what to suggest that's better, though - are we using it in any particular situation?

src/main/resources/css/custom.css
```diff
@@ -39,3 +41,24 @@
...
+    margin-bottom: 30px;
+}
+
+.prsd-fill-container {
+    width: 100%;
+}
```
rowanhill, line 64: This is just a thought, but I wonder if a nicer approach might be to define something like `.stretch-justify-buttons .govuk-button` and then apply `stretch-justify-buttons` on the parent of the relevant buttons - then you could avoid plumbing the `fullWidth` parameter through to the fragment?

src/main/resources/templates/localAuthorityDashboard.html
```diff
@@ -0,0 +1,53 @@
...
+                'localAuthority.dashboard.action.aboutPilot.linkText',
+                ${aboutPilotUrl},
+                'localAuthority.dashboard.action.aboutPilot.descriptionText')}">
+                    View About pilot action
+                </div>
+            </div>
+        </section>
+
+
+
```
rowanhill, line 50: Bit of unnecessary whitespace here

rowanhill (approved):
I've made some relatively minor suggestions, but it broadly LGTM :)

PR #238, opened 2025-02-26
PRSD-NONE: Add useful to string on step

AEPR:
This was annoying me while I was debugging with steps.


isobel-softwire (approved):
LGTM :)

PR #239, opened 2025-02-26
PRSD-689: Creates UpdateJourney base class

isobel-softwire:
- Renames landlord update properties where not specific enough
- Creates base `UpdateJourney` base class
- Uses `journeyDataKey` to index `JourneyData` 
- Makes `Journey` responsible for determining its path segment (rather than `JourneyType`)

TODO:
- Update affected tests


AEPR (changes requested):
My immediate thought is that we're passing around the `journeyDataKey` far too much - it's ended up everywhere. As it stands it's going to make the whole journey logic so much harder to understand. Especially because we're often passing both the data and the key around together - which shouldn't be necessary. The purpose of the key is to be able to get the data, after all!

I think for this all to work, we should basically only retrieve the journeyData from the session once per request and then just pass it around. 

I've not looked through this whole review thoroughly, but I think that will need to be addressed before we can really look at everything else. If the above ends up requiring anything more than a minimal refactor you should check in with @Travis-Softwire before starting on it, and it could probably get its own PR.

Edit: See other comment!

AEPR:
I've had another thought overnight about how we could make this all work without having to do a big refactor. Since the `JourneyDataService` is already request scoped, we could cache the key on a class variable on `JourneyDataService` (with some appropriate guards to fail early if something unexpected happens) and then we can continue to use the `JourneyDataService` as a source of journey data throughout the code without needing to pass the key around everywhere.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -17,24 +17,27 @@ import java.security.Principal
...
+
     val steps: Set<Step<T>>
         get() = sections.flatMap { section -> section.tasks }.flatMap { task -> task.steps }.toSet()
 
     abstract val sections: List<JourneySection<T>>
 
-    open fun getUnreachableStepRedirect(journeyData: JourneyData) = "/${journeyType.urlPathSegment}/${initialStepId.urlPathSegment}"
+    abstract val journeyPathSegment: String
+
+    open fun getUnreachableStepRedirect(journeyData: JourneyData) = initialStepId.urlPathSegment
```
rowanhill, line 33: Could this be protected, to avoid cluttering the public API?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
```diff
@@ -88,11 +90,18 @@ class LandlordRegistrationJourney(
...
             JourneyTask.withOneStep(checkAnswersStep(addressDataService)),
-            JourneyTask.withOneStep(declarationStep(journeyDataService, landlordService, addressDataService, emailNotificationService)),
+            JourneyTask.withOneStep(
+                declarationStep(
+                    journeyDataService,
+                    landlordService,
+                    addressDataService,
+                    emailNotificationService,
+                ),
+            ),
```
rowanhill, line 100: This is just a whitespace change, right? I only ask because it's going to conflict with some of Alex's changes, I think.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/JourneyWithTaskList.kt
```diff
@@ -8,16 +8,38 @@ import uk.gov.communities.prsdb.webapp.forms.tasks.TaskListViewModelFactory
...
     journeyDataService: JourneyDataService,
 ) : Journey<T>(journeyType, validator, journeyDataService) {
     abstract val taskListFactory: TaskListViewModelFactory<T>
     abstract val taskListUrlSegment: String
 
-    final override fun getUnreachableStepRedirect(journeyData: JourneyData) = "/${journeyType.urlPathSegment}/$taskListUrlSegment"
+    final override fun getUnreachableStepRedirect(journeyData: JourneyData) = taskListUrlSegment
 
-    fun populateModelAndGetTaskListViewName(model: Model): String {
+    fun initialiseJourneyDataIfNotInitialised(
```
rowanhill, line 20: How come this belongs on JourneyWithTaskList and not on Journey?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataService.kt
```diff
@@ -23,10 +23,29 @@ class JourneyDataService(
     private val oneLoginUserRepository: OneLoginUserRepository,
     private val objectMapper: ObjectMapper,
 ) {
-    fun getJourneyDataFromSession(): JourneyData = objectToStringKeyedMap(session.getAttribute("journeyData")) ?: mapOf()
+    var journeyDataKey: String? = null
```
rowanhill, line 26: IMO this is weird enough that it's worth a comment to point out that it's okay for this service to have state because it's request scoped.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
isobel-softwire: Yes - I haven't gotten my linting issue sorted yet ☹️

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/JourneyWithTaskList.kt
isobel-softwire: I put it here because we're only using it in `/task-list` endpoints at the moment, but I do think putting it in `Journey` makes more sense

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/JourneyWithTaskList.kt
```diff
@@ -29,10 +33,18 @@ abstract class JourneyWithTaskList<T : StepId>(
...
         rootId: String,
     ) = TaskListViewModelFactory(
-        "registerProperty.title",
-        "registerProperty.taskList.heading",
-        "registerProperty.taskList.subtitle",
-        "register-property-task",
+        titleKey,
+        headingKey,
+        subtitleKey,
+        rootId,
```
AEPR, line 39: Oops - good catch.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/JourneyWithTaskList.kt
```diff
@@ -29,10 +33,18 @@ abstract class JourneyWithTaskList<T : StepId>(
...
         sections,
     ) { task, journeyData -> getTaskStatus(task, journeyData) }
+
+    private fun getTaskStatus(
+        task: JourneyTask<T>,
+        journeyData: JourneyData,
+    ): TaskStatus {
+        val canTaskBeStarted = isStepReachable(task.steps.single { it.id == task.startingStepId })
+        return task.getTaskStatus(journeyData, validator, canTaskBeStarted)
+    }
```
AEPR, line 49: What's the thought behind moving this here from the base task?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -135,6 +141,34 @@ abstract class Journey<T : StepId>(
...
+            sections.size,
+        )
+    }
+
+    private fun getStep(stepId: StepId) =
+        steps.singleOrNull { step -> step.id == stepId }
+            ?: throw ResponseStatusException(
+                HttpStatus.NOT_FOUND,
+                "Step $stepId not valid for journey ${journeyType.name}",
+            )
```
AEPR, line 170: Pulling this out as separate function is a good change - I know you've just taken the method signature from what was there before, but this could actually be narrowed to `getStep(stepId: T)` so I think that would be better.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -17,41 +16,57 @@ import java.security.Principal
...
 ) : Iterable<StepDetails<T>> {
     abstract val initialStepId: T
-    val steps: Set<Step<T>>
-        get() = sections.flatMap { section -> section.tasks }.flatMap { task -> task.steps }.toSet()
 
     abstract val sections: List<JourneySection<T>>
 
-    open fun getUnreachableStepRedirect(journeyData: JourneyData) = "/${journeyType.urlPathSegment}/${initialStepId.urlPathSegment}"
+    protected val steps: Set<Step<T>>
+        get() = sections.flatMap { section -> section.tasks }.flatMap { task -> task.steps }.toSet()
```
AEPR, line 28: I like that you've protected this here!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -72,36 +87,24 @@ abstract class Journey<T : StepId>(
...
         principal: Principal,
+        journeyDataKey: String? = null,
     ): String {
-        val currentStep =
-            steps.singleOrNull { step -> step.id == stepId } ?: throw ResponseStatusException(
-                HttpStatus.NOT_FOUND,
-                "Step ${stepId.urlPathSegment} not valid for journey ${journeyType.urlPathSegment}",
-            )
+        val journeyDataKeyOrDefault = journeyDataKeyOrDefault(journeyDataKey)
+        val journeyData = journeyDataService.getJourneyDataFromSession(journeyDataKeyOrDefault)
```
AEPR, line 99: I think this can just be `journeyDataKey` - we don't need to hold onto the fact that it could be the default key once it's been assigned so adding that to the variable name just adds to our mental load while reading this code.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -135,6 +141,34 @@ abstract class Journey<T : StepId>(
...
     }
 
+    protected open fun getUnreachableStepRedirect(journeyData: JourneyData) = initialStepId.urlPathSegment
+
+    protected fun <T : StepId> createSingleSectionWithSingleTaskFromSteps(
+        initialStepId: T,
+        steps: Set<Step<T>>,
+    ): List<JourneySection<T>> = listOf(JourneySection.withOneTask(JourneyTask(initialStepId, steps)))
+
+    protected fun journeyDataKeyOrDefault(journeyDataKey: String?) = journeyDataKey ?: journeyType.name
```
AEPR, line 151: I think this is maybe a little confusing/misleading - what this actually does is "return the string you pass in unless it's null, in which case return the default datakey value.

I think it would be better to make this a protected property `defaultJourneyDataKey` and just do the elvis operation inline.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LaUserRegistrationJourney.kt
```diff
@@ -42,13 +43,15 @@ class LaUserRegistrationJourney(
             ),
         )
 
+    override val journeyPathSegment = REGISTER_LA_USER_JOURNEY_URL
+
     fun initialiseJourneyData(token: String) {
-        val journeyData = journeyDataService.getJourneyDataFromSession()
+        val journeyData = journeyDataService.getJourneyDataFromSession(journeyType.name)
```
AEPR, line 49: Making a `defaultJourneyDataKey` would also let you use it here, making it clear why we're passing a value in.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateJourney.kt
```diff
@@ -0,0 +1,41 @@
...
+import uk.gov.communities.prsdb.webapp.services.JourneyDataService
+
+abstract class UpdateJourney<T : StepId>(
+    journeyType: JourneyType,
+    validator: Validator,
+    journeyDataService: JourneyDataService,
+) : Journey<T>(journeyType, validator, journeyDataService) {
+    abstract val updateStepId: StepId
+
+    protected val originalDataKey = "ORIGINAL_$journeyType.name"
```
AEPR, line 17: Why `.name`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateJourney.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    override fun getUnreachableStepRedirect(journeyData: JourneyData) =
+        if (journeyData[originalDataKey] == null) {
```
AEPR, line 20: Another thing I must have changed in parallel - this should use `!journeyData.containsKey`.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateJourney.kt
```diff
@@ -0,0 +1,41 @@
...
+        // For any fields where the data is updated, replace the original value with the new value
+        val updatedData =
+            journeyData.keys
+                .union(originalData?.keys ?: setOf())
+                .map { key ->
+                    key to if (journeyData.containsKey(key)) journeyData[key] else originalData?.get(key)
+                }.associate { it }
+
+        return ReachableStepDetailsIterator(updatedData, steps, initialStepId, validator)
+    }
```
AEPR, line 40: I think there's a problem here with what this abstract class is responsible for and what it delegates to its children. This says:
* If `originalDataKey` has not been set, redirect to the step with the `UpdateStepId`
* Whenever we traverse the journey, combine the `originalDataKey` data with the main journey data.

This works at the moment because the only subclass __does__ get the `originalDataKey` set exactly when you redirect to the update step. I think that, conceptually, this abstract class needs to be aware of the data being set and so it can complete the loop of:
* If `originalDataKey` has not been set, redirect to where...
* We set data for the `originalDataKey` so that...
* Whenever we traverse the journey, we can combine the `originalDataKey` data with the main journey data.

What I think that looks like is: 
* moving `initialiseJourneyDataIfNotInitialised` to the abstract `UpdateJourney`
* making what is currently `createOriginalLandlordJourneyData` an override of an abstract method on `UpdateJourney`
* (potentially) creating a function here `populateModelAndGetTemplateForUpdateStep`, which calls `initialise` and then `populateModel...` with the `updateStepId`

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/JourneyWithTaskList.kt
isobel-softwire: We're only using it in `JourneyWithTaskList` so I think its cleaner to have a private method here than a protected one in `Journey`

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateJourney.kt
isobel-softwire: This is a typo (I forgot the curly braces)!

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataServiceTests.kt
```diff
@@ -176,6 +176,7 @@ class JourneyDataServiceTests {
             val journeyType = JourneyType.LANDLORD_REGISTRATION
```
AEPR, line 176: I think we need to write some new tests covering the new, basic functionality of this service, something like:
* Once a key is set is is used
* If a different key is used, it throws
* If you try to get/set/clear without first setting the key, it throws

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/PropertyRegistrationJourneyTests.kt
```diff
@@ -1,334 +0,0 @@
...
-        ) {
-            testJourney.updateJourneyDataAndGetViewNameOrRedirect(
-                stepId = stepId,
-                pageData = pageData,
-                subPageNumber = null,
-                principal = mock(),
-                model = mock(),
-            )
-        }
-    }
```
AEPR, line 245: Please don't delete all the tests I wrote for PRSD-804!

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/JourneyTests.kt
```diff
@@ -912,4 +940,94 @@ class JourneyTests {
...
         }
     }
+
+    @Nested
+    inner class JourneyDataManipulationTests {
+        @Test
+        fun `when there is no journey data in the session or the database, journey data is not loaded`() {
+            val principalName = "principalName"
+            val testJourney =
+                PropertyRegistrationJourney(
```
AEPR, line 950: Appreciate you've moved the logic for this into `Journey` so you've moved the tests too, which makes sense. Can you update these tests to use the TestJourney from above instead of the PropertyRegistrationJourney?

AEPR (commented):
A couple of comments on the tests!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateLandlordDetailsJourney.kt
```diff
@@ -275,12 +275,13 @@ class UpdateLandlordDetailsJourney(
     }
 
     override fun initialiseJourneyDataIfNotInitialised(updateEntityId: String) {
-        super.initialiseJourneyDataIfNotInitialised(updateEntityId)
-
-        val landlord = landlordService.retrieveLandlordByBaseUserId(updateEntityId)!!
-        if (addressDataService.getAddressData(landlord.address.singleLineAddress) == null) {
+        val journeyData = journeyDataService.getJourneyDataFromSession(defaultJourneyDataKey)
+        if (!isJourneyDataInitialised(journeyData)) {
+            val landlord = landlordService.retrieveLandlordByBaseUserId(updateEntityId)!!
```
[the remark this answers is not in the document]
isobel-softwire: We spoke about this and decided to go with option two

PR #240, opened 2025-02-26
Feat/prsd 916 setup sass

JasminConterioSW:
- Switch custom css classes from directly writing in css to instead writing in scss, which will be compiled when the project is run
- Installed `rollup-plugin-sass` and add config to `rollup.config.mjs`
- Replace the code in the existing  `src/main/resources/css/custom.css` file with `.scss` files
- Update the padding on the LA user dashboard to more closely match the designs

image


src/main/js/index.js
```diff
@@ -2,6 +2,7 @@ import {initAll as initGDS} from 'govuk-frontend';
 import {initSelectAutocomplete} from "./autocomplete";
 import $ from 'jquery'
 import {initAll as initMoJDS} from '@ministryofjustice/frontend'
+import '../resources/css/custom.scss'
```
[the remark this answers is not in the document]
Travis-Softwire: Ah, this will be because our input for rollup is set to `src/main/js/index.js`. We could use another plugin to allow multiple inputs, but tbh I think importing it in index.js makes the most sense here.

rollup.config.mjs
```diff
@@ -17,6 +18,9 @@ export default {
         del({
             targets: 'dist/*',
         }),
+        sass({
+            output: 'dist/css/custom.css',
+        }),
```
Travis-Softwire, line 23: I _think_ the import point mentioned on the ticket can be resolved by adding: 
```
options: {
    includePaths: ['node_modules/govuk-frontend/dist'],
}
```
to the `sass` block here

Travis-Softwire (approved):
One thought about the `node_modules` import issue but otherwise LGTM :)

rollup.config.mjs
JasminConterioSW: That seems to be working - I thought I had tried that, but maybe didn't have it exactly!
Either that or it did work and I thought it hadn't because it's adding lots of blank lines to the compiled .css file making it look like it's not found everything...

PR #241, opened 2025-02-26
PRSD-NONE: move all path segments to UrlSegmentConstants.kt

JasminConterioSW:
Consolidate our path segment constants into one file


src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/UrlSegmentConstants.kt
```diff
@@ -1,5 +1,15 @@
...
-const val LANDLORD_DETAILS_SEGMENT = "landlord-details"
+const val LANDLORD_DETAILS_PATH_SEGMENT = "landlord-details"
+const val LOCAL_AUTHORITY_PATH_SEGMENT = "local-authority"
+const val LANDLORD_PATH_SEGMENT = "landlord"
+const val PROPERTY_PATH_SEGMENT = "property"
+const val DASHBOARD_PATH_SEGMENT = "dashboard"
+const val SEARCH_PATH_SEGMENT = "search"
+
+// TODO PRSD-670: Set to landlord dashboard
+val LANDLORD_DASHBOARD_URL = ServletUriComponentsBuilder.fromCurrentContextPath().build().toUriString()
```
AEPR, line 15: Fine to leave this for this PR, but this being not a constant and coming from a static context actually causes us a bit of a headache in the tests I'm writing at the moment, but that's out of the scope of this PR!

AEPR (approved):
Looks good to me.

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/UrlSegmentConstants.kt
JasminConterioSW: It's going to change on my next ticket...

JasminConterioSW: It doesn't actually exist yet, but it will be 
`const val LANDLORD_DASHBOARD_URL = "/$LANDLORD_PATH_SEGMENT/$DASHBOARD_PATH_SEGMENT"`
in the companion object of the (currently non-existed) LandlordDashboardController.

You might be able to just change it for now if it's causing problems?

JasminConterioSW: I'll change it here for now and move it to the controller as part of PRSD-670.
No point in having extra headaches if you don't need them :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/PRSDWebAppConstants.kt
```diff
@@ -1,16 +0,0 @@
-package uk.gov.communities.prsdb.webapp.constants
-
-import org.springframework.web.servlet.support.ServletUriComponentsBuilder
-
-// TODO PRSD-670: Set to landlord dashboard
-val LANDLORD_DASHBOARD_URL = ServletUriComponentsBuilder.fromCurrentContextPath().build().toUriString()
```
isobel-softwire, line 6: This returns an absolute URL for use in emails, rather than being a URL segment - I don't think it should've been deleted.

JasminConterioSW: Hah, is this is one of the tests that gets ignored?

isobel-softwire: I don't think there's a test for checking that email links are right, unfortunately ☹️

AEPR: There is actually one for the LaInvite URL IIRC - it's an MVC test.

In any case, the problem is that it's getting used by e.g. a journey directly and can't be mocked - which then means you can't do unit tests.

I think this probably has a couple of consequences:
* There should be integration tests for all of our email links.
* We should maybe have some sort of `AbsoluteUrlProvider` service class that makes these calls, so that it can be mocked in tests that don't care.

JasminConterioSW: I can have a look at adding this in properly as part of PRSD-670, I'll look tomorrow and see if I can get a proper fix in quickly or whether I should try and add this back in.

PR #243, opened 2025-02-26
Fix/prsd 804 remove extraneous data

AEPR:
Creates unit tests that fail because we store data from disused paths in the journey, then uses the iterable nature of `Journey` to get the filtered `JourneyData` and use that for database submissions.


src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/PropertyRegistrationJourneyTests.kt
```diff
@@ -6,22 +6,244 @@ import org.junit.jupiter.api.Test
...
+    val alwaysTrueValidator: AlwaysTrueValidator = AlwaysTrueValidator()
+
+    class AlwaysTrueValidator : Validator {
+        override fun supports(clazz: Class<*>): Boolean = true
+
+        override fun validate(
+            target: Any,
+            errors: Errors,
+        ) {}
+    }
```
rowanhill, line 70: Worth pulling this out into a helper file, rather than copy and paste it?

rowanhill (approved):
Making the services properties rather than just constructor params on the journey is such a help! We should have done that before.

I was going to ask whether we definitely wanted to have all the unit tests you've written (i.e. the different ways you might end up with journey data that doesn't belong the currently taken path), given they're all exercising the same code, but I think on balance it's probably sensible to be exhaustive like this in the tests.

AEPR:
> Making the services properties rather than just constructor params on the journey is such a help! We should have done that before.

This was originally done with these being passed around because all of the step generating functions lived in a static context - I refactored that a while ago but missed some of these services (until now!)

> I was going to ask whether we definitely wanted to have all the unit tests you've written
I've probably gone a little overboard, but these are unit tests so they're fast anyway and the mechanism that they're all passing is a little different - only one of them was actually failing when I wrote them because of how we're handling the data!

PR #244, opened 2025-02-27
Feat/prsd 670 landlord dashboard absolute url

JasminConterioSW:
This does not complete PRSD-670, but the aim is to get a working absolute landlord url for use in emails, in a way that can be mocked if necessary in tests.

- Create a skeleton landlord dashboard controller / template / endpoint
- Add AbsoluteUrlProvider to generate the absolute urls that are sent in emails
- Move the existing buildInvitationUri() function to the new provider
- Add a new buildLandlordDashboardUri() function to the new provider

This does not yet add tests of the url but I have manually checked that the links in property registration confirmation, landlord registration confirmation and la user invitation emails all work (I'm working on it but these can be in a later PR)


rowanhill (approved):
:+1: Seems fairly uncontroversial, so far!

PR #245, opened 2025-02-27
Feat/prsd 792 update landlord date of birth

Kiran-R-K:
Created the landlord update date of birth step
Ensured that a landlord with their identity verified can't update their name or date of birth


src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/LandlordViewModel.kt
Kiran-R-K, line 1: I refactored this file to be in line with the pattern in `PropertyDetailsLandlordViewModel`

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/JourneyDataHelper.kt
```diff
@@ -82,6 +82,11 @@ open class JourneyDataHelper {
             )
         }
 
+        fun getValueByKey(
+            journeyData: JourneyData,
+            key: String,
+        ): String? = journeyData[key]?.toString()
+
```
AEPR, line 89: I think you can probably just do this inline in `getIsIdentityVerified`, but if you want to keep it I think it should be `getStringValueByKey`

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateLandlordDetailsJourney.kt
```diff
@@ -99,6 +106,33 @@ class UpdateLandlordDetailsJourney(
...
                 ),
             handleSubmitAndRedirect = { _, _ -> UpdateDetailsStepId.UpdateDetails.urlPathSegment },
+            nextAction = { journeyData, _ ->
+                if (UpdateLandlordDetailsJourneyDataHelper.getIsIdentityVerified(journeyData)) {
+                    Pair(UpdateDetailsStepId.UpdatePhoneNumber, null)
+                } else {
+                    Pair(UpdateDetailsStepId.UpdateDateOfBirth, null)
+                }
+            },
+            saveAfterSubmit = false,
```
AEPR, line 116: I don't think we need to repeat the check here - this can just go straight to date of birth because we can't reach the name step if the identity is verified.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/LandlordViewModel.kt
AEPR: Bit of a nitpick, but I've had a look at the `addRow` extension. Couldn't we make the "addRow" logic live on a companion factory function for `SummaryListRowViewModel`? 

Not really the focus of this PR though, so happy to leave it.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/LandlordViewModel.kt
```diff
@@ -5,127 +5,116 @@ import uk.gov.communities.prsdb.webapp.controllers.LandlordDetailsController.Com
...
+                addRow(
                     "landlordDetails.personalDetails.dateOfBirth",
                     landlord.dateOfBirth,
-                    // TODO: PRSD-792 toggleChangeLink("$UPDATE_ROUTE/date-of-birth"),
-                    null,
-                ),
-                SummaryListRowViewModel(
+                    if (!landlord.isVerified) "$UPDATE_ROUTE/date-of-birth" else null,
+                    withChangeLinks,
+                )
```
AEPR, line 63: Doing this whole refactor really pays off in making this more readable!

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/LandlordViewModel.kt
```diff
@@ -5,127 +5,116 @@ import uk.gov.communities.prsdb.webapp.controllers.LandlordDetailsController.Com
...
+                        // TODO: PRSD-688 toggleChangeLink("$UPDATE_ROUTE/address"),
+                        null,
+                        withChangeLinks,
+                    )
+                    addRow(
+                        "landlordDetails.personalDetails.englandOrWalesAddress",
+                        landlord.address.singleLineAddress,
+                        // TODO: PRSD-688 toggleChangeLink("$UPDATE_ROUTE/contact-address"),
+                        null,
+                        withChangeLinks,
```
AEPR, line 116: Can you update these comments to remove the now defunct `toggleChangeLink`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/LandlordViewModel.kt
```diff
@@ -5,127 +5,116 @@ import uk.gov.communities.prsdb.webapp.controllers.LandlordDetailsController.Com
...
-                    toggleChangeLink("$UPDATE_ROUTE/phone-number"),
-                ),
-                SummaryListRowViewModel(
+                    "$UPDATE_ROUTE/phone-number",
+                    withChangeLinks,
+                )
+                addRow(
                     "landlordDetails.personalDetails.englandOrWalesResident",
                     isEnglandOrWalesResident,
                     // TODO: PRSD-688 toggleChangeLink("$UPDATE_ROUTE/country-of-residence"),
```
AEPR, line 85: And here too.

AEPR (commented):
Couple of very small suggestions on main code. I've not looked through tests yet, but I will now as none of my suggestions would stop me approving!

src/test/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/LandlordViewModelTests.kt
```diff
@@ -323,15 +323,43 @@ class LandlordViewModelTests {
...
+
+    @Test
+    fun `LandlordViewModel populates change links in rows that should have them for a verified landlord`() {
+        // Arrange
+        val testLandlord = MockLandlordData.createLandlord(isVerified = true)
+        val changeablePersonalDetailKeys =
+            listOf(
+                "landlordDetails.personalDetails.emailAddress",
+                "landlordDetails.personalDetails.telephoneNumber",
+                "landlordDetails.personalDetails.contactAddress",
```
AEPR, line 362: I think it's probably worth commonising and parameterising these tests to reduce repetition. You can use something like:
```kotlin
@ValueSource(booleans = [true, false])
fun `Test name when` (isVerified: Boolean) {
```

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordServiceTests.kt
```diff
@@ -382,9 +386,18 @@ class LandlordServiceTests {
...
         val newAddress = createAddress("new address")
-        val updateModel = LandlordUpdateModel("newEmail", "newName", "new phone number", AddressDataModel.fromAddress(newAddress))
+        val newDateOfBirth = LocalDate.of(1992, 2, 2)
+        val updateModel =
+            LandlordUpdateModel(
+                "newEmail",
+                "newName",
+                "new phone number",
+                AddressDataModel.fromAddress(newAddress),
+                newDateOfBirth,
```
AEPR, line 399: You can declare this date inline in the `LandlordUpdateModel` constructor because we reference the `updateModel` in the assertion below.

AEPR (approved):
Two more nitpicks on the tests - other than that and my comments from earlier, LGTM.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateLandlordDetailsJourney.kt
Kiran-R-K: We just discussed this, and I will add tests to ensure that date of birth is not reachable by a verified user and then will remove the conditional here

Kiran-R-K: So doing this has re-raised an issue of `"/landlord-details/update/email"` not redirecting to `"/landlord-details/update/details"` without that redirection the 'journeyData' isn't being initialized and the check to see if the landlord is verified is always returning false. 

I think the right solution is to add a check to `getJourneyStep()` of the `LandlordDetailsController`  that will  check the journey data is initialized and if not initialize it. But I know you and @Travis-Softwire have already discussed this issue so I want to make sure this seems like the right changes to the two of you?

PR #246, opened 2025-02-28
PRSD-724: Removes Change Links from Property Details - Landlord Tab

isobel-softwire:
image


src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extenstions/SummaryListViewModelExtensionHelper.kt
```diff
@@ -12,3 +12,11 @@ fun MutableList<SummaryListRowViewModel>.addRow(
     val changeLinkOrNull = if (withChangeLinks) changeLink else null
     add(SummaryListRowViewModel(key, value, changeLinkOrNull, valueUrl))
 }
+
+fun MutableList<SummaryListRowViewModel>.addRowWithoutChangeLink(
```
Travis-Softwire, line 16: Would this not work as an overload of `addRow`? Feels a bit odd to have `addRow` and `addRowWithoutChangeLinks` when you and also add a row without change links using `addRow`

Travis-Softwire (approved):
One thought about the new `addRowWithoutChangeLinks` function otherwise LGTM 👍

PR #247, opened 2025-03-03
PRSD-NONE: Add ReadMe documentation for the journey framework

AEPR:
This honestly ended up a little longer than I was expecting, but I think I've spent enough time on it and it covers pretty much everything.

I've based it on Isobel's branch so it will be true after than merges in! (E.g. the update journey being a separate class.

For the main function names on Journey I've suggested two titles for them. One is what they currently are and the other is what will make this documentation make more sense and therefore what I'm suggesting we refactor them too when it won't cause conflicts! I'll comment there and we should resolve  that and update the documentation to match the decision before merging in!

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/ReadMe.md
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/ReadMe.md
@@ -10,12 +10,12 @@ A journey is made up of a collection of steps, each of which represents a single
 some information and the user provides it. Each has a unique ID, which is an enum value where each journey has its own 
 enum. The step is also responsible for the journey flow via the `nextAction` and `handleSubmitAndRedirect` functions. 
 
-Most of the flow is managed by the former, but the decision to save the journey to the database is determined by the 
-presence of the latter. For a simpler journeys, each step's `nextAction` would just return the fixed next step in the 
-flow with the last having a `handleSubmitAndRedirect`. 
+Most of the flow is managed by the former, but the decision to complete the overall submission of the form is determined
+by the presence of the latter. For a simpler journeys, each step's `nextAction` would just return the fixed next step in
+the flow with the last having a `handleSubmitAndRedirect`.
 Most of the journeys in the service are not that simple and have branching paths based on the users previous answers. 
 In this case the `nextAction` function looks at the data the user has submitted so far to determine which step should be
-visited next. 
+visited next.
 The `handleSubmitAndRedirect` function can also be used if the user should be "jumped" to another point in the journey
 when completing a `Step`. The main use of this currently is to skip steps that already have valid data (in update 
 journeys, see below). 
@@ -23,6 +23,12 @@ journeys, see below).
 Each step also has a `Page` class which represents the step's form page. It is responsible for providing the Thymeleaf 
 template name, the content attributes for the template and for validating the data submitted on that page. 
 The `Step` by default delegates all data validation to the `Page`, but custom validation can be set on a `Step`.
+
+### Step Details
+Where a `Step` represents a point in a journey that a user can visit, a `StepDetails` represents a concrete point in a
+particular user's traversal of a journey.
+It fully specifies both the step and subpage (see below), but also all necessary data the user has submitted to reach 
+that point in the journey.
 
 ### Subpages
 Some steps are repeatable an indeterminate number of times, for example if you need to specify all interested parties
@@ -44,22 +50,24 @@ to display tasks in logical groupings and on the pages of the journey to show th
 belongs to.
 
 ## The Journey base class
-The `Journey` class is responsible for marshalling the steps. The base class has two main functions for progressing 
-through the multi-page form:
+The `Journey` class is responsible for marshalling the steps. 
+Each final child of `Journey` represents a particular multi-page form in the service. The actual steps that make up that
+journey are specified, including the `nextAction` and `handleSubmitAndRedirect` logic for each.
+The base class has two main functions for progressing through the multi-page form:
 
-### `getPageForStep`/`populateModelAndGetViewName`
+### `populateModelAndGetViewName`
 returns the name of the ThymeLeaf template for a step and populates the model for it, or returns a 
 redirect if the requested step is unreachable. This includes validation errors for the page if data has been submitted
 for it.
 
-### `completeStep`/`updateJourneyDataAndGetViewNameOrRedirect`
+### `updateJourneyDataAndGetViewNameOrRedirect`
 validates the data submitted for the step and either adds it to the journey data and redirects to the next step or
 returns the page for the current step with validation errors. If a `handleSubmitAndRedirect` has been set for the step
 that will be called instead of moving to the next step.
 
 ### Saving journey progress
-A step can be set to `saveAfterSubmit` - if this is set to true, then whenever that step is completed a copy of the 
-`JourneyData` is saved to the database as a JSON encoded string. 
+A step can be set to `saveAfterSubmit` - if this is set to true, then whenever that step is completed a copy of the
+data representing the user's progress through the journey so far is saved to the database as a JSON encoded string. 
 That journey can be restored by calling `loadJourneyDataIfNotLoaded` on the `Journey`, which will return the data to the
 session to allow the user to resume the in progress journey.
 
@@ -75,6 +83,7 @@ each page reached in turn.
 
 ## Journey's Subclasses
 There are a couple of abstract subclasses of journey:
+
 ### The JourneyWithTaskList
 the JourneyWithTaskList adds a task-list page, which is like
 a contents page for the journey - it shows the list of all tasks in the journey along with their current status. If the
@@ -86,7 +95,3 @@ proceeds through all the steps in the usual way, the iterator takes into account
 new journey data to determine whether a step is reachable. Many of the steps set a `handleSubmitAndRedirect` to redirect
 the user to the equivalent of the "check your answers" page in other journeys. This allows us to check the new combined 
 data is valid when multiple changes are made, and to direct the user to the correct page if there is a problem. 
-
-### Final children
-Each final child of `Journey` represents a particular multi-page form in the service. The actual steps that make up that
-journey are specified, including the `nextAction` and `handleSubmitAndRedirect` logic for each. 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/ReadMe.md
```diff
@@ -0,0 +1,91 @@
...
+
+### `getPageForStep`/`populateModelAndGetViewName`
+returns the name of the ThymeLeaf template for a step and populates the model for it, or returns a 
+redirect if the requested step is unreachable. This includes validation errors for the page if data has been submitted
+for it.
+
+### `completeStep`/`updateJourneyDataAndGetViewNameOrRedirect`
+validates the data submitted for the step and either adds it to the journey data and redirects to the next step or
+returns the page for the current step with validation errors. If a `handleSubmitAndRedirect` has been set for the step
+that will be called instead of moving to the next step.
```
AEPR, line 57: Currently we use the second of these in both cases - which is descriptive of what it does internally. I think it would be better to describe what it's doing at the abstraction level above the journey (e.g. not referring to the existence of `JourneyData`).

Either way, this documentation should be consistent with code.

rowanhill (commented):
Thanks for doing this, Alex! Looks good to me, but probably worth someone else more familiar with the Journey code to review it too.

Refactoring to rename the two main public Journey methods also sounds good to me - the names at the moment aren't particularly memorable; it's probably better in this case to be more succinct even if slightly less exhaustively accurate.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/ReadMe.md
```diff
@@ -0,0 +1,91 @@
...
+enum. The step is also responsible for the journey flow via the `nextAction` and `handleSubmitAndRedirect` functions. 
+
+Most of the flow is managed by the former, but the decision to save the journey to the database is determined by the 
+presence of the latter. For a simpler journeys, each step's `nextAction` would just return the fixed next step in the 
+flow with the last having a `handleSubmitAndRedirect`. 
+Most of the journeys in the service are not that simple and have branching paths based on the users previous answers. 
+In this case the `nextAction` function looks at the data the user has submitted so far to determine which step should be
+visited next. 
+The `handleSubmitAndRedirect` function can also be used if the user should be "jumped" to another point in the journey
+when completing a `Step`.
```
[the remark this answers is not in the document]
AEPR: At the moment we only use this for the `UpdateJourney`, where you are "jumped" to change details summary page on completing a single change. It's a pretty open ended way of bypassing the "usual" flow of the journey, that we could need again in the future.

I guess the main use case I can think of would be to skip steps in journey that already have data in them. For example, if we wanted to make it so that when a user goes back using the check your answers page then only completing that task you skip back to the end if everything was complete.

I'll put that and leave it to future devs to figure out if there are any other creative applications to solve a problem they are facing.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/ReadMe.md
```diff
@@ -0,0 +1,92 @@
...
+The entire multi-page form is represented by the `Journey` class, and each point of interaction with the user is a 
+`Step`.
+
+## Steps
+A journey is made up of a collection of steps, each of which represents a single transaction where we ask the user for
+some information and the user provides it. Each has a unique ID, which is an enum value where each journey has its own 
+enum. The step is also responsible for the journey flow via the `nextAction` and `handleSubmitAndRedirect` functions. 
+
+Most of the flow is managed by the former, but the decision to save the journey to the database is determined by the 
+presence of the latter. For a simpler journeys, each step's `nextAction` would just return the fixed next step in the
```
isobel-softwire, line 14: I thought saving the journey to the database was determined by `saveAfterSubmit`? Or were you referring to creating a `Landlord` entity at the end of the landlord journey etc.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/ReadMe.md
isobel-softwire: I like the idea of using more succinct names, but I think `completeStep`'s name should indicate that its getting the next page or a redirect

isobel-softwire (approved):
This does a great job of capturing the Journey framework's behaviours - LGTM!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/ReadMe.md
Travis-Softwire: I think it's the latter, but agree it's worth clarifying as it's more that the 'result' of the journey is saved to the db by `handleSubmitAndRedirect`, whereas saving the journey itself to the db is governed by `saveAfterSubmit` as you say

AEPR: Yes, I guess that's a little confusing given that we can also save progress to the database (see "Saving journey progress" below). I'll rephrase this to something like "complete the overall submission of the form", since in theory we could be doing something other than saving to the database anyway (say if submitting the form actually sent an email to a support email address instead!).

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/ReadMe.md
Travis-Softwire: Also fine with some briefer names (and it felt gross from the start having an 'And' in the function name...) - although in the first example is 'Page' the right word here given we're not returning a Page class? Maybe `getViewForStep`? Also wonder if we should avoid the `getX` convention for a function with side effects - so maybe `prepareViewForStep`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/ReadMe.md
```diff
@@ -0,0 +1,92 @@
...
+for it.
+
+### `completeStep`/`updateJourneyDataAndGetViewNameOrRedirect`
+validates the data submitted for the step and either adds it to the journey data and redirects to the next step or
+returns the page for the current step with validation errors. If a `handleSubmitAndRedirect` has been set for the step
+that will be called instead of moving to the next step.
+
+### Saving journey progress
+A step can be set to `saveAfterSubmit` - if this is set to true, then whenever that step is completed a copy of the 
+`JourneyData` is saved to the database as a JSON encoded string.
```
Travis-Softwire, line 62: First mention of `JourneyData` - if we want to refer to it by name we should probably introduce it earlier on in the docs? Alternatively could just draft around it, e.g. 'a copy of the data representing the user's progress through the journey so far' or similar.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/ReadMe.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Saving journey progress
+A step can be set to `saveAfterSubmit` - if this is set to true, then whenever that step is completed a copy of the 
+`JourneyData` is saved to the database as a JSON encoded string. 
+That journey can be restored by calling `loadJourneyDataIfNotLoaded` on the `Journey`, which will return the data to the
+session to allow the user to resume the in progress journey.
+
+### The iterator
+Journey implements `Iterable<StepDetails>`, which represents the series of steps taken in order based on the current
+journey data. The first `StepDetails` is always the same, then the iterator validates that step and calculates the
```
Travis-Softwire, line 68: Similarly we probably need to say somewhere what a `StepDetails` is if we want to refer to it here.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/ReadMe.md
```diff
@@ -0,0 +1,92 @@
...
+
+### The UpdateJourney
+The `UpdateJourney` represents a journey for updating details and has a few idiosyncrasies. While the `nextAction` flow
+proceeds through all the steps in the usual way, the iterator takes into account a merge of original data and the
+new journey data to determine whether a step is reachable. Many of the steps set a `handleSubmitAndRedirect` to redirect
+the user to the equivalent of the "check your answers" page in other journeys. This allows us to check the new combined 
+data is valid when multiple changes are made, and to direct the user to the correct page if there is a problem. 
+
+### Final children
+Each final child of `Journey` represents a particular multi-page form in the service. The actual steps that make up that
```
Travis-Softwire, line 91: I'd move this closer to where we describe the Journey class - I think it helps give context to everything that follows that.

Travis-Softwire (approved):
Have left a few thoughts but overall looking good :) Also naming things are hard(TM)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/ReadMe.md
AEPR: I think this one is worth explaining.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/ReadMe.md
AEPR: OK, maybe this is mad, but what if we make this a type signature problem and make it so that `get***ForStep` doesn't have side effects?

I could create a data class that represents a (generic) page for the service, which can by default have a map for attributes, and a function that can apply all values from the map to a model and return the string? 

Then we don't need to pass the `Model` into the `Journey`, we get rid of some side effects and `completeStep` clearly returns a page for the controller based on typing (rather than returning a string and we need the function name to know what it's for)

rowanhill: How about we just use Spring's `ModelAndView` as the return type? I.e. we stop injecting a `Model` in the controller actions and we have `Journey` create a new `ModelAndView`, populate the attributes and set the view (or redirect) name? The controller can then just return the `ModelAndView` it receives from `Journey`.

AEPR: > How about we just use Spring's `ModelAndView`

This seems like the best choice - it's basically what I was describing but not home-rolled!

Travis-Softwire: > How about we just use Spring's `ModelAndView`

This seems sensible - if we're going to do that we should probably mirror that refactor in the `Page` class and it's subclasses, since `Journey` delegates to a very similarly named function there.

AEPR: I'm going to merge this in with original names and raise a new PR for this change.

PR #248, opened 2025-03-03
feat/PRSD-638 remove united kingdom from place names

Kiran-R-K:
Removed United Kingdom from the list of international place names

image


JasminConterioSW (approved):
LGTM :)

PR #251, opened 2025-03-04
PRSD-NONE: Switch to model and view for journey steps

AEPR:
Switches to using `ModelAndView` for journey functions and renames them.


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/Page.kt
```diff
@@ -20,55 +20,55 @@ open class Page(
     private val content: Map<String, Any>,
     val shouldDisplaySectionHeader: Boolean = false,
 ) {
-    open fun populateModelAndGetTemplateName(
+    open fun getModelAndView(
```
rowanhill, line 23: Whilst we're refactoring here, one thought: what's  your opinion on Page having a protected function called something like `enrichViewModel(model: ModelAndView, journeyData: JourneyData)`, which is either abstract or open with an empty body on Page, and which is invoked from `getModelAndView` here. That might make subclassing Page a tiny bit tidier - subclasses would override `enrichViewModel` instead of `getModelAndView`, so they wouldn't need the call to super (or the return).

rowanhill (approved):
Basically looks good to me :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/Page.kt
AEPR: Yeah, that's probably nicer - I think I'd probably go for making an `AbstractPage` class with it abstract, then a `final Page` with it having an empty body - that way anyone making a new page needs to make an explicit decision about how to impement it.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateJourney.kt
```diff
@@ -30,15 +30,14 @@ abstract class UpdateJourney<T : StepId>(
         }
     }
 
-    fun populateModelAndGetViewNameForUpdateStep(
+    fun getModelAndViewNameForUpdateStep(
```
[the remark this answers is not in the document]
AEPR: Yep, I've just misread/missed that when refactoring.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PropertyRegistrationCheckAnswersPage.kt
```diff
@@ -28,29 +28,30 @@ class PropertyRegistrationCheckAnswersPage(
...
+        model: ModelAndView,
         journeyData: JourneyData,
     ) {
         val propertyName = getPropertyName(journeyData)
         val propertyDetails = getPropertyDetailsSummary(journeyData)
 
-        model.addAttribute("propertyDetails", propertyDetails)
-        model.addAttribute("propertyName", propertyName)
-        model.addAttribute("showUprnDetail", !DataHelper.isManualAddressChosen(journeyData))
+        model.addObject("propertyDetails", propertyDetails)
```
Travis-Softwire, line 52: Minor, but for consistency should probably rename `model` to `modelAndView` here

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
```diff
@@ -37,13 +38,12 @@ class LandlordDetailsController(
     fun getUpdateUserLandlordDetails(
         model: Model,
         principal: Principal,
-    ): String {
+    ): ModelAndView {
         addLandlordDetailsToModel(model, principal, includeChangeLinks = true)
         // TODO: PRSD-355 Remove this way of showing submit button
         model.addAttribute("shouldShowSubmitButton", true)
-        return updateDetailsJourney.populateModelAndGetViewNameForUpdateStep(
+        return updateDetailsJourney.getModelAndViewNameForUpdateStep(
```
Travis-Softwire, line 45: Think Rowan has commented elsewhere, but agree we should drop the 'Name' bit of this for consistency

PR #252, opened 2025-03-04
Feat/prsd 670 landlord dashboard

JasminConterioSW:
- Populates the Landlord dashboard
    - Only included buttons when the page they link to already exists
    - Link text is always included but it's only a link if the content it goes to exists
   
- Link to the dashboard from 
    - Register property confirmation page
    - Register landlord confirmation page
    - Landlord Details (back link)
    - Property details (back link)

- Add tests
    - Controller tests (added a few more for the LA version too)
    - Integration tests checking content and links
    - MVC tests checking the links to the dashboard from the landlord and property registration confirmation emails
    
    
image


src/test/kotlin/uk/gov/communities/prsdb/webapp/mockObjects/JourneyDataBuilder.kt
```diff
@@ -91,8 +91,7 @@ class JourneyDataBuilder(
                 LandlordRegistrationStepId.Email.urlPathSegment to mapOf("emailAddress" to "test@example.com"),
                 LandlordRegistrationStepId.PhoneNumber.urlPathSegment to mapOf("phoneNumber" to "07123456789"),
                 LandlordRegistrationStepId.CountryOfResidence.urlPathSegment to mapOf("livesInEnglandOrWales" to true),
-                LandlordRegistrationStepId.LookupAddress.urlPathSegment to
-                    mapOf("address" to mapOf("houseNameOrNumber" to "44", "postcode" to "EG1 1GE")),
+                LandlordRegistrationStepId.LookupAddress.urlPathSegment to mapOf("houseNameOrNumber" to "44", "postcode" to "EG1 1GE"),
```
JasminConterioSW, line 94: The extra layer in here was making the LookupAddress step fail validation, so it didn't go to the next page (SelectAddress) and add it to the filtered journey data

src/test/kotlin/uk/gov/communities/prsdb/webapp/urlProviders/LandlordDashboardUrlTests.kt
```diff
@@ -0,0 +1,190 @@
...
+    lateinit var mockAddressDataService: AddressDataService
+
+    @MockBean
+    lateinit var mockLocalAuthorityService: LocalAuthorityService
+
+    @MockBean
+    lateinit var mockPropertyOwnershipService: PropertyOwnershipService
+
+    @MockBean
+    lateinit var mockPropertyRegistrationService: PropertyRegistrationService
```
isobel-softwire, line 85: super minor but these can all be private

src/test/kotlin/uk/gov/communities/prsdb/webapp/urlProviders/LandlordDashboardUrlTests.kt
```diff
@@ -0,0 +1,190 @@
...
+                name = anyOrNull(),
+                email = anyOrNull(),
+                phoneNumber = anyOrNull(),
+                addressDataModel = anyOrNull(),
+                countryOfResidence = anyOrNull(),
+                isVerified = anyOrNull(),
+                nonEnglandOrWalesAddress = anyOrNull(),
+                dateOfBirth = anyOrNull(),
+            ),
+        ).thenReturn(landlord)
```
isobel-softwire, line 108: Why do we need to use `anyOrNull` here?

src/test/kotlin/uk/gov/communities/prsdb/webapp/urlProviders/LandlordDashboardUrlTests.kt
```diff
@@ -0,0 +1,190 @@
...
+
+        whenever(mockAbsoluteUrlProvider.buildLandlordDashboardUri())
+            .thenCallRealMethod()
```
isobel-softwire, line 111: Do we need to mock the absolute url provider if we're using its real methods?

src/main/resources/templates/fragments/dashboardAction.html
```diff
@@ -1,7 +1,13 @@
...
-    class="govuk-heading-m govuk-link">linkTextKey</a>
+    <a th:unless="${linkUrl} == '#'"
+       th:href="${linkUrl}"
+       th:text="#{${linkTextKey}}"
+        class="govuk-heading-m govuk-link">linkTextKey</a>
+    <span th:if="${linkUrl} == '#'"
+          th:text="#{${linkTextKey}}"
+          class="govuk-heading-m">
+        linkTextKey
+    </span>
```
rowanhill, line 11: Are you just checking for "#" to avoid making it a link if it's a TODO? I'm not sure I'd bother with that - we'll just forget that this is here and leave in some complexity in the code that we don't need.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordDashboardTests.kt
```diff
@@ -0,0 +1,51 @@
...
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.LandlordDetailsPage
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.basePages.BasePage.Companion.assertPageIs
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.propertyRegistrationJourneyPages.RegisterPropertyStartPage
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+@Sql("/data-local.sql")
+class LandlordDashboardTests : IntegrationTest() {
+    @Test
+    fun `the dashboard loads displaying the user's name and lrn`(page: Page) {
```
rowanhill, line 16: Does this actually need the page param?

src/test/kotlin/uk/gov/communities/prsdb/webapp/urlProviders/LandlordDashboardUrlTests.kt
```diff
@@ -0,0 +1,190 @@
...
+        mvc
+            .post("/$REGISTER_LANDLORD_JOURNEY_URL/${LandlordRegistrationStepId.Declaration.urlPathSegment}") {
+                contentType = MediaType.APPLICATION_FORM_URLENCODED
+                content = encodedDeclarationContent
+                with(csrf())
+            }.andExpect { status { is3xxRedirection() } }
+
+        mvc
+            .get(confirmationCaptor.firstValue.prsdURL)
+            .andExpect { status { is3xxRedirection() } }
```
rowanhill, line 132: Is expecting a 3xx redirection response good enough? Does it not matter where the redirect is to? E.g. isn't that what we'd get if we requested any URL if we're not authenticated?

JasminConterioSW: True, I'll add in a check for the redirected url.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordDashboardTests.kt
JasminConterioSW: No, they almost all do so it tends to just go in by default... I'll take it out.

src/test/kotlin/uk/gov/communities/prsdb/webapp/urlProviders/LandlordDashboardUrlTests.kt
JasminConterioSW: anyOrNull() is safer...

any() doesn't include null, so if you used any() when defining your stub but get a null parameter passed to your function as the test runs, it doesn't match it with the stub so doesn't return the value you want.

If the argument to the function is not nullable we could use any(), but anyOrNull() will work either way.

src/test/kotlin/uk/gov/communities/prsdb/webapp/urlProviders/LandlordDashboardUrlTests.kt
JasminConterioSW: We need a bean, but could use a Spybean here instead of a Mockbean

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
```diff
@@ -54,6 +56,7 @@ class LandlordDetailsController(
         principal: Principal,
     ): String {
         addLandlordDetailsToModel(model, principal, includeChangeLinks = false)
+        model.addAttribute("registeredPropertiesUrlSegment", REGISTERED_PROPERTIES_PATH_SEGMENT)
```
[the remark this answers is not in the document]
JasminConterioSW: Good catch

src/main/resources/messages.properties
```diff
@@ -398,6 +398,27 @@ localAuthority.dashboard.action.rentersRightsBill.descriptionText=View the guide
...
+landlord.dashboard.action.policyUpdates.linkText=Policy updates TODO: Link to content
+landlord.dashboard.action.policyUpdates.descriptionText=See how the Renters' Rights Bill is progressing and what that means for the database
+landlord.dashboard.action.privacyNotice.linkText=How your information is used TODO PRSD-676: Link to content
+landlord.dashboard.action.privacyNotice.descriptionText=Read the Privacy Notice to understand how your information is used on the pilot
+landlord.dashboard.guidance.heading=Guidance for landlords
+landlord.dashboard.action.howToLet.linkText=How to let TODO: link to content
+landlord.dashboard.action.howToLet.descriptionText=Understand what responsibilities you have as a landlord
+landlord.dashboard.action.rentersRightsBill.linkText=Renters' Rights Bill
+landlord.dashboard.action.rentersRightsBill.descriptionText=View the guidelines related to the database and the Renters' Rights Bill
+landlord.dashboard.action.keepingPropertyCompliant.linkText=Keep you property compliant TODO: link to content
```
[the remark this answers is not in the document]
JasminConterioSW: There aren't for most of them, hopefully it'll be clear from the page that we need to return to it!

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDashboardController.kt
```diff
@@ -6,27 +6,58 @@ import org.springframework.ui.Model
...
         model: Model,
-        principale: Principal,
+        principal: Principal,
     ): String {
-        model.addAttribute("title", "Landlord Dashboard")
+        val landlord =
+            landlordService.retrieveLandlordByBaseUserId(principal.name)
+                ?: throw PrsdbWebException("User ${principal.name} is not registered as a landlord")
+
+        model.addAttribute("title", "dashboard.title")
```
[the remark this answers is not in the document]
JasminConterioSW: Yep, I'll update the localAuthorityDashboard to match

src/main/resources/templates/fragments/dashboardAction.html
JasminConterioSW: That was the plan, @AEPR suggested making the links spans if they didn't go anywhere so he might also have a preference!

Is it nicer to make them not be links now that it goes on an integration environment where MadeTech / MHCLG can play with the site? It makes it very clear that we don't expect it to go anywhere. Although maybe the big "TODO..." text is enough!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
```diff
@@ -610,7 +610,7 @@ class PropertyRegistrationJourney(
 
             journeyDataService.deleteJourneyData()
 
-            return CONFIRMATION_PAGE_PATH_SEGMENT
+            return "/$REGISTER_PROPERTY_JOURNEY_URL/$CONFIRMATION_PAGE_PATH_SEGMENT"
```
Travis-Softwire, line 613: Did the relative URL not work here?

src/main/resources/templates/fragments/dashboardAction.html
```diff
@@ -1,7 +1,13 @@
-<div th:fragment="dashboardAction(linkTextKey, linkUrl,descriptionKey)"
+<div th:fragment="dashboardAction(linkText, linkUrl,descriptionKey)"
```
Travis-Softwire, line 1: V. minor, inconsistent space vs no space between args.

Travis-Softwire (approved):
A couple of v. minor points but LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
JasminConterioSW: It did, but we had the full path for the LandlordRegistrationJourney, I could make them both just go to CONFIRMATION_PAGE_PATH_SEGMENT instead?

src/main/resources/templates/landlordDetailsView.html
```diff
@@ -34,7 +35,7 @@ <h2 class="govuk-heading-m" th:text="#{landlordDetails.personalDetails.optionalC
                     </th:block>
                 </div>
 
-                <div th:replace="~{fragments/tabs/tabsPanel :: tabsPanel('registered-properties', ~{:: #registered-properties-panel-content})}">
+                <div th:replace="~{fragments/tabs/tabsPanel :: tabsPanel(${registeredPropertiesUrlSegment}, ~{:: #registered-properties-panel-content})}">
```
AEPR, line 38: I don't think we should refer to this as a URL segment - it's actually an html ID. I think this could either be left as a string literal or in the controller this model attribute should have a more appropriate narrow name.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
AEPR: I think we should prefer relative paths where possible, so yes please.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDashboardController.kt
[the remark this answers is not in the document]
AEPR: Looks like the attribute is still being set here even though it isn't being used in the template.

src/main/resources/templates/localAuthorityLandlordDetailsView.html
```diff
@@ -17,7 +17,7 @@ <h1 class="govuk-heading-l" th:text="${name}">name</h1>
                     <th:block id="tab-headings">
                         <li th:replace="~{fragments/tabs/tabsListItem :: tabsListItem('personal-details', #{landlordDetails.personalDetails.heading})}">
                         </li>
-                        <li th:replace="~{fragments/tabs/tabsListItem :: tabsListItem('registered-properties', #{landlordDetails.registeredProperties.heading})}">
+                        <li th:replace="~{fragments/tabs/tabsListItem :: tabsListItem(${registeredPropertiesUrlSegment}, #{landlordDetails.registeredProperties.heading})}">
```
AEPR, line 20: Same note here about this not being a url segment.

AEPR (commented):
Couple of quibbles, but otherwise looking good.

Certainly very happy with the invitation URL tests.

src/main/resources/templates/fragments/dashboardAction.html
JasminConterioSW: I've kept it in but added a comment in the template to explain why it is there

src/main/resources/templates/landlordDetailsView.html
JasminConterioSW: I'd prefer it not to be a string literal because we do link directly to it from the View Properties button, which could be easy to miss if it changed.

I've gone with `registeredPropertiesTabId`

PR #255, opened 2025-03-04
PRSD-NONE: Passes Step Path Segments into Journey Methods

isobel-softwire:
- Passes step path segments instead of IDs into journey methods
- Updates affected tests

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
@@ -108,7 +108,7 @@ abstract class Journey<T : StepId>(
         }
 
         if (currentStep.handleSubmitAndRedirect != null) {
-            return "redirect:${currentStep.handleSubmitAndRedirect!!(newJourneyData, subPageNumber)}"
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -119,7 +108,7 @@ abstract class Journey<T : StepId>(
         }
 
         if (currentStep.handleSubmitAndRedirect != null) {
-            return "redirect:${currentStep.handleSubmitAndRedirect!!(newJourneyData, subPageNumber)}"
+            return "redirect:${currentStep.handleSubmitAndRedirect.invoke(newJourneyData, subPageNumber)}"
```
Travis-Softwire, line 111: Is this change related to the refactor?

Travis-Softwire (approved):
One Q otherwise LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
isobel-softwire: No - Intellij kept giving a warning about the not-null assertion so I changed it

PR #256, opened 2025-03-05
PRSD-NONE: Creates Journey Factories

isobel-softwire:
- Moves abstract attributes into `Journey` constructors where possible
- Uses `journeyPathSegment` as each journey's `journeyDataKey`
- Bases `UpdateJourney.originalDataKey` on `journeyPathSegment` to allow for parallel update journeys of the same kind
- Creates a factory for each journey type
- Updates `ReadMe` to mention journey factories
- Updates affected tests


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -16,26 +16,19 @@ import java.security.Principal
 import java.util.Optional
 
 abstract class Journey<T : StepId>(
-    protected val journeyType: JourneyType,
+    private val journeyType: JourneyType,
+    protected val journeyPathSegment: String,
+    val initialStepId: T,
```
AEPR, line 21: Not hugely against it, but how come we're moving this `initialStep` to the constructor over an abstract value?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
```diff
@@ -33,6 +33,8 @@ class RegisterLAUserController(
         // see https://github.com/spring-projects/spring-hateoas/issues/155 for details
         if (invitationService.tokenIsValid(token)) {
             invitationService.storeTokenInSession(token)
+
+            val laUserRegistrationJourney = laUserRegistrationJourneyFactory.create()
             laUserRegistrationJourney.initialiseJourneyData(token)
```
AEPR, line 39: The factory should be responsible for all initialisation. Having looked at was this does, it's just pre-populating the email address in the journey data rather than really "initialising" the journey data.

I think there are a couple of things we could reasonably do:
* Rename the function to make it clearer that it's not crucial for the journey to work correctly - it's just for convenience
* Make the factory take the token as a parameter on its `create` function, using the token from the session for all other steps (and make the function idempotent on session, so if you change the email address that still works)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateJourney.kt
```diff
@@ -34,10 +32,9 @@ abstract class UpdateJourney<T : StepId>(
         updateEntityId: String,
         subPageNumber: Int? = null,
         submittedPageData: PageData? = null,
-        journeyDataKey: String? = null,
     ): ModelAndView {
-        initialiseJourneyDataIfNotInitialised(updateEntityId, journeyDataKey)
-        return super.getModelAndViewForStep(updateStepId.urlPathSegment, subPageNumber, submittedPageData, journeyDataKey)
+        initialiseJourneyDataIfNotInitialised(updateEntityId)
```
AEPR, line 36: This is initialisation, so it should be happening in the journey constructor - the update entity can be a parameter passed to the factory's `create`.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateJourney.kt
```diff
@@ -65,8 +62,8 @@ abstract class UpdateJourney<T : StepId>(
 
     private fun isJourneyDataInitialised(journeyData: JourneyData): Boolean = journeyData.containsKey(originalDataKey)
 
-    protected fun isJourneyDataInitialised(journeyDataKey: String? = null): Boolean {
-        val journeyData = journeyDataService.getJourneyDataFromSession(journeyDataKey ?: defaultJourneyDataKey)
+    protected fun isJourneyDataInitialised(): Boolean {
+        val journeyData = journeyDataService.getJourneyDataFromSession(journeyPathSegment)
         return isJourneyDataInitialised(journeyData)
```
AEPR, line 67: We also shouldn't need to be checking if the journey data is initialised any more - it always is! (Once you do it in the constructor)

AEPR (changes requested):
Couple of small notes.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
isobel-softwire: If it's an abstract value, we have to make it final in final journey children to prevent warnings when we use it in `sections`. This avoids that (not that it makes any real difference).

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateJourney.kt
isobel-softwire: I think we'll still need this so we don't overwrite journey data once it has been initialized once

AEPR: Yes, you are of course completely correct. Oops.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
isobel-softwire: I've gone with the second option here

Travis-Softwire:
Came to make the same comment as Alex about moving initialisation into constructors and having any params be arguments of `create()`. Won't submit a separate review, but LGTM once those points have been addressed :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LaUserRegistrationJourneyFactory.kt
```diff
@@ -0,0 +1,17 @@
+package uk.gov.communities.prsdb.webapp.forms.journeys
```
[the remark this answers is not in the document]
AEPR: I think this seems sensible.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
```diff
@@ -108,6 +114,15 @@ class RegisterLAUserController(
...
 
+    private fun getValidTokenFromSessionOrNull(): String? {
+        val token = invitationService.getTokenFromSession()
+        return if (token == null || !invitationService.tokenIsValid(token)) {
+            null
+        } else {
+            token
+        }
+    }
+
```
AEPR, line 125: This function certainly belongs on the invitation service, and all this token manipulation can probably be a little less convoluted. For instance - why are we passing this token around everywhere and constantly converting back and forth to an invitation.

In my opinion:
* the `LaUserRegistrationJourney` should take a `LocalAuthorityInvitation` as a constructor parameter (indeed as a `val` parameter)
* The logic currently in the `initializeJourneyDataIfNotInitialized` function should be called in the constructor of the journey, not the factory - and can just use the invitation passed into the constructor
* `checkAnswersHandleSubmitAndRedirect` can just use the invitation passed in the constructor 

I think there's also room for a small refactor of the `LocalAuthorityInvitationService` itself to simplify its API, but that's probably a little beyond the scope of this PR. (And will be simpler after it's been merged.)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LaUserRegistrationJourney.kt
```diff
@@ -43,15 +42,19 @@ class LaUserRegistrationJourney(
...
-        journeyDataService.setJourneyDataInSession(newJourneyData)
+    fun initializeJourneyDataIfNotInitialized(token: String) {
+        val journeyData = journeyDataService.getJourneyDataFromSession(journeyDataKey)
+        if (!isJourneyDataInitialized(journeyData)) {
+            val emailStep = steps.single { step -> step.id == RegisterLaUserStepId.Email }
+            val emailFormData = mapOf("emailAddress" to invitationService.getEmailAddressForToken(token))
+            val newJourneyData = emailStep.updatedJourneyData(journeyData, emailFormData, subPageNumber = null)
+            journeyDataService.setJourneyDataInSession(newJourneyData)
+        }
     }
```
AEPR, line 53: This type of method should be called in the constructor, not the factory.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordDetailsUpdateJourneyFactory.kt
```diff
@@ -15,12 +15,16 @@ class LandlordDetailsUpdateJourneyFactory(
...
+    fun create(landlordBaseUserId: String): UpdateLandlordDetailsJourney {
+        val landlordDetailsUpdateJourney =
+            UpdateLandlordDetailsJourney(
+                validator,
+                journeyDataService,
+                addressLookupService,
+                landlordService,
+                addressDataService,
+            )
+        landlordDetailsUpdateJourney.initializeJourneyDataIfNotInitialized(landlordBaseUserId)
```
AEPR: This call should happen in the constructor in the Journey subclass, rather than here in the factory.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourneyFactory.kt
```diff
@@ -24,16 +24,20 @@ class PropertyRegistrationJourneyFactory(
...
+                journeyDataService,
+                addressLookupService,
+                addressDataService,
+                propertyRegistrationService,
+                localAuthorityService,
+                landlordService,
+                absoluteUrlProvider,
+                confirmationEmailSender,
+            )
+        propertyRegistrationJourney.loadJourneyDataIfNotLoaded(principalName)
```
AEPR: This should be in the constructor in  the Journey subclass, not here in the factory.

AEPR (changes requested):
A few more small changes around where we call initialisation logic (and some stuff about the invitation token).

We don't want the journey classes to need to be initialised separately to being constructed, even if the factory is responsible for both - it means that some future uninformed dev could construct them and not initiatise without using the factory.

The constructor should specify all of the information and services necessary to initialise the journey, and the factory is responsible for pulling together the services (from dependency injection) and information (from the parameters of `create`).

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LaUserRegistrationJourney.kt
```diff
@@ -86,7 +85,7 @@ class LaUserRegistrationJourney(
                             "fieldSetHint" to "forms.name.fieldSetHint",
                             "label" to "forms.name.label",
                             "submitButtonText" to "forms.buttons.continue",
-                            "backUrl" to "/$journeyPathSegment/",
+                            "backUrl" to "/$REGISTER_LA_USER_JOURNEY_URL/",
```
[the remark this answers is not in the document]
isobel-softwire: You're right - removed 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
isobel-softwire: This all sounds sensible to me. Would you mind writing a ticket for the `LocalAuthorityInvitationService` refactor?

AEPR: Yep agree with this - the refactor I'm imagining (if it happens) would just mean doing the checks in the `InvitationService`, without exposing the token. Basically make most of the functions in there private and have a nice `getValidInvitationOrNull` function that we then use in every request

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
```diff
@@ -66,13 +68,22 @@ class RegisterLAUserController(
...
-            formData,
-            subpage,
-            principal,
-        )
+    ): ModelAndView {
+        val token = getValidTokenFromSessionOrNull()
+        if (token == null) {
+            invitationService.clearTokenFromSession()
+            return ModelAndView("redirect:$INVALID_LINK_PAGE_PATH_SEGMENT")
+        }
```
JasminConterioSW, line 76: This bit can probably be a little function `redirectIfTokenIsInvalid`, I think it's the same as we call in getJourneyStep.

(It might be overkill to call it when we get and post every page in the journey - the most important is when the final page gets submitted, the second most important is when the user loads the landing page)

isobel-softwire: I agree that it's overkill. Ideally there'll be an `invitationService.getInvitationFromTokenInSession` method we can use instead of checking the token in each endpoint

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
AEPR: Ticket raised as PRSD-972 (draft).

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
AEPR: Yeah, I think this can fold into PRSD-972 rather than this PR!

PR #257, opened 2025-03-05
PRSD-NONE: Restructures Test Helpers

isobel-softwire:
- Creates `testHelpers` folder with the following structure:
  - `testHelpers`
    - `builders`
    - `extensions`
    - `mockObjects`
    - `validators`
- Gives main and test `form` folders same structure


src/test/kotlin/uk/gov/communities/prsdb/webapp/testHelpers/validators/AlwaysTrueValidator.kt
```diff
@@ -1,4 +1,4 @@
-package uk.gov.communities.prsdb.webapp.helpers
+package uk.gov.communities.prsdb.webapp.testHelpers.validators
```
AEPR, line 1: I think this can go in `mockObjects` - it's a "real" validator, but it's just there to be passed into journeys to not worry about validation! That way we can get rid of the `validators` package here which I don't think is a separate category for `testHelpers`.

PR #258, opened 2025-03-05
PRSD-670: Update back link from property registration

JasminConterioSW:
Point the back link from the property registration start page to the landlord dashboard


Kiran-R-K (approved):
LGTM

PR #259, opened 2025-03-05
feat/PRSD-896 update number of households validation

Kiran-R-K:
changed the validation on number of household for the property registration to be a positive-integer validator rather than a positive-or-zero-integer validator

Note: the message key does not need to be updated

image

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/PositiveIntegerValidator.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/PositiveIntegerValidator.kt
@@ -4,11 +4,7 @@ import org.hibernate.validator.internal.constraintvalidators.bv.number.sign.Posi
 
 class PositiveIntegerValidator : PropertyConstraintValidator {
     override fun isValid(value: Any?): Boolean {
-        try {
-            val integerValue = value.toString().toInt()
-            return PositiveValidatorForInteger().isValid(integerValue, null)
-        } catch (e: NumberFormatException) {
-            return false
-        }
+        val integerValue = value.toString().toIntOrNull() ?: return false
+        return PositiveValidatorForInteger().isValid(integerValue, null)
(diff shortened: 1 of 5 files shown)


isobel-softwire (approved):
LGTM!

Travis-Softwire:
Worth adding a test (or adding to an existing test) to check for the `0` edge case?

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/PositiveIntegerValidator.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.hibernate.validator.internal.constraintvalidators.bv.number.sign.PositiveValidatorForInteger
+
+class PositiveIntegerValidator : PropertyConstraintValidator {
+    override fun isValid(value: Any?): Boolean {
+        try {
+            val integerValue = value.toString().toInt()
+            return PositiveValidatorForInteger().isValid(integerValue, null)
+        } catch (e: NumberFormatException) {
+            return false
+        }
```
AEPR, line 12: I think it would be better to avoid using a `try-catch` block to map to `true` or `false` if possible. I'd use `toIntOrNull` and check if it's returned null instead.

AEPR (changes requested):

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/PositiveIntegerValidator.kt
isobel-softwire: That sounds good to me. I think the same change should be made in the other numerical validators (`DateValidator` and `PositiveOrZeroIntegerValidator`) as well

AEPR:
I accidentally pushed a change to this branch that I didn't intend to - so I force-pushed to revert it back to the correct commit.

PR #260, opened 2025-03-07
PRSD-NONE: Clarifies when JourneyData is filtered

isobel-softwire:


AEPR (approved):
This seems like a good change to make, and all LGTM.

If we wanted to go further we could maybe do something to be a little stricter with the types of JourneyData, PageData and FilteredJourneyData. If we updated JourneyData to have:
```kotlin
interface JourneyData : Map<String, Any?>
interface FilteredJourneyData : Map<String, Any?>
interface PageData : Map<String, Any?>
```

Instead of type aliases, then we could still freely convert between them with casts but we'd have to be explicit about it.

Certainly beyond the scope of this PR, but an idea inspired by it - if it's important that we keep it clear which of these we're using at a given time we should maybe use the type system over variable names. Maybe one for tech catch up this week?

isobel-softwire:
> This seems like a good change to make, and all LGTM.
> 
> If we wanted to go further we could maybe do something to be a little stricter with the types of JourneyData, PageData and FilteredJourneyData. If we updated JourneyData to have:
> 
> ```kotlin
> interface JourneyData : Map<String, Any?>
> interface FilteredJourneyData : Map<String, Any?>
> interface PageData : Map<String, Any?>
> ```
> 
> Instead of type aliases, then we could still freely convert between them with casts but we'd have to be explicit about it.
> 
> Certainly beyond the scope of this PR, but an idea inspired by it - if it's important that we keep it clear which of these we're using at a given time we should maybe use the type system over variable names. Maybe one for tech catch up this week?

I like this idea - I'll raise it in the next catch-up 👍

PR #261, opened 2025-03-10
PRSD-670: Replace link to external content

JasminConterioSW:
The original link contains a uuid (this was a link from a ticket), but clicking on it redirects you to a page without a uuid.
It feels a bit cleaner to just use the link without the uuid!


isobel-softwire (approved):
LGTM!

PR #263, opened 2025-03-10
Feat/prsd 696 property deregistration are you sure

JasminConterioSW:
- Creates the property deregistration journey and controller
    - Check that the logged in user is allowed to delete this property before showing them the first step of the journey
    - Include the propertyOwnershipId in the journeyDataKey to isolate deregistration journeys for different properties
    - **Some of the content for one of the steps (i.e. the address) comes from the database, so the database was getting called every time the journey was created (which is now fairly often). I've added something to cache the address in the session and use that if it's available instead of calling the database every time**
        - The cached address is cleared from the session if the user exits the journey
- Creates the "are you sure" step
    - If the user submits "yes", redirect to next (placeholder) step
    - If the user submits "no", return to property details page
    - If the user submits without selecting an answer, show a validation error
- Link the delete button on the property details page to the initial step of the property deregistration journey

Tests
- Adds the usual controller and integration tests
- **Refactor integration tests so that we can check that variable parameters (e.g. propertyOwnershipId) appear in the url.** Included this check in property details tests as well as the property deregistration journey tests

image
image


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyController.kt
```diff
@@ -0,0 +1,90 @@
...
+    fun getJourneyStep(
+        @PathVariable("stepName") stepName: String,
+        @PathVariable("propertyOwnershipId") propertyOwnershipId: Long,
+        @RequestParam(value = "subpage", required = false) subpage: Int?,
+        model: Model,
+        principal: Principal,
+    ): ModelAndView {
+        if (stepName == initialStepId.urlPathSegment) {
+            if (!propertyOwnershipService.getIsAuthorizedToDeleteRecord(propertyOwnershipId, principal.name)) {
+                addressDataService.clearCachedSingleLineAddressForPropertyOwnershipId(propertyOwnershipId)
```
JasminConterioSW, line 42: I clear the cache here because if an unauthorised user navigates directly to a later step, the journey gets created and address is stored in the cache, then the user gets redirected to the first step, then they hit this if block and get the "Not found" error.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
```diff
@@ -0,0 +1,127 @@
...
+            id = DeregisterPropertyStepId.AreYouSure,
+            page =
+                Page(
+                    formModel = PropertyDeregistrationAreYouSureFormModel::class,
+                    templateName = "forms/areYouSureForm",
+                    content =
+                        mapOf(
+                            "title" to "deregisterProperty.title",
+                            "fieldSetHeading" to "deregisterProperty.areYouSure.fieldSetHeading",
+                            "propertyAddress" to retrieveAddressFromCacheOrDatabase(),
```
AEPR, line 54: I think adding a session cache for this is excessive in this case and adds a lot of extra complexity. We needed it especially for the "property already registered" case because it was being accessed every time we interacted with what is now the `StepDetailsIterator`. In general, we shouldn't be afraid of retrieving the data we need to service a request from the database in the request in which we need it.

I think it would be quite a lot cleaner to have this `Page` be responsible for retrieving the address from the database.

You can subclass abstract page (passing in either the `propertYOwnershipService` or a lambda that calls it) and override `enrichModel` to retrieve the single line address and add it to the model. That way we retrieve the address data when we need it.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -101,18 +102,26 @@ abstract class Journey<T : StepId>(
...
+        val redirectUrl = getRedirectForNextStep(currentStep, newJourneyData, subPageNumber)
+        return ModelAndView("redirect:$redirectUrl")
+    }
+
+    protected fun getRedirectForNextStep(
+        currentStep: Step<T>,
+        newJourneyData: JourneyData,
+        subPageNumber: Int?,
+    ): String {
+        val (newStepId: T?, newSubPageNumber: kotlin.Int?) = currentStep.nextAction(newJourneyData, subPageNumber)
```
AEPR, line 115: We shouldn't need to specify `kotlin.Int?` - just `Int?` should be enough.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
```diff
@@ -0,0 +1,127 @@
...
+                            "title" to "deregisterProperty.title",
+                        ),
+                ),
+        )
+
+    private fun areYouSureContinueToNextActionOrExitJourney(
+        newJourneyData: JourneyData,
+        subPageNumber: Int?,
+    ): String {
+        val currentStep = areYouSureStep()
```
AEPR, line 93: Because this is a function call you actually have a different instance of the step to the one that has just called this function during `handleSubmitAndRedirect`, which is a bit convoluted. There are a couple of fixes for this, but we might be able to just avoid it.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
```diff
@@ -0,0 +1,127 @@
...
+
+        if (getWantsToProceed(newJourneyData)!!) {
+            return getRedirectForNextStep(currentStep, newJourneyData, subPageNumber)
```
AEPR, line 96: This is always going to redirect to the same place, right? So why are we creating and evaluating the `nextAction` function when we can just return `"redirect:${DeregisterPropertyStepId.Reason.urlPathSegment"`. 

I appreciate that in this case we want to ensure that changing one changes both, but I think there are better and clearer ways to do that than this!

Happy to have a call about it to come up with something better if you like.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
```diff
@@ -0,0 +1,127 @@
...
+            ?.property
+            ?.address
+            ?.singleLineAddress ?: throw ResponseStatusException(
+            HttpStatus.NOT_FOUND,
+            "Address for property ownership id $propertyOwnershipId not found",
+        )
+
+    companion object {
+        val initialStepId = DeregisterPropertyStepId.AreYouSure
+    }
```
AEPR, line 126: Why are we specifying this here in the companion? For all the other journeys we just pass the literal into the `super.constructor`?

(And also reference the literal directly when needing to send someone to the start of the journey)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyController.kt
```diff
@@ -0,0 +1,90 @@
...
+        fun getPropertyDeregistrationPath(
+            propertyOwnershipId: Long,
+            propertyDeregistrationJourneyFactory: PropertyDeregistrationJourneyFactory,
+        ): String {
+            val initialStepPathSegment = propertyDeregistrationJourneyFactory.create(propertyOwnershipId).initialStepId.urlPathSegment
+
+            return UriTemplate("$PROPERTY_DEREGISTRATION_ROUTE/$initialStepPathSegment")
+                .expand(propertyOwnershipId)
+                .toASCIIString()
+        }
```
isobel-softwire, line 90: I don't think we should be creating a journey here just to get its initial step. Can't we just access the enum directly?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
```diff
@@ -47,7 +50,7 @@ class PropertyDetailsController(
 
         model.addAttribute("propertyDetails", propertyDetails)
         model.addAttribute("landlordDetails", landlordViewModel.landlordsDetails)
-        model.addAttribute("deleteRecordLink", "delete-record")
+        model.addAttribute("deleteRecordLink", getPropertyDeregistrationPath(propertyOwnershipId, propertyDeregistrationJourneyFactory))
```
isobel-softwire, line 53: This is just personal preference, but I think we should qualify methods from other controllers

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
isobel-softwire: Is it worth creating a new `Page` class if all we're adding to the model is an address? We could just retrieve the address from `propertyOwnershipService` here (that is, if the steps stay as methods instead of attributes).

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
```diff
@@ -0,0 +1,127 @@
...
+                            "radioOptions" to
+                                listOf(
+                                    RadiosButtonViewModel(
+                                        value = true,
+                                        labelMsgKey = "forms.radios.option.yes.label",
+                                    ),
+                                    RadiosButtonViewModel(
+                                        value = false,
+                                        labelMsgKey = "forms.radios.option.no.label",
+                                    ),
```
isobel-softwire, line 64: Might be worth setting `valueStr` for these `RadioButtonViewModel`s so the html IDs say yes/no instead of true/false

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
```diff
@@ -0,0 +1,127 @@
...
+                                ),
+                            "backUrl" to getPropertyDetailsPath(propertyOwnershipId),
```
isobel-softwire, line 66: You can use the `BACK_URL_ATTR_NAME` constant here

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
JasminConterioSW: Ah you're right, we're not using this any more - it went in before the journey refactor that used factories

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
```diff
@@ -0,0 +1,127 @@
...
+                    templateName = "forms/deregistrationReasonForm",
+                    content =
+                        mapOf(
+                            "title" to "deregisterProperty.title",
+                        ),
+                ),
+        )
+
+    private fun areYouSureContinueToNextActionOrExitJourney(
+        newJourneyData: JourneyData,
```
isobel-softwire, line 90: why call this `newJourneyData` instead of `journeyData`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
```diff
@@ -0,0 +1,127 @@
...
+        subPageNumber: Int?,
+    ): String {
+        val currentStep = areYouSureStep()
+
+        if (getWantsToProceed(newJourneyData)!!) {
+            return getRedirectForNextStep(currentStep, newJourneyData, subPageNumber)
+        }
+
+        addressDataService.clearCachedSingleLineAddressForPropertyOwnershipId(propertyOwnershipId)
+        return getPropertyDetailsPath(propertyOwnershipId)
```
isobel-softwire, line 100: Same point here, about qualifying controller companion methods

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/PropertyDeregistrationJourneyDataHelper.kt
```diff
@@ -0,0 +1,15 @@
...
+class PropertyDeregistrationJourneyDataHelper : JourneyDataHelper() {
+    companion object {
+        fun getWantsToProceed(journeyData: JourneyData) =
+            getFieldBooleanValue(
+                journeyData,
+                DeregisterPropertyStepId.AreYouSure.urlPathSegment,
+                "wantsToProceed",
+            )
+    }
+}
```
isobel-softwire, line 15: We're moving from helper to extension methods for journey data, so it would be helpful if new helpers were implemented as extensions.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -88,6 +88,11 @@ class PropertyOwnershipService(
         baseUserId: String,
     ): Boolean = propertyOwnership.primaryLandlord.baseUser.id == baseUserId
 
+    fun getIsAuthorizedToDeleteRecord(
+        propertyOwnershipId: Long,
+        baseUserId: String,
+    ): Boolean = getIsPrimaryLandlord(getPropertyOwnership(propertyOwnershipId), baseUserId)
```
isobel-softwire, line 94: We're going to need this method for the property update journey as well. Would you mind making it an overload of `getIsPrimaryLandlord` so we can use it more generally?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
JasminConterioSW: I think we need to define `nextAction` for this step because when we load the next (`Reason`) step, it looks for a step where `Reason `is defined as the nextAction when deciding whether `Reason` is reachable, so if you don't have nextAction defined as `Reason` on the `areYouSure` step, you never reach the `Reason`  page, just keep getting redirected back to `AreYouSure`

isobel-softwire (commented):
Left some (minor) thoughts. You might want to keep an eye on #262 and #265 as they affect `AddressDataService` and `JourneyDataService`

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyController.kt
```diff
@@ -0,0 +1,90 @@
...
+        @RequestParam(value = "subpage", required = false) subpage: Int?,
+        model: Model,
+        principal: Principal,
+    ): ModelAndView {
+        if (stepName == initialStepId.urlPathSegment) {
+            if (!propertyOwnershipService.getIsAuthorizedToDeleteRecord(propertyOwnershipId, principal.name)) {
+                addressDataService.clearCachedSingleLineAddressForPropertyOwnershipId(propertyOwnershipId)
+                throw ResponseStatusException(
+                    HttpStatus.NOT_FOUND,
+                    "Property ownership $propertyOwnershipId not found",
```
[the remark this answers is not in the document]
JasminConterioSW: This is the message we currently use in PropertyOwnershipService getPropertyOwnershipIfAuthorisedUser if the current user is not a local authority user or the primary landlord on the property. 

We were returning "Forbidden" rather than "Not found" so the user didn't see hints that they had found a real property, just one they didn't have access too! But it's probably safe to make the message more specific

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyController.kt
JasminConterioSW: Do you mean the StepId enum? I prefer not to do that in case the initial path segment changes at any point, we could easily miss having to update it elsewhere

JasminConterioSW: I did at one point have it in a companion object on the journey, but took it out after a comment from Alex.

I'll have another look in a bit, I might put it back...
Is it a problem to be creating the journeys though?

isobel-softwire: Not necessarily, I just feel like it's a bit awkward for this method dependent on `propertyDeregistrationJourneyFactory`

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/PropertyDeregistrationJourneyDataHelper.kt
JasminConterioSW: Something like the JourneyDataExtensions file from [/262](https://github.com/communitiesuk/prsdb-webapp/pull/262/files#diff-bd8dbe3483df602f7e270fd7309873719e261a809e4652fbecb664f806105be7)?

isobel-softwire: Yep - I think we're still going to split the methods by journey (i.e. `PropertyDeregistrationJourneyDataExtensions`)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
AEPR: > We could just retrieve the address from propertyOwnershipService here 

This will mean that we're retrieving the address from the database for every request to populate the `areYouSure` page content. That isn't the end of the world, but I also think that subclassing `AbstractPage` is a pretty small cost - it's set up to be easily done!

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyController.kt
AEPR: I think I agree with Isobel here, that we shouldn't be instantiating (and initiating) the journey just to get the link to the start of the journey. It would mean that as things stand, we'd be caching the addresses of a property whenever this link is present on a page.

I do respect the desire to "keep it clean" and prevent this going out of date.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
JasminConterioSW: Yes I think the getting the address on the page is going to be the way to go, thanks :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
JasminConterioSW: We decided to stick with calling `getRedirectForNextStep` so that the next step in the journey is defined in one place only.

We'll get the step from the existing `steps `rather than creating a new one though!

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyController.kt
```diff
@@ -0,0 +1,84 @@
...
+        principal: Principal,
+    ): ModelAndView {
+        if (stepName == PropertyDeregistrationJourney.initialStepId.urlPathSegment) {
+            if (!propertyOwnershipService.getIsAuthorizedToDeleteRecord(propertyOwnershipId, principal.name)) {
+                throw ResponseStatusException(
+                    HttpStatus.NOT_FOUND,
+                    "The current user is not authorised to delete property ownership $propertyOwnershipId",
+                )
+            }
+        }
```
AEPR, line 45: As discussed, this check probably needs to be repeated for each step, so we should run this regardless of the step we're on.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PageWithSingleLineAddress.kt
```diff
@@ -0,0 +1,21 @@
...
+import uk.gov.communities.prsdb.webapp.forms.JourneyData
+import uk.gov.communities.prsdb.webapp.models.requestModels.formModels.FormModel
+import kotlin.reflect.KClass
+
+class PageWithSingleLineAddress(
+    formModel: KClass<out FormModel>,
+    templateName: String,
+    content: Map<String, Any>,
+    shouldDisplaySectionHeader: Boolean = false,
+    val addressProvider: () -> String,
```
AEPR, line 13: I think we should get someone else to review this since I'm the one who wrote it!

Basically, just this file and lines 45-75 of the `PropertyDeregistrationJourney`.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -69,7 +69,7 @@ class PropertyOwnershipService(
         if (!isLocalAuthority && !isPrimaryLandlord) {
             throw ResponseStatusException(
                 HttpStatus.NOT_FOUND,
-                "Property ownership $propertyOwnershipId not found",
+                "The current user is not authorised to view property ownership $propertyOwnershipId",
             )
```
AEPR, line 73: Can we have a chat about when these error messages will be returned and ensure that they couldn't be useful to a malicious actor?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -88,6 +88,16 @@ class PropertyOwnershipService(
...
 
+    fun getIsPrimaryLandlord(
+        propertyOwnershipId: Long,
+        baseUserId: String,
+    ): Boolean = getIsPrimaryLandlord(getPropertyOwnership(propertyOwnershipId), baseUserId)
+
+    fun getIsAuthorizedToDeleteRecord(
+        propertyOwnershipId: Long,
+        baseUserId: String,
+    ): Boolean = getIsPrimaryLandlord(propertyOwnershipId, baseUserId)
```
AEPR, line 99: I think this `getIsPrimaryLandlord` overload is a little confusing - I don't think it should belong on this service really. Either propertyOwnerships should have this function, or one that returns the primary landlord base user id or the comparison can just be done by the calling class.

As it stands, the property ownership service is doing a lot of reaching into the ownership to do logic on it. 

The same probably applies to the `getIsAuthorizedToDeleteRecord`.

AEPR (commented):
Couple of minor changes, but mainly someone else should review the changes I suggested I pushed before I approve!

AEPR (commented):
One thought on the change to the BasePage, otherwise looks good.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
JasminConterioSW: Removed the private function but I have kept the others as for now I think the named functions help describe what is going on (i.e. checking if the user is authorised, which for the moment happens to be true if and only if they are the primary landlord)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
JasminConterioSW: It looks like the error message is not going to be visible, as it looks like the message does not get included in the response to the browser.

@Travis-Softwire do you know if that's correct? Are we ok to give specific / useful error messages on the assumption that they won't get seen by the public?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PageWithSingleLineAddress.kt
```diff
@@ -0,0 +1,21 @@
+package uk.gov.communities.prsdb.webapp.forms.pages
+
+import org.springframework.web.servlet.ModelAndView
+import uk.gov.communities.prsdb.webapp.forms.JourneyData
+import uk.gov.communities.prsdb.webapp.models.requestModels.formModels.FormModel
+import kotlin.reflect.KClass
+
+class PageWithSingleLineAddress(
```
isobel-softwire, line 8: Are there other steps that will use `PageWithSingleLineAddress`? If not I feel like it should be called something along the lines of `PropertyDeregistrationAreYouSurePage`

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyControllerTests.kt
```diff
@@ -0,0 +1,100 @@
...
+        mvc
+            .get("/deregister-property/1/are-you-sure")
+            .andExpect {
+                status { isForbidden() }
+            }
+    }
+
+    @Test
+    @WithMockUser(roles = ["LANDLORD"])
+    fun `getJourneyStep for the initial step returns 403 for a landlord user who does not own this property`() {
```
isobel-softwire, line 52: I think this should be `...returns 404...`

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyControllerTests.kt
```diff
@@ -0,0 +1,100 @@
...
+    @MockBean
+    private lateinit var propertyDeregistrationJourneyFactory: PropertyDeregistrationJourneyFactory
+
+    @MockBean
+    private lateinit var propertyOwnershipService: PropertyOwnershipService
+
+    @Test
+    fun `getJourneyStep for the initial step returns a redirect for an unauthenticated user`() {
+        mvc
+            .get("/deregister-property/1/are-you-sure")
```
isobel-softwire, line 34: You could use the controller constants in these tests

isobel-softwire (commented):
All the new stuff LGTM! Left some more minor thoughts

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PageWithSingleLineAddress.kt
JasminConterioSW: There aren't at the moment, but it's been written in a way that it could be reused (passing in an addressProvider function rather than just an address, for example) which is why I went with the more generic name

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PageWithSingleLineAddress.kt
JasminConterioSW: Isobel has had a look :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PageWithSingleLineAddress.kt
AEPR: I think leaving it generic is fine - there's nothing in this file that links it to the PropertyDeregistration or AreYouSure.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/basePages/BasePage.kt
```diff
@@ -14,9 +14,15 @@ abstract class BasePage(
         fun <T : BasePage> createValidPage(
             page: Page,
             expectedPageClass: KClass<T>,
+            urlArguments: Map<String, String>? = null,
```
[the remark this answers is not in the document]
AEPR: Yeah, I think that's fine then. It feels a little clunky because you need to know the key string to create the page, but these are just tests afterall so a little boilerplate isn't the end of the world.

I suppose if we wanted to we could move the keys to companions or have a function on the companion that takes the parameter value and returns the map, but I'm not convinced it's worth it.

AEPR (approved):
I'm happy with this!

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
Travis-Softwire: We discussed on slack - we're comfortable that error messages will only end up in the logs and will not get sent back to the client. Response codes are effectively exposed to the client via which template they see, and so we'll want to be careful that we don't accidentally expose the existence of a hidden record that we didn't mean to by showing a forbidden error page vs a not found.

PR #266, opened 2025-03-10
Feat/prsd 926 remove example code

Kiran-R-K:
Removed example code from the code base


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDashboardController.kt
```diff
@@ -19,7 +19,7 @@ import java.security.Principal
 
 @PreAuthorize("hasAnyRole('LANDLORD')")
 @Controller
-@RequestMapping(LANDLORD_BASE_URL)
+@RequestMapping(LANDLORD_BASE_URL, "/")
```
Travis-Softwire, line 22: I think we want to redirect the user to the dashboard instead of adding a request mapping - so that the user's browser is updated e.g. to show the url of the dashboard

Travis-Softwire (approved):
One comment (which will become academic after PRSD-458 is complete but that might take a little while) otherwise LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDashboardController.kt
Kiran-R-K: The index method in this controller redirects to the dashboard, so the browser will be updated

PR #271, opened 2025-03-12
Fix/prsd 893 registration requires log out to take effect

AEPR:
This fixes PRSD-893 and PRSD-921. It works by clearing the security context, which will cause the service to reauthenticate with one-login, but without logging the user out of one-login first. This means that they arrive back and are re-assigned roles as though they have just logged in.

There was a second approach I considered, so I'll note it here in case you prefer it:
Instead of using a `Resetter` service, which gets the request and response from the `RequestContextHolder`, I could create a thread local object with a boolean flag. Resetting the security context from a service would mean calling a function on that object (e.g. `SecurityContextResetter.reset()`) which would set the flag on the object. I would create a filter in the filter chain which would read the value of the flag out of the object (with e.g. `SecurityContextResetter.shouldResetContext()`) and call the same function I do on the `SecurityContextResository` here, but since it is in the filter chain it will have direct access to the request and the response (rather than needing to use the `RequestContextHolder`).

Overall I think this approach is more convoluted and harder to follow cause and effect, so I prefer the option in this PR!

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LaUserRegistrationJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LaUserRegistrationJourney.kt
@@ -18,7 +18,7 @@ import uk.gov.communities.prsdb.webapp.models.requestModels.formModels.NoInputFo
 import uk.gov.communities.prsdb.webapp.services.JourneyDataService
 import uk.gov.communities.prsdb.webapp.services.LocalAuthorityDataService
 import uk.gov.communities.prsdb.webapp.services.LocalAuthorityInvitationService
-import uk.gov.communities.prsdb.webapp.services.SecurityContextResetter
+import uk.gov.communities.prsdb.webapp.services.SecurityContextService
 
 class LaUserRegistrationJourney(
     validator: Validator,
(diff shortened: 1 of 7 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LaUserRegistrationJourneyTests.kt
```diff
@@ -94,7 +94,9 @@ class LaUserRegistrationJourneyTests : IntegrationTest() {
 
         // Return to dashboard button
         confirmationPage.returnToDashboardButton.clickAndWait()
-        assertPageIs(page, LocalAuthorityDashboardPage::class)
+        val dashboard = assertPageIs(page, LocalAuthorityDashboardPage::class)
+
+        assertThat(dashboard.bannerSubHeading).containsText("Local authority")
```
Travis-Softwire, line 99: Out of interest, what was the reason for adding this assertion (and the one below)? I would have though checking the url would be enough as you'd hit a redirect if you didn't have the correct role by this point?

Travis-Softwire (approved):
This is a really elegant solution (better than what I had in mind, which was to manually add the role to the security context) and a really nice bug to get fixed.

One question about the tests - but otherwise all LGTM!

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LaUserRegistrationJourneyTests.kt
Travis-Softwire: Ah ofc, it's an internal call to `/error` rather than a redirect. One for us to think about whether that's the behaviour we want or if we want to redirect the client, but not something we need to worry about now - fine with this as-is in that case :)

PR #272, opened 2025-03-13
PRSD-NONE: Creates PageWithContentProviders

isobel-softwire:
Replaces `PageWithSingleLineAddress` with `PageWithContentProviders`, which can be used more generally.

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
@@ -10,7 +10,7 @@ import uk.gov.communities.prsdb.webapp.controllers.LandlordDashboardController.C
 import uk.gov.communities.prsdb.webapp.controllers.PropertyDetailsController
 import uk.gov.communities.prsdb.webapp.forms.JourneyData
 import uk.gov.communities.prsdb.webapp.forms.pages.Page
-import uk.gov.communities.prsdb.webapp.forms.pages.PageWithContentProviders
+import uk.gov.communities.prsdb.webapp.forms.pages.PageWithContentProvider
 import uk.gov.communities.prsdb.webapp.forms.steps.DeregisterPropertyStepId
 import uk.gov.communities.prsdb.webapp.forms.steps.Step
 import uk.gov.communities.prsdb.webapp.helpers.extensions.journeyDataExtensions.PropertyDeregistrationJourneyDataExtensions.Companion.getWantsToProceed
@@ -46,7 +46,7 @@ class PropertyDeregistrationJourney(
         Step(
             id = DeregisterPropertyStepId.AreYouSure,
             page =
-                PageWithContentProviders(
+                PageWithContentProvider(
                     formModel = PropertyDeregistrationAreYouSureFormModel::class,
                     templateName = "forms/areYouSureForm",
                     content =
@@ -68,8 +68,7 @@ class PropertyDeregistrationJourney(
                                 ),
                             BACK_URL_ATTR_NAME to PropertyDetailsController.getPropertyDetailsPath(propertyOwnershipId),
                         ),
-                    contentProviders = listOf { providePropertyAddress() },
-                ),
+                ) { providePropertyAddress() },
             // handleSubmitAndRedirect will execute. It does not have to redirect to the step specified in nextAction.
             handleSubmitAndRedirect = { newJourneyData, subPage -> areYouSureContinueToNextActionOrExitJourney(newJourneyData, subPage) },
             // This gets checked when determining whether the next step is reachable
@@ -109,7 +108,7 @@ class PropertyDeregistrationJourney(
         return PropertyDetailsController.getPropertyDetailsPath(propertyOwnershipId)
     }
 
-    private fun providePropertyAddress(): Pair<String, String> {
+    private fun providePropertyAddress(): Map<String, String> {
         val propertyAddress =
             propertyOwnershipService
                 .retrievePropertyOwnershipById(propertyOwnershipId)
@@ -120,7 +119,7 @@ class PropertyDeregistrationJourney(
                 "Address for property ownership id $propertyOwnershipId not found",
             )
 
-        return ("singleLineAddress" to propertyAddress)
+        return mapOf("singleLineAddress" to propertyAddress)
     }
 
     private fun deregisterPropertyAndRedirectToConfirmation(): String {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
```diff
@@ -63,9 +62,10 @@ class PropertyDeregistrationJourney(
                                         labelMsgKey = "forms.radios.option.no.label",
                                     ),
                                 ),
-                            BACK_URL_ATTR_NAME to getPropertyDetailsPath(propertyOwnershipId),
+                            BACK_URL_ATTR_NAME to PropertyDetailsController.getPropertyDetailsPath(propertyOwnershipId),
```
AEPR, line 65: Sneaky - but I approve of the namespacing.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PageWithContentProviders.kt
```diff
@@ -5,17 +5,20 @@ import uk.gov.communities.prsdb.webapp.forms.JourneyData
 import uk.gov.communities.prsdb.webapp.models.requestModels.formModels.FormModel
 import kotlin.reflect.KClass
 
-class PageWithSingleLineAddress(
+class PageWithContentProviders(
     formModel: KClass<out FormModel>,
     templateName: String,
     content: Map<String, Any>,
+    private val contentProviders: List<() -> Pair<String, Any?>>,
```
AEPR, line 12: I think that this should be a single content provider that provides all of the content, probably as a map. That mirrors the way that `content` above works.

It will also let us use the expression body syntax when initialising this page!

AEPR (commented):
I really like this change, but I'd prefer that the provided content is a map so as to be the same as the parameter `content`

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PageWithContentProviders.kt
isobel-softwire: That makes sense - implemented this change!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PageWithContentProviders.kt
```diff
@@ -5,17 +5,15 @@ import uk.gov.communities.prsdb.webapp.forms.JourneyData
...
-class PageWithSingleLineAddress(
+class PageWithContentProviders(
     formModel: KClass<out FormModel>,
     templateName: String,
     content: Map<String, Any>,
     shouldDisplaySectionHeader: Boolean = false,
-    val addressProvider: () -> String,
-) : AbstractPage(formModel, templateName, content, shouldDisplaySectionHeader) {
+    contentProvider: () -> Map<String, Any>,
+) : AbstractPage(formModel, templateName, content + contentProvider(), shouldDisplaySectionHeader) {
```
AEPR: ~~Unfortunately, this doesn't do what we want as it will mean that the content is provided when the page is constructed and the point of this is that we only provide the content when the page is being shown.~~

~~What you had before was good - I just wanted it to loop through a `Map<String, Any?>` provided by a function rather than looping through a list of functions each providing a pair.~~

Edit: You've caught the issue here yourself, but it's easily done so I think it's worth writing a test for this page that ensures it doesn't happen again.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
```diff
@@ -99,15 +98,19 @@ class PropertyDeregistrationJourney(
...
+                .retrievePropertyOwnershipById(propertyOwnershipId)
+                ?.property
+                ?.address
+                ?.singleLineAddress ?: throw ResponseStatusException(
+                HttpStatus.NOT_FOUND,
+                "Address for property ownership id $propertyOwnershipId not found",
+            )
+
+        return mapOf("singleLineAddress" to propertyAddress)
+    }
```
AEPR, line 113: I would also make this function return the String `propertyAddress` and give it an appropriate name (`getSingleLineAddress` or something). Then you can define the provider inline above in the expression body (see my next comment).

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDeregistrationJourney.kt
```diff
@@ -63,9 +62,9 @@ class PropertyDeregistrationJourney(
                                         labelMsgKey = "forms.radios.option.no.label",
                                     ),
                                 ),
-                            BACK_URL_ATTR_NAME to getPropertyDetailsPath(propertyOwnershipId),
+                            BACK_URL_ATTR_NAME to PropertyDetailsController.getPropertyDetailsPath(propertyOwnershipId),
                         ),
-                ) { retrieveAddressFromDatabase() },
+                ) { providePropertyAddress() },
```
AEPR, line 67: (continued from other comment)

Then you can make this expression body:
```kotlin
{ mapOf("singleLineAddress" to getPropertySingleLineAddress()) }
```

I don't think it really makes a difference in this specific case, but my thought here is that it makes it's a better pattern to follow for the future because it makes it easier to provide two unrelated values to the same page:
```kotlin
{ 
    mapOf(
        "first-value" to getFirstValue(),
        "second-value" to getSecondValue(),
    )
}

**Comment by Travis-Softwire** [comment]
Nothing to add - agree this is a nice improvement in a nice compact PR :) Will let @AEPR approve when he's happy with it

File: src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PageWithContentProviderTests.kt [hunk 6]
```diff
@@ -0,0 +1,26 @@
...
+                templateName = "any",
+                content = emptyMap(),
+            ) { providedContent }
+
+        val modelAndView = ModelAndView()
+
+        testPage.enrichModel(modelAndView, filteredJourneyData = null)
+
+        assertEquals(providedContent, modelAndView.modelMap.toMap())
+    }
```
AEPR, line 25: Please can we also have a test `constructing the PageWithContentProvider` does not call the provider`?

Basically there so that our code is self documenting and someone who comes along later knows that it is deliberate and important that the provider isn't called until the model is enriched.

PR #273, opened 2025-03-14
PRSD-799: Update Property - Occupation Status

isobel-softwire:
### Features
- Adds `PropertyDetailsUpdateJourneyDataExtensions` methods for retrieving the original and updated occupation statuses
- Updates `PropertyDetailsUpdateJourneyDataExtensions` households and people number retrieval methods to return 0 if the property's occupation status has been updated to vacant
- Updates `PropertyDetailsUpdateJourney`
  - Adds occupation step 
  - Makes the number of people step follow the number of households one
  - Updates the pages of the number of people and households steps to determine their `fieldSetHeading` via the original occupation status and their `backUrl` via the updated one.

### Tests
- Adds tests for the new/updated `PropertyDetailsUpdateJourneyDataExtensions` methods
- Adds integration tests for new `PropertyDetailsUpdateJourney` functionality

### Screenshots
- Update details page
image

- Update occupancy page (originally occupied)
image

- Update occupancy page (originally vacant)
image

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
@@ -7,15 +7,19 @@ import uk.gov.communities.prsdb.webapp.constants.enums.OwnershipType
 import uk.gov.communities.prsdb.webapp.controllers.PropertyDetailsController
 import uk.gov.communities.prsdb.webapp.forms.JourneyData
 import uk.gov.communities.prsdb.webapp.forms.pages.Page
+import uk.gov.communities.prsdb.webapp.forms.pages.PageWithContentProvider
 import uk.gov.communities.prsdb.webapp.forms.steps.Step
 import uk.gov.communities.prsdb.webapp.forms.steps.UpdatePropertyDetailsStepId
+import uk.gov.communities.prsdb.webapp.helpers.extensions.journeyDataExtensions.PropertyDetailsUpdateJourneyDataExtensions.Companion.getIsOccupiedUpdateIfPresent
 import uk.gov.communities.prsdb.webapp.helpers.extensions.journeyDataExtensions.PropertyDetailsUpdateJourneyDataExtensions.Companion.getNumberOfHouseholdsUpdateIfPresent
 import uk.gov.communities.prsdb.webapp.helpers.extensions.journeyDataExtensions.PropertyDetailsUpdateJourneyDataExtensions.Companion.getNumberOfPeopleUpdateIfPresent
+import uk.gov.communities.prsdb.webapp.helpers.extensions.journeyDataExtensions.PropertyDetailsUpdateJourneyDataExtensions.Companion.getOriginalIsOccupied
 import uk.gov.communities.prsdb.webapp.helpers.extensions.journeyDataExtensions.PropertyDetailsUpdateJourneyDataExtensions.Companion.getOwnershipTypeUpdateIfPresent
 import uk.gov.communities.prsdb.webapp.models.dataModels.updateModels.PropertyOwnershipUpdateModel
 import uk.gov.communities.prsdb.webapp.models.requestModels.formModels.NoInputFormModel
 import uk.gov.communities.prsdb.webapp.models.requestModels.formModels.NumberOfHouseholdsFormModel
 import uk.gov.communities.prsdb.webapp.models.requestModels.formModels.NumberOfPeopleFormModel
+import uk.gov.communities.prsdb.webapp.models.requestModels.formModels.OccupancyFormModel
 import uk.gov.communities.prsdb.webapp.models.requestModels.formModels.OwnershipTypeFormModel
 import uk.gov.communities.prsdb.webapp.models.viewModels.formModels.RadiosButtonViewModel
 import uk.gov.communities.prsdb.webapp.services.JourneyDataService
@@ -40,9 +44,11 @@ class PropertyDetailsUpdateJourney(
 
     override fun createOriginalJourneyData(): JourneyData {
         val propertyOwnership = propertyOwnershipService.getPropertyOwnership(propertyOwnershipId)
+        val isOccupied = propertyOwnership.currentNumTenants > 0
 
         return mapOf(
             UpdatePropertyDetailsStepId.UpdateOwnershipType.urlPathSegment to mapOf("ownershipType" to propertyOwnership.ownershipType),
+            UpdatePropertyDetailsStepId.UpdateOccupancy.urlPathSegment to mapOf("occupied" to isOccupied),
             UpdatePropertyDetailsStepId.UpdateNumberOfHouseholds.urlPathSegment to
                 mapOf("numberOfHouseholds" to propertyOwnership.currentNumHouseholds),
             UpdatePropertyDetailsStepId.UpdateNumberOfPeople.urlPathSegment to
@@ -92,7 +98,39 @@ class PropertyDetailsUpdateJourney(
                         ),
                 ),
             handleSubmitAndRedirect = { _, _ -> UpdatePropertyDetailsStepId.UpdateDetails.urlPathSegment },
-            nextAction = { _, _ -> Pair(UpdatePropertyDetailsStepId.UpdateNumberOfHouseholds, null) },
+            nextAction = { _, _ -> Pair(UpdatePropertyDetailsStepId.UpdateOccupancy, null) },
+            saveAfterSubmit = false,
+        )
+
+    private val occupancyStep =
+        Step(
+            id = UpdatePropertyDetailsStepId.UpdateOccupancy,
+            page =
+                PageWithContentProvider(
+                    formModel = OccupancyFormModel::class,
+                    templateName = "forms/propertyOccupancyForm",
+                    content =
+                        mapOf(
+                            "title" to "propertyDetails.update.title",
+                            "radioOptions" to
+                                listOf(
+                                    RadiosButtonViewModel(
+                                        value = true,
+                                        valueStr = "yes",
+                                        labelMsgKey = "forms.radios.option.yes.label",
+                                        hintMsgKey = "forms.occupancy.radios.option.yes.hint",
+                                    ),
+                                    RadiosButtonViewModel(
+                                        value = false,
+                                        valueStr = "no",
+                                        labelMsgKey = "forms.radios.option.no.label",
+                                        hintMsgKey = "forms.occupancy.radios.option.no.hint",
+                                    ),
+                                ),
+                            BACK_URL_ATTR_NAME to UpdatePropertyDetailsStepId.UpdateDetails.urlPathSegment,
+                        ),
+                ) { mapOf("fieldSetHeading" to getOccupancyStepFieldSetHeading()) },
+            nextAction = { journeyData, _ -> occupancyNextAction(journeyData) },
             saveAfterSubmit = false,
         )
 
@@ -100,18 +138,20 @@ class PropertyDetailsUpdateJourney(
         Step(
             id = UpdatePropertyDetailsStepId.UpdateNumberOfHouseholds,
             page =
-                Page(
+                PageWithContentProvider(
                     formModel = NumberOfHouseholdsFormModel::class,
                     templateName = "forms/numberOfHouseholdsForm",
                     content =
                         mapOf(
                             "title" to "propertyDetails.update.title",
-                            "fieldSetHeading" to "forms.update.numberOfHouseholds.fieldSetHeading",
                             "label" to "forms.numberOfHouseholds.label",
-                            BACK_URL_ATTR_NAME to UpdatePropertyDetailsStepId.UpdateDetails.urlPathSegment,
                         ),
-                ),
-            handleSubmitAndRedirect = { _, _ -> UpdatePropertyDetailsStepId.UpdateDetails.urlPathSegment },
+                ) {
+                    mapOf(
+                        "fieldSetHeading" to getNumberOfHouseholdsStepFieldSetHeading(),
+                        BACK_URL_ATTR_NAME to getNumberOfHouseholdsStepBackUrl(),
+                    )
+                },
             nextAction = { _, _ -> Pair(UpdatePropertyDetailsStepId.UpdateNumberOfPeople, null) },
             saveAfterSubmit = false,
         )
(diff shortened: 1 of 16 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
```diff
@@ -40,9 +44,11 @@ class PropertyDetailsUpdateJourney(
 
     override fun createOriginalJourneyData(): JourneyData {
         val propertyOwnership = propertyOwnershipService.getPropertyOwnership(propertyOwnershipId)
+        val isOccupied = propertyOwnership.currentNumTenants > 0
```
JasminConterioSW, line 47: I'm sure we do this check on the registration journey, maybe in multiple places. Maybe we can add `getIsOccupied ` to PropertyOwnershipService and use that everywhere?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
```diff
@@ -92,26 +98,60 @@ class PropertyDetailsUpdateJourney(
...
+                                    RadiosButtonViewModel(
+                                        value = false,
+                                        valueStr = "no",
+                                        labelMsgKey = "forms.radios.option.no.label",
+                                        hintMsgKey = "forms.occupancy.radios.option.no.hint",
+                                    ),
+                                ),
+                            BACK_URL_ATTR_NAME to UpdatePropertyDetailsStepId.UpdateDetails.urlPathSegment,
+                        ),
+                ) { mapOf("fieldSetHeading" to getOccupancyStepFieldSetHeading()) },
```
JasminConterioSW, line 132: I think it's ok for this to go into `content ` rather than being passed through as a provider (so this can also just be a `Page`)
If it is just getting data from the session, it doesn't matter if they are being called every time the journey is created.

(The address one that was used on `PropertyDeregistrationJourney ` was calling the database, so we wanted to only call the get function when we actually needed the information i.e. when the page loaded, rather than every time the journey was created, which is what happens if it is part of `content`).

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
```diff
@@ -92,26 +98,60 @@ class PropertyDetailsUpdateJourney(
...
+            nextAction = { journeyData, _ -> occupancyNextAction(journeyData) },
             saveAfterSubmit = false,
         )
 
     private val numberOfHouseholdsStep =
         Step(
             id = UpdatePropertyDetailsStepId.UpdateNumberOfHouseholds,
             page =
-                Page(
+                PageWithContentProvider(
```
JasminConterioSW, line 141: I think this step can also just be a `Page` with everything in the content rather than needing content providers, as long it as it not calling the database

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
```diff
@@ -120,19 +160,21 @@ class PropertyDetailsUpdateJourney(
         Step(
             id = UpdatePropertyDetailsStepId.UpdateNumberOfPeople,
             page =
-                Page(
+                PageWithContentProvider(
```
JasminConterioSW, line 163: Same as above, this can just be a `Page`

JasminConterioSW (changes requested):
Mostly looks good to me! Just a couple of things:

- We might want to commonise the isOccupied check so that the registration and update journeys use the same implementation. 
- Also I think we should be ok with using Page rather than PageWithContentProvider as long as the content is not coming from the database.

AEPR (commented):
Ending review so I can comment!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
AEPR: I think that we shouldn't add it to the `PropertyOwnershipService` - it's about the contents of the property ownership itself.

Commonising this makes sense, but it should be a function on the `PropertyOwnership` entity (e.g. `isOccupied()` or as a getter for a get only property `isOccupied`).

src/main/resources/messages.properties
```diff
@@ -682,6 +682,7 @@ forms.update.lookupAddress.fieldSetHeading=Update your contact address
 forms.update.phoneNumber.fieldSetHeading=What is your updated phone number?
 forms.update.dateOfBirth.fieldSetHeading=Update your date of birth on your landlord record
 forms.update.ownershipType.fieldSetHeading=Update the ownership type for your property
+forms.update.occupancy.fieldSetHeading=Is your property still occupied by tenants?
```
AEPR, line 685: I think this name is a little confusing, since this is specifically the heading if previously occupied not if you're updating the occupancy in general.

Maybe `forms.update.occupancy.occupied.fieldSetHeading`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
isobel-softwire: I agree with Alex - I've given `PropertyOwnership` an `isOccupied` attribute

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
```diff
@@ -141,9 +182,56 @@ class PropertyDetailsUpdateJourney(
...
+        when (journeyData.getIsOccupiedUpdateIfPresent()) {
+            true -> Pair(UpdatePropertyDetailsStepId.UpdateNumberOfHouseholds, null)
+            false -> Pair(UpdatePropertyDetailsStepId.UpdateDetails, null)
+            null ->
+                if (journeyData.getOriginalIsOccupied(originalDataKey)!!) {
+                    Pair(UpdatePropertyDetailsStepId.UpdateNumberOfHouseholds, null)
+                } else {
+                    Pair(UpdatePropertyDetailsStepId.UpdateDetails, null)
+                }
+        }
```
AEPR, line 233: I don't think this nested complexity is necessary here as we're always acting on the merged journey data. You should be able to do:
```
if (journeyData.getIsOccupiedUpdateIfPresent()!!) {
```

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
JasminConterioSW: Hmm, actually I think you're right. Fair enough :)

AEPR: I think I agree with Jasmin here. We ARE using factories at the moment, which means that journeys are scoped to single request. If that changes then things that rely on that will break and we'll need to refactor them, but I don't think it's good practice to implement things just in case we do need to refactor later.

In this case, I think it's also a little bit surprising that we're passing in a content provider that uses the `JourneyDataService` to get the journey data to determine the content to add, wheras the `enrichModel` method already has the `filteredJourneyData` passed in. Now, we can't actually use the filtered journey data for any of these because they're all determined by whether the user has updated changed the occupation status or not (which we can't tell from the `filteredJourneyData` as it's the combined values).

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
```diff
@@ -141,9 +182,56 @@ class PropertyDetailsUpdateJourney(
...
+        } else {
+            UpdatePropertyDetailsStepId.UpdateDetails.urlPathSegment
+        }
+
+    private fun getNumberOfPeopleStepBackUrl() =
+        if (journeyDataService.getJourneyDataFromSession().getIsOccupiedUpdateIfPresent() != null) {
+            UpdatePropertyDetailsStepId.UpdateNumberOfHouseholds.urlPathSegment
+        } else {
+            UpdatePropertyDetailsStepId.UpdateDetails.urlPathSegment
+        }
```
AEPR, line 221: I understand why they're are all here, but there's something I can't quite put my finger on that I don't like about this battery of functions. Thinking aloud:
* They feel like they're all following the same logic, but expressed slightly differently
* Their names don't really tell you what they're doing, or why these values need to come from functions

I guess it feels like these are all expressing "Do thing A if you're updating occupancy, do thing B if you're just updating the tennants". I don't quite know how to describe that neatly, though.

AEPR (commented):
Left some thoughts on the `main` code.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
AEPR: I've ruminated over lunch, and I think that my main issue here is that after reading all of these functions I'm not entirely sure what they're really doing. I think the answer is perhaps that the two boolean statements should be pulled out into helpfully named little private functions that communicate what the difference is (e.g. `wasOriginallyOccupied` and `hasBeenSetToOccupied`). 

At that point we might consider inlining the if statements, but I feel much less strongly about that once the reasoning behind the if statements is easy to understand. (And it might make the page section harder to read)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
isobel-softwire: That's fair enough - I'll make them `Page`s

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
isobel-softwire: Ah ok - so `ifPresent` refers to present in the original or updated journey data?

AEPR: Well, the `ifPresent` refers to present in the journey data you run it on as an extension. In the `UpateJourney.iterator`, we take the original data then apply any updates and use that as the journey data. In the `handleSubmitAndRedirect` function on a page, we pass in the current journey data without including any of the original data.

What that means is that if you're dealing with filtered journey data, it will always be present. If you run it on the unfiltered journey data, it's only present if the user has set something for that value.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
isobel-softwire: I've abstracted the boolean statements. I tried inlining the if-statements but I feel like it decreased readability.

src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/journeyDataExtensions/PropertyDetailsUpdateJourneyDataExtensionsTests.kt
```diff
@@ -33,19 +37,85 @@ class PropertyDetailsUpdateJourneyDataExtensionsTests {
...
+    @Test
+    fun `getOriginalIsOccupied returns null if the corresponding page is not in original journeyData`() {
+        val originalJourneyKey = "original-key"
+        val originalJourneyData = journeyDataBuilder.build()
+        val testJourneyData = mapOf(originalJourneyKey to originalJourneyData)
+
+        val originalOccupancy = testJourneyData.getOriginalIsOccupied(originalJourneyKey)
+
+        assertNull(originalOccupancy)
+    }
```
AEPR, line 65: I'm slightly surprised by this - I would have assumed that this should always return a boolean since there should always be an original ocupied value. I don't know that it matters enough to enforce though.

This is a great example of tests acting as documentation - I missed the behaviour when I reviewed the extension but have caught it because the test spels it out.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDetailsUpdateJourneyTests.kt
```diff
@@ -114,29 +93,128 @@ class PropertyDetailsUpdateJourneyTests : IntegrationTest() {
...
+            // Check changes have occurred
+            assertThat(propertyDetailsUpdatePage.propertyDetailsSummaryList.occupancyRow.value).containsText("Yes")
+            assertThat(propertyDetailsUpdatePage.propertyDetailsSummaryList.numberOfHouseholdsRow.value).containsText(
+                newNumberOfHouseholds.toString(),
+            )
+            assertThat(propertyDetailsUpdatePage.propertyDetailsSummaryList.numberOfPeopleRow.value).containsText(
+                newNumberOfPeople.toString(),
+            )
+        }
+    }
```
AEPR, line 219: Not worth writing tests for these cases, but there is also the case of updating from occupied to occupied, which will have the fieldset headings of was originally occupied and the back buttons of occupancy has been updated.

AEPR (approved):
That looks good to me.

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/journeyDataExtensions/PropertyDetailsUpdateJourneyDataExtensions.kt
```diff
@@ -14,18 +14,36 @@ class PropertyDetailsUpdateJourneyDataExtensions {
                 "ownershipType",
             )
 
-        fun JourneyData.getNumberOfHouseholdsUpdateIfPresent() =
-            JourneyDataHelper.getFieldIntegerValue(
+        fun JourneyData.getOriginalIsOccupied(originalJourneyKey: String) =
+            JourneyDataHelper.getPageData(this, originalJourneyKey)?.getIsOccupiedUpdateIfPresent()
```
JasminConterioSW, line 18: The name of `getOriginalIsOccupied` sounds to me like it shouldn't depend on the value of the occupancy update, is this doing what we expect?

JasminConterioSW (changes requested):
It's maybe more of a clarification than a change I'm requesting here, the `getOriginalIsOccupied` function may want renaming!

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/journeyDataExtensions/PropertyDetailsUpdateJourneyDataExtensions.kt
AEPR: Yep - this function is working as expected. The `getPageData(this, originalJourneyKey)` is getting the original journey data, then we're checking whether that has occupied or not. It's a little confusing because `getIsOccupiedUpdate` isn't getting an update.

Perhaps worth either not calling `getIsOccupiedUpdateIfPresent` or creating another function that gets the occupied value without implying it's an update.

AEPR: Just seen this update - I think the original name is clearer to me. This is checking whether occupied was set before updates started, which is explicitly not an update - it's just the original state.

isobel-softwire: Ok - I'll revert that change and make a private `getIsOccupied` method that both `getOriginalIsOccupied` and `getIsOccupiedUpdateIfPresent` can use

JasminConterioSW (approved):
LGTM :)

PR #274, opened 2025-03-17
Chore/prsd none use confirmation path segment

JasminConterioSW:
We were defining new CONFIRMATION_PATH_SEGMENTs in various journeys (which were all "confirmation"), instead this uses one common constant for all of them.

Also picks up a couple of other places where constants should be used.


isobel-softwire (approved):
Nice refactor! `PropertyRegistrationJourney` currently passes `TASK_LIST_PATH_SEGMENT` as a literal, so that can be updated as well, but apart from that LGTM :)

PR #275, opened 2025-03-17
PRSD-697: Add hint text to property deregistration reason page

JasminConterioSW:
Adds hint text to the property deregistration reason page

image


isobel-softwire (approved):
LGTM!

PR #277, opened 2025-03-18
PRSD-930: Adds file uploads ADR

Travis-Softwire:
ADR for file uploads as a starting point for discussing the available options


adrs/0025-file-uploads.md
```diff
@@ -0,0 +1,98 @@
+# ADR-0025: file-uploads
```
rowanhill, line 1: Can we make this title case (i.e. "File Uploads")?

PR #278, opened 2025-03-18
Feat/prsd 699 property deregistration confirmation email

JasminConterioSW:
Sends a confirmation email a property has been deregistered.

image


AEPR (commented):
I imagine this will be worth a re-review after you've merged with the other PR, but happy with it for now!

JasminConterioSW:
> I imagine this will be worth a re-review after you've merged with the other PR, but happy with it for now!

They could really be merged in either order, if you are happy - it would just mean that for a while on the test environment you would get a confirmation email but not see the confirmation page...

AEPR (approved):
Well, in that case I'm happy with it!

PR #279, opened 2025-03-18
feat/PRSD-896 added coherence valisation to to PropertyRegistrationSe…

Kiran-R-K:
…rvice before saving property

Added validation to check number of people is not less than number of households when updating a property in the database.

Added validation to check number of people is 0 if number of households is 0 when updating a property in the database.

Added unit tests to cover both changes above

Per this [comment](https://github.com/communitiesuk/prsdb-webapp/pull/264#discussion_r1998556137) I also added tests to cover the case where a landlord updates a properties number of households to be higher than the number of people


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -160,6 +161,8 @@ class PropertyOwnershipService(
     ) {
         val propertyOwnership = getPropertyOwnership(id)
 
+        validateNumberOfPeopleWithNumberOfHouseholds(update, propertyOwnership)
+
```
AEPR, line 165: It's surprising to me that we're validating this here in the service when we update, but not validating it in the service when we create the record.

Generally, I think we're doing validation as part of journeys, so I think we should stick with that here rather than mixing where it happens.

Kiran-R-K: I put validation here following the discussion in this thread https://softwire.slack.com/archives/C07M9ND7U66/p1741178960880609?thread_ts=1741177528.176689&cid=C07M9ND7U66
My understanding is that because all the property data comes from the session when creating the record that we don't have the same risk and that's why we don't need the validation there

AEPR: I think that the way we've implemented the update journey the same should be true - it will merge the original data with the update and validate it before submitting to the database. 

Mostly, we should check how the validation actually interacts with the update journey as is and make sure the journey validation for the update catches this problem and handles it appropriately.

Also, for what it's worth I'd be happy with it being a separate PR so you can get the main validation PR merged in then come back to check how it interacts with the update journey.

Kiran-R-K:
As a result of the changes described in this [comment](https://github.com/communitiesuk/prsdb-webapp/pull/264#discussion_r2002923455) the tests and validation in this PR are no longer needed so I have closed it

PR #280, opened 2025-03-19
PRSD-929: Address Lookups Revisited ADR-0026

rowanhill:
Adds an ADR that revisits address lookups (superseding the original one), to address the issue of address data becoming stale after we save it from OS Places API.

### DIFF
--- a/adrs/0026-address-lookups-revisited.md
+++ b/adrs/0026-address-lookups-revisited.md
@@ -1,0 +1,82 @@
+# ADR-0026: Address Lookups (Revisited)
+
+## Status
+
+Proposed
+
+Date of decision: {date - TODO update after PR approved}
+
+## Context and Problem Statement
+
+{Describe the context and problem statement, e.g. in free-form text using two or three sentences. You may want to
+articulate the problem in the form of a question.}
+
+ADR-0010 decided upon OS Places API as our address lookup solution: users search for an address, and we cache the
+result in our local database, for later searches. A key part of the information saved is the local authority. LA
+boundaries can change surprisingly often, however, leaving our cached data stale; the same is true for other types
+of address changes.
+
+How should we deal with the problem of cached address data going stale?
+
+## Considered Options
+
+* Refresh via OS Places API
+* OS AddressBase product
+* OS NGD Address
+* OS NGD Address + OS Places API
+
+## Decision Outcome
+
+OS NGD Address, because this will minimise stale data and ongoing complexity, as well as the risk of needing replacement
+in the future.
+
+## Pros and Cons of the Options
+
+### Refresh via OS Places API
+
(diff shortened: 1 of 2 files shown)


adrs/0026-address-lookups-revisited.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+### OS NGD Address
+
+Regularly ingest OS NGD (National Geographic Database - the newest data product for addresses (and much more) that OS encourage
+services to use) and search on that (instead of via OS Places API).
+
+* Bad, because existing address lookup journeys will need to be reworked.
+* Neutral, because extra complexity is needed to regularly download and ingest the NGD data.
+* Good, because data updates are automatic and frequent (we believe daily).
```
Travis-Softwire, line 69: Is this definitely automatic given we can only get it from a Select + Build recipe? It seemed to me from the documentation that we'd need to request a resupply each time we want an updated version of the package, which is something that requires admin powers on the datahub...

Travis-Softwire (commented):
A couple of thoughts on the NGD option. The biggest one being I'm not sure how automated we can currently make the downloads, unless I'm reading the docs on Datahub wrong

adrs/0026-address-lookups-revisited.md
rowanhill: Good point - we still need to investigate if we can access NGD Address data via the Downloads API (as our OS contact suggested might be possible). It surely must be automatable to some degree, though!

adrs/0026-address-lookups-revisited.md
```diff
@@ -0,0 +1,82 @@
...
+* Neutral, because we may need to handle addresses moving to a new LA before we have migrated our LA users.
```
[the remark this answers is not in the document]
rowanhill: Yes, correct. I'm wondering if we might need some more complicated "history" of local authority for addresses, or something - or, alternatively we only ingest updates from NGD on some manual trigger, or something. Do you think we need to resolve that question in this ADR?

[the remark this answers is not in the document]
Travis-Softwire: No, I think we can cover it separately when we decide how we're going to handle updates

Travis-Softwire (approved):

adrs/0026-address-lookups-revisited.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+### OS NGD Address + OS Places API
+
+Regularly ingest OS NGD (National Geographic Database - the newest data product for addresses (and much more) that OS encourage
+services to use) but continue to use OS Places API for address lookups.
+
+* Bad, because existing although address lookup journeys will stay the same, we'll still need to update the LA search
```
isobel-softwire, line 78: Should this be '...because although existing address...'?

PR #283, opened 2025-03-20
PRSD-946: Creates ADR for Non-deterministic Journey Support

isobel-softwire:
Creates an ADR for choosing not to support non-deterministic journeys

### DIFF
--- a/adrs/0027-non-deterministic-journey-support.md
+++ b/adrs/0027-non-deterministic-journey-support.md
@@ -1,4 +1,4 @@
-# ADR-0026: Non-deterministic Journey Support
+# ADR-0027: Non-deterministic Journey Support
 
 ## Status
 
--- a/adrs/0027-non-deterministic-journey-support.md
+++ b/adrs/0027-non-deterministic-journey-support.md
@@ -9,15 +9,19 @@ Date of decision: 20/03/2025
 ## Context and Problem Statement
 
 Users of the Private Rented Sector Database will need to complete a range of multi-page forms, like landlord registration 
-or compliance. We represent multi-page forms as journeys, with each one consisting of an ordered collection of steps. 
-At the moment, journeys can only be deterministic, i.e. steps can only be completed in one order (given the user's answers).
-However, steps that don't depend on each other could be completed in any order if we supported non-deterministic journeys.
+or compliance. We represent multi-page forms as journeys, with each one consisting of an ordered collection of reachable 
+steps. Related steps are grouped into tasks. Step order is determined by a journey's iterator, which uses each step's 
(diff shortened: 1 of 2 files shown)


adrs/0027-non-deterministic-journey-support.md
```diff
@@ -0,0 +1,51 @@
...
+Date of decision: 20/03/2025
+
+## Context and Problem Statement
+
+Users of the Private Rented Sector Database will need to complete a range of multi-page forms, like landlord registration 
+or compliance. We represent multi-page forms as journeys, with each one consisting of an ordered collection of steps. 
+At the moment, journeys can only be deterministic, i.e. steps can only be completed in one order (given the user's answers).
+However, steps that don't depend on each other could be completed in any order if we supported non-deterministic journeys.
+
+Should we support non-deterministic journeys?
```
Travis-Softwire, line 16: I think it would be useful to include "and if so, how?" and then add the other options that were considered alongside the one that we ended up with

adrs/0027-non-deterministic-journey-support.md
```diff
@@ -0,0 +1,51 @@
...
+
+No, because the product owner has decided that at this time, the benefit of being able to complete multi-page forms in 
+different orders isn't worth the cost of the effort it will take to implement. We will re-consider supporting 
+non-deterministic journeys once all the journeys have been implemented.
+
+## Pros and Cons of the Options
+
+### Yes, by using DFS to order steps
+
+If we keep track of the steps reachable from each step, we can perform a depth-first search to order them.
```
Travis-Softwire, line 33: Can we add a little more detail here, e.g. the fact that we're performing the DFS using the iterator, that we'd made the assumption that by default the step reachable from a given step is just the single result of calling `nextAction` to support linear journeys etc. Basically anything that would be useful reminders if you came back to implement this again later.

adrs/0027-non-deterministic-journey-support.md
```diff
@@ -0,0 +1,51 @@
...
+
+* Good, because this will allow us to support non-deterministic journeys.
```
Travis-Softwire, line 35: As part of adding in the options, would focus here on why we went with this option over, e.g. just adding support for multiple initial steps, doing something with tasks etc.

Travis-Softwire (commented):
Thanks for this Isobel. Can we refocus this slightly to include details of the other options that were considered and why we went with DFS in the end? Basically the goal is to allow us to skip (or at least speedrun) the conversations we already had on this if we have to revisit it again later

Travis-Softwire (approved):

Travis-Softwire:
Happy with this :)

PR #285, opened 2025-03-20
PRSD-940: Compliance Start Page

isobel-softwire:
### Features:
- Creates `provideComplianceStartPage` template
- Creates `ProvideComplianceController`
  - Sets index to return `provideComplianceStartPage` if the user is authorized

### Tests:
- Adds `ProvideComplianceControllerTests` for index

### Screenshot:
image


src/main/resources/templates/provideComplianceStartPage.html
```diff
@@ -0,0 +1,52 @@
...
+                </h2>
+                <p class="govuk-body" th:text="#{provideCompliance.whenToUpload.paragraph.one}">
+                    provideCompliance.whenToUpload.paragraph.one
+                </p>
+                <p class="govuk-body" th:text="#{provideCompliance.whenToUpload.paragraph.two}">
+                    provideCompliance.whenToUpload.paragraph.two
+                </p>
+                <p class="govuk-body prsd-warning-text" th:text="#{provideCompliance.whenToUpload.paragraph.three}">
+                    provideCompliance.whenToUpload.paragraph.three
+                </p>
```
AEPR, line 22: Worth noting that this is now black in Figma, even though it's red in the screenshot. I'd suspect that this may be because they're (correctly) moving away from using colour to communicate emphasis. If so, we can dispose with the css!

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/UrlSegmentConstants.kt
```diff
@@ -12,3 +12,4 @@ const val SEARCH_PATH_SEGMENT = "search"
 const val REGISTERED_PROPERTIES_PATH_SEGMENT = "registered-properties"
 const val CONFIRMATION_PATH_SEGMENT = "confirmation"
 const val TASK_LIST_PATH_SEGMENT = "task-list"
+const val PROVIDE_COMPLIANCE_PATH_SEGMENT = "provide-compliance"
```
AEPR, line 15: I think this should be `provide-compliance-certificates` even though it's a mouthful, just because that's what you're actually going to upload and what the journey is called!

src/main/resources/templates/demoStart.html
```diff
@@ -6,6 +6,6 @@ <h1 class="govuk-heading-xl">Welcome to the Private Rental Sector Database</h1>
         <span>Example Landlord Registration Number: </span>
         <span th:text="${landlordRegNum}">Example Landlord Registration Number</span>
     </p>
-    <div th:replace="~{fragments/buttons/startButton :: startButton( @{/registration}, #{startButtonText} )}"></div>
+    <div th:replace="~{fragments/buttons/startButton :: startButton( @{/registration}, #{startNowButtonText} )}"></div>
```
AEPR, line 9: I think this template is unused and can be deleted.

AEPR (commented):
Only very minor things (including a bit of figma drift). I think removing the warning text class is worth doing though.

src/main/resources/templates/provideComplianceStartPage.html
isobel-softwire: Nice catch!

src/main/resources/templates/demoStart.html
isobel-softwire: I think you're right

src/main/resources/templates/provideComplianceStartPage.html
Travis-Softwire: Will post on slack as well, but in general red text is the designers' way of signalling that this is tbc or to be added later (we should never be showing the user actual red text outside of validation errors). If it's turned black in the designs that means they've now confirmed that it's needed.

PR #286, opened 2025-03-20
Feat/prsd 703 landlord deregistration are you sure

JasminConterioSW:
This is not completing a ticket, but is setting up the LandlordDeregistrationJourney to work for landlords both with and without registered properties.

In this PR:
- Set up the new controller and journey
- Set the initial step to be a "ghost" step - this saves whether or not the landlord has registered properties in the journey data so we can access this information in later steps (getting this information from the database will be in a later PR).  Note that `getModelAndViewForStep` is never called for this step, instead the controller calls `completeStep` in the get endpoint (`checkForRegisteredProperties`).
- Create the "Are you sure" step - this is a custom page where some of the content changes depending on whether or not the user has registered properties.

Also added some tests:
- Controller tests
- Integration tests
- Page tests

image
image
image


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordDeregistrationJourney.kt
```diff
@@ -0,0 +1,102 @@
...
+                            BACK_URL_ATTR_NAME to LANDLORD_DETAILS_PATH_SEGMENT,
+                        ),
+                ) { getAreYouSureFormModel() },
+            handleSubmitAndRedirect = { newJourneyData, _ -> areYouSureContinueOrExitJourney(newJourneyData) },
+            saveAfterSubmit = false,
+        )
+
+    // TODO: PRSD-703 - check in journeyData if the user has registered properties and return this version if they have none
+    // TODO: PRSD-705 - return a "with properties" version if the user has registered properties
+    private fun getAreYouSureFormModel(): KClass<out FormModel> = LandlordWithNoPropertiesDeregistrationAreYouSureFormModel::class
```
JasminConterioSW, line 77: Let me know if you have a suggestion on just using one form model for all of them, I think we can't do this because they have different error messages...

src/main/resources/messages.properties
```diff
@@ -701,3 +708,4 @@ pagination.pageText=page
 pagination.nextText=Next
 
 beforeYouRegister.text=Before you register
+blank=
```
JasminConterioSW, line 711: This is to make it easier to use the same areYouSureForm for the version with and without hint text!

`#{${fieldSetHint}}` doesn't cope with null `fieldSetHint`

I did also try using `th:with="fieldSetHintMessage=${fieldSetHint == null ? null : #{${fieldSetHint}}}"` and then passing fieldSetHintMessage to the fieldSet fragment, but it didn't cope with curly brackets in` #{${fieldSetHint}}`

src/main/resources/templates/forms/propertyDeregistrationAreYouSureForm.html
```diff
@@ -0,0 +1,17 @@
...
+<!--/*@thymesVar id="fieldSetHeading" type="java.lang.String"*/-->
+<!--/*@thymesVar id="fieldSetHint" type="java.lang.String"*/-->
+<!--/*@thymesVar id="singleLineAddress" type="java.lang.String"*/-->
+<!--/*@thymesVar id="backUrl" type="java.lang.String"*/-->
+<!--/*@thymesVar id="radioOptions" type="java.lang.Object"*/-->
+<!DOCTYPE html>
+<html th:replace="~{fragments/forms/questionPage :: questionPage(#{${title}}, ${formModel}, ~{::#form-content}, ${backUrl}, null)}">
+<th:block id="form-content">
+    <th:block id="fieldset-content"
+              th:replace="~{fragments/forms/fieldSet :: fieldSet(~{::#fieldset-content/content()}, 'wantsToProceed', #{${fieldSetHeading}(${singleLineAddress})}, null)}">
```
isobel-softwire, line 12: If the only difference between this 'are you sure' form and the other one is that this one's field set heading takes a parameter, you could try using [`#messages.msgWithParams()`](https://www.thymeleaf.org/doc/tutorials/3.0/usingthymeleaf.html#appendix-b-expression-utility-objects) to make a general form that supports landlords and properties.

src/main/resources/templates/fragments/forms/fieldSet.html
```diff
@@ -7,7 +7,7 @@ <h1 class="govuk-fieldset__heading" th:text="${fieldSetHeading}">
                 Heading for set of form fields
             </h1>
         </legend>
-        <div id="field-hint" th:if="${fieldSetHint != null}" th:id="${fieldName}+'-hint'" class="govuk-hint"
+        <div id="field-hint" th:if="${fieldSetHint != null and fieldSetHint != ''}" th:id="${fieldName}+'-hint'" class="govuk-hint"
```
isobel-softwire, line 10: If using `messages.msgOrNull()` works, this change can be reverted

isobel-softwire (dismissed):
I had some thoughts over where checking if the user has properties happens and how to avoid needing two landlord deregistration 'are you sure' models. Not sure how well I explained it so let me know if you'd like to chat about it!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordDeregistrationJourney.kt
isobel-softwire: You could add a `doesLandlordHaveProperties` property to `LandlordDeregistrationAreYouSureFormModel`, then put two `DelegatedPropertyConstraintValidator`s on `wantsToProceed` (one for the landlord having properties and one for the user not having properties).

To populate `doesLandlordHaveProperties`, you could override `bindDataToFormModel` in `LandlordDeregistrationAreYouSurePage`.

`NumberOfPeopleFormModel` and `PropertyRegistrationNumberOfPeoplePage` do something similar.

src/main/resources/messages.properties
isobel-softwire: Have you tried th:with="fieldSetHintMessage=${#messages.msgOrNull(fieldSetHint)}"? If that doesn't work try th:with="fieldSetHintMessage=${#messages.msgOrNull(fieldSetHint) ?: null}" as well.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordDeregistrationJourney.kt
JasminConterioSW: Ooh that sounds promising, thanks I'll take a look!

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterLandlordController.kt
```diff
@@ -0,0 +1,86 @@
...
+
+        return landlordDeregistrationJourneyFactory
+            .create()
+            .completeStep(
+                CHECK_FOR_REGISTERED_PROPERTIES_PATH_SEGMENT,
+                newFormData,
+                subpage,
+                principal,
+            )
+    }
```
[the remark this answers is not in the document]
isobel-softwire: In that case I agree with you about the ghost step being a better fit. 

I'm not sure that `formData` should be passed to the endpoint as an argument here though. It looks like it'll be empty unless the user manually sets it as a query parameter.

src/main/resources/templates/forms/propertyDeregistrationAreYouSureForm.html
JasminConterioSW: It still works if you do something like `#{${fieldSetHeading}(${optionalFieldSetHeadingParam})}`, I didn't want to include the `singleLineAddress` parameter on the landlord version so have made the parameter name more generic

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterLandlordController.kt
```diff
@@ -0,0 +1,86 @@
...
+                principal,
+            )
+
+    companion object {
+        const val CHECK_FOR_REGISTERED_PROPERTIES_PATH_SEGMENT = "check-user-properties"
+
+        const val USER_HAS_REGISTERED_PROPERTIES_JOURNEY_DATA_KEY = "userHasRegisteredProperties"
+
+        fun getLandlordDeregistrationPath(): String =
+            "/$DEREGISTER_LANDLORD_JOURNEY_URL/${LandlordDeregistrationJourney.initialStepId.urlPathSegment}"
```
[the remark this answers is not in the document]
JasminConterioSW: I think it can't be a constant if we use `LandlordDeregistrationJourney.initialStepId` because the initialStepId is not a const val (and can't be because it's not a primitive type)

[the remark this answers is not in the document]
isobel-softwire: You're completely right - sorry when I said constant I meant a `val`!

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/journeyDataExtensions/JourneyDataExtensions.kt
```diff
@@ -40,5 +40,13 @@ class JourneyDataExtensions {
             }
             return journeyDataValue ?: 0
         }
+
+        @JvmStatic
```
Travis-Softwire, line 44: What makes us need this vs. just the journey-specific extension?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/LandlordDeregistrationAreYouSureFormModel.kt
```diff
@@ -0,0 +1,35 @@
...
+                messageKey = "forms.areYouSure.landlordDeregistration.withProperties.radios.error.missing",
+                validatorType = DelegatedPropertyConstraintValidator::class,
+                targetMethod = "isNotNullWhenLandlordHasRegisteredProperties",
+            ),
+        ],
+    )
+    var wantsToProceed: Boolean? = null,
+    var userHasRegisteredProperties: Boolean? = null,
+) : FormModel {
+    fun isNotNullWhenLandlordHasRegisteredProperties(): Boolean =
```
Travis-Softwire, line 28: If we're assuming that `userHasRegisteredProperties` is never null in practice, then I think this is equivalent to just `userHasRegisteredProperties == false || NotNullValidator().isValid(wantsToProceed, null)` due to short circuiting.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/LandlordDeregistrationAreYouSureFormModel.kt
```diff
@@ -0,0 +1,35 @@
...
+        (NotNullValidator().isValid(wantsToProceed, null) && userHasRegisteredProperties == true) ||
+            userHasRegisteredProperties == false
+
+    fun isNotNullWhenLandlordHasNoProperties(): Boolean =
+        (NotNullValidator().isValid(wantsToProceed, null) && userHasRegisteredProperties == false) ||
```
Travis-Softwire, line 33: Same point here.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/LandlordDeregistrationCheckUserPropertiesPage.kt
```diff
@@ -0,0 +1,23 @@
+package uk.gov.communities.prsdb.webapp.forms.pages
+
+import org.springframework.validation.BindingResult
+import org.springframework.web.servlet.ModelAndView
+import uk.gov.communities.prsdb.webapp.forms.JourneyData
+import uk.gov.communities.prsdb.webapp.models.requestModels.formModels.NoInputFormModel
+
+class LandlordDeregistrationCheckUserPropertiesPage : AbstractPage(NoInputFormModel::class, "", mapOf()) {
```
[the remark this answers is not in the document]
JasminConterioSW: A new form model should work, I'll do that!

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/journeyDataExtensions/JourneyDataExtensions.kt
JasminConterioSW: Yes, it's so they use same fieldName but different urlPathSegments

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/LandlordDeregistrationAreYouSureFormModel.kt
JasminConterioSW: Yes, you're right!

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordService.kt
```diff
@@ -111,4 +111,7 @@ class LandlordService(
             landlordPage.totalElements,
         )
     }
+
+    // TODO: PRSD-703 - check the database to see if the principal has any registered properties
+    fun getLandlordHasRegisteredProperties(baseUserId: String): Boolean = false
```
JasminConterioSW, line 116: Implementing this will be in the next PR!

JasminConterioSW: Thinking about it, this might end up moving to a new LandlordDeregistrationService as it won't just be looking at the landlord table, but again this can go into the next PR (I needed it somewhere mockable for tests!)

JasminConterioSW: This was less fiddly than I thought it might be, so I'm adding it to this PR

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordDeregistrationJourneyTests.kt
```diff
@@ -0,0 +1,95 @@
...
+
+@Sql("/data-local.sql")
+class LandlordDeregistrationJourneyTests : IntegrationTest() {
+    @MockBean
+    lateinit var landlordDeregistrationService: LandlordDeregistrationService
+
+    @Test
+    fun `User with properties can navigate the whole journey`(page: Page) {
+        whenever(landlordDeregistrationService.getLandlordHasRegisteredProperties(anyString())).thenReturn(true)
+        val areYouSurePage = navigator.goToLandlordDeregistrationAreYouSurePage()
```
isobel-softwire, line 27: You could move the mocking logic into `goToLandlordDeregistrationAreYouSurePage()` and add a `landlordHasProperties` parameter

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordDeregistrationServiceTests.kt
```diff
@@ -0,0 +1,53 @@
...
+            .thenReturn(landlordWithListedPropertyCount)
+
+        // Act, Assert
+        assertTrue(landlordDeregistrationService.getLandlordHasRegisteredProperties(baseUserId))
+    }
+
+    @Test
+    fun `getLandlordHasRegisteredProperties returns false true if listedPropertyCount is 0`() {
+        // Arrange
+        val landlordWithListedPropertyCount = createLandlordWithListedPropertyCount(0)
```
isobel-softwire, line 45: I think this would be even easier to read if the `MockLandlordData` methods were qualified

isobel-softwire (commented):
Mostly LGTM - left some small comments

src/main/resources/templates/forms/areYouSureForm.html
```diff
@@ -1,14 +1,17 @@
...
+<!--/*@thymesVar id="optionalFieldSetHeadingParam" type="java.lang.String"*/-->
+<!--/*@thymesVar id="fieldSetHintList" type="java.lang.List<String>"*/-->
 <!--/*@thymesVar id="singleLineAddress" type="java.lang.String"*/-->
 <!--/*@thymesVar id="backUrl" type="java.lang.String"*/-->
 <!--/*@thymesVar id="radioOptions" type="java.lang.Object"*/-->
 <!DOCTYPE html>
 <html th:replace="~{fragments/forms/questionPage :: questionPage(#{${title}}, ${formModel}, ~{::#form-content}, ${backUrl}, null)}">
-<th:block id="form-content">
+<th:block id="form-content"
+          th:with="fieldSetHint=${#lists.isEmpty(fieldSetHintList) ? null : #messages.msgOrNull(fieldSetHintList[0])}" >
```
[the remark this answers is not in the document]
JasminConterioSW: Ah sorry I missed that, that version does worth thanks!

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordDeregistrationJourneyTests.kt
```diff
@@ -0,0 +1,95 @@
...
+    @ValueSource(booleans = [true, false])
+    fun `User is returned to the landlord details page if they click the back link`(
+        userHasRegisteredProperties: Boolean,
+        page: Page,
+    ) {
+        whenever(landlordDeregistrationService.getLandlordHasRegisteredProperties(anyString())).thenReturn(userHasRegisteredProperties)
+        val areYouSurePage = navigator.goToLandlordDeregistrationAreYouSurePage()
+        areYouSurePage.backLink.clickAndWait()
+        assertPageIs(page, LandlordDetailsPage::class)
+    }
```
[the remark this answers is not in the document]
JasminConterioSW: It feels safer to me to check that the behaviour is exactly the same for the two versions, even though it is maybe overkill

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordDeregistrationJourney.kt
```diff
@@ -0,0 +1,96 @@
...
+            // TODO: PRSD-704 - continue to reason page if user has registered properties
+        }
+        return "/$LANDLORD_DETAILS_PATH_SEGMENT"
+    }
+
+    private fun deregisterLandlord(): String {
+        // TODO: PRSD-703 - implement this
+        //      delete from landlord table
+        //      delete from one-login table (check if they are another type of user first)
+        //      refresh user roles to remove landlord permissions
```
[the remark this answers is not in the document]
JasminConterioSW: Yes, that'll be in the next one (sorry I should have commented, just didn't want to put everything in in one go)

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/journeyDataExtensions/LandlordDeregistrationJourneyDataExtensions.kt
```diff
@@ -0,0 +1,20 @@
...
+import uk.gov.communities.prsdb.webapp.models.requestModels.formModels.LandlordDeregistrationCheckUserPropertiesFormModel.Companion.USER_HAS_REGISTERED_PROPERTIES_JOURNEY_DATA_KEY
+
+class LandlordDeregistrationJourneyDataExtensions : JourneyDataExtensions() {
+    companion object {
+        fun JourneyData.getWantsToProceed() = this.getWantsToProceed(DeregisterLandlordStepId.AreYouSure.urlPathSegment)
+
+        fun JourneyData.getLandlordUserHasRegisteredProperties(): Boolean? =
+            JourneyDataHelper.getFieldBooleanValue(
+                this,
+                CHECK_FOR_REGISTERED_PROPERTIES_PATH_SEGMENT,
```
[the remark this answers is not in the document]
JasminConterioSW: I can't do that everywhere as `DeregisterLandlordStepId.CheckForUserProperties.urlSegment` cannot be used in the` @GetMapping` annotation because it's not a compile time constant, but I'll update it in everything referring to journey data

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordDeregistrationJourneyTests.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    @ParameterizedTest
+    @CsvSource(
+        "true, Select whether you want to delete your landlord record and properties",
+        "false, Select whether you want to delete your account from the database",
+    )
```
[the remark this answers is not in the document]
JasminConterioSW: Ah, I chose the wrong version to crib from!

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordDeregistrationJourneyTests.kt
JasminConterioSW: I did think about that, but we mostly don't have mocking logic in the Navigator. There are other examples where we setup mocks in the integration tests files (e.g. PropertyRegistrationJourneyTests, LandlordRegistrationJourneyTests) so if feels more consistent to me to keep the mocking here.

Travis-Softwire (approved):
Happy with this 👍

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordDeregistrationJourneyTests.kt
isobel-softwire: That makes sense. You could put the logic in a private method on `LandlordDeregistrationJourneyTests`, but if you'd prefer to keep it inline that's fine.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/LandlordDeregistrationAreYouSurePage.kt
```diff
@@ -0,0 +1,57 @@
...
+                    )
+                )
+
+        if (!userHasRegisteredProperties) {
+            modelAndView.addObject("fieldSetHeading", "forms.areYouSure.landlordDeregistration.noProperties.fieldSetHeading")
+        } else {
+            modelAndView.addObject("fieldSetHeading", "forms.areYouSure.landlordDeregistration.hasProperties.fieldSetHeading")
+            modelAndView.addObject("fieldSetHintList", listOf("forms.areYouSure.landlordDeregistration.hasProperties.fieldSetHint"))
+        }
+    }
```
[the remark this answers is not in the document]
JasminConterioSW: I was thinking this lived quite naturally on the page as it is page content, but then again is is headings which are often passed into the page for the journey... 

I think it could be either, I've moved the logic to a provider method on the journey for now but have a look and see if you think that makes more sense!

PR #287, opened 2025-03-21
Feat/prsd 999 example file upload template

AEPR:
Creates the file upload page and an endpoint that accepts the file, but no streaming currently occurs. (Will be in a future PR, I expect)

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/ExampleFileUploadController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/ExampleFileUploadController.kt
@@ -16,7 +16,7 @@ class ExampleFileUploadController {
     fun getFileUploadForm() = "example/fileUpload"
 
     @PostMapping(produces = [MediaType.APPLICATION_JSON_VALUE])
-    fun postmappedvalue(
+    fun uploadFile(
         @RequestParam("uploaded-file") file: MultipartFile,
         model: Model,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/ExampleFileUploadController.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.multipart.MultipartFile
+
+@Controller
+@RequestMapping("example/file-upload")
+class ExampleFileUploadController {
+    @GetMapping
+    fun getFileUploadForm() = "example/fileUpload"
+
+    @PostMapping(produces = [MediaType.APPLICATION_JSON_VALUE])
```
isobel-softwire, line 18: Is the `produces` argument needed?

isobel-softwire (approved):
This LGTM!

src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/ExampleFileUploadController.kt
AEPR: Nope, it shouldn't be there any more.

Travis-Softwire:
Also LGTM once the produces arg is sorted :)

PR #289, opened 2025-03-21
PRSD-999: Process uploaded file

AEPR:
Adds a `FileUploader` interface and two implementations. When running locally the file is saved to your file system for inspection. When running deployed it will either echo back the content of the stream as text (if the file uploaded has a `.txt` extension or read the stream and reply with the number of bytes read.

This follows on from [the other PR for PRSD-999](https://github.com/communitiesuk/prsdb-webapp/pull/287), and is probably best reviewed after it.


src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/FileUploader.kt
```diff
@@ -0,0 +1,13 @@
+package uk.gov.communities.prsdb.webapp.examples
+
+import java.io.InputStream
+
+// TODO PRSD-1001 - ensure this interface is updated to match what is needed for the AWS file upload service
```
Travis-Softwire, line 5: Was about to ask for a comment along these lines in the implementation, but then found it here :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/ExampleFileUploadController.kt
```diff
@@ -20,10 +22,13 @@ class ExampleFileUploadController {
         @RequestParam("uploaded-file") file: MultipartFile,
         model: Model,
     ): String {
+        val extension = file.originalFilename?.let { it.substring(it.lastIndexOf('.') + 1) }
```
Travis-Softwire, line 25: There are couple of commonly used util libraries that handle all the edge cases of filenames e.g. Apache's `FilenameUtils` - might be overkill if we only care about the extension though

Travis-Softwire (approved):
👍 One musing but def doesn't matter for this PR since it's just bootstrapping

PR #290, opened 2025-03-21
PRSD-939: Start Compliance after Property Registration

isobel-softwire:
### Features
- Adds `isOccupied` and `provideComplianceUrl` to model in `RegisterPropertyController` confirmation endpoint.
- Updates existing `registerPropertyConfirmation` content to match the new designs.
- Updates `registerPropertyConfirmation` template to show different content based on whether property is occupied. 

### Tests
- Consolidates `PropertyRegistrationJourneyTests` into one where the property is occupied and one where it's vacant. 
- Checks that a valid link to the compliance journey only appears when the property is occupied.

### Screenshots:
- Property is occupied
image

- Property is vacant
image


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationJourneyTests.kt
```diff
@@ -197,10 +200,115 @@ class PropertyRegistrationJourneyTests : IntegrationTest() {
...
         )
 
+        // Go to compliance journey
+        confirmationPage.addComplianceButton.clickAndWait()
+        assertPageIs(page, StartPageComplianceProvision::class, mapOf("propertyOwnershipId" to propertyOwnershipCaptor.value.id.toString()))
+    }
+
+    @Test
+    @Suppress("ktlint:standard:max-line-length")
+    fun `User can navigate the whole journey if pages are correctly filled in (manual address, custom property type, no license, unoccupied)`(
```
Travis-Softwire, line 210: I think this makes sense in this particular case because we have so many independent branches to the journey. I'd be wary of this becoming a pattern though - I'm still intending we'll set up single page testing properly...

Travis-Softwire (approved):
All LGTM :)

PR #292, opened 2025-03-24
PRSD-1020: Github Action for manually resetting the integration DB

isobel-softwire:
- Cleans integration DB
  - Checks out `main`
  - Starts SSM port-forwarding session
  - Retrieves integration DB parameters/secrets from SSM/Secrets manager and adds them to the environment
  - Runs `flywayClean`
  - Terminates the SSM session
- Restarts ECS

### DIFF
--- a/.github/workflows/reset-database.yml
+++ b/.github/workflows/reset-database.yml
@@ -1,0 +1,75 @@
+name: Reset Database
+on: workflow_dispatch
+
+permissions:
+  id-token: write
+  contents: read
+
+env:
+  AWS_REGION: eu-west-2
+
+jobs:
+  clean-database:
+    name: Clean Database
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::794038239680:role/github-actions-terraform-admin
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Start SSM port-forwarding Session
+        run: |
+          BASTION_ID=$(aws ec2 describe-instances --output text --filters "Name=tag:Name,Values=integration-bastion-1" "Name=instance-state-name,Values=running" --query "Reservations[*].Instances[].InstanceId | [0]")
+
+          DB_URL=$(aws ssm get-parameter --output text --name "integration-prsdb-database-url" --query "Parameter.Value")
+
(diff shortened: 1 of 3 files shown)


.github/workflows/reset-database.yml
```diff
@@ -0,0 +1,56 @@
...
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::794038239680:role/github-actions-terraform-admin
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Set DB ENV Variables
+        uses: aws-actions/aws-secretsmanager-get-secrets@v2.0.8
+        with:
+          secret-ids: |
+            FLYWAY_URL, integration-prsdb-database-url
```
Travis-Softwire, line 28: I'm afraid only the password is in secrets manager, the other two are in AWS SSM Parameter store. From a quick search, it doesn't look like AWS provides an action for that, oddly, so we'll want to use the AWS CLI. Have a look at how we set up the CLI in the build & deploy pipeline, and then we'll want to use `aws ssm get-parameter --name "<parameter name>"` to retrieve it and then put it into an environment variable

.github/workflows/reset-database.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::794038239680:role/github-actions-terraform-admin
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Set DB URL & username in ENV
+        run: |
+          echo "FLYWAY_URL=$(aws ssm get-parameter --name "integration-prsdb-database-url")" >> $GITHUB_ENV
```
Travis-Softwire, line 29: Is it clever enough to allow nested double quotes? If you want to test, you can add an on push trigger, and comment out the flywayClean and restart ecs tasks

Travis-Softwire:
I forgot to put this on the ticket, but before merging this can you find time with one of me, @AEPR, or @Kiran-R-K once she has setup AWS access so we can extract MHCLG & Made Tech email One Login IDs from the database on integration & add them to the seed data?

Travis-Softwire:
This LGTM - I think the order of events from here should be:
- Get changes to `data-*.sql` made and merged so that we retain users in the db for everyone
- Test this action in full by temporarily changing the trigger
- Once it's working, merge it :)

Travis-Softwire (approved):
All looks good, but let's hold off merging for now. I've added a comment on proposed next steps

PR #293, opened 2025-03-24
Feat/prsd 999 example file upload session token

AEPR:
Adds a token to ensure that you are calling the post having been issued with a token in a cookie during a get. As currently implemented this means that you cannot actually post twice in a row without a get in the middle. That will be addressed in PRSD-1003.


src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/FileUploadTokenService.kt
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.stereotype.Service
+import java.util.UUID
+
+// TODO: PRSD-945/PRSD-395 Move this to service folder and manage multiple tokens
+// It may be that these session attributes should/will need store additional data, e.g. property ownership id
+@Service
+class FileUploadTokenService(
+    private val session: HttpSession,
+) {
+    fun issueTokenFor(value: Any): String {
```
Travis-Softwire, line 13: I'd imagine in practice this will always be a string?

AEPR: Could be - especially if you just pass in the full route. I guess I thought you might decide to pass in a `data class`, e.g. 
```kotlin
data class ComplianceUploadContext(val ownershipId: Int, val uploadType: UploadType)
```

I didn't want to be particularly opinionated in the example, so I just left it as any.

PR #297, opened 2025-03-25
PRSD-1020: Github Actions for manually resetting lower environment DBs

isobel-softwire:
- Makes `reset-database.yml` a reusable workflow
- Parameterizes `reset-database.yml` on target environment
- Creates `reset-database-integration.yml` and `reset-database-test.yml` workflows, which call `reset-database.yml`

### DIFF
--- a/.github/workflows/reset-database.yml
+++ b/.github/workflows/reset-database.yml
@@ -42,6 +42,7 @@ jobs:
           aws-region: ${{ env.AWS_REGION }}
 
       - name: Start SSM port-forwarding Session
+        id: start-session
         run: |
           BASTION_ID=$(aws ec2 describe-instances --output text --filters "Name=tag:Name,Values=${{ inputs.environment }}-bastion-1" --query "Reservations[*].Instances[].InstanceId | [0]")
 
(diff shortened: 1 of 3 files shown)


.github/workflows/reset-database.yml
```diff
@@ -57,7 +76,7 @@ jobs:
         run: ./gradlew flywayClean --no-daemon
 
       - name: Terminate SSM port-forwarding session
-        if: success() || failure()
+        if: env.SSM_SESSION_ID != ''
```
Travis-Softwire, line 79: If you're not using an environment variable in string interpolation I think you can just use e.g. `$SSM_SESSION_ID != ''` in cases like this

Travis-Softwire, line 79: Q: what happens if the flyway command fails? I don't think this would be run, in which case we'd have a dangling port forwarding session?

Travis-Softwire (approved):
Looks really good :) One substantive comment about under what conditions the session termination will happen, but once that's resolved I think this is good to go

.github/workflows/reset-database.yml
isobel-softwire: `SSM_SESSION_ID` is an environment variable - it gets added to the environment in the `Start SSM port-forwarding Session` step. I'll change it to be an output from that step instead so it's clearer.

Travis-Softwire: Sorry, it was my mistake - you're completely right that you need to use the `env` context here, it's only inside `run` commands you can access them just using the `$` syntax. You can still make it an output if you like, I don't mind either way, but you can ignore my original point!

isobel-softwire: You're right - I didn't realise I'd still need the status check. Added it back 👍

PR #299, opened 2025-03-25
PRSD-NONE: Adds minimal changes required to upgrade to Spring 3.4

Travis-Softwire:


build.gradle.kts
```diff
@@ -67,7 +67,6 @@ dependencies {
     testImplementation("org.springframework.boot:spring-boot-starter-test")
     testImplementation("org.springframework.boot:spring-boot-testcontainers")
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit5")
-    testImplementation("net.sourceforge.htmlunit:htmlunit")
```
Travis-Softwire, line 70: This was causing an error, but we weren't actually using it any more

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordRegistrationJourneyTests.kt
```diff
@@ -55,7 +55,7 @@ class LandlordRegistrationJourneyTests : IntegrationTest() {
     @Autowired
     private lateinit var landlordService: LandlordService
 
-    @SpyBean
```
Travis-Softwire, line 58: One of the updates in Spring boot (haven't been able to pin down if it's Mockito or something else) mean that `SpyBean`s or even `MockitoSpyBean`s no longer work when the bean is a generic interface

isobel-softwire (approved):
LGTM :)

PR #300, opened 2025-03-25
PRSD-968: Copy integration seed data with fixes

AEPR:
Copies the seed data from `data-integration.sql` with a couple of fixes. Specifically:
* UPRN 1 is a real UPRN, so changed the two used to 2 and 3 which are not currently in use.
* The single property ownership referenced `registration_number` 6, which does not exist. Updated to be 2 which does.

I've applied these fixes to the integration seed data as well.


isobel-softwire (approved):
LGTM!

PR #301, opened 2025-03-26
PRSD-1001: Add AwsS3FileUploader and update interface

AEPR:
Uploads to S3 when not running locally.

We will need to set the quarantine bucket string for the build before it will work! I'm not sure the best way to configure that.


src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/LocalFileUploader.kt
```diff
@@ -1,20 +1,28 @@
...
+        streamSize: Long,
     ): String {
-        File(".local-uploads").mkdir()
-        val extensionWithDot = if (extension != null) ".$extension" else ""
-        val destinationRoute = ".local-uploads/destination$extensionWithDot"
+        val cleanObjectKey =
+            objectKey
+                .map { char -> if (char in forbiddenFilenameCharacters) "" else char }
+                .joinToString("")
+        File(".local-uploads/${cleanObjectKey.substring(0, cleanObjectKey.lastIndexOf('/'))}").mkdirs()
```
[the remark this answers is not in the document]
AEPR: It's already there! (I added it in a previous PR).

src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/LocalFileUploader.kt
```diff
@@ -1,20 +1,28 @@
...
 import org.springframework.context.annotation.Profile
 import org.springframework.stereotype.Service
 import java.io.File
 import java.io.InputStream
 
 @Service
+@Primary
 @Profile("local")
 class LocalFileUploader : FileUploader {
+    private val forbiddenFilenameCharacters = listOf(':', '<', '>', '"', '?', '*', '&')
```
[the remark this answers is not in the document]
AEPR: I'm actually allowing "/" deliberately as the `mkDirs()` creates all the directories in the path so (on windows) this creates a folder structure mirroring the string passed in. That said, since this is just for running locally, stripping out all '/' and '\' so its compeltely flat might make more sense.

src/test/resources/application.yml
```diff
@@ -10,3 +10,7 @@ notify:
 
 os-places:
   api-key: ${OS_PLACES_API_KEY}
+
+aws:
+  s3:
+    quarantineBucket: 'quarantine-bucket'
```
AEPR, line 16: I'm not sure if this is the best way to add a `@Value` to the tests. It doesn't actually matter here because the integration tests have a mock s3Client so this bucket isn't getting used for anything, but it needs to be passed in to the `FileUploader` to make Spring happy.

src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/AwsS3FileUploader.kt
```diff
@@ -0,0 +1,34 @@
...
+@Service
+class AwsS3FileUploader(
+    private val s3Client: S3Client,
+    @Value("\${aws.s3.quarantineBucket}")
+    private val bucketName: String,
+) : FileUploader {
+    override fun uploadFile(
+        objectKey: String,
+        inputStream: InputStream,
+        streamSize: Long,
```
Travis-Softwire, line 19: Just to check - is the expectation that this is just an intermediate "does uploading to s3 work" implementation, and we'll replace it with one more akin to [this one](https://docs.aws.amazon.com/code-library/latest/ug/s3_example_s3_Scenario_UploadStream_section.html) later? I ask as I don't think we'll know the length of the stream in the final version

src/main/resources/application.yml
```diff
@@ -59,6 +59,9 @@ server:
   error:
     path: /error
 
+aws:
+  s3:
+    quarantineBucket: ${AWS_QUARANTINE_BUCKET}
```
Travis-Softwire, line 64: Obviously not for this PR, but seeing this has made me realise we should probably pop this in an SSM parameter so that we can inject it into the task definition

src/test/resources/application.yml
Travis-Softwire: I'm ok with this - alternatively if it's only needed in a small number of places you could look at `@TestPropertySource`

Travis-Softwire (approved):
Just double checking my understanding of the plan with the file uploader, but otherwise LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/AwsS3FileUploader.kt
AEPR: I think we will know the size from the `Content-Length` header - as long as the body of the upload only contains the file.

AEPR: In any case - this can be resolved in PRSD-1002. This is a sensible intermediate point, showing a successful upload to S3.

PR #302, opened 2025-03-26
Feat/prsd 703 landlord deregistration submit are you sure

JasminConterioSW:
Landlord deregistration "are you sure" page

If the landlord has no properties, deregister them when they submit this page.
Refresh the user roles to reset their permissions.

I also re-arranged the journey tests so that the with / without properties version use different data....sql files


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordDeregistrationJourney.kt
```diff
@@ -81,15 +86,19 @@ class LandlordDeregistrationJourney(
     }
 
     private fun deregisterLandlord(): String {
-        // TODO: PRSD-703 - implement this
-        //      delete from landlord table
-        //      delete from one-login table (check if they are another type of user first)
-        //      refresh user roles to remove landlord permissions
+        val baseUserId = SecurityContextHolder.getContext().authentication.name
```
Travis-Softwire, line 89: Elsewhere I think we've passed in the principal name rather than getting out of the security context inside the journey. I think I'd probably prefer that so that the journey isn't directly coupled to our decision to use the sub from One Login as the id for our base users.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordDeregistrationService.kt
```diff
@@ -0,0 +1,27 @@
...
+    private val localAuthorityUserRepository: LocalAuthorityUserRepository,
+) {
+    @Transactional
+    fun deregisterLandlord(baseUserId: String) {
+        landlordRepository.deleteByBaseUser_Id(baseUserId)
+        deleteFromOneLoginIfNotAnotherTypeOfUser(baseUserId)
+    }
+
+    private fun deleteFromOneLoginIfNotAnotherTypeOfUser(baseUserId: String) {
+        val userIsLocalAuthorityUser = localAuthorityUserRepository.findByBaseUser_Id(baseUserId) != null
```
Travis-Softwire, line 22: We're doing potentially 2 round trips to the db here in addition to the one deleting the landlord record (unless hibernate/ jpa is optimising them away). 

I'm fine with this given we're not expecting deletion to happen in large volumes and we'll be replacing it by marking records as inactive in the mandatory service, but in general worth considering if we can reduce the number of db calls with a custom query or by using cascading relationships.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordDeregistrationJourneyTests.kt
```diff
@@ -3,106 +3,98 @@ package uk.gov.communities.prsdb.webapp.integration
...
 import uk.gov.communities.prsdb.webapp.integration.pageObjects.components.BaseComponent.Companion.assertThat
 import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.LandlordDetailsPage
 import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.basePages.BasePage.Companion.assertPageIs
-import uk.gov.communities.prsdb.webapp.services.LandlordService
 
-@Sql("/data-local.sql")
 class LandlordDeregistrationJourneyTests : IntegrationTest() {
-    @MockBean
-    lateinit var landlordService: LandlordService
+ /*   @MockBean
```
Travis-Softwire, line 14: Is this intentionally left in but commented out?

Travis-Softwire (approved):
I've left a couple of Qs/ comments but overall happy with this :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordDeregistrationJourney.kt
JasminConterioSW: At the moment we are getting this from the context in a few journeys (LaUserRegistration, LandlordRegistration, PropertyRegistration and UpdateLandlordDetails), shall I raise a ticket to change those too? Or I could just change it in another PR

We can instead pass the principal.name into the journey when we create() it, like we do with the propertyOwnershipId in the PropertyDeregistrationJourney, is that what you were thinking of?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordDeregistrationJourneyTests.kt
JasminConterioSW: Nope, that should go...

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordDeregistrationService.kt
JasminConterioSW: I've had a go with a custom query, which seems to be working.

The behaviour is not really covered by tests though as I don't think we have any testing of the repository layer set up - should I be looking into that if we might be doing more in the way of custom queries?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordDeregistrationJourney.kt
Travis-Softwire: Ah, so we do - I had a quick look but didn't spot those. In that case fine to leave it as it is, and we can pick up at the next dev catchup. It may be that this is too deeply imbedded and too unlikely to change for it to be worth the effort.

PR #303, opened 2025-03-27
PRSD-603: Adds new filter to handle trailing forward slashes

Travis-Softwire:
- Creates new OncePerRequestFilter subclass that intercepts any request with a url - excluding query strings - that ends in a '/', wraps it in a new request without the trailing slash, and delegates handling that new request to the rest of the filter chain
  - this leverages the new `UrlHandlerFilter` added in Spring boot 3.4
 - Adds this as the highest precedence filter to ensure that the rest of the filter chain operates on the correct url
 - Adds two tests to ensure this all works - one for a public endpoint and one for one behind auth

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/filters/TrailingSlashFilter.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/filters/TrailingSlashFilter.kt
@@ -1,0 +1,27 @@
+package uk.gov.communities.prsdb.webapp.config.filters
+
+import jakarta.servlet.FilterChain
+import jakarta.servlet.ServletException
+import jakarta.servlet.http.HttpServletRequest
+import jakarta.servlet.http.HttpServletResponse
+import org.springframework.core.Ordered
+import org.springframework.core.annotation.Order
+import org.springframework.stereotype.Component
+import org.springframework.web.filter.OncePerRequestFilter
+import org.springframework.web.filter.UrlHandlerFilter
+import java.io.IOException
+
+@Component
+@Order(Ordered.HIGHEST_PRECEDENCE)
+class TrailingSlashFilter : OncePerRequestFilter() {
+    @Throws(ServletException::class, IOException::class)
+    override fun doFilterInternal(
+        request: HttpServletRequest,
+        response: HttpServletResponse,
+        filterChain: FilterChain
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/HealthCheckControllerTests.kt
```diff
@@ -10,6 +11,9 @@ import kotlin.test.Test
 class HealthCheckControllerTests(
     @Autowired val webContext: WebApplicationContext,
 ) : ControllerTest(webContext) {
+    @Autowired
+    private lateinit var mockMvc: MockMvc
+
```
AEPR, line 16: I'm slightly surprised we need to declare this separately to the mvc already in `ControllerTest`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/filters/TrailingSlashFilter.kt
```diff
@@ -0,0 +1,27 @@
...
+class TrailingSlashFilter : OncePerRequestFilter() {
+    @Throws(ServletException::class, IOException::class)
+    override fun doFilterInternal(
+        request: HttpServletRequest,
+        response: HttpServletResponse,
+        filterChain: FilterChain,
+    ) {
+        // Wraps any request with a url that ends in a '/' in a new request without that trailing slash
+        val filter: UrlHandlerFilter = UrlHandlerFilter.trailingSlashHandler("/**").wrapRequest().build()
+        filter.doFilter(request, response, filterChain)
```
AEPR, line 25: Why are we not just using a `UrlHandlerFilter` directly? E.g. declaring a bean = `UrlHandlerFilter.trailingSlashHandler("/**").wrapRequest().build()`

Does this do something different?

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/HealthCheckControllerTests.kt
Travis-Softwire: No you're completely right - that's left over from a failed experiment and I missed it during clean up!

AEPR (approved):
LGTM

PR #305, opened 2025-03-27
Chore/prsd 1018 use reflection for fieldnames

Kiran-R-K:
Switches to using reflection to get field names from Form Models instead of inputting them as raw strings.

E.g. instead of:
```
getFieldStringValue(
                journeyData,
                RegisterLaUserStepId.Name.urlPathSegment,
                "name",
            )
```

we now have:
```
  getFieldStringValue(
                journeyData,
                RegisterLaUserStepId.Name.urlPathSegment,
                NameFormModel::class.memberProperties.first().name,
            )
```

### DIFF
--- a/.run/local.run.xml
+++ b/.run/local.run.xml
@@ -5,7 +5,7 @@
       <option value="$PROJECT_DIR$/.env" />
     </option>
     <option name="FRAME_DEACTIVATION_UPDATE_POLICY" value="UpdateClassesAndResources" />
-    <module name="uk.gov.communities.prsdb.prsdb-webapp.main" />
+    <module name="prsdb-webapp.main" />
     <selectedOptions>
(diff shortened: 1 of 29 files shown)


rowanhill (changes requested):
I don't think we should be using `MyClass::class.memberProperties` and then getting first / last / elementAt. That relies on the ordering of the member properties, and I don't even know if that's guaranteed to be stable (e.g. might an update of the Kotlin compiler break our app? 😱)

Why not instead use `MyClass::property.name` directly? It's shorter and more legible, too.

(I'm setting aside the question of whether it's sensible / necessary to have these strings match the Form class properties exactly - I assume that has already been determined to be the case)

Travis-Softwire:
I think we can close this now in favour of #315

PR #307, opened 2025-03-27
Feat/prsd 1002 upload streaming without memory buffer

AEPR:
Switches to using an async, size agnostic `S3TransferManager` for the upload to S3 and to using Apache FileUpload2 for the controller. 

Still a draft, I think I still need to do (tomorrow):
* clean up the controller method to be 
    * easier to understand what I'm doing
    * more encapsulated
    * simpler to transfer to another context (i.e. a real upload endpoint!)
* add a "max size" step to the stream that throws after a certain number of bytes are read.

Mostly putting up a draft because I'm away after tomorrow so I should be oversharing rather than risking my work being wasted.

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/CustomSecurityConfig.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/CustomSecurityConfig.kt
@@ -55,7 +55,7 @@ class CustomSecurityConfig(
             .logout { logout ->
                 logout.logoutSuccessHandler(oidcLogoutSuccessHandler())
             }.csrf { requests ->
-                requests.disable()
+                requests.ignoringRequestMatchers("/local/**")
             }
 
         return http.build()
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/CustomSecurityConfig.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/CustomSecurityConfig.kt
@@ -55,7 +55,7 @@ class CustomSecurityConfig(
             .logout { logout ->
                 logout.logoutSuccessHandler(oidcLogoutSuccessHandler())
             }.csrf { requests ->
-                requests.ignoringRequestMatchers("/local/**")
+                requests.disable()
             }
 
         return http.build()
--- a/src/main/resources/application-local.yml
+++ b/src/main/resources/application-local.yml
@@ -38,10 +38,6 @@ spring:
       ssl:
         enabled: false
 
-  servlet:
-    multipart:
-      enabled: false
-
 one-login:
   jwt:
     # We provide a default for the keys if the env vars are missing: this allows us to keep the keys in uncommitted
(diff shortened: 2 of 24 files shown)


Travis-Softwire:
This all sounds sensible plan-wise.

On:
> add a "max size" step to the stream that throws after a certain number of bytes are read.
This might be a bit more involved as I think you'll need a custom InputStream subclass to do it - If you have time in this PR then great, but I wouldn't prioritise it over getting everything in a working and understandable state.

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/CustomSecurityConfig.kt
```diff
@@ -55,7 +55,7 @@ class CustomSecurityConfig(
             .logout { logout ->
                 logout.logoutSuccessHandler(oidcLogoutSuccessHandler())
             }.csrf { requests ->
-                requests.ignoringRequestMatchers("/local/**")
+                requests.disable()
```
isobel-softwire, line 58: Why does this need to change?

src/main/resources/application-local.yml
```diff
@@ -38,6 +38,10 @@ spring:
       ssl:
         enabled: false
 
+  servlet:
+    multipart:
+      enabled: false
```
isobel-softwire, line 43: Why are we disabling this locally?

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/CustomSecurityConfig.kt
AEPR: Definitely shouldn't in the non-draft version - I haven't resolved csrf for the fileupload, but it causes issues and I was isolating it!

src/main/resources/application-local.yml
AEPR: This will be applied everywhere in the non-draft version. Basically, if it's enabled spring loads the entire file into memory before letting the controller access it which will cause issues for large files.

src/main/resources/templates/example/fileUpload.html
```diff
@@ -1,5 +1,5 @@
 <th:block id="main-content" th:replace="~{fragments/layout :: layout('Example Form Page', ~{::#main-content/content()}, null)}">
-    <form th:action="@{''}" novalidate method="post" enctype="multipart/form-data">
+    <form th:action="@{'?' + ${_csrf.parameterName} + '=' + ${_csrf.token}}" novalidate method="post" enctype="multipart/form-data">
```
isobel-softwire, line 2: Might be neater to use the `csrfToken` fragment here

isobel-softwire (approved):
LGTM - left a small suggestion on the `fileUpload` template

src/main/resources/templates/example/fileUpload.html
AEPR: Good thought, but unfortunately I can't! Because I've switched off multipart processing it can't read the hidden inputs in the form. We need to keep it off though to prevent caching the whole file on the server.

Travis-Softwire (approved):
👍

PR #309, opened 2025-03-28
Feat/prsd 1002 add stream size capping

AEPR:
Allows for a maximum size of an input stream, after which it throws an exception. The tests are split into two sections - mostly boilerplate that checks the delegating methods continue to delegate and the three interesting ones that check that reading the stream works how we like.


AEPR:
That sounds much better that writing our own wrapper for all of this - and pretty simple too. 

I do like it being creatable from an extension method though.

PR #311, opened 2025-03-31
PRSD-603: Use permanent redirects instead of wrapping

Travis-Softwire:


isobel-softwire (approved):
LGTM :)

rowanhill:
Are we confident a 308 is what we want to use, here? I'm just always a bit nervous about permanent redirects, in case any client ends up caching them and we actually change our minds in future.

Travis-Softwire:
> Are we confident a 308 is what we want to use, here? I'm just always a bit nervous about permanent redirects, in case any client ends up caching them and we actually change our minds in future.

We could make them 307s if we're worried about it (although I can't imagine a case where we'd ever need to specifically allow a request to a url with a trailing `/`). It feels semantically a bit odd to say they're temporary redirects but I don't think there's any other practical difference. It needs to be a 307 or a 308 though to work with POST requests.

rowanhill:
Is there a meaningful downside to a 307? I think the main issue is around SEO - with a 308 search engine crawlers will treat the two URLs as basically the same thing, but with a 307 they'll score them separately, so I guess we might end up lower down search results. Feels like that'll be unimportant, though (and it'll be the gov.uk landing page that needs good SEO scores)?

I agree it's not the likely to matter, but you never know :)

PR #313, opened 2025-03-31
Feat/prsd 891 deregister landlord submit reason page

JasminConterioSW:
This covers deregistering a landlord user with properties when they submit the "Reason" page (PRSD-891)

- Moves property deregistration functions to a new PropertyDeregistrationService (from PropertyRegistrationService)
- Deletes the license when deregistering a property (or list of properties)
- Add new method to delete multiple properties
- When deregistering a landlord, always check for properties before deleting (we need all properties, not just the active ones) and delete these first


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordDeregistrationJourney.kt
```diff
@@ -107,25 +107,27 @@ class LandlordDeregistrationJourney(
     ): String {
         if (journeyData.getWantsToProceed()!!) {
             if (!journeyData.getLandlordUserHasRegisteredProperties()!!) {
-                return deregisterLandlord()
+                // journeyData.getLandlordUserHasRegisteredProperties() only checked for active, registered properties.
```
JasminConterioSW, line 110: getLandlordUserHasRegisteredProperties looks for active, registered properties (this is the properties that the landlord will see on their property record and it's the properties the LA user see in a landlord's listed property count.). This is used to determine which version of the journey they see.

When it comes to deleting the landlord, we need to look again and make sure we get ALL their properties even the ones that are hidden (inactive) as they all link to the landlord table via primary_landlord_id

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordDeregistrationService.kt
```diff
@@ -9,9 +9,16 @@ import uk.gov.communities.prsdb.webapp.database.repository.OneLoginUserRepositor
 class LandlordDeregistrationService(
     private val landlordRepository: LandlordRepository,
     private val oneLoginUserRepository: OneLoginUserRepository,
+    private val propertyOwnershipService: PropertyOwnershipService,
+    private val propertyDeregistrationService: PropertyDeregistrationService,
 ) {
     @Transactional
-    fun deregisterLandlord(baseUserId: String) {
+    fun deregisterLandlordAndTheirProperties(baseUserId: String) {
```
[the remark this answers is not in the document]
JasminConterioSW: We could reduce the calls to the database by using cascade delete. So we could set it up so that deleting the landlord cascade deletes all the property ownerships, which cascade deletes all the properties and licences. This feels dangerous and might not be what we want once we start including joint landlords etc.

A bit less drastic would be to set it up to cascade delete the properties and licenses when the property ownership is deleted. This might be good to do unless we expect properties or licenses to be transferred between property ownerships?

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyControllerTests.kt
```diff
@@ -28,17 +28,17 @@ import kotlin.test.assertEquals
 class DeregisterPropertyControllerTests(
     @Autowired val webContext: WebApplicationContext,
 ) : ControllerTest(webContext) {
-    @MockBean
+    @MockitoBean
```
JasminConterioSW, line 31: Updated this as I was already updating this file - it's giving me red lines saying `@MockBean` is deprecated now we have updated Spring!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordDeregistrationJourney.kt
Travis-Softwire: Yes, this makes sense. I'd expect the properties/ no-properties distinction only to make a difference in terms of copy or the pages they see, not how we approach actually deleting the records.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordDeregistrationJourney.kt
```diff
@@ -107,25 +107,27 @@ class LandlordDeregistrationJourney(
...
         return "/$LANDLORD_DETAILS_PATH_SEGMENT"
     }
 
-    private fun deregisterLandlordAndProperties(): String {
-        // TODO: PRSD-891
-        return "/${REGISTER_LANDLORD_JOURNEY_URL}"
-    }
-
-    private fun deregisterLandlord(): String {
+    private fun deregisterLandlordAndProperties(userHadRegisteredProperties: Boolean): String {
```
Travis-Softwire, line 120: Can we make this something like `userHadActiveProperites` or `userHadActiveRegisteredProperties` (not sure we need the 'registered' bit?) to make clear that it's only active properties that this flag relates to? Otherwise it could be a bit surprising that we delete properties even when the flag is false.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordDeregistrationService.kt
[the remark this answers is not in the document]
Travis-Softwire: This is a tricky one because as you say it will change when we go from deleting -> marking as inactive. It's also possible that JPA will at least bundle the delete queries to reduce the number of round trips to the DB. So I'm fine to leave this as-is for now but we'll want to make sure we cover deletion during performance testing.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyDeregistrationService.kt
```diff
@@ -0,0 +1,45 @@
...
+import uk.gov.communities.prsdb.webapp.database.entity.PropertyOwnership
+
+@Service
+class PropertyDeregistrationService(
+    private val propertyService: PropertyService,
+    private val licenseService: LicenseService,
+    private val propertyOwnershipService: PropertyOwnershipService,
+    private val session: HttpSession,
+) {
+    fun deregisterProperty(propertyOwnershipId: Long) {
```
Travis-Softwire, line 16: Does this not also need to be transactional to avoid the possibility of orphaned licenses or properties?

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyControllerTests.kt
Travis-Softwire: Yep, we have a ticket to go through and fix all of these, but happy for you to update these here since you were already touching the file

Travis-Softwire (approved):
A couple of minor thoughts but LGTM overall 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyDeregistrationService.kt
JasminConterioSW: Oh yes, we do

PR #314, opened 2025-03-31
Feat/prsd 705 deregister landlord confirmation page

JasminConterioSW:
Creates the confirmation endpoint for landlord deregistration PRSD-705, including the template for landlords with no properties

image

PR #316, opened 2025-03-31
PRSD-707: Add confirmation page for landlord with properties deregist…

JasminConterioSW:
Add the confirmation page for deregistered a landlord with properties (PRSD-707)

image

Note this one is branching off (https://github.com/communitiesuk/prsdb-webapp/pull/314), 
and the place where we redirect to the confirmation page changes in (https://github.com/communitiesuk/prsdb-webapp/pull/313). This one also just throws an error at the moment as the landlord is not actually deleted until (https://github.com/communitiesuk/prsdb-webapp/pull/313) is merged!


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterLandlordController.kt
```diff
@@ -104,7 +104,6 @@ class DeregisterLandlordController(
                 .getLandlordUserHasRegisteredProperties()!!
 
         if (landlordHadRegisteredProperties) {
-            // TODO PRSD-707: create this template
             return "deregisterLandlordWithRegisteredPropertiesConfirmation"
         }
         return "deregisterLandlordWithNoPropertiesConfirmation"
```
isobel-softwire, line 109: Minor but this could be `return if ...`

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordDeregistrationJourney.kt
```diff
@@ -109,25 +108,20 @@ class LandlordDeregistrationJourney(
...
+                return deregisterLandlordAndProperties()
             }
             val areYouSureStep = steps.single { it.id == DeregisterLandlordStepId.AreYouSure }
             return getRedirectForNextStep(areYouSureStep, journeyData, subPageNumber)
         }
         return "/$LANDLORD_DETAILS_PATH_SEGMENT"
     }
 
-    private fun deregisterLandlordAndProperties(userHadActiveProperties: Boolean): String {
+    private fun deregisterLandlordAndProperties(): String {
```
isobel-softwire, line 119: As it's at the end of the journey, should this method clear the journey data?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordDeregistrationJourneyTests.kt
```diff
@@ -26,12 +26,12 @@ class LandlordDeregistrationJourneyTests : IntegrationTest() {
...
+            val confirmationPage = assertPageIs(page, ConfirmationPageLandlordDeregistration::class)
+            assertThat(confirmationPage.confirmationBanner).containsText("You have deleted your account from the database")
             assertTrue(
-                areYouSurePage.page
-                    .url()
-                    .toString()
-                    .contains("register-as-a-landlord"),
+                confirmationPage.page
+                    .content()
+                    .contains("You have deleted your landlord information and all your properties from the database"),
```
isobel-softwire, line 34: Might be worth verifying that the landlord (and their properties) have been deleted

isobel-softwire (approved):
Left a few thoughts, but happy to approve. LGTM!

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordDeregistrationJourneyTests.kt
JasminConterioSW: Checking that they can't access the landlord dashboard should cover that - if they are still a landlord they would be able to access it. They also can't be deleted as a landlord until their properties have been deleted (because otherwise their landlord.id is referenced in the property_ownership table).

There are also tests in the service layer checking that the right delete command get called on the repo when the landlord is deregistered but those were merged in in a different PR.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordDeregistrationJourney.kt
JasminConterioSW: Good catch, this is missing in the PropertyDeregistrationJourney as well so I'll add that in

JasminConterioSW: Ah, actually we still need this because on the confirmation page we check whether the user had registered properties as they went through the journey. I could clear the session in the confirmation endpoint but then you wouldn't be able to refresh the page, so I think I'll leave it!

isobel-softwire: I'm slightly worried about what could happen if we don't clear journeys after completing them. `userHadActiveProperties` could be added to the session separately before clearing journey data?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordDeregistrationService.kt
```diff
@@ -22,4 +25,12 @@ class LandlordDeregistrationService(
...
         oneLoginUserRepository.deleteIfNotLocalAuthorityUser(baseUserId)
     }
+
+    fun addLandlordHadActivePropertiesToSession(hadActiveProperties: Boolean) =
+        session.setAttribute(
+            LANDLORD_HAD_ACTIVE_PROPERTIES,
+            hadActiveProperties,
+        )
+
+    fun getLandlordHadActivePropertiesFromSession(): Boolean = (session.getAttribute(LANDLORD_HAD_ACTIVE_PROPERTIES) == true)
```
isobel-softwire, line 35: Are the brackets here needed?

isobel-softwire (approved):
LGTM!

PR #320, opened 2025-04-02
Feat/prsd 706 deregister landlord no properties confirmation email

JasminConterioSW:
Sends a confirmation email when a landlord with no properties is deregistered

image


Travis-Softwire (approved):
👍

PR #326, opened 2025-04-03
Fix/prsd 895 handle no addresses found lookup in lookup step

JasminConterioSW:
This is no-where near ready for review!
We may also not want to use this approach, but if we do this is a start....

This branch just has the front end template, which should be useful whether we go for this approach or not:
https://github.com/communitiesuk/prsdb-webapp/tree/fix/PRSD-895-handle-no-addresses-found

## Approach
- General changes
    - Give Step the option of having an `autocompleteAndRedirect` function.
    - `autocompleteAndRedirect` (if it is not null) gets called once in Journey's `getModelAndViewForStep` once we have checked that the step is reachable. 
        - This can return null, in which case the rest of `getModelAndViewForStep` executes as normal and the step's page is rendered.
        - Or it can return a url string, in which case the user is redirected without rendering the step.
- When submitting the LookupAddress page (performLookupCacheResultsAndDirectToNextStep)
    - Do the lookup on OS places and store the results in the JourneyData. 
    - Redirect to the next step
- When getting the SelectAddress page `autocompleteSelectAddressWithManualAddressChosen` is called:
    - it returns null if address lookup found addresses. SelectAddress should then be rendered as normal.
    - If addressLookup found no addresses, this sets `address` to `MANUAL_ADDRESS_CHOSEN` (I think we need this as I think we use it later), then redirects to the NoAddressFound step

## Some bits that still need doing (if this is used)
Add the implementation for other journeys (LandlordRegistration and UpdateLandlordDetails) - at the moment these won't work because the lookup functionality has been taken out of SelectAddressPage but not added into those journeys)

Testing - check that we can get to the steps we should be able to get to, and can't get to those we shouldn't!


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/SelectAddressPage.kt
```diff
@@ -37,11 +35,7 @@ class SelectAddressPage(
                 lookupAddressPathSegment,
             )!!
 
-        val addressLookupResults = addressLookupService.search(houseNameOrNumber, postcode)
-
-        val journeyData = journeyDataService.getJourneyDataFromSession()
-        val updatedJourneyData = journeyData.withUpdatedLookedUpAddresses(addressLookupResults)
-        journeyDataService.setJourneyDataInSession(updatedJourneyData)
+        val addressLookupResults = journeyDataService.getJourneyDataFromSession().getLookedUpAddresses()
```
JasminConterioSW, line 38: It _may_ be easier to redirect here if there are no results, but currently this doesn't feel like the place to do it.
- This function is just adding to the model
- The page doesn't know about where it is in the journey, and doesn't know about other steps that it may need to redirect to (although this could be passed in?)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
```diff
@@ -164,12 +169,32 @@ class PropertyRegistrationJourney(
                         ),
                     shouldDisplaySectionHeader = true,
                 ),
+            handleSubmitAndRedirect = { journeyData, subPage -> performLookupCacheResultsAndDirectToNextStep(journeyData, subPage) },
             nextAction = { _, _ -> Pair(RegisterPropertyStepId.SelectAddress, null) },
```
Travis-Softwire, line 173: I wonder if we could instead:
- in the `nextAction` function for the lookup address step, check the looked-up addresses and route to your new no results page if they are empty
- in your new no results page, add a formModel that just has a hard-coded value to say that a manual address has been chosen (just to record in the journey data that the page has been visited)
- in our journeyData extensions/ helpers, check both the select address page and the no results page to see if a manual address has been selected
- modify the logic in our journeyDataService extensions/ helpers to check 

I think that means our journeys still reflect the path that the user takes, rather than pretending that they've visited the select address page, and doesn't require us to alter the journey framework itself. What do you think?

JasminConterioSW: That sounds sensible, I'll take a look!

JasminConterioSW:
I'm going to close this review and open a new one with a modified approach (this one was partly for handover purposes anyway...)

PR #330, opened 2025-04-07
Feat/prsd 1002 read csrf from apache upload

AEPR:
Raising a draft with something that technically works, still needs some fixing (e.g. the "_csrf" string literal needs to come from somewhere and the precedance shouldn't be "minimum + 2")

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/filters/MultipartFormDataFilter.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/filters/MultipartFormDataFilter.kt
@@ -79,21 +79,21 @@ class MultipartFormDataFilter(
         val csrfToken = deferredCsrfToken.get()
         return csrfToken.parameterName
     }
-}
 
-class CsrfProvidingRequestWrapper(
-    private val request: HttpServletRequest,
-    private val csrfToken: ParameterCsrfTokenDetails,
-) : HttpServletRequest by request {
-    override fun getParameter(p0: String?): String =
-        if (p0 == csrfToken.parameterName) {
-            csrfToken.token
-        } else {
-            request.getParameter(p0)
-        }
-}
+    class CsrfProvidingRequestWrapper(
+        private val request: HttpServletRequest,
+        private val csrfToken: ParameterCsrfTokenDetails,
+    ) : HttpServletRequest by request {
+        override fun getParameter(p0: String?): String =
+            if (p0 == csrfToken.parameterName) {
+                csrfToken.token
+            } else {
+                request.getParameter(p0)
+            }
+    }
 
-class ParameterCsrfTokenDetails(
-    val parameterName: String,
-    val token: String,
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/config/filters/MultipartFormDataFilter.kt
```diff
@@ -0,0 +1,97 @@
...
+            csrfToken.token
+        } else {
+            request.getParameter(p0)
+        }
+}
+
+class ParameterCsrfTokenDetails(
+    val parameterName: String,
+    val token: String,
+)
```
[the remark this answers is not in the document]
Travis-Softwire: Yes, I think nest them in the filter class - they're essentially private utility classes.

src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/ExampleFileUploadController.kt
```diff
@@ -47,7 +47,7 @@ class ExampleFileUploadController(
         }
 
         val file =
-            getFirstFileItem(request) ?: throw ResponseStatusException(HttpStatus.BAD_REQUEST, "Not a valid multipart file upload request")
+            getFirstFileItem(iterator) ?: throw ResponseStatusException(HttpStatus.BAD_REQUEST, "Not a valid multipart file upload request")
```
Travis-Softwire, line 50: Fine for the purposes of the example endpoint. Would be good to understand what this looks like in the case of 'valid' failure conditions, e.g. the user submits the form without staging a file.

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/filters/MultipartFormDataFilter.kt
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    private fun doFilterOnMultipartRequest(
+        multipartRequest: HttpServletRequest,
+        response: HttpServletResponse,
+        chain: FilterChain,
+    ) {
+        val upload = JakartaServletFileUpload()
+        val multipartItemIterator = upload.getItemIterator(multipartRequest)
+        multipartRequest.setAttribute("multipartItemIterator", multipartItemIterator)
```
Travis-Softwire, line 44: Minor, but should probably be a constant somewhere that we can reference from the controller.

Travis-Softwire (approved):
Nice :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/ExampleFileUploadController.kt
AEPR: That atually doesn't hit this point at all - it will very happily just create an empty file instead. I'm going to leave it as is for now, and address in error handling or dealing with metadata.

PR #344, opened 2025-04-09
Feat/prsd 1028 run migrations in the deployment pipeline

Kiran-R-K:
## Ticket number

PRSD-1028

## Goal of change

Switch to running DB migrations in deployment pipeline

Added a Run Database Migration job to the build and deploy pipeline, between Update Task Definition and Update ECS Service jobs

Set Update ECS Service to rely upon the new Run Database Migration job

Set Run Database Migration to:

- Checkout the code
- Connect to AWS
- Retrieve the SSM parameters and SecretsManager secret for the database and set the relevant flyway environment variables
- Setup port forwarding via the bastion host using SSM
- Run the flywayMigrate gradle task
- End SSM port forwarding

Updated the Reset Database workflow to include running migrations before restarting the ECS service

Updated Spring config so that migrations are not run when the app starts up except for when the app is running locally

A lot of the additions to `build-and-deploy.yml` are directly copied from `reset-database.yml` I was considering pulling them out into a reusable workflow but I also think having the steps clear and readable in each pipeline file is helpful - and there is a lot of duplication in different pipeline steps elsewhere so I wasn't sure of the value of it. I'd appreciate opinions.

## Checklist


.github/workflows/build-and-deploy.yml
```diff
@@ -113,9 +116,69 @@ jobs:
...
+      id-token: write
+      contents: read
+    env:
+      AWS_REGION: eu-west-2
+      IAM_ROLE: arn:aws:iam::${{ inputs.account-id }}:role/github-actions-terraform-admin
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.branch }}
```
Travis-Softwire, line 133: I might be wrong, but I'm not sure we need the branch here - our triggers for the job will only fire on `main` or `test`, and by default the checkout action will checkout the branch that triggered it

src/main/resources/application.yml
```diff
@@ -19,7 +19,7 @@ spring:
       defer-datasource-initialization: true
 
   flyway:
-    enabled: true
+    enabled: false
     locations: classpath:db/migrations
```
Travis-Softwire, line 23: I don't think we need the other properties here if we've set enabled to false?

Travis-Softwire (approved):
Couple of minor comments but otherwise looks good :)

.github/workflows/build-and-deploy.yml
```diff
@@ -113,9 +116,69 @@ jobs:
...
+      - name: Set DB password in ENV
+        uses: aws-actions/aws-secretsmanager-get-secrets@v2.0.8
+        with:
+          secret-ids: |
+            FLYWAY_PASSWORD, tf-${{ inputs.environment }}-prsdb-database-password
+
+      - name: Run gradle flywayMigrate
+        env:
+          FLYWAY_URL: jdbc:postgresql://localhost:5432/prsdb
+        run: ./gradlew flywayMigrate
```
isobel-softwire, line 173: We're using the --no-daemon flag when calling flywayClean - would we want it here too?

src/main/resources/application.yml
Kiran-R-K: I just manually tested that and agree

PR #356, opened 2025-04-15
PRSD-945: Certificate Upload Page

isobel-softwire:
## Ticket number

PRSD-945

## Goal of change

Creates certificate upload page

- Upgrades GDS so 'improved' file upload component can be used
- Modifies `questionPage` template to allow for file upload pages
- Creates certificate upload page template
- Tests template by using it in `ExampleFileUploadController`

- Certificate upload page template currently uses placeholder text. This will be updated when it's used in a journey step.
- `questionPage` uses an implicit parameter to enable file upload pages. This method of updating a fragment's parameters differs from how we've done it in the past (explicitly adding them to fragment signatures), but has the advantage of backwards compatibility.

## Checklist

image
image


src/main/resources/templates/forms/uploadCertificateForm.html
```diff
@@ -0,0 +1,16 @@
+<!--/*@thymesVar id="title" type="java.lang.String"*/-->
+<!--/*@thymesVar id="formModel" type="java.lang.Object"*/-->
+<!--/*@thymesVar id="backUrl" type="java.lang.String"*/-->
+<!--/*@thymesVar id="sectionHeaderInfo" type="uk.gov.communities.prsdb.webapp.models.viewModels.SectionHeaderViewModel"*/-->
+<!--/*@thymesVar id="fieldSetHeading" type="java.lang.String"*/-->
+<!--/*@thymesVar id="fieldSetHint" type="java.lang.String"*/-->
+<!--/*@thymesVar id="label" type="java.lang.String"*/-->
+<!DOCTYPE html>
+<html th:with="enctype='multipart/form-data'">
```
Travis-Softwire, line 9: Not sure I like relying on part of the template that's not rendered to pass in this value. I think you can instead:
- use named parameters in the call to `questionPage()`, which *magically* lets you pass in parameters that are not in the signature
- add `th:with="enctype=${enctype} ?: 'application/x-www-form-urlencoded'"` to the end of the `th:block` tag at the start of `questionPage` (making use of the default form encoding so we don't need to append an empty attribute)
- just use `th:attrappend="enctype=${enctype}"` in the `form` tag

Travis-Softwire (approved):
A couple of Qs/ thoughts, but happy with this once they're dealt with :)

AEPR (approved):
Looks good to me.

PR #357, opened 2025-04-16
PRSD-945: Refactors File Upload Functionality

isobel-softwire:
## Ticket number

PRSD-945

## Goal of change

Refactors example file upload functionality to prepare for use in compliance journey.

- Moves example file upload classes into appropriate folders
- Moves some of `ExampleFileUploadController`'s functionality into services/helper extensions (and adds tests for them)
- Updates `ExampleFileUploadController` based on changes

N/A

## Checklist


ReadMe.md
```diff
@@ -148,3 +148,42 @@ logging in.
...
+```
+
+Then run the service as usual, it will pick up the profile provided by the `aws-vault exec` command.
+
+When you have finished running the service, run `exit` in the server terminal to close the server.
+
+#### Connecting to AWS S3 locally
+By default, when the service is run locally, it uses the `LocalFileUploader` instead of the `AwsS3FileUploader`.
+You can manually switch by manipulating the profiles and attributes on those classes.
+Currently, there isn't a profile which connects to AWS with an otherwise local build.
```
isobel-softwire, line 189: This is taken from the ReadMe that has been deleted from the examples folder

src/main/kotlin/uk/gov/communities/prsdb/webapp/local/services/StubFileUploader.kt
```diff
@@ -1,15 +1,16 @@
...
 import org.springframework.stereotype.Service
+import uk.gov.communities.prsdb.webapp.services.FileUploader
 import java.io.File
 import java.io.InputStream
 
 @Service
 @Primary
 @Profile("local")
-class LocalFileUploader : FileUploader {
+class StubFileUploader : FileUploader {
```
Travis-Softwire, line 13: V. minor, but I'd argue this isn't a stub, it does allow uploading from the client filesystem to the webserver, and it then saves the file server-side. It just so happens that all 3 are on the same machine :)

Travis-Softwire (approved):
One minor comment - it all LGTM but suggest waiting for Alex to review as well as he's closer to it

src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/ExampleFileUploadController.kt
```diff
@@ -55,12 +57,21 @@ class ExampleFileUploadController(
         principal: Principal,
         redirectAttrs: RedirectAttributes,
     ): String {
-        if (!tokenService.checkTokenIsFor(token, freeSegment)) {
+        if (!cookieService.isTokenForCookieValue(token, streamlessRequest.requestURI)) {
             throw ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid upload token")
+        } else {
+            cookieService.useToken(token)
         }
```
AEPR, line 64: I think I'd flip this if statement now that there's an action when the token is valid. I think it's clearer to say "If the token's valid, use it, otherwise reject the request" than to say "If the token's invalid, reject the request, otherwise use the token."

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/CookieService.kt
```diff
@@ -0,0 +1,54 @@
+package uk.gov.communities.prsdb.webapp.services
+
+import jakarta.servlet.http.Cookie
+import jakarta.servlet.http.HttpSession
+import org.springframework.stereotype.Service
+import uk.gov.communities.prsdb.webapp.constants.COOKIE_TOKENS
+import java.util.UUID
+
+@Service
+class CookieService(
```
AEPR, line 10: I think I would call this something with a bit more detail than `CookieService` - it specifically deals in a kind of cookie where we store a value in the session which can be retrieved using the token issued in the cookie. Maybe a `TokenCookieService`? Or a `SessionTokenCookieService`? IDK, open to something else as well.

AEPR (approved):
One little nitpick. This all looks good to me.

PR #359, opened 2025-04-17
Feat/prsd 1079 allow landlords to resume incomplete property registrations

Kiran-R-K:
## Ticket number

PRSD-1079 (part 2/2)

## Goal of change

Allow landlords to resume incomplete property registrations

Created a ghost resume page for the Property Registration Journey that takes a conextId as a parameter, uses that to get and load the journeyData from the database and then redirects the user to the task list in the registration flow so that they can continue the property registration.

## Checklist


Travis-Softwire (commented):
Happy with this approach overall, but I've left a few comments on the approach in some of the tests, happy to chat through them when we're all back :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -43,7 +43,7 @@ abstract class Journey<T : StepId>(
                 context to an overload of journeyDataService.loadJourneyDataIntoSession().*/
```
isobel-softwire, line 43: Might be worth updating this comment to reflect the change in `loadJourneyDataIntoSession`'s name

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataService.kt
```diff
@@ -28,6 +30,11 @@ class JourneyDataService(
         session.setAttribute(journeyDataKey, null)
     }
 
+    fun removeJourneyDataAndContextIdFromSession() {
+        session.removeAttribute(CONTEXT_ID)
+        session.removeAttribute(journeyDataKey)
+    }
```
isobel-softwire, line 36: Should we be using this instead of `clearJourneyDataFromSession`? Are there cases where we'd want to clear the journey data but keep the context id in session?

isobel-softwire: Also, `deleteJourneyData()` could be refactored to use this method

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataService.kt
```diff
@@ -67,15 +74,32 @@ class JourneyDataService(
...
+        contextId: Long,
+        baseUserId: String,
+        journeyType: JourneyType,
+    ) {
+        val formContext =
+            formContextRepository.findByIdAndUser_IdAndJourneyType(contextId, baseUserId, journeyType) ?: throw ResponseStatusException(
+                HttpStatus.NOT_FOUND,
+                "Form context with ID: $contextId and journey type: ${journeyType.name} not found for base user: $baseUserId",
+            )
+        loadJourneyDataIntoSession(formContext)
```
isobel-softwire, line 102: Minor but I think these method names should align as they're all trying to do the same thing. `loadJourneyDataIntoSession` could be renamed to `loadJourneyDataFromFormContext` or all the methods could be called `loadJourneyDataIntoSession`

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataServiceTests.kt
```diff
@@ -223,32 +225,94 @@ class JourneyDataServiceTests {
...
+            // Act
+            spiedOnJourneyDataService.loadJourneyDataFromFormContext(formContext.id, formContext.user.id, formContext.journeyType)
+
+            // Assert
+            verify(mockFormContextRepository).findByIdAndUser_IdAndJourneyType(formContext.id, formContext.user.id, formContext.journeyType)
+            verify(spiedOnJourneyDataService).loadJourneyDataIntoSession(formContext)
+        }
+
+        @Test
+        fun `throws an response status exception NOT_FOUND if form context is missing`() {
```
isobel-softwire, line 293: Super minor but this should be "throws **a** response..."

isobel-softwire (commented):
LGTM :) - left a few small suggestions and a question about clearing journey data/context IDs

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataServiceTests.kt
```diff
@@ -223,32 +225,94 @@ class JourneyDataServiceTests {
...
-            journeyDataService.loadJourneyDataIntoSession(contextId)
+            journeyDataService.loadJourneyDataIntoSession(formContext)
             val formContextCaptor = captor<JourneyData>()
             verify(mockHttpSession).setAttribute(eq(journeyDataKey), formContextCaptor.capture())
             val contextIdCaptor = captor<Long>()
             verify(mockHttpSession).setAttribute(eq(CONTEXT_ID), contextIdCaptor.capture())
 
             // Assert
-            assertEquals(journeyData, formContextCaptor.value)
-            assertEquals(contextId, contextIdCaptor.value)
```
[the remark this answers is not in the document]
Kiran-R-K: That was a mistake - I've added it back in

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/JourneyDataServiceTests.kt
```diff
@@ -223,32 +225,94 @@ class JourneyDataServiceTests {
...
+            val spiedOnJourneyDataService = spy(journeyDataService)
+            val formContext = FormContext(JourneyType.PROPERTY_REGISTRATION, "{\"testPage\":\"noData\"}", OneLoginUser())
+            whenever(mockFormContextRepository.findById(formContext.id)).thenReturn(Optional.ofNullable(formContext))
+
+            // Act
+            spiedOnJourneyDataService.loadJourneyDataFromFormContext(formContext.id)
+
+            // Assert
+            verify(mockFormContextRepository).findById(formContext.id)
+            verify(spiedOnJourneyDataService).loadJourneyDataIntoSession(formContext)
```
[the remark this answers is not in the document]
Kiran-R-K: I could, the reason I didn't was because I was thinking of this as a strict unit test that should only test the behaviour of this method, which is to call `loadJourneyDataIntoSession`.  But happy to change it to checking the http session instead?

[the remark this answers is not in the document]
Travis-Softwire: It's a tricky balance to ensure that the scope of your tests are narrow enough to isolate one bit of functionality vs just testing the specific implementation. In this case, I think the key behaviour to test is that the data does in fact get loaded into the session - I don't think we need to test that it does that via j`ourneyDataService.loadJourneyDataIntoSession`.

isobel-softwire (approved):
LGTM!

PR #364, opened 2025-04-23
PRSD-945: Property Compliance - Gas Safety Upload Page

isobel-softwire:
## Ticket number

PRSD-945

## Goal of change

Adds gas safety certificate upload step to property compliance journey

- Creates form model for validating staged/uploaded files
- Creates helper method for naming gas safety certificate files
- Adds gas safety certificate upload step to property compliance journey
- Modifies existing property compliance controller journey step endpoints and creates new POST endpoint to support file uploads

- Gas safety certificate filename format is `property_<property_ownership_id>_gas_safety_certificate.<file_extension>`

## Checklist

- Controller and integration tests will make up a separate PR 

image
image
image


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyComplianceController.kt
```diff
@@ -96,10 +167,28 @@ class PropertyComplianceController(
         }
     }
 
+    private fun uploadFileIfValid(
+        file: FileItemInput,
+        fileLength: Long,
+        propertyOwnershipId: Long,
+    ): Boolean? {
```
Travis-Softwire, line 174: I don't love us passing around a `Boolean?` as a basically a trinary boolean - I think I'd prefer we keep the two checks separate, and have a flag that denotes whether the file was valid, and another that denotes whether it was uploaded successfully, with a little bit of additional logic in the form model to handle the fact that we don't want the 'upload unsuccessful' error to show if there are file validation errors.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/UploadCertificateFormModel.kt
```diff
@@ -0,0 +1,71 @@
...
+    )
+    val certificate = null
+
+    var name: String = ""
+
+    var contentType: String = ""
+
+    var contentLength: Long = 0
+
+    var isUploadSuccessful: Boolean? = null
```
Travis-Softwire, line 45: See other comments about making this `Boolean` rather than `Boolean?`, but if that's really not possible then I think it should at least be `isUploadSuccessfulOrNull` to make the explicit comparisons with `true` and `false` slightly easier to make sense of.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/UploadCertificateFormModel.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    fun isNameNotBlank() = name.isNotBlank()
+
+    fun isFileTypeValid() = !isNameNotBlank() || FilenameUtils.getExtension(name) in validExtensions && contentType in validMimeTypes
+
+    fun isContentLengthValid() = !isNameNotBlank() || !isFileTypeValid() || contentLength <= maxContentLength
+
+    fun isUploadNotUnsuccessful() = isUploadSuccessful != false
```
Travis-Softwire, line 53: I think this illustrates why `Boolean?`s can get confusing, as it means that you can be NotUnsuccessful without being Successful.... My suggestion would be that this becomes `isUploadSuccessfulOrInvalid` which I think is what we're going for.

Travis-Softwire (approved):
Broadly happy with this - some suggestions about using avoiding needing to use a nullable boolean to pass around a trinary value, but otherwise LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/UploadCertificateFormModel.kt
isobel-softwire: I agree that making it a `Boolean` would make things clearer. The model needs to be aware of whether an upload has been attempted so I've gone with `isUploadSuccessfulOrNull`, which (to me) feels like the cleanest way of doing this.

PR #366, opened 2025-04-24
Fix/prsd 895 handle no address found part 1

JasminConterioSW:
## Ticket number

PRSD-895

## Goal of change

When looking up addresses, direct to a "No address found" page if there are no results.
This PR does not complete the ticket.

- Adds a template for the "No address found" page and include it as a step on the Property Registration journey (using hardcoded values for now)
- Perform address lookup when the address lookup page is submitted and use the result to determined whether to proceed to the "Select address" or "No address found" step.
    - This is done on a new LookupAddressStep which inherits from Step as we will want to re-use it in the Landlord Registration and Update Landlord journey
    - 
image

An alternative approach might be to make an AbstractStep which Step and LookupAddressStep can both inherit from (there is a branch which does this [here](https://github.com/communitiesuk/prsdb-webapp/tree/fix/PRSD-895-create-abstract-step-and-lookup-step)). I've gone with inheriting from Step as Step contains the default behaviour.

### Still to do
At least one more PR will be needed to complete the ticket, to include:
- Update the No Address Found pages to take the looked up postcode and house number
- On SelectAddressPage, get the looked up addresses from journeyData instead of doing the lookup here
- Update journeyData extensions/helpers checking whether a manual address is being used
- Update Landlord registration and update landlord journeys to handle No Address found
- Tests

## Checklist

(These are relevant but will be done in a later PR rather than here)


src/main/resources/templates/noAddressFoundPage.html
```diff
@@ -0,0 +1,32 @@
...
+<!--/*@thymesVar id="formModel" type="java.lang.Object"*/-->
+<!--/*@thymesVar id="sectionHeaderInfo" type="java.lang.Object"*/-->
+<!--/*@thymesVar id="postcode" type="java.lang.String"*/-->
+<!--/*@thymesVar id="houseNameOrNumber" type="java.lang.String"*/-->
+<!--/*@thymesVar id="searchAgainUrl" type="java.lang.String"*/-->
+<!DOCTYPE html>
+<th:block id="main-content"
+          th:replace="~{fragments/layout :: layout(#{${title}}, ~{::#main-content/content()}, false)}">
+    <a href="#" th:replace="~{fragments/forms/backLink :: backLink(${backUrl})}">Back
+        link</a>
```
isobel-softwire, line 11: Minor, but could these elements be single lines?

src/main/resources/templates/noAddressFoundPage.html
```diff
@@ -0,0 +1,32 @@
+<!--/*@thymesVar id="title" type="java.lang.String"*/-->
+<!--/*@thymesVar id="formModel" type="java.lang.Object"*/-->
+<!--/*@thymesVar id="sectionHeaderInfo" type="java.lang.Object"*/-->
+<!--/*@thymesVar id="postcode" type="java.lang.String"*/-->
+<!--/*@thymesVar id="houseNameOrNumber" type="java.lang.String"*/-->
+<!--/*@thymesVar id="searchAgainUrl" type="java.lang.String"*/-->
+<!DOCTYPE html>
+<th:block id="main-content"
+          th:replace="~{fragments/layout :: layout(#{${title}}, ~{::#main-content/content()}, false)}">
```
isobel-softwire, line 9: Although it's not really a form page, we've been using the `questionPage` template for content pages that are part of journeys like this one (e.g. `gasSafetyExemptionConfirmationForm`).

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -103,7 +102,7 @@ abstract class Journey<T : StepId>(
         }
 
         if (currentStep.handleSubmitAndRedirect != null) {
-            return ModelAndView("redirect:${currentStep.handleSubmitAndRedirect.invoke(newJourneyData, subPageNumber)}")
+            return ModelAndView("redirect:${currentStep.handleSubmitAndRedirect!!.invoke(newJourneyData, subPageNumber)}")
```
isobel-softwire, line 105: Is this change needed?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
```diff
@@ -549,7 +575,9 @@ class PropertyRegistrationJourney(
         journeyData: JourneyData,
         propertyRegistrationService: PropertyRegistrationService,
     ): Pair<RegisterPropertyStepId, Int?> =
-        if (PropertyRegistrationJourneyDataHelper.isManualAddressChosen(journeyData)) {
+        if (journeyData.getLookedUpAddresses().isEmpty()) {
+            Pair(RegisterPropertyStepId.NoAddressFound, null)
+        } else if (PropertyRegistrationJourneyDataHelper.isManualAddressChosen(journeyData)) {
```
isobel-softwire, line 580: Can the select address page be reached if there are no looked up addresses?

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/StepUrlBuilder.kt
```diff
@@ -0,0 +1,20 @@
+package uk.gov.communities.prsdb.webapp.helpers
+
+import org.springframework.web.util.UriComponentsBuilder
+import uk.gov.communities.prsdb.webapp.forms.steps.StepId
+import java.util.Optional
+
+class StepUrlBuilder {
+    companion object {
+        fun <T : StepId> getStepUrl(
+            stepId: T,
```
isobel-softwire, line 10: Could `stepId`'s type be `StepId` instead of a generic?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/steps/LookupAddressStep.kt
```diff
@@ -0,0 +1,60 @@
...
+    private val journeyDataService: JourneyDataService,
+) : Step<T>(id, page, isSatisfied, saveAfterSubmit) {
+    override var nextAction: (JourneyData, Int?) -> Pair<T?, Int?> =
+        { _: JourneyData, subPageNumber: Int? -> Pair(getNextStep(), subPageNumber) }
+
+    override var handleSubmitAndRedirect: ((JourneyData, Int?) -> String)? = { journeyData: JourneyData, subPageNumber: Int? ->
+        performAddressLookupCacheResultsAndGetRedirect(
+            journeyData,
+            subPageNumber,
+        )
```
isobel-softwire, line 30: This method name is a bit long - maybe it should be split into two methods? E.g. `addLookedUpAddressesToJourneyData` and `getRedirect` or something along those lines

isobel-softwire (commented):
Looking good so far - left a few thoughts :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
JasminConterioSW: No, but the manual address page can be reached

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/StepUrlBuilder.kt
```diff
@@ -0,0 +1,20 @@
+package uk.gov.communities.prsdb.webapp.helpers
+
+import org.springframework.web.util.UriComponentsBuilder
+import uk.gov.communities.prsdb.webapp.forms.steps.StepId
+import java.util.Optional
+
+class StepUrlBuilder {
```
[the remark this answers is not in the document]
JasminConterioSW: Ah yes, that's a better place for it

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
isobel-softwire: In that case could `journeyData.getLookedUpAddresses().isEmpty()` ever be true at this point?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/steps/Step.kt
```diff
@@ -7,19 +7,38 @@ import uk.gov.communities.prsdb.webapp.forms.objectToStringKeyedMap
...
-class Step<T : StepId>(
+open class Step<T : StepId>(
     val id: T,
     val page: AbstractPage,
-    val handleSubmitAndRedirect: ((journeyData: JourneyData, subPageNumber: Int?) -> String)? = null,
     val isSatisfied: (bindingResult: BindingResult) -> Boolean = { bindingResult -> page.isSatisfied(bindingResult) },
-    val nextAction: (journeyData: JourneyData, subPageNumber: Int?) -> Pair<T?, Int?> = { _, _ ->
+    val saveAfterSubmit: Boolean = true,
+) {
+    constructor(
```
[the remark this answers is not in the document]
JasminConterioSW: Hmm, actually I don't need two constructors on Step because handleSubmitAndRedirect and nextAction both have defaults...

I'm not sure about passing functions from LookupAddressStep into the parent constructor though, maybe we can have a look when you are free? I might be missing something!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
JasminConterioSW: Ah true, this is left over from a previous iteration I think!

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
JasminConterioSW: At the moment yes, because handleSubmitAndRedirect is mutable and nullable.
If we can pass it into Step's constructor from LookupStep and make it a val, we might not need this...

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/steps/LookupAddressStep.kt
JasminConterioSW: I think it should stay as one function, as this is the handleSubmitAndRedirect for this step.
It could maybe be renamed to something shorter but that might be less useful! (E.g. it could just be `lookupSubmitAndRedirect`, similar to the `checkAnswersSubmitAndRedirect` in the PropertyRegistrationJourney )

isobel-softwire: That makes sense :)

PR #370, opened 2025-04-24
Feat/prsd 1065 integrate with epc register

AEPR:
## Ticket number

PRSD-1065

## Goal of change

Proof of concept for integrating with the EPC register API.

Adds an example controller to retrieve EPC certificate JSON. Includes adding OAuth client to connect to the register. 

Nothing in particular

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/EpcRegisterClient.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/EpcRegisterClient.kt
@@ -1,0 +1,39 @@
+package uk.gov.communities.prsdb.webapp.clients
+
+import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.stereotype.Service
+import org.springframework.web.client.RestClient
+import org.springframework.web.client.body
+import org.springframework.web.util.UriComponentsBuilder
+import java.net.URLEncoder
+
+@Service
+class EpcRegisterClient(
+    @Qualifier("epc-client") private val client: RestClient,
+) {
+    fun getByUprn(uprn: Int): String {
+        val uprnString = uprn.toString().padStart(12, '0')
+        val query = URLEncoder.encode(uprnString, "UTF-8")
+        return searchFor("uprn", "UPRN-$query")!!
+    }
+
+    fun getByRrn(rrn: String): String {
+        val query = URLEncoder.encode(rrn, "UTF-8")
+        return searchFor("rrn", query)!!
+    }
+
+    private fun searchFor(
+        parameterName: String,
+        parameterValue: String,
+    ): String? =
+        client
+            .get()
+            .uri(
+                UriComponentsBuilder
+                    .fromUriString("/api/prsdatabase/assessments/search")
+                    .queryParam(parameterName, parameterValue)
+                    .build()
+                    .toUri(),
+            ).retrieve()
+            .body<String>()
+}
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/EpcRegisterClient.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/EpcRegisterClient.kt
@@ -1,10 +1,13 @@
 package uk.gov.communities.prsdb.webapp.clients
 
+import org.json.JSONObject
 import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.http.HttpStatus
 import org.springframework.stereotype.Service
 import org.springframework.web.client.RestClient
 import org.springframework.web.client.body
 import org.springframework.web.util.UriComponentsBuilder
+import uk.gov.communities.prsdb.webapp.exceptions.PrsdbWebException
 import java.net.URLEncoder
 
 @Service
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/EpcRegisterClient.kt
```diff
@@ -0,0 +1,72 @@
...
+                    }
+                }.body<String>()
+
+        return bodyString ?: errorString
+    }
+
+    private fun getErrorCode(bodyString: String): String? {
+        val jsonObject = JSONObject(bodyString)
+        val errors = jsonObject.getJSONArray("errors")
+        return if (errors.count() == 1) {
```
Travis-Softwire, line 66: Will they only ever return us 1 error?

src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/EpcRegisterClient.kt
```diff
@@ -0,0 +1,72 @@
...
+                    val bodyString = response.body.reader().readText()
+                    val code = getErrorCode(bodyString)
+                    if (code == "INVALID_REQUEST" || code == "BAD_REQUEST") {
+                        errorString = bodyString
+                    } else {
+                        throw PrsdbWebException("Example fallback error: $bodyString")
+                    }
+                }.body<String>()
+
+        return bodyString ?: errorString
```
Travis-Softwire, line 60: Are there real cases where we'd want to return the underlying error or is this just for the purposes of the example?

src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/EpcRegisterClient.kt
```diff
@@ -0,0 +1,72 @@
...
+                    UriComponentsBuilder
+                        .fromUriString("/api/prsdatabase/assessments/search")
+                        .queryParam(parameterName, parameterValue)
+                        .build()
+                        .toUri(),
+                ).retrieve()
+                // This ensures that we naively treat NOT_FOUND responses as valid, for searches that do not
+                // match a property/certificate. This silently ignores the error and the response body is treated
+                // like a successful response.
+                .onStatus({ it == HttpStatus.NOT_FOUND }) { request, response -> }
```
Travis-Softwire, line 46: What does the body look like when it returns a 404?

Travis-Softwire (approved):
Happy with this as  PoC - enjoyed how succinct it all is :) 

I've left a few Qs but they're more about how the API works and how we'll handle responses in reality than they are about the PoC

src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/ExampleEpcController.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    @GetMapping("uprn/{uprn}")
+    @ResponseBody
+    fun getEpcByUprn(
+        @PathVariable uprn: Int,
+    ): String = client.getByUprn(uprn)
+
+    @GetMapping("rrn/{rrn}")
+    @ResponseBody
+    fun getEpcByUprn(
```
isobel-softwire, line 27: should this be `getEpcByRrn`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/EpcRegisterClient.kt
AEPR: The same as the other error case but with a different error code.

src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/EpcRegisterClient.kt
AEPR: This is fundementally just for the purposes of the example - in real life we probably wouldn't want to return the body string either. This is basically just showing how to extract info out of the `onStatus` without throwing an exception.

src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/EpcRegisterClient.kt
AEPR: They return an array - I've only ever seen one error but the only cases I've tried are:
 - UPRN/RRN not found
 - Not a valid format for UPRN/RRN
 - UPRN for a property that does not have a certificate of the correct type

src/main/kotlin/uk/gov/communities/prsdb/webapp/examples/ExampleEpcController.kt
AEPR: Yep, absolutely.

src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/EpcRegisterClient.kt
Travis-Softwire: Fair enough - doesn't matter for the example, but think we'll want to check for `>= 1` errors rather than `== 1` just to be on the safe side

PR #384, opened 2025-04-30
PRSD-1099: Split change for properties POC

AEPR:
## Ticket number
PRSD-1099

## Goal of change
Proof of concept for allowing separate details updates, specifically for the property details update journey.

Uses the Journey Factory and Journey Data Service Factory to set a different data key for each "sub-journey". This ensures these can be modified in parallel without untoward effects. I've added check your answer pages for licensing and occupancy to make those journey sections work.

The magic switch statement in `PropertyDetailsUpdateJourneyFactory` which is the heart of this whole scheme. I imagine we'd probably want a comment explaining why we're using it, but for now I just want to get your thoughts!

The tests will be failing, specifically the `PropertyDetailsUpdateJourneyTests`, as I don't want to change the testing behaviour unless you're happy with the functionality.

## Checklist
As noted, several of these are outstanding.

{89174AAA-4B1E-4B80-AC8A-6C057B01AC0E}

{45018735-1D00-4CEA-A310-41ECE2356B67}


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
```diff
@@ -51,13 +51,13 @@ class PropertyDetailsController(
...
+    ): String {
         addPropertyDetailsToModelIfAuthorizedUser(model, principal, propertyOwnershipId, withPropertyChangeLinks = true)
         // TODO: PRSD-355 Remove this way of showing submit button
         model.addAttribute("shouldShowSubmitButton", true)
-        return propertyDetailsUpdateJourneyFactory
-            .create(propertyOwnershipId)
-            .getModelAndViewForStep(UpdatePropertyDetailsStepId.UpdateDetails.urlPathSegment, subPageNumber = null)
+        model.addAttribute(BACK_URL_ATTR_NAME, PROPERTY_DETAILS_ROUTE)
+
+        return "propertyDetailsView"
```
isobel-softwire, line 60: Do we still need this endpoint?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
```diff
@@ -70,6 +69,8 @@ class PropertyDetailsUpdateJourney(
                 UpdatePropertyDetailsStepId.UpdateNumberOfHouseholds toPageData NumberOfHouseholdsFormModel::fromPropertyOwnership,
                 UpdatePropertyDetailsStepId.UpdateNumberOfPeople toPageData NumberOfPeopleFormModel::fromPropertyOwnership,
                 UpdatePropertyDetailsStepId.UpdateLicensingType toPageData LicensingTypeFormModel::fromPropertyOwnership,
+                UpdatePropertyDetailsStepId.CheckYourLicensing.urlPathSegment to mapOf<String, Any>() as PageData,
+                UpdatePropertyDetailsStepId.CheckYourOccupancy.urlPathSegment to mapOf<String, Any>() as PageData,
```
isobel-softwire, line 73: I appreciate that this PR is just a POC, but this would read better if it was after the update number of people step

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/factories/PropertyDetailsUpdateJourneyFactory.kt
```diff
@@ -13,11 +15,41 @@ class PropertyDetailsUpdateJourneyFactory(
...
+            ) + "-LICENSING"
+        UpdatePropertyDetailsStepId.UpdateOccupancy.urlPathSegment,
+        UpdatePropertyDetailsStepId.UpdateNumberOfPeople.urlPathSegment,
+        UpdatePropertyDetailsStepId.UpdateNumberOfHouseholds.urlPathSegment,
+        UpdatePropertyDetailsStepId.CheckYourOccupancy.urlPathSegment,
+        ->
+            PropertyDetailsController.getUpdatePropertyDetailsPath(
+                propertyOwnershipId,
+            ) + "-OCCUPANCY"
+        else -> throw PrsdbWebException("Invalid step name $stepName")
```
isobel-softwire, line 53: I think it would helpful if the error message mentioned that the step name's invalid for the property update journey

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/factories/PropertyDetailsUpdateJourneyFactory.kt
```diff
@@ -13,11 +15,41 @@ class PropertyDetailsUpdateJourneyFactory(
...
+        propertyOwnershipId: Long,
+        stepName: String,
+    ) = PropertyDetailsUpdateJourney(
+        validator,
+        journeyDataServiceFactory.create(getJourneyDataKey(propertyOwnershipId, stepName)),
+        propertyOwnershipService,
+        propertyOwnershipId,
+    )
+
+    fun getJourneyDataKey(
```
isobel-softwire, line 28: Minor but could this be private/static?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/factories/PropertyDetailsUpdateJourneyFactory.kt
```diff
@@ -13,11 +15,41 @@ class PropertyDetailsUpdateJourneyFactory(
...
+                propertyOwnershipId,
+            ) + "-LICENSING"
+        UpdatePropertyDetailsStepId.UpdateOccupancy.urlPathSegment,
+        UpdatePropertyDetailsStepId.UpdateNumberOfPeople.urlPathSegment,
+        UpdatePropertyDetailsStepId.UpdateNumberOfHouseholds.urlPathSegment,
+        UpdatePropertyDetailsStepId.CheckYourOccupancy.urlPathSegment,
+        ->
+            PropertyDetailsController.getUpdatePropertyDetailsPath(
+                propertyOwnershipId,
+            ) + "-OCCUPANCY"
```
isobel-softwire, line 52: This is a little hard to read - maybe group the related steps into lists before the `when`

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/CheckLicensingPage.kt
```diff
@@ -0,0 +1,52 @@
...
+class CheckLicensingPage :
+    AbstractPage(
+        NoInputFormModel::class,
+        "forms/checkAnswersForm",
+        mapOf(
+            "title" to "registerProperty.title",
+            "submitButtonText" to "forms.buttons.confirm",
+            "summaryName" to "registerLaUser.checkAnswers.summaryName",
+        ),
+        shouldDisplaySectionHeader = false,
```
isobel-softwire, line 20: Minor but I don't think this line is needed? Same goes for the `CheckOccupancyPage`

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/CheckLicensingPage.kt
```diff
@@ -0,0 +1,52 @@
...
+        val licensingType = PropertyRegistrationJourneyDataHelper.getLicensingType(filteredJourneyData)!!
+        val licenceNumber = PropertyRegistrationJourneyDataHelper.getLicenseNumber(filteredJourneyData)!!
+        val licensingSummaryValue = getLicensingSummaryValue(licenceNumber, licensingType)
+
+        val formData =
+            listOf(
+                SummaryListRowViewModel(
+                    "forms.checkPropertyAnswers.propertyDetails.licensing",
+                    licensingSummaryValue,
+                    RegisterPropertyStepId.LicensingType.urlPathSegment,
```
isobel-softwire, line 36: Should this be `UpdatePropertyDetailsStepId...` (same for the step IDs in `CheckOccupancyPage`)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/steps/UpdatePropertyDetailsStepId.kt
```diff
@@ -13,5 +11,6 @@ enum class UpdatePropertyDetailsStepId(
     UpdateSelectiveLicence("selective-licence"),
     UpdateHmoMandatoryLicence("hmo-mandatory-licence"),
     UpdateHmoAdditionalLicence("hmo-additional-licence"),
-    UpdateDetails(DETAILS_PATH_SEGMENT),
+    CheckYourLicensing("check-licensing"),
+    CheckYourOccupancy("check-occupancy"),
```
isobel-softwire, line 15: Same point here about putting `CheckYourOccupancy` after the other occupancy steps

isobel-softwire (commented):
This all LGTM! Left a few (mostly minor) thoughts, but they're probably only worth acting on if this is the method we end up going for.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
Travis-Softwire: I agree we won't need it once this is all implemented, but imagine it's still here to avoid updating the property record page as part of the POC

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/factories/PropertyDetailsUpdateJourneyFactory.kt
```diff
@@ -13,11 +15,41 @@ class PropertyDetailsUpdateJourneyFactory(
...
+        journeyDataServiceFactory.create(getJourneyDataKey(propertyOwnershipId, stepName)),
+        propertyOwnershipService,
+        propertyOwnershipId,
+    )
+
+    fun getJourneyDataKey(
+        propertyOwnershipId: Long,
+        stepName: String,
+    ) = when (stepName) {
+        UpdatePropertyDetailsStepId.UpdateOwnershipType.urlPathSegment ->
```
Travis-Softwire, line 32: It feels a bit gross having the factory need to know about how the steps are grouped. Would it be crazy to instead have something like:
- a new interface that inherits from `StepId`, e.g. `GroupedStepId`
- add a 'group name' property to it
- have `UpdatePropertyDetailsStepId` inherit from `GroupedStepId`
- a sensible way for the factory to get at the GroupedStepId for the journey based on the step name
That way every step we can enforce at build time that every step has a group rather than relying on catching it in the exception, the factory doesn't need to know what steps we have, and we don't need the switch statement :) 

But happy to chat it through - there may well be problems I've not thought of with that approach

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/factories/PropertyDetailsUpdateJourneyFactory.kt
```diff
@@ -13,11 +15,41 @@ class PropertyDetailsUpdateJourneyFactory(
...
-    fun create(propertyOwnershipId: Long) =
-        PropertyDetailsUpdateJourney(
-            validator,
-            journeyDataServiceFactory.create(PropertyDetailsController.getUpdatePropertyDetailsPath(propertyOwnershipId)),
-            propertyOwnershipService,
-            propertyOwnershipId,
-        )
+    fun create(
+        propertyOwnershipId: Long,
+        stepName: String,
```
Travis-Softwire, line 20: I've not thought this all the way through, but if we're defining step name here can we, for these journeys, make it a class variable and provide it when we construct the journey in the factory (with some wrappers around the functions that take it as a parameter)? Basically just to avoid having to pass `stepName` into both the factory and the journey's functions when we need them to be the same.

Travis-Softwire (commented):
I've left some thoughts about a slightly different way to implement this approach, but v. happy to discuss it!

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
AEPR: Yep, what travis said.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateLandlordDetailsJourney.kt
```diff
@@ -48,8 +48,8 @@ class UpdateLandlordDetailsJourney(
         initialStepId = UpdateLandlordDetailsStepId.UpdateEmail,
         validator = validator,
         journeyDataService = journeyDataService,
-        updateStepId = UpdateLandlordDetailsStepId.UpdateDetails,
-        updateEntityId = landlordBaseUserId,
+        // TODO PRSD-XXX Use the actual step name
```
Travis-Softwire, line 51: Is this XXX because we haven't got a ticket for this yet?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/CheckOccupancyPage.kt
```diff
@@ -0,0 +1,63 @@
...
+        filteredJourneyData: JourneyData?,
+    ) {
+        filteredJourneyData!!
+        val formData = getTenancyDetails(filteredJourneyData)
+
+        modelAndView.addObject("formData", formData)
+    }
+
+    private fun getTenancyDetails(journeyData: JourneyData): List<SummaryListRowViewModel> {
+        val occupied = PropertyRegistrationJourneyDataHelper.getIsOccupied(journeyData)!!
```
Travis-Softwire, line 31: Are there extension methods for extracting these bits of info from the journeyData? Or are they not appropriate here?

Travis-Softwire (approved):
Couple of minor Qs but otherwise LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/CheckOccupancyPage.kt
AEPR: Actually, both of these have been directly copy and pasted from the other CYA pages as I don't think I should spend time on the operation of them when there are specific tickets coming up! I'll add todos noting that this shouldn't let left as is.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateLandlordDetailsJourney.kt
AEPR: I think just that I didn't know it! PRSD-1100 or PRSD-1101 will do the job.

PR #390, opened 2025-04-30
Chore/prsd none restart local redis

Travis-Softwire:
Ensures that containers are restarted when webapp is run locally, including Redis

### DIFF
--- a/.run/local_resources_up.run.xml
+++ b/.run/local_resources_up.run.xml
@@ -1,11 +1,13 @@
 <component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="local_resources_up" type="docker-deploy" factoryName="docker-compose.yml" server-name="Docker">
-    <deployment type="docker-compose.yml">
-      <settings>
-        <option name="envFilePath" value="" />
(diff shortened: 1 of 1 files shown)


.run/local_resources_up.run.xml
```diff
@@ -1,11 +1,13 @@
 <component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="local_resources_up" type="docker-deploy" factoryName="docker-compose.yml" server-name="Docker">
+  <configuration default="false" name="local_resources_up" type="docker-deploy" factoryName="docker-compose.yml"
+                 server-name="Docker">
```
AEPR, line 3: Have you changed your max line length or something?

AEPR (approved):
Looks good to me - there seems to be a line length reformat that is fine to merge but might imply different settings in a linter.

.run/local_resources_up.run.xml
Travis-Softwire: No, this was because I edited the file rather than just letting Intellij sort it for me (Intellij was having a moment and refusing to write changes to the project run configs). Editing the file meant that our formatting rules were applied to it for the first time.

PR #393, opened 2025-05-01
PRSD-NONE: Clean Up Deprecations

isobel-softwire:
## Ticket number

PRSD-NONE

## Goal of change

Cleans up deprecations caused by upgrading Springboot

- Replaces deprecated classes 
- Makes any necessary changes caused by changed dependencies

N/A

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/OneLoginConfig.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/OneLoginConfig.kt
@@ -63,7 +63,7 @@ class OneLoginConfig {
     @Bean
     fun oneloginAuthorizationCodeTokenResponseClient() =
         RestClientAuthorizationCodeTokenResponseClient().apply {
-            setParametersConverter(NimbusJwtClientAuthenticationParametersConverter(::oneloginJWKResolver))
+            addParametersConverter(NimbusJwtClientAuthenticationParametersConverter(::oneloginJWKResolver))
         }
 
     @Bean
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/OneLoginConfig.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/OneLoginConfig.kt
@@ -13,9 +13,8 @@ import org.json.JSONObject
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
-import org.springframework.security.oauth2.client.endpoint.DefaultAuthorizationCodeTokenResponseClient
 import org.springframework.security.oauth2.client.endpoint.NimbusJwtClientAuthenticationParametersConverter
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/config/OneLoginConfig.kt
```diff
@@ -62,17 +61,10 @@ class OneLoginConfig {
...
-        )
-
-        val tokenResponseClient = DefaultAuthorizationCodeTokenResponseClient()
-        tokenResponseClient.setRequestEntityConverter(requestEntityConverter)
-
-        return tokenResponseClient
-    }
+    fun oneloginAuthorizationCodeTokenResponseClient() =
+        RestClientAuthorizationCodeTokenResponseClient().apply {
+            setParametersConverter(NimbusJwtClientAuthenticationParametersConverter(::oneloginJWKResolver))
```
Travis-Softwire, line 66: QQ - why `setParametersConverter` rather than `addParametersConverter`? I'm not going to pretend to know anything about the inner workings of the client, but do we know it doesn't have a default converter/ we want to replace it rather than compose it with our JWK resolver?

src/test/kotlin/uk/gov/communities/prsdb/webapp/urlProviders/LandlordDashboardUrlTests.kt
```diff
@@ -51,59 +50,57 @@ import kotlin.test.Test
 class LandlordDashboardUrlTests(
     context: WebApplicationContext,
 ) : ControllerTest(context) {
-    @SpyBean
-    private lateinit var landlordRegistrationJourneyFactory: LandlordRegistrationJourneyFactory
+    @MockitoBean
+    private lateinit var mockLandlordRegistrationJourneyFactory: LandlordRegistrationJourneyFactory
```
Travis-Softwire, line 54: Can we split out the refactoring of these tests from removing the deprecations please? Not saying we shouldn't refactor it (I imagine there are quite a few tests we could refactor if we reviewed them all) but good to keep the PR focused on just this ticket

Travis-Softwire (approved):
A couple of thoughts from me, but LGTM once they're addressed 👍

src/test/kotlin/uk/gov/communities/prsdb/webapp/urlProviders/LandlordDashboardUrlTests.kt
isobel-softwire: Updating to `@MockitoSpyBean` broke these tests (attempting to spy on the journey factories now throws an error) so I refactored to use `@MockitoBean` instead

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/OneLoginConfig.kt
isobel-softwire: It does have a default converter. Using add vs set doesn't seem to make a difference but I'll change it just in case 👍

PR #397, opened 2025-05-02
Spike/prsd 1063 scoped cya pages

AEPR:
## Ticket number

PRSD-1063

## Goal of change
POC for check your answers not requiring a user to re-traverse the whole journey.

Adds a URL parameter to each change link, marking the start point for that change. Each has an allowed list of pages and any navigation outside of that instead goes back to the CYA page.

Draft review, no new tests, currently any step relying on `handleSubmitAndRedirect` erases the "changingAnswerFor" marker, but that will be easily fixed. I want to make sure we want to before I continue down this path!

## Checklist

- [ ] A new journey integration test has been added for any new journeys


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
```diff
@@ -31,6 +31,7 @@ abstract class Journey<T : StepId>(
         get() = sections.flatMap { section -> section.tasks }.flatMap { task -> task.steps }.toSet()
 
     protected open val unreachableStepRedirect = initialStepId.urlPathSegment
+    protected open val checkYourAnswersStepId: T? = null
```
Travis-Softwire, line 34: Does having this on the base journey cause us any problems for update journeys where we're expecting to have multiple CYA pages?

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyRegistrationJourney.kt
```diff
@@ -79,6 +80,48 @@ class PropertyRegistrationJourney(
             listOf("registerProperty.taskList.subtitle"),
         )
 
+    override fun isStepAllowedForChangingAnswersTo(
+        destinationStep: RegisterPropertyStepId?,
+        changingAnswersFor: RegisterPropertyStepId?,
+    ): Boolean =
+        when (changingAnswersFor) {
```
Travis-Softwire, line 87: I'm guessing this doesn't quite map nicely to grouped steps in the update journeys? Appreciate we'll never get away from a switch statement entirely, but if theyr'e essentially the same (i.e. a group of dependent steps) then might be able to avoid having to list them out again?

Travis-Softwire (commented):
A couple of thoughts - happy to chat about them, but overall think it's the right approach

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/Journey.kt
AEPR: Maybe - I think it should be fine, because we can override the getter since it's open. That said - it's going to be another switch statement mapping groups to CYA pages. That said, I don't think we can get away from that since we'll need to explicitly map them somewhere.

AEPR:
Closing draft request - will create a new one.

PR #402, opened 2025-05-06
Feat/prsd 1096 send invitation email

JasminConterioSW:
## Ticket number

PRSD-1096

## Goal of change

Sends the email inviting a local authority admin.

- Modifies the local authority invitation table to include whether or not the invitee should be added as an admin
- Sends the invitation email
image

There aren't any designs for this, but it is heavily based on the existing Invite LA User email

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/InviteLocalAuthorityAdminController.kt
```diff
@@ -38,14 +47,34 @@ class InviteLocalAuthorityAdminController(
...
+                    admin = true,
+                )
+            val invitationLinkAddress = absoluteUrlProvider.buildInvitationUri(token)
+            invitationEmailSender.sendEmail(
+                inviteLocalAuthorityAdminModel.email,
+                LocalAuthorityAdminInvitationEmail(localAuthority, invitationLinkAddress),
+            )
+
+            redirectAttributes.addFlashAttribute("invitedEmailAddress", inviteLocalAuthorityAdminModel.email)
+            return "redirect:/$SYSTEM_OPERATOR_PATH_SEGMENT/$INVITE_LA_ADMIN_PATH_SEGMENT/$CONFIRMATION_PATH_SEGMENT"
```
isobel-softwire, line 73: Do all the endpoints in this class start with `/$SYSTEM_OPERATOR_PATH_SEGMENT/$INVITE_LA_ADMIN_PATH_SEGMENT`? If so it's probably worth making that the top-level mapping.

Also might be worth making a `CONFIRMATION_ROUTE` constant (probably with a better name!)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/LocalAuthorityInvitation.kt
```diff
@@ -24,15 +24,24 @@ class LocalAuthorityInvitation(
     lateinit var invitedEmail: String
         private set
 
+    @Column
+    var invitedAsAdmin: Boolean = false
```
isobel-softwire, line 28: Should this be non-nullable?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityInvitationService.kt
```diff
@@ -17,9 +17,10 @@ class LocalAuthorityInvitationService(
     fun createInvitationToken(
         email: String,
         authority: LocalAuthority,
+        admin: Boolean = false,
```
isobel-softwire, line 20: Minor but a name like `isForAdmin` (or `invitedAsAdmin` like you've used elsewhere) would make it clearer that this should be a Boolean

src/main/resources/db/migrations/V1_18_0__modify_la_invitations_table.sql
isobel-softwire, line 1: This migration name could be a bit more descriptive, i.e mention that the `invited_as_admin` column is being added

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/InviteLocalAuthorityAdminControllerTests.kt
```diff
@@ -57,18 +74,55 @@ class InviteLocalAuthorityAdminControllerTests(
...
                 with(csrf())
             }.andExpect {
                 status { is3xxRedirection() }
                 redirectedUrl("${InviteLocalAuthorityAdminController.INVITE_LA_ADMIN_ROUTE}/$CONFIRMATION_PATH_SEGMENT")
             }
     }
+
+    @Test
+    @WithMockUser(roles = ["SYSTEM_OPERATOR"])
+    fun `sendInvitation sends invitation for valid form submission`() {
```
isobel-softwire, line 98: Does this test need to be separate from the one before? It looks like the only difference between them is what they're asserting

isobel-softwire (approved):
LGTM!

src/main/resources/db/migrations/V1_18_0__modify_la_invitations_table.sql
```diff
@@ -0,0 +1,19 @@
+ALTER TABLE local_authority_invitation
+    ADD COLUMN invited_as_admin BOOLEAN;
```
Travis-Softwire, line 2: If we want to ensure that the migration doesn't fail if there are existing invites, we could  create the column with a default value of false, and then drop that default in a subsequent statement.

Travis-Softwire (approved):
One comment from me on the migration, but overall LGTM

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/InviteLocalAuthorityAdminControllerTests.kt
JasminConterioSW: I did originally combine them, but thought it might make it more obvious which bit was failing if they do start failing (and controller tests are quick!)

Although thinking about it, it is unlikely that only one of them will fail (if emailing sends an error it returns to the same page rather than going to the confirmation page)

PR #431, opened 2025-05-14
PRSD-623: Use Navigator to Set-up Session

isobel-softwire:
## Ticket number

PRSD-623

## Goal of change

Uses navigator to set session data in tests 

- Creates custom journey data deserializer to support LocalDates
- Creates endpoint for storing LA user invitation tokens
- Refactors navigator methods to set-up tests via session setting endpoints

- This PR enforces a new naming scheme for navigator methods:
  - `navigateToX` - navigate to `X`
  - `goToX` - navigate to `X` and return a page object
  - `skipToX` - perform some set-up, then navigate to `X` and return a page object
- The next (and hopefully final) PR for this ticket will split integration tests into journey and single page test classes

## Checklist


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/Navigator.kt
```diff
@@ -147,168 +146,261 @@ class Navigator(
...
 
-    fun goToLandlordRegistrationCountryOfResidencePage(): CountryOfResidenceFormPageLandlordRegistration {
-        val phoneNumberPage = goToLandlordRegistrationPhoneNumberFormPage()
-        phoneNumberPage.submitPhoneNumber("07456097576")
+    fun skipToLandlordRegistrationPhoneNumberPage(): PhoneNumberFormPageLandlordRegistration {
+        setJourneyDataInSession(
+            LandlordRegistrationJourneyFactory.JOURNEY_DATA_KEY,
+            JourneyDataBuilder()
+                .withVerifiedUser()
+                .withEmailAddress()
```
Travis-Softwire, line 194: One nice thing about the previous version is that if we added a new pages in a journey we'd only have to change the navigator once, because each page only depended on the previous one. Could we replicate that witht some helper functions to create a builder for the journey data on a given page? And then subsequent pages can just call the builder for the previous page and add their extra bit of data to it? Hopefully that makes sense, but happy to chat about it

src/test/kotlin/uk/gov/communities/prsdb/webapp/testHelpers/JourneyDataDeserializer.kt
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    private fun JsonNode.deserializeValue(
+        parser: JsonParser,
+        context: DeserializationContext,
+    ): Any? =
+        when {
+            this.isTextual -> {
+                try {
+                    LocalDate.parse(this.asText())
```
Travis-Softwire, line 27: I think we can live with this in the tests - we'll just need to be careful to not include any strings that can be successfully parsed as a LocalDate in our test data!

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/Navigator.kt
isobel-softwire: I've put the helper methods in `JourneyPageDataBuilder`  (not sure about this name so if you have a better idea let me know). Copilot was suprisingly helpful here!

PR #438, opened 2025-05-15
Fix/prsd 1131 js progressive enhancement

AEPR:
## Ticket number

PRSD-1131

## Goal of change
Progressively enhance file uploads to prevent worst UI for ~99.98% of users.

Add a JS callback to the submit button on the file upload fragment to apply basic client side validation. On failure, complete the step with metadata only to elicit full validation errors. Also disables the submit button to prevent retries.

Also prevents metadata only completion of file upload steps from ever succeeding.

## Checklist

- [ ] Screenshots of any UI changes have been added 
    - The only UI change is the new no-script warning component, which is only a placeholder until we have designs


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/EditLaUserPage.kt
```diff
@@ -25,7 +25,7 @@ class EditLaUserPage(
 
     class EditLaUserForm(
         page: Page,
-    ) : Form(page) {
+    ) : JourneyForm(page) {
```
isobel-softwire, line 28: Editing LA users isn't really part of a journey. Maybe it should be called `PostForm` instead?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/basePages/UploadCertificateFormPage.kt
```diff
@@ -17,9 +20,33 @@ abstract class UploadCertificateFormPage(
         form.submit()
     }
 
+    fun metadataOnlySubmission(
```
isobel-softwire, line 23: I don't think this is being used anywhere? Is it going to be used in a future PR's tests?

src/main/js/fileUploadScript.js
```diff
@@ -0,0 +1,76 @@
...
+    const fileInput = form.querySelector('.govuk-file-upload');
+    return fileInput.files[0];
+}
+
+function isFileTooLarge(file) {
+    const megaByteInBytes = 1024 * 1024;
+    return file.size > 15 * megaByteInBytes;
+}
+
+export { isFileTooLarge, submitSyntheticForm, createHiddenForm, getInputFileFromForm };
```
isobel-softwire, line 76: Do all of these methods need to be exported?

package.json
```diff
@@ -3,9 +3,13 @@
   "private": true,
   "version": "0.1.0",
   "main": "src/javascript/app.js",
+  "type": "module",
+  "imports": {
+    "main-javascript": "./src/js"
+  },
```
isobel-softwire, line 9: What does this do?

isobel-softwire (approved):
This LGTM! I left some small thoughts.
I think it's worth adding an assertion that a synthetic form has been sent in the integration tests that try to upload a large file.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyComplianceController.kt
```diff
@@ -109,9 +109,13 @@ class PropertyComplianceController(
     ): ModelAndView {
         throwErrorIfUserIsNotAuthorized(principal.name, propertyOwnershipId)
 
+        // We must ensure that we can distinguish between a metadata-only file upload and a normal file upload when
+        // postJourneyData() is used for a file upload endpoint.
+        val annotatedFormData = formData + ("metadataOnly" to true)
```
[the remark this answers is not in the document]
AEPR: No, it explicitly needs to be done in the service not the javascript - because otherwise it is spoof-able. The point of this is that if a client sends in a post claiming to not be `metadataOnly` then this endpoint will overwrite that so we still treat it correctly.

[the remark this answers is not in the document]
isobel-softwire: In the form model this field's called `isMetadataOnly` - is `metadataOnly` being bound to that or is it just using the default value?

package.json
AEPR: Good catch! It's a slightly faulty implementation of import aliasas. When it's fixed, it means that I can replace the import statement at the top of the test (`'../../main/js/fileUploadScript.js'`) with a nicer one (`'#main-javascript/fileUploadScript'`)

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/EditLaUserPage.kt
AEPR: Yeah, good suggestion.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyComplianceController.kt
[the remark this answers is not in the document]
isobel-softwire: Fair enough. I think it's worth using `UploadCertificateFormModel::isMetadataOnly.name` (probably won't matter much though)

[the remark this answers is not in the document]
AEPR: Yeah, definitely agree that's much nicer.

src/test/js/fileUploadScript.test.js
```diff
@@ -0,0 +1,104 @@
+import assert from 'node:assert/strict';
+import { describe, test } from 'node:test';
+import jsdom from 'global-jsdom';
+import {createFileUploadIntercepter} from '#main-javascript/fileUploadScript.js';
```
Travis-Softwire, line 4: Minor but inconsistent formatting in imports. Not as part of this PR, but we should probably add prettier since we'll probably end up adding more progressive enhancements in future.

Travis-Softwire (approved):
LGTM :)

Travis-Softwire (approved):
👍Config changes look right to me

src/main/js/fileUploadScript.js
```diff
@@ -0,0 +1,76 @@
...
+
+function getInputFileFromForm(form){
+    const fileInput = form.querySelector('.govuk-file-upload');
+    return fileInput.files[0];
+}
+
+function isFileTooLarge(file) {
+    const megaByteInBytes = 1024 * 1024;
+    return file.size > 15 * megaByteInBytes;
+}
```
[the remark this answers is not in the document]
AEPR: I've (separately) had a suggestion with Travis about this and we want them to align given we're doing the check as otherwise we're getting users to upload unnecesary and potentially very large files that we know we're going to reject.

PR #442, opened 2025-05-19
Feat/prsd 1138 epc lookup service

JasminConterioSW:
## Ticket number

PRSD-1138

## Goal of change

First part of PRSD-1138 - service layer for the EPC lookup page (lookup by certificate number)

Creates an EpcLookupService that calls the existing EpcRegisterClient and maps the results to an EpcDataModel. This includes the information needed for subsequent epc pages in the compliance journey.

Note this not complete PRSD-1138

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/EpcLookupService.kt
```diff
@@ -0,0 +1,69 @@
...
+                addressLineOne = epcDataAddress.getString("addressLine1"),
+                townOrCity = epcDataAddress.getString("town"),
+                postcode = epcDataAddress.getString("postcode"),
+                addressLineTwo = epcDataAddress.optString("addressLine2"),
+            )
+
+        return EpcDataModel(
+            certificateNumber = epcData.getString("epcRrn"),
+            singleLineAddress = singleLineAddress,
+            energyRating = epcData.getString("currentEnergyEfficiencyBand"),
```
JasminConterioSW, line 44: There is also a "currentEnergyEfficiencyRating" field but this gives a number. The design for the Check Matched EPC page (which is what will use this field of this data model) wants the band

Travis-Softwire: Yeah this is fine - we asked for the rating as well in case we need to do something with the score later rather than just the band

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/EpcLookupService.kt
```diff
@@ -0,0 +1,69 @@
...
+
+        if (jsonResponse.has("errors")) {
+            val errorCode = getErrorCode(jsonResponse)
+            if (errorCode == "NOT_FOUND") {
+                return null
+            }
+            if (errorCode == "INVALID_REQUEST" || errorCode == "BAD_REQUEST") {
+                throw ResponseStatusException(HttpStatus.BAD_REQUEST, getErrorMessage(jsonResponse))
+            }
+        }
```
isobel-softwire, line 29: Rather than repeating this in the `EpcRegisterClient`, could the client handle all the error throwing and returning null for 404s? If you go with this the `getErrorCode` and `getErrorMessage` could be private `EpcRegisterClient` methods.

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/EpcLookupServiceTests.kt
```diff
@@ -0,0 +1,172 @@
...
+        whenever(mockEpcRegisterClient.getByRrn(certificateNumber)).thenReturn(mockResponse)
+
+        val result = epcLookupService.getEpcByCertificateNumber(certificateNumber)
+
+        assertNotNull(result)
+        assertEquals(certificateNumber, result?.certificateNumber)
+        assertEquals("C", result?.energyRating)
+        assertEquals(LocalDate(2027, 1, 5), result?.expiryDate)
+        assertEquals("123 Test Street, Flat 1, Test Town, TT1 1TT", result?.singleLineAddress)
+        assertTrue(result?.isLatestCertificateForThisProperty() == true)
```
isobel-softwire, line 57: Could this test's assertions compare the result against an `expectedEpcDataModel` instead of looking at individual properties?

isobel-softwire (commented):
LGTM - left a few refactoring thoughts

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/EpcLookupService.kt
```diff
@@ -0,0 +1,69 @@
...
+        val epcDataAddress = epcData.getJSONObject("address")
+        val singleLineAddress =
+            AddressDataModel.manualAddressDataToSingleLineAddress(
+                addressLineOne = epcDataAddress.getString("addressLine1"),
+                townOrCity = epcDataAddress.getString("town"),
+                postcode = epcDataAddress.getString("postcode"),
+                addressLineTwo = epcDataAddress.optString("addressLine2"),
+            )
+
+        return EpcDataModel(
```
Travis-Softwire, line 41: I don't feel super strongly about it - but this might be a bit nicer as a `.fromJSONObject` static method on EpcDataModel?

Travis-Softwire (approved):
LGTM overall - just a few minor suggestions, but don't need to see it again :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/EpcDataModel.kt
```diff
@@ -0,0 +1,27 @@
...
+    val energyRating: String,
+    val expiryDate: LocalDate,
+    val latestCertificateNumberForThisProperty: String? = null,
+) {
+    fun isLatestCertificateForThisProperty() = certificateNumber == latestCertificateNumberForThisProperty
+
+    companion object {
+        fun formatCertificateNumber(certificateNumber: String): String {
+            val certNumberNoHyphens = certificateNumber.replace("-", "")
+            require(certNumberNoHyphens.all { it.isDigit() }) { "Input must contain only digits and hyphens" }
```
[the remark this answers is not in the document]
JasminConterioSW: We actually shouldn't get to here because there will be validation checks on the page's form model, so maybe these checks aren't necessary.

I might go with Isobel's suggestion of returning null if certificateNumber is not of the right format

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/EpcLookupService.kt
JasminConterioSW: I think I prefer to leave it as the client returning the response and the service layer deciding what to do with it (e.g. there may be times when we want to throw an error for a NOT_FOUND response rather than returning null)

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/EpcLookupServiceTests.kt
JasminConterioSW: Yep, co-pilot had done it with individual properties. Either works but you're probably right that the data model version is nicer

PR #444, opened 2025-05-21
Feat/prsd 1138 epc lookup page

JasminConterioSW:
## Ticket number

PRSD-1138

## Goal of change

Add the EPC lookup page

image
- Add the EpcLookup step
   - This calls the epd endpoint and depending on the response, sets the nextAction as EpcSuperseded, EpcNotFound or CheckMatchedEpc
- Add a skeleton CheckMatchedEpc step where the nextAction can be modified as required to avoid an infinite loop of nextActions.

I ran into a problem where the nextAction for Check Matched Epc was Epc Lookup, but then the nextAction for Epc Lookup can be Check Matched Epc. When you then load any journey step and it tries to calculate whether steps are reachable, zipWithNext() ends up in an infinite loop (it doesn't error or timeout, just hangs...)

This is now fixed by setting the nextAction to null if it should be Check Matched Epc, but still redirecting to that page - the user sees the correct journey but the Journey doesn't hit the Infinite Loop of Doom. This does mean that unlike almost everywhere else, completing this step won't call Journey's getRedirectForNextStep to get the redirect from nextAction.

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyComplianceJourney.kt
```diff
@@ -860,6 +935,54 @@ class PropertyComplianceJourney(
...
+        } else {
+            Pair(PropertyComplianceStepId.EpcLookup, null)
+        }
+
+    private fun checkMatchedEpcStepHandleSubmitAndRedirect(journeyData: JourneyData): String {
+        val epcLookupStep = steps.single { it.id == PropertyComplianceStepId.EpcLookup }
+
+        val newJourneyData =
+            if (epcLookupStep.nextAction(journeyData, null).first == PropertyComplianceStepId.CheckMatchedEpc) {
+                journeyData.withEpcDetails(null)
```
JasminConterioSW, line 951: This might need some tweaking, thinking about it...
CheckMatchedEpc will still need to use the looked up EPC details if they exist so that it can display those instead of lookup up the EPC by UPRN again.

We could also save the EpcDetails somewhere specifically on the EpcLookup page, and make the Lookup page check there to determine it's next action. I'll have a bit of a think!

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/EpcLookupFormModel.kt
```diff
@@ -0,0 +1,35 @@
...
+import jakarta.validation.constraints.NotNull
+import uk.gov.communities.prsdb.webapp.validation.ConstraintDescriptor
+import uk.gov.communities.prsdb.webapp.validation.DelegatedPropertyConstraintValidator
+import uk.gov.communities.prsdb.webapp.validation.IsValidPrioritised
+import uk.gov.communities.prsdb.webapp.validation.NotBlankConstraintValidator
+import uk.gov.communities.prsdb.webapp.validation.ValidatedBy
+
+@IsValidPrioritised
+class EpcLookupFormModel : FormModel {
+    @NotNull(message = "forms.epcLookup.error.missing")
```
Travis-Softwire, line 12: A couple of points:
- Did you mean to make `certificateNumber` nullable so that a null check makes sense?
- I don't think this will play well in general with `@IsValidPrioritised` and `ValidatedBy` - if both the `NotNull` validation and the other validations trigger I think you'll get multiple validation errors on the same field (and if you don't , I don't think Spring guarentees their order)

Travis-Softwire (approved):
One question about the validation, but otherwise LGTM :) As discussed - suspect we'll want to come back to the journey logic RE preventing loops when you add the page for checking the automatic UPRN EPC lookup, but that's fine.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/EpcLookupFormModel.kt
```diff
@@ -0,0 +1,35 @@
...
+        ],
+    )
+    var certificateNumber: String = ""
+
+    fun isEpcCertificateNumberFormatValid(): Boolean {
+        // a blank certificate number does not have an invalid format
+        if (certificateNumber.isBlank()) return true
+
+        val certNumberNoHyphens = certificateNumber.replace("-", "")
+        return (certNumberNoHyphens.all { it.isDigit() } && certNumberNoHyphens.length == 20)
```
isobel-softwire, line 33: Using `EpcDataModel.parseCertificateNumberOrNull` here would allow us to have one source of truth as to what a valid certificate number looks like

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/EpcLookupFormModel.kt
JasminConterioSW: Oops, I think I meant to take line 12 out when I added the other validation messages...

PR #461, opened 2025-05-30
PRSD-1202: Save Compliance Record

isobel-softwire:
## Ticket number

PRSD-1202

## Goal of change

Saves a record to the database at the end of the property compliance journey

- Creates `PropertyCompliance` entity
- Updates the property compliance journey's final `handleSubmitAndRedirect` to create a property compliance record

N/A

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
```diff
@@ -0,0 +1,109 @@
...
+    )
+    lateinit var propertyOwnership: PropertyOwnership
+        private set
+
+    @Column(name = "gas_safety_cert_s3_key")
+    var gasSafetyCertS3Key: String? = null
+
+    var gasSafetyCertIssueDate: LocalDate? = null
+
+    var gasSafetyCertExpiryDate: LocalDate? = null
```
Travis-Softwire, line 39: I'm not convinced we should be storing expiry dates rather than just calculating from issue dates. It's just a cache of the result of issue date + how long they're valid, and the latter could potentially change in the future.

Travis-Softwire (approved):
One thought about storing expiry dates or not, but otherwise LGTM :) I did also pause on storing the s3 keys rather than storing the file extensions and computing the keys when we need them, but I think in practice there's not much benefit

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
isobel-softwire: That's fair enough. I'll make the expiry dates properties with getters instead

PR #463, opened 2025-06-02
PRSD-1040: Use db-specific roles for migrations

Travis-Softwire:
## Ticket number

PRSD-1040

## Goal of change

Use the new rds access role for jobs that run migrations

Updates our github actions to use our new role which has more restricted permissions to run the migrations - following the principle of least privilege.


isobel-softwire (approved):
LGTM :)

PR #465, opened 2025-06-03
PRSD-1219: Isolate Occupancy Update Sub-journeys

isobel-softwire:
## Ticket number

PRSD-1219

## Goal of change

Isolates journey data for the property occupancy update sub-journeys

- Adds copies of occupancy steps for each sub-journey
- Creates a factory to determine which copy of steps to include in the journey
- Pre-loads skipped sub-journey step data (e.g. when updating number of households, occupation status is loaded)

N/A

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyDetailsUpdateJourney.kt
```diff
@@ -404,12 +389,8 @@ class PropertyDetailsUpdateJourney(
...
     }
 
-    private fun wasPropertyOriginallyOccupied() = journeyDataService.getJourneyDataFromSession().getOriginalIsOccupied(originalDataKey)!!
-
-    private fun hasPropertyOccupancyBeenUpdated() = journeyDataService.getJourneyDataFromSession().getIsOccupiedUpdateIfPresent() != null
-
-    private fun hasNumberOfHouseholdsBeenUpdated() =
-        journeyDataService.getJourneyDataFromSession().getNumberOfHouseholdsUpdateIfPresent() != null
+    private fun wasPropertyOriginallyOccupied() =
+        journeyDataService.getJourneyDataFromSession().getOriginalIsOccupied(stepFactory.getOccupancyStepId(), originalDataKey)!!
```
AEPR, line 393: I think, given the new set up, we should move this and `getOccupancyStepFieldSetHeading` to the `StepFactory` as a `createOccupancyStep`

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordDetailsUpdateJourney.kt
```diff
@@ -82,12 +83,12 @@ class LandlordDetailsUpdateJourney(
...
+        val journeyData = journeyDataService.getJourneyDataFromSession()
+        val originalJourneyData = JourneyDataHelper.getPageData(journeyData, originalDataKey)
 
-            val journeyData = journeyDataService.getJourneyDataFromSession()
-            val lookedUpAddresses = JourneyDataHelper.getPageData(journeyData, originalDataKey)!!.getSerializedLookedUpAddresses()!!
+        if (journeyData.getSerializedLookedUpAddresses() == null && originalJourneyData != null) {
+            val lookedUpAddresses = originalJourneyData.getSerializedLookedUpAddresses()!!
             val journeyDataWithLookedUpAddresses = journeyData.withUpdatedLookedUpAddresses(lookedUpAddresses)
             journeyDataService.setJourneyDataInSession(journeyDataWithLookedUpAddresses)
         }
```
AEPR, line 94: So this initialises the `lookedUpAddresses` if the original journey data has been initalised but the looked up addresses haven't been added. I think its current name doesn't communicate that. Maybe something like `addLookedUpAddressesToOriginalJourneyData` or `initialiseLookedUpAddressesInOriginalJourneyData`?

AEPR (commented):
I've not reviewed the extensions or tests yet - I will continue looking now, but I wanted to share the feedback I had so far sooner rather than later. 

I started this review on Friday afternoon, and some of struggles to understand what's going on may be Friday afternoon brain (I've definitely got my head round it better this morning!) but I still think it's worth trying to make it all as clear as possible given that the journey framework is already somewhat unapproachable.

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/journeyExtensions/PropertyDetailsUpdateJourneyExtensions.kt
```diff
@@ -26,32 +26,50 @@ class PropertyDetailsUpdateJourneyExtensions {
                 OwnershipTypeFormModel::ownershipType.name,
             )
 
-        fun JourneyData.getOriginalIsOccupied(originalJourneyKey: String) =
-            JourneyDataHelper.getPageData(this, originalJourneyKey)?.getIsOccupied()
+        fun JourneyData.getOriginalIsOccupied(
+            occupancyStepId: UpdatePropertyDetailsStepId,
+            originalJourneyKey: String,
+        ) = JourneyDataHelper.getPageData(this, originalJourneyKey)?.getIsOccupied(occupancyStepId)
```
AEPR, line 32: I don't love that we now need to pass in the step id here that we actually want the extension to retrieve (since hiding that is part of the point of the extensions). That said I can't see a neater way to access the data without making assumptions about journey data structure that might not be true. You could pass in the group id instead of the actual step id, but that sounds like making more work that it's worth. 

I haven't looked at the tests yet, but it's probably worth adding a test that explictly calls out what happens if you pass an incorrect step if  here (you get no update back).

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDetailsUpdateSinglePageTests.kt
```diff
@@ -0,0 +1,40 @@
...
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.basePages.BasePage.Companion.assertPageIs
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.propertyDetailsUpdateJourneyPages.HouseholdsOccupancyFormPagePropertyDetailsUpdate
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.propertyDetailsUpdateJourneyPages.PeopleNumberOfHouseholdsFormPagePropertyDetailsUpdate
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.propertyDetailsUpdateJourneyPages.PeopleOccupancyFormPagePropertyDetailsUpdate
+import kotlin.test.assertEquals
+
+class PropertyDetailsUpdateSinglePageTests : SinglePageTestWithSeedData("data-local.sql") {
+    private val propertyOwnershipId = 1L
+    private val urlArguments = mapOf("propertyOwnershipId" to propertyOwnershipId.toString())
+
```
AEPR, line 15: I think its worth having a test for back link behaviour - specifically that if you go to a skipped update page then go forward you can get back to it again with the back links.

I'm not sure if it should live here or in the Journey tests though.

Travis-Softwire (changes requested):

Travis-Softwire:
Agree w/ Alex's comments - I think we need to do our best to make it clear what's happening here, probably more so than usual, because this is essentially a big workaround for a non-obvious problem, and so it's never going to be logical or intuitive to someone reading it without that context. Also wouldn't be afraid of some repetition if it makes it easier to understand in this case.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordDetailsUpdateJourney.kt
isobel-softwire: This adds the `lookedUpAddresses` from the original journey data to the current journey data, so I'll change it's name to `initializeLookedUpAddressesIfNotInitialized`.

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/journeyExtensions/PropertyDetailsUpdateJourneyExtensions.kt
isobel-softwire: I agree that passing in the group ID would make it a bit nicer. It would only take a call to a step factory method to get the corresponding `occupancyStepId` so I think it's worth doing. Using group id also removes the possibility of passing an incorrect step.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDetailsUpdateSinglePageTests.kt
isobel-softwire: I agree - I'll put them in here as they won't involve updating the database

AEPR (approved):
Couple of minor comments. I'm off tomorrow, so I'm leaving an approving review subject to the couple of minor remaining issues being resolved and Travis no longer requesting changes!

Travis-Softwire (approved):
Agree the block comment makes this much easier to follow :)

PR #466, opened 2025-06-03
PRSD-1231: Draft ADR for handling virus scanning results

Travis-Softwire:
## Ticket number

PRSD-1231

## Goal of change

Evaluate our options for handling the results of our virus scanning

Adds an ADR discussing the options for how we might handle the results of our virus scan, with a focus on how we might address it during the pilot.


rowanhill (approved):
Looks good to me, as written.

I'm in favour of the EventBridge approach, for all the reasons you've listed :)

PR #473, opened 2025-06-05
PRSD-734: Change initial journey step back links

AEPR:
## Ticket number
PRSD-734

## Goal of change
Update back links on the first page of journeys where it is not set.

* Updated the back url for the first page of registering a property to go to the landlord dashboard.
* Updated the back url for the first page of registering as a landlord to go to the journey intro page.

I've left the first page of the register as an LA user without a back link as you have reached it from an email so it's not very useful.
I've currently not added any tests for these back links. Do you think I need to add them?

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
@@ -210,6 +210,7 @@ class LandlordRegistrationJourney(
                             "fieldSetHeading" to "forms.confirmDetails.heading",
                             "fieldSetHint" to "forms.confirmDetails.summary",
                             "submitButtonText" to "forms.buttons.confirmAndContinue",
+                            "backUrl" to "/$REGISTER_LANDLORD_JOURNEY_URL",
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/LandlordRegistrationJourney.kt
```diff
@@ -210,6 +210,7 @@ class LandlordRegistrationJourney(
                             "fieldSetHeading" to "forms.confirmDetails.heading",
                             "fieldSetHint" to "forms.confirmDetails.summary",
                             "submitButtonText" to "forms.buttons.confirmAndContinue",
+                            "backUrl" to "/$REGISTER_LANDLORD_JOURNEY_URL",
```
isobel-softwire, line 213: You could use `BACK_URL_ATTR_NAME` here (and in `nameStep`)

isobel-softwire (approved):
This LGTM. The links are static so I feel like it's fine to leave out tests.

Travis-Softwire (approved):
👍Comfortable with not testing these back links specifically, provided we're happy we've got enough tests for the back link override functionality more generally

PR #489, opened 2025-06-12
PRSD-NONE: Clean up (Filtered) Journey Data usage in Property Compliance Journey

isobel-softwire:
## Ticket number

PRSD-NONE

## Goal of change

Use filtered or session journey data where appropriate in the property compliance journey

- Modifies property compliance journey to use filtered or session journey data where appropriate

N/A

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyComplianceJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyComplianceJourney.kt
@@ -1252,19 +1252,12 @@ class PropertyComplianceJourney(
         return getRedirectForNextStep(currentStep, newFilteredJourneyData, null)
     }
 
-    private fun resetCheckMatchedEpcInSession(
-        filteredJourneyData: JourneyData,
-        epcDetails: EpcDataModel?,
-    ): JourneyData {
-        // Removes the answer to checkMatchedEpc if the EPC details have changed
-        if (epcDetails != filteredJourneyData.getEpcDetails(autoMatched = false)) {
-            val newJourneyData = journeyDataService.getJourneyDataFromSession().withResetCheckMatchedEpc()
+    private fun resetCheckMatchedEpcInSessionIfChangedEpcDetails(newEpcDetails: EpcDataModel?) {
+        val journeyData = journeyDataService.getJourneyDataFromSession()
+        if (newEpcDetails != journeyData.getEpcDetails(autoMatched = false)) {
+            val newJourneyData = journeyData.withResetCheckMatchedEpc()
             journeyDataService.setJourneyDataInSession(newJourneyData)
-
-            val newFilteredJourneyData = filteredJourneyData.withResetCheckMatchedEpc()
-            return newFilteredJourneyData
         }
-        return filteredJourneyData
     }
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyComplianceJourney.kt
```diff
@@ -1297,8 +1290,8 @@ class PropertyComplianceJourney(
     private fun epcLookupStepHandleSubmitAndRedirect(filteredJourneyData: JourneyData): String {
         val certificateNumber = filteredJourneyData.getEpcLookupCertificateNumber()!!
         val lookedUpEpc = epcLookupService.getEpcByCertificateNumber(certificateNumber)
-        val newFilteredJourneyData = resetCheckMatchedEpcInSession(filteredJourneyData, lookedUpEpc)
-        return updateEpcDetailsInSessionAndRedirectToNextStep(epcLookupStep, newFilteredJourneyData, lookedUpEpc, autoMatchedEpc = false)
+        resetCheckMatchedEpcInSessionIfChangedEpcDetails(lookedUpEpc)
```
JasminConterioSW, line 1293: In case anyone else is wondering:

- checkMatchedEpc gets updated in the session so that the answer is not populated when that page loads. It is always retrieved from the session so shouldn't matter if it is somehow in the filteredJourneyData for this step here.
- checkMatchedEpc won't be in the filteredJourneyData here anyway - it may have actually been the user's previous step, but epcLookup is not set as the nextAction for checkMatchedEpc (we just tell it to redirect to here) in order to avoid an infinite loop of previous steps.

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyComplianceJourneyTests.kt
```diff
@@ -592,17 +593,19 @@ class PropertyComplianceJourneyTests {
...
                     .withEpcLookupCertificateNumber(SUPERSEDED_EPC_CERTIFICATE_NUMBER)
                     .withLookedUpEpcDetails(supersededEPC)
                     .build()
             whenever(mockJourneyDataService.getJourneyDataFromSession()).thenReturn(originalJourneyData)
 
             val updatedJourneyData =
-                JourneyDataBuilder()
+                JourneyDataBuilder(initialJourneyData = originalJourneyData)
                     .withEpcLookupCertificateNumber(SUPERSEDED_EPC_CERTIFICATE_NUMBER)
                     .withLookedUpEpcDetails(supersededEPC)
```
JasminConterioSW, line 607: I think if you are already including all the originalJourneyData you shouldn't need these?
The only difference here would be the checkMatchedEpc page being removed.

(It will also change the saved epc details but that comes after resetting the checkMatchedEpc page data)

JasminConterioSW (approved):
One minor comment, but LGTM :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyComplianceJourneyTests.kt
isobel-softwire: You're right - I'll change this 👍

PR #498, opened 2025-06-16
PRSD-1299: Adds draft ADR for restricting access to the pilot

Travis-Softwire:
## Ticket number

PRSD-1299

## Goal of change

Add an ADR covering how we should restrict access to the pilot

No decision currently - the aim is to use this to nail down the options and then take it to Product & Service owners for a decision

### DIFF
--- a/adrs/0032-restricting-pilot-access.md
+++ b/adrs/0032-restricting-pilot-access.md
@@ -1,0 +1,92 @@
+# ADR-0032: Restricting pilot access
+
+## Status
+
+Draft
+
+Date of decision: {yyyy-MM-dd}
+
+## Context and Problem Statement
+
+We need to restrict access to the pilot for two reasons:
+
+- to ensure that no one can accidentally mistake the pilot for a live service,
+- to manage the risk of the application being overloaded with too many users
+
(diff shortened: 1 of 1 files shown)


rowanhill:
All looks sensible to me. My vote is for the single shared passcode, FWIW.

I did think of a couple more possible options:
1. IP allowlisting. Lots of issues with this, though, most notably the admin overhead and people without fixed IPs.
2. Landlord registration pauses immediately after One Login, requires LA to click "allow" on some dashboard to continue. Avoids the need for passcodes, so maybe is simpler in some sense? But means anyone can (start to) create a LL account, and we'd have to fiddle with the LL registration journey.

rowanhill:
> So I didn't include IP allowlisting as most domestic ISPs use dynamic IPs, which I think makes it unworkable.

Yep, I agree - I was just mentioning it for completeness!

> We could probably address both of those by still having a passcode, but instead issuing one per LA, so we know where to route the registration requests. As you say though, it would interrupt the registration journey, and is probably at least as fiddly as adding an invite flow.

Also agreed: I don't think we need to include this in the ADR since we have an item that achieves something similar and probably with less work...

rowanhill (approved):

PR #501, opened 2025-06-17
Feat/PRSD-601 redirect registered la users from registration flow to dashboard

Kiran-R-K:
## Ticket number

PRSD-601

## Goal of change

Redirect registered local authority users from the registration flow to the local authority dashboard

Added a check to the landing-page step in the la user registration controller that redirects the user to the la dashboard if they are already registered
Added a method to the UserRolesService to check if they are a registered la user
Added unit and integration tests

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
```diff
@@ -125,5 +155,6 @@ class RegisterLAUserController(
 
     companion object {
         const val INVALID_LINK_PAGE_PATH_SEGMENT = "invalid-link"
+        const val LANDING_PAGE_PATH_SEGMENT = "landing-page"
```
JasminConterioSW, line 158: I wonder if we should change this to be START_PAGE_PATH_SEGMENT to match the Landlord and Property registration journeys?

I don't think there is a reason for this one to be different

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
```diff
@@ -42,11 +44,39 @@ class RegisterLAUserController(
...
+    ): ModelAndView {
+        val token = getValidTokenFromSessionOrNull()
+        if (token == null) {
+            invitationService.clearTokenFromSession()
+            return ModelAndView("redirect:$INVALID_LINK_PAGE_PATH_SEGMENT")
+        }
+
+        val invitation = invitationService.getInvitationFromToken(token)
+
+        if (userRolesService.getHasLocalAuthorityUserRole(principal.name)) {
```
JasminConterioSW, line 60: I think here we should also check if they have a `LocalAuthorityAdminRole `- but maybe the service can just have one function `getHasLocalAuthorityRole` which looks for both.

In practice, it shouldn't make a difference because we add the LA_USER role to LA_ADMIN anyway, but I think it would make it clearer than it applies to anyone in the local authority table.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/UserRolesService.kt
```diff
@@ -43,4 +43,9 @@ class UserRolesService(
         val roles = getRolesForSubjectId(subjectId)
         return roles.contains(ROLE_LANDLORD)
     }
+
+    fun getHasLocalAuthorityUserRole(subjectId: String): Boolean {
+        val roles = getRolesForSubjectId(subjectId)
+        return roles.contains(ROLE_LA_USER)
```
JasminConterioSW, line 49: We could make this check for `ROLE_LA_ADMIN `as well and call it `getHasLocalAuthorityRole`?

JasminConterioSW (commented):
Looking good! Just a couple of comments :)

PR #511, opened 2025-06-18
PRSD-1219: Starting Update Clears Journey Context

isobel-softwire:
## Ticket number

PRSD-1219

## Goal of change

Clear journey context when starting an update from the property details page

- Adds `startingUpdate` query parameter to change links on the property details page
- Clears journey data and context ID from session for requests where `startingUpdate` is true

This PR is motivated by an issue that came up during QA (see ticket for details)

## Checklist


Travis-Softwire:
Hmm, not keen on this solution. We should at least do a redirect to remove the query parameter once we've cleared the session, but mixing this together with the other multiple steps workaround + the CYA check altogether all feels very brittle and hard to predict exactly what is going to happen in any given set of scenarios. E.g. I think this means that just clicking on the same change link in a different tab will be enough to invalidate a CYA in a different tab?

Travis-Softwire (approved):
Nice :)

PR #519, opened 2025-06-20
chore/PRSD-NONE-update-manage-local-authority-users-urls-to-constants

Kiran-R-K:
## Ticket number

PRSD-NONE

## Goal of change

Update the url path segments in the `ManageLocalAuthorityUsersController` to use constants instead of strings

As above

## Checklist


isobel-softwire (commented):
I think it would be worth creating constants (or methods for ones with parameters) for the routes instead of constructing them each time they're needed. Apart from that LGTM :)

PR #523, opened 2025-06-23
PRSD-965: Placeholder Compliance View

isobel-softwire:
## Ticket number

PRSD-965

## Goal of change

Adds a placeholder compliance view tab to property records in preparation for partially compliant confirmation emails

- Adds placeholder compliance info tab to property details view
- Adds method for generating absolute URL for property compliance info tab

This will be used/tested in the next PR for PRSD-965, which should complete the ticket.

## Checklist

image


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
```diff
@@ -40,11 +42,9 @@ class PropertyDetailsController(
     @GetMapping(PROPERTY_DETAILS_ROUTE)
     fun getPropertyDetails(
         @PathVariable propertyOwnershipId: Long,
-        model: Model,
-        principal: Principal,
-    ): String {
-        val propertyOwnership =
-            propertyOwnershipService.getPropertyOwnershipIfAuthorizedUser(propertyOwnershipId, principal.name)
+    ): ModelAndView {
+        val baseUserId = SecurityContextHolder.getContext().authentication.name
```
Travis-Softwire, line 46: Why are we using this in favour of using the `Principal`? I know we're using this elsewhere to avoid having to pass a `Principal` down through a lot of function calls, but here we have direct access to the injected `Principal`

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
```diff
@@ -65,12 +65,13 @@ class PropertyDetailsController(
                 landlordDetailsUrl,
             )
 
-        model.addAttribute("propertyDetails", propertyDetails)
-        model.addAttribute("landlordDetails", landlordViewModel)
-        model.addAttribute("deleteRecordLink", DeregisterPropertyController.getPropertyDeregistrationPath(propertyOwnershipId))
-        model.addAttribute("backUrl", LANDLORD_DASHBOARD_URL)
-
-        return "propertyDetailsView"
+        val modelAndView = ModelAndView("propertyDetailsView")
```
Travis-Softwire, line 68: I don't think we need to switch to a `ModelAndView` here - they're useful where we're delegating populating the model and deciding on the template to another function, but here we're just adding some properties to a model and returning the template, so I think we can leave this as the simpler version that just returns a string.

Travis-Softwire (approved):
Couple of minor points where I don't think we need the refactors, but otherwise LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
isobel-softwire: We call this method in `AbsoluteUrlProvider.buildComplianceInformationUri`, so I thought it made more sense to deal with the `Principal` and `ModelAndView` in the controller than to make them parameters (of `buildComplianceInformationUri`).

PR #527, opened 2025-06-24
PRSD-965: Compliance Confirmation Email

isobel-softwire:
## Ticket number

PRSD-965

## Goal of change

Sends confirmation email and the end of the property compliance journey

- Creates email templates and models for fully and partially compliant confirmation emails
- Uploads templates to Notify
- Sends relevant confirmation email at end of the property compliance journey

N/A

## Checklist

## Screenshots
- Full compliance
image
- Partial compliance
image


src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/MessageSourceExtensions.kt
```diff
@@ -0,0 +1,13 @@
+package uk.gov.communities.prsdb.webapp.helpers.extensions
+
+import org.springframework.context.MessageSource
+import java.util.Locale
+
+class MessageSourceExtensions {
+    companion object {
+        fun MessageSource.getMessageForKey(
```
JasminConterioSW, line 8: I don't think we have explicit tests for these (unless I missed them which is entirely possible!), could you add one please?

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyComplianceJourneyTests.kt
```diff
@@ -775,6 +799,97 @@ class PropertyComplianceJourneyTests {
...
+                .sendEmail(compliantPropertyCompliance.propertyOwnership.primaryLandlord.email, expectedEmailModel)
+        }
+
+        @Test
+        fun `checkAndSubmitHandleSubmitAndRedirect sends a partially compliant confirmation email for non-compliant property`() {
+            val nonCompliantPropertyCompliance = PropertyComplianceBuilder.createWithMissingCerts()
+            setUpMocks(nonCompliantPropertyCompliance, isPropertyCompliant = false)
+
+            val expectedMessageKeys = PropertyComplianceConfirmationMessageKeys(nonCompliantPropertyCompliance)
+            val expectedCompliantMessages = expectedMessageKeys.compliantMsgKeys.map { mockMessageSource.getMessageForKey(it) }
```
JasminConterioSW, line 835: I am wondering if in the tests it's better to hardcode the messages rather than get them from mockMessageSource.

Using getMessageForKey feels nice because it will still work if the content gets updated, but I think the test _should_ fail if the content is updated. (E.g. we might use the same message somewhere else, update it for there, are not realise it's getting used here?).

It also means if the test is failing because a message is wrong, it will be more obvious at a glance what it is supposed to be.

JasminConterioSW (commented):
Just a couple of comments on getting message keys, but generally looking good!

src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyComplianceJourneyTests.kt
isobel-softwire: `mockMessageSource` doesn't actually use the messages from `messages.properties` as it's a `MockMessageSource` rather than a mocked `MessageSource` (`MessageSource` can't be mocked as it has `final` methods). So we can't check the content here, but it is being checked in the integration tests which is probably sufficient.

JasminConterioSW: Ah I see.
It might still be clearer that the test is not getting the real message if make this something like 
`val expectedCompliantMessages = expectedMessageKeys.compliantMsgKeys.map { "Message for $it" }`

and update the methods in MockMessageSource to return `"Message for $code"`

PR #533, opened 2025-06-25
Chore/prsd 458 remove unused controllers

JasminConterioSW:
## Ticket number

PRSD-458

## Goal of change

Remove unused controllers

Remove CheckHomeController and RegisterHomeController

This does not complete the ticket

## Checklist


isobel-softwire (approved):
LGTM - probably worth also removing any messages that are only used in these controllers

JasminConterioSW:
> LGTM - probably worth also removing any messages that are only used in these controllers

Good point! I think we're ok though, they both just return the "index" template which is still used in SignOutController :)

JasminConterioSW:
> > > LGTM - probably worth also removing any messages that are only used in these controllers
> > 
> > 
> > Good point! I think we're ok though, they both just return the "index" template which is still used in SignOutController :)
> 
> I meant messages from `messages.properties`, like `checkHome` and `registerAHome`

Oops... quick PR here! https://github.com/communitiesuk/prsdb-webapp/pull/535

PR #550, opened 2025-07-09
Feat/prsd 1283 update user compliance certificate files

AEPR:
## Ticket number
PRSD-1283

## Goal of change
When a compliance certificate virus scan is processed after the compliance record is created, then processing should add it to the record.

When the virus scan processing happens for NO_THREATS_FOUND, the service checks whether the appropriate record exists and if it does it sets the appropriate compliance certificate on the record to match the newly safe file.

## Checklist

### DIFF
--- a/src/main/resources/data-local.sql
+++ b/src/main/resources/data-local.sql
@@ -293,7 +293,7 @@ VALUES (1, true, 0, 1, 1, 2, 6, 1, 1, '01/15/25', '02/02/25', null, 2),
        (5, true, 0, 1, 1, 2, 37, 1, 5, '01/15/25', '01/15/25', null, null),
        (6, false, 0, 1, 1, 2, 38, 1, 6, '01/15/25', '01/15/25', null, null),
        (7, true, 0, 1, 0, 0, 39, 1, 7, '02/02/25', '02/02/25', 1, 4),
-       (8, true, 0, 1, 0, 0, 40, 1, 8, '05/02/25', '01/15/25', null, null),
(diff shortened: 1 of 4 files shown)


src/main/resources/data-local.sql
```diff
@@ -328,4 +328,9 @@ INSERT INTO system_operator (id, created_date, last_modified_date, subject_ident
...
        (2,'2025-05-01 12:01:07.575927+00',null,'urn:fdc:gov.uk:2022:GzFopg--2AyE6XtssVWwQTPELVQFupHJOjpONWS2uz0');
 
-SELECT setval(pg_get_serial_sequence('system_operator', 'id'), (SELECT MAX(id) FROM system_operator));
\ No newline at end of file
+SELECT setval(pg_get_serial_sequence('system_operator', 'id'), (SELECT MAX(id) FROM system_operator));
+
+INSERT INTO property_compliance (id, property_ownership_id, created_date, last_modified_date, has_fire_safety_declaration, has_keep_property_safe_declaration, has_responsibility_to_tenants_declaration)
+VALUES (1, 8, '01/01/25', '01/01/25', true, true, true);
+
+SELECT setval(pg_get_serial_sequence('property_compliance', 'id'), (SELECT MAX(id) FROM property_compliance));
```
AEPR, line 336: This adds a minimal acceptable compliance record without uploaded file names to the local data to allow local virus scan processing to occur. Specifically, you can locally run targetting this record and add the files to it.

PR #551, opened 2025-07-09
Feat/prsd 517 expire magic links

JasminConterioSW:
## Ticket number

PRSD-517

## Goal of change

Expire magic links for LA user invitations

If an LA user invitation is more than 48 hours old, when it is used, delete the invitation and redirect the user to the invalid-link page

N/A

## Checklist

- [x] Any special release instructions (e.g. the database will need resetting) have been added as checklist items to a draft PR (merging `main` into `test`) for the next release

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/InvitationLifetimeConstants.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/InvitationLifetimeConstants.kt
@@ -1,0 +1,3 @@
+package uk.gov.communities.prsdb.webapp.constants
+
+const val LOCAL_AUTHORITY_INVITATION_LIFETIME_IN_HOURS: Int = 48
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityInvitationService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityInvitationService.kt
@@ -1,12 +1,15 @@
 package uk.gov.communities.prsdb.webapp.services
 
 import jakarta.servlet.http.HttpSession
+import kotlinx.datetime.toKotlinInstant
 import uk.gov.communities.prsdb.webapp.annotations.PrsdbWebService
 import uk.gov.communities.prsdb.webapp.constants.LA_USER_INVITATION_TOKEN
+import uk.gov.communities.prsdb.webapp.constants.LOCAL_AUTHORITY_INVITATION_LIFETIME_IN_HOURS
 import uk.gov.communities.prsdb.webapp.database.entity.LocalAuthority
 import uk.gov.communities.prsdb.webapp.database.entity.LocalAuthorityInvitation
 import uk.gov.communities.prsdb.webapp.database.repository.LocalAuthorityInvitationRepository
 import uk.gov.communities.prsdb.webapp.exceptions.TokenNotFoundException
+import uk.gov.communities.prsdb.webapp.helpers.DateTimeHelper
 import java.util.UUID
 
 @PrsdbWebService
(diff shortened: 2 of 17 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/InvitationLifetimeConstants.kt
```diff
@@ -0,0 +1,3 @@
+package uk.gov.communities.prsdb.webapp.constants
+
+const val LOCAL_AUTHORITY_INVITATION_LIFETIME_IN_HOURS: Int = 48
```
AEPR, line 3: We could consider moving this to configuration instead of a constant - I'm not sure if it would be worth it though.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityInvitationService.kt
```diff
@@ -64,4 +67,13 @@ class LocalAuthorityInvitationService(
...
 
     fun getInvitationById(id: Long): LocalAuthorityInvitation = invitationRepository.getReferenceById(id)
+
+    fun getInvitationHasExpired(invitation: LocalAuthorityInvitation): Boolean {
+        val expiresAtTime =
+            DateTimeHelper.getDateTimeNHoursFromInstant(
+                invitation.createdDate.toKotlinInstant(),
+                LOCAL_AUTHORITY_INVITATION_LIFETIME_IN_HOURS,
+            )
+        return DateTimeHelper().getCurrentDateTimeInUK() > expiresAtTime
```
AEPR, line 77: This helper method and comparison is a little clunky and is doing more than it needs to. We don't need to worry about what the UK is doing for either of these - we just want to know if one instant is more than a fixed time after the other.

You need something that will return `Clock.System.now()`, which we could add to the non-static context of the `DateTimeHelper` or just invoke directly, then you can make this whole thing something like:

```kotlin
val expiresAtInstant = 
     invitation.createdDate.toKotlinInstant()
         .plus(LOCAL_AUTHORITY_INVITATION_LIFETIME_IN_HOURS.hours)
return Clock.System.now() < expiresAtInstant
```

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityInvitationServiceTests.kt
```diff
@@ -87,20 +92,57 @@ class LocalAuthorityInvitationServiceTests {
...
         assertTrue(inviteService.tokenIsValid(testUuid.toString()))
     }
 
+    @Test
+    fun `tokenIsValid returns false if the token is in the database but has expired`() {
+        val testUuid = UUID.randomUUID()
+        val createdDate =
+            LocalDateTime(2024, 1, 1, 0, 0)
+                .toInstant(TimeZone.of("Europe/London"))
+                .toJavaInstant()
```
AEPR, line 138: This and the test above that relies on expired tokens could break for large enough values of `LOCAL_AUTHORITY_INVITATION_LIFETIME_IN_HOURS` - I think we should set the created date as `Clock.System.now().minus(LOCAL_AUTHORITY_INVITATION_LIFETIME_IN_HOURS.hours).minus(1.minutes)`

AEPR (changes requested):
I reckon we can simplify the time comparisons a little here, strip out the new `DateTimeHelper` tests and just directly test the time differences in the `LocalAuthorityInvitationServiceTests`.

Good thinking with adding pre-expired invitations to the database - that's exactly the sort of thing that would be really annoying because of having to wait 48 hours!

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/InvitationLifetimeConstants.kt
JasminConterioSW: Maybe we leave it for now, we can always revisit if it turns out to be something that is likely to change

PR #553, opened 2025-07-09
Feat/prsd 1284 handle unsuccessful virus scans

AEPR:
## Ticket number
PRSD-1284

## Goal of change
Handle failed virus scans.

When a virus scan returns a result other than "NO_THREATS_FOUND", the file should not be added to the safe bucket. For most other statuses the service deletes the dangerous file after sending an alert email to the primary landlord of that property and to the support team to allow them to track any dangerous files, or patterns within the dangerous files. In the case that the result is "ACCESS_DENIED", no email is sent and the file is not deleted. Instead the entire processing fails with an exception as this means GuardDuty has the wrong permission configuration.

## Checklist

image


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalFileDequarantiner.kt
```diff
@@ -13,4 +13,13 @@ class LocalFileDequarantiner : FileDequarantiner {
         val localFile = File(".local-uploads/$objectKey")
         return localFile.exists()
     }
+
+    override fun delete(objectKey: String): Boolean {
```
Travis-Softwire, line 17: Not sure why it feels wrong to have a plain `delete` function (probably my C++ days coming back to haunt me) but should it maybe be `deleteFile`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/VirusAlertSender.kt
```diff
@@ -0,0 +1,49 @@
...
+            absoluteUrlProvider.buildComplianceInformationUri(propertyOwnership.id),
+        )
+
+    private fun certificateDescriptionForHeading(category: FileCategory): String =
+        when (category) {
+            FileCategory.GasSafetyCert -> "gas safety certificate"
+            FileCategory.Eirc -> "Electrical Installation Condition Report (EICR)"
+        }
+
+    private fun certificateDescriptionForBody(category: FileCategory): String =
```
Travis-Softwire, line 44: As above, we shouldn't need this with the new template

src/test/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/emailModels/EmailTemplateModelsTests.kt
```diff
@@ -98,11 +112,13 @@ class EmailTemplateModelsTests {
     @MethodSource("templateList")
     fun `EmailTemplateModels hashmaps have keys that match the parameters in their markdown templates`(testData: EmailTemplateTestData) {
         // Arrange
-        var storedBody = javaClass.getResource(testData.markdownLocation)?.readText() ?: ""
-        var parameters = getParametersFromBody(storedBody).distinct()
+        val storedBody = javaClass.getResource(testData.markdownLocation)?.readText() ?: ""
+        val storedMetadata = EmailTemplateMetadata.metadataList.single { metadata -> metadata.id == testData.model.templateId.idValue }
+
+        val parameters = getParametersFromBody(storedMetadata.subject + "\n\n" + storedBody).distinct()
```
Travis-Softwire, line 118: Not sure I care that much about it, but prepending the subject but still having the function refer to 'body' doesn't feel quite right. We could make it `getParametersFromString` or `getParametersFromTemplate` and then do `(getParametersFromString(subject) + getParametersFromString(body)).distinct()` ?

Travis-Softwire (approved):
One very minor comment which you should feel to ignore :)

PR #554, opened 2025-07-10
Feat/PRSD-1033 property record compliance tab back end

Kiran-R-K:
## Ticket number

PRSD-1033

## Goal of change

Create the view models for the property compliance tab on the property record

As above and created tests for all the view models (and their variations)

## Checklist


src/test/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/PropertyComplianceViewModelTests.kt
```diff
@@ -0,0 +1,172 @@
...
+        expectedNotificationMessages: List<String>,
+    ) {
+        val result = PropertyComplianceViewModel(propertyCompliance)
+
+        assertEquals(result.notificationMessages, expectedNotificationMessages)
+    }
+
+    companion object {
+        @JvmStatic
+        private fun providePropertyComplianceAndExpectedMessages() =
```
Travis-Softwire, line 25: Where we're just listing the inputs like this (vs combinations of inputs as with the other tests) I think it's clearer just to have separate tests for each rather than a parameterised test

Travis-Softwire (approved):
Couple of minor points, but otherwise LGTM

PR #557, opened 2025-07-14
PRSD-NONE: Absolute url provider uses explicit base url

AEPR:
## Ticket number
PRSD-NONE

## Goal of change
Email urls have configurable base urls and drop the "landlord" and "local-authority" first path segment.

## Checklist

This will currently break test when deployed there by making all emails sent from the test environment link to integration. This can be easily fixed with an infra PR to create the env vars holding the two domains.


src/main/resources/application.yml
```diff
@@ -97,6 +97,9 @@ spring:
           epc-register:
             token-uri: ${EPC_REGISTER_TOKEN_URI}
 
+base-url:
+  landlord: ${LANDLORD_BASE_URL:https://integration.register-home-to-rent.test.communities.gov.uk}
```
Travis-Softwire, line 101: I'd prefer this to fail to deploy if we haven't set the environment variable rather than running the risk of sending emails pointing at integration from production.

Travis-Softwire (approved):
One comment about referencing integration in the production config - which I think we should avoid even as a fallback. Happy to chat about it though - and good to merge once that's sorted :)

PR #547, opened 2025-07-04
PRSD-1244: Sets up Property Compliance Update Journey

isobel-softwire:
## Ticket number

PRSD-1244

## Goal of change

Sets up property compliance update journey

## Description of main change(s)

- Creates skeleton compliance update journey
- Creates compliance update journey factory

## Anything you'd like to highlight to the reviewer?

TODO:
- Add (and test) endpoints for compliance updates to `PropertyDetailsController`. You can use the endpoints in `PropertyComplianceController` as a guide.

## Checklist

Delete any that are not applicable, and add explanation below for any that are applicable but haven't been done

- [ ] Screenshots of any UI changes have been added
- [ ] Unit tests for new logic (e.g. new service methods) have been added
- [ ] Controller tests for any new endpoints, including testing the relevant permissions
- [ ] Single page integration tests have been added for any unhappy-flow UI features, e.g. validation errors
- [ ] New journey steps have been added to the appropriate journey integration test(s)
- [ ] A new journey integration test has been added for any new journeys
- [ ] New email templates have been added to `/src/main/kotlin/resources/emails/emailTemplates.json`
- [ ] Test suite has been run in full locally and is passing
- [ ] Branch has been rebased onto main and run locally, with everything working as expected (both for your new feature
  and any related functionality)
- [ ] Seed data has been updated as needed for your feature to be tested without having to e.g. register a new property
- [ ] Any special release instructions (e.g. the database will need resetting) have been added as checklist items to a draft PR (merging `main` into `test`) for the next release

PR #559, opened 2025-07-14
Feat/prsd 1244 set up compliance update journey

JasminConterioSW:
## Ticket number

PRSD-1244

## Goal of change

Building on https://github.com/communitiesuk/prsdb-webapp/pull/547, set up the `PropertyComplianceUpdateJourney `using common steps with the `PropertyComplianceJourney `where possible

- Refactors `PropertyComplianceJourney `to take steps that are shared with `PropertyComplianceUpdateJourney `from a new step factory
- Commonises the `handleSubmitAndRedirect `methods between `PropertyComplianceJourney `and `PropertyComplianceUpdateJourney `where possible
- Adds placeholder steps for `PropertyComplianceUpdateJourney `which are not shared with `PropertyComplianceJourney `
- Adds endpoints to the `PropertyComplianceController` for PropertyComplianceUpdateJourney  (commonising behaviour from the PropertyComplianceJourney endpoints)
- Adds tests for the new controller endpoints

- This does not complete the ticket (which is for adding the GasSafetyUpdate step)
- I need to add tests for the `PropertyComplianceUpdateJourney `but can do those in a separate PR

## Checklist

The following tasks will be completed in a separate PR
- Single page integration tests have been added for any unhappy-flow UI features, e.g. validation errors
- New journey steps have been added to the appropriate journey integration test(s)
- A new journey integration test has been added for any new journeys

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyComplianceJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyComplianceJourney.kt
@@ -222,7 +222,7 @@ class PropertyComplianceJourney(
 
     private val epcLookupStep
         get() =
-            propertyComplianceSharedStepFactory.createGetEpcLookupStep(
+            propertyComplianceSharedStepFactory.createEpcLookupStep(
                 handleSubmitAndRedirect = { filteredJourneyData ->
                     epcLookupStepHandleSubmitAndRedirect(filteredJourneyData)
                 },
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyComplianceUpdateJourney.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/PropertyComplianceUpdateJourney.kt
@@ -118,7 +118,7 @@ class PropertyComplianceUpdateJourney(
 
     private val epcLookupStep
         get() =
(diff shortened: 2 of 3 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/JourneyContextHelper.kt
```diff
@@ -0,0 +1,7 @@
+package uk.gov.communities.prsdb.webapp.helpers
+
+class JourneyContextHelper {
+    companion object {
+        fun isCheckingAnswers(checkingAnswersForStep: String?) = checkingAnswersForStep != null
```
JasminConterioSW, line 5: This gets used in a few places (controllers and journeys) so I wasn't really sure where to put it. This seemed sensible?

Travis-Softwire: As above - I'd leave this check in line rather than having this function

Travis-Softwire (approved):
A few thoughts, but otherwise LGTM (if you saw previous comments - I think I worked through why we have things the way we do. Ideally we'd remove the seemingly circular dependency but we can live with it for now)

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/JourneyContextHelper.kt
JasminConterioSW: Fair enough, I'll put it back inline

PR #564, opened 2025-07-16
Feat/prsd 1244 compliance update gas safety journey tests

JasminConterioSW:
## Ticket number

PRAS-1244 - set up compliance update journey

## Goal of change

Set up compliance update journey. Here, add integration tests for gas safety task.

- Add journey tests for the Gas Safety task (these will need update as part of PRSD-1244 and PRSD-1245)
- Use generic submit button text on the last page of a update task

## Checklist

- [X] A new journey integration test has been added for any new journeys


AEPR (approved):
This all looks good to me - I don't know if there's anywhere that we'd want to use the new single page tests over full integration/journey tests, but for all the cases covered it's probably worth covering the whole journey (as you have!)

PR #575, opened 2025-07-21
PRSD-NONE: Add default bucket names to local config

AEPR:
## Ticket number
PRSD-NONE

## Goal of change
Fix local development without adding bucket names to `.env`

## Checklist


JasminConterioSW (approved):
LGTM :)

PR #570, opened 2025-07-17
Chore/prsd none compliance update eicr journey tests

JasminConterioSW:
## Ticket number

PRSD-NONE - Setup Property Compliance Update Journey

## Goal of change

PRSD-NONE - Setup Property Compliance Update Journey

## Description of main change(s)

Add journey tests (with some placeholders steps) for the EICR update task.
These should cover all the steps that are shared with the PropertyComplianceJourney EICR task (just making sure we can get through the task on the update journey as well)

## Checklist

Delete any that are not applicable, and add explanation below for any that are applicable but haven't been done

- [x] A new journey integration test has been added for any new journeys
- [x] Test suite has been run in full locally and is passing
- [x] Branch has been rebased onto main and run locally, with everything working as expected (both for your new feature
  and any related functionality)

PR #576, opened 2025-07-21
PRSD-1245: Fix compliance update original journey data

AEPR:
## Ticket number
PRSD-1244

## Goal of change
Adds in compliance data that actually works for gas safety to "unlock" EICR tests.

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/AwsS3FileDequarantiner.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/AwsS3FileDequarantiner.kt
@@ -9,13 +9,11 @@ import software.amazon.awssdk.transfer.s3.S3TransferManager
 class AwsS3FileDequarantiner(
     private val transferManager: S3TransferManager,
     private val s3Client: S3Client,
-) : FileDequarantiner {
     @Value("\${aws.s3.quarantineBucket}")
-    lateinit var quarantineBucketName: String
-
+    val quarantineBucketName: String,
     @Value("\${aws.s3.safeBucket}")
-    lateinit var safeBucketName: String
-
+    val safeBucketName: String,
+) : FileDequarantiner {
     override fun dequarantineFile(objectKey: String): Boolean {
         val copyResponse =
             transferManager
--- a/src/main/resources/application-local.yml
+++ b/src/main/resources/application-local.yml
@@ -66,6 +66,11 @@ local:
 base-url:
   landlord: http://localhost:8080/landlord
   local-authority: http://localhost:8080/local-authority
+
+aws:
+  s3:
+    quarantineBucket: ${AWS_QUARANTINE_BUCKET:prsdb-quarantine-integration}
+    safeBucket: ${S3_SAFE_BUCKET_KEY:prsdb-uploaded-files-integration}
 ---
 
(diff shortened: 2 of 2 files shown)


JasminConterioSW (commented):
I haven't checked the tests carefully as they have been reviewed and approved in [this PR](https://github.com/communitiesuk/prsdb-webapp/pull/570).

This is what I think this does:

- If the property compliance has a gas certificate, this will set up the journey as though the user has added a new gas certificate (making the eicr and epc task reachable with the existing compliance data for gas safety).
- Similarly if the property has a gas exemption, this will set up the journey as though the user has added a new gas exemption (making the eicr and epc task reachable with the existing compliance data for gas safety).
- If there is no certificate and no exemption, it treats as though the user has added a dummy "Other" reason.

When we save the updated compliance data to the database, do we need to make sure we are only updating the task we have just been working on? So we don't change e.g. the gas safety exemption reason to "Other" "No reason provided" or are we already doing that?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/GasSafetyExemptionOtherReasonFormModel.kt
```diff
@@ -27,10 +27,8 @@ class GasSafetyExemptionOtherReasonFormModel : FormModel {
 
     companion object {
         fun fromComplianceRecordOrNull(record: PropertyCompliance): GasSafetyExemptionOtherReasonFormModel? =
-            record.gasSafetyCertExemptionOtherReason?.let {
-                GasSafetyExemptionOtherReasonFormModel().apply {
-                    this.otherReason = it
-                }
+            GasSafetyExemptionOtherReasonFormModel().apply {
+                this.otherReason = record.gasSafetyCertExemptionOtherReason ?: "No reason provided"
```
Travis-Softwire, line 31: It feels a bit brittle to use 'Other' and then a special 'No reason provided' string to me - what if a user entered that exact reason in the free text field? I think I'd prefer adding a new reason to the `GasSatefyExemptionReason` enum of `NO EXEMPTION` or similar, which then gets mapped to null when saving to the database, but can be sensibly displayed on the Gas Safety CYA page if the user skips directly to it.

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/factories/PropertyComplianceUpdateJourneyFactory.kt
```diff
@@ -32,7 +32,9 @@ class PropertyComplianceUpdateJourneyFactory(
         checkingAnswersForStep = checkingAnswersFor,
     )
 
-    private fun getJourneyDataKey(propertyOwnershipId: Long) =
-        PropertyComplianceController
-            .getUpdatePropertyComplianceBasePath(propertyOwnershipId)
+    companion object {
+        fun getJourneyDataKey(propertyOwnershipId: Long) =
+            PropertyComplianceController
+                .getUpdatePropertyComplianceBasePath(propertyOwnershipId)
```
isobel-softwire, line 38: Should the journey data key include the step group ID, like in the property update journey?

PR #583, opened 2025-07-23
PRSD-NONE: Never add original data to submitted journey data

AEPR:
## Ticket number
PRSD-NONE

## Goal of change
Ensure that we do not add original journey data to the submitable data while still ensuring that we handle filtered data and do not consider the data from branches that the use has filled in but not taken.


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/journeys/UpdateJourney.kt
```diff
@@ -62,4 +62,19 @@ abstract class UpdateJourney<T : StepId>(
...
+        val iterableJourney =
+            object : Iterable<StepDetails<T>> {
+                override fun iterator() = ReachableStepDetailsIterator(combinedJourneyData, steps, initialStepId, validator)
+            }
+
+        val lastStep = iterableJourney.last()
+        val subPageData = JourneyDataHelper.getPageData(combinedJourneyData, lastStep.step.name, lastStep.subPageNumber)
+        val bindingResult = lastStep.step.page.bindDataToFormModel(validator, subPageData)
+        return subPageData != null && lastStep.step.isSatisfied(bindingResult)
+    }
```
AEPR, line 79: An alternative validation would be to check the result of `createOriginalJourneyData` matches the stored original journey data to check no update to the database has happened since the journey started.

JasminConterioSW: This looks like it is checking that originalJourneyData + submittedData is valid, but could we end up with something where that combinedJourneyData is valid (so it passes this check) but not what the user is expecting (because it could include data from multiple updates)?

AEPR:
@JasminConterioSW - this disables a number of EPC and EICR tests because there are several tests where on main the partial compliance journey tries to make what is now an illegal update.

src/test/kotlin/uk/gov/communities/prsdb/webapp/testHelpers/builders/JourneyDataBuilder.kt
```diff
@@ -423,18 +423,46 @@ class JourneyDataBuilder(
...
+        return this
+    }
+
     fun withIsOccupiedUpdate(isOccupied: Boolean): JourneyDataBuilder {
         journeyData[UpdatePropertyDetailsStepId.UpdateOccupancy.urlPathSegment] =
             mutableMapOf("occupied" to isOccupied)
         return this
     }
 
+    fun withIsOccupiedNumberOfHouseholdsUpdate(households: Int = 1): JourneyDataBuilder {
```
Travis-Softwire, line 442: Should this and the one below not just delegate to the versions that don't take the default arg?

Travis-Softwire (approved):
One minor point but otherwise LGTM - nice work making it understandable :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/testHelpers/builders/JourneyDataBuilder.kt
AEPR: Unless I've missed them, there aren't equivalent functions that dont take a default arg - the two below are for the Households update journey whereas these are for the Occupation update journey.

I'm going to resolve and merge given that you've approved, so please grab me if I've mis-understood and I'll throw up a mini-PR with a fix.

AEPR:
>nice work making it understandable 

Thank you!

src/test/kotlin/uk/gov/communities/prsdb/webapp/testHelpers/builders/JourneyDataBuilder.kt
Travis-Softwire: Ah you're quite right, I missed the different journey data key

PR #611, opened 2025-08-01
PRSD-1356: Fire Safety - switch to checkbox

isobel-softwire:
## Ticket number

PRSD-1356

## Goal of change

Switch to using checkboxes on the fire safety declaration page

- Updates the fire safety declaration model, template and step to use checkboxes instead of radios
- Deletes the fire safety risk step and template
- Gives the fire safety row of the property compliance CYA page a 'View' link instead of a 'Change' one
- Makes having a fire safety declaration always true in property compliance records

N/A

## Checklist

## Screenshots
- Fire safety declaration page
image

- Fire safety declaration validation message
image

- Landlord responsibilities section of property compliance CYA page
image


AEPR (approved):
I've added a point to the dev catch up thread that I'm not sure whether it's ok for us to set all declaration values to true in the database without it being drawn directly from the journey data (even though we logically know it is true). Either way I think that's beyond the scope of this PR, which all looks good to me!

PR #614, opened 2025-08-04
Feat/prsd 1352 properly track upload status updates

AEPR:
## Ticket number
PRSD-1352

## Goal of change
Track file uploads in certificate upoads and create a certificate upload on submitting a certificate form step.

When a file is uploaded in the compliance journey it is saved in the database as a `FileUpload` immediately when the upload to S3 occurs, then when the step succeeds that is attached to a `CertificateUpload` which links to the ownership. When the virus scan is complete, the processor uses the file details to find the certificate upload and uses that to update the status in the database.

This means that you can save the compliance record independently of the virus scanning - you just have a `FileUpload` in some state that will change when the file is scanned.

This has got a little out of hand. I think it's now done, but feel free to grab me for a chat if you think it would be helpful to understand the PR.

## Checklist

- [ ] Any special release instructions (e.g. the database will need resetting) have been added as checklist items to a draft PR (merging `main` into `test`) for the next release


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/VirusScanProcessingService.kt
```diff
@@ -1,67 +1,91 @@
...
     fun processScan(
-        fileNameInfo: PropertyFileNameInfo,
+        locator: UploadedFileLocator,
+        scanResultStatus: ScanResult,
+    ) {
+        val certificateUpload = getCertificateUpload(locator)
+
+        if (certificateUpload != null) {
+            processCertificateScanResult(certificateUpload, scanResultStatus)
+        } else {
```
Travis-Softwire, line 28: I think it's probably fine in practice, certainly for pilot, but I think there are some implicit race conditions here if the scan + the processing of the result happens before we can create the DB records. Only scenario I can think of is if RDS was hanging for a few seconds.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/VirusScanProcessingService.kt
```diff
@@ -1,67 +1,91 @@
...
         scanResultStatus: ScanResult,
     ) {
-        val ownership = getOwnershipForFileInfo(fileNameInfo)
         when (scanResultStatus) {
             ScanResult.NoThreats -> {
-                if (dequarantiner.dequarantineFile(fileNameInfo.toString())) {
-                    addFileToComplianceRecordIfPresent(fileNameInfo, ownership)
-                } else {
-                    throw PrsdbWebException("Failed to dequarantine file: $fileNameInfo")
+                // TODO PRSD-976 - We might need to dequarantine to a more contextual object key if that is visible to users
```
Travis-Softwire, line 39: Not sure I follow this comment?

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/UploadDequarantinerTests.kt
```diff
@@ -0,0 +1,176 @@
...
+        // Act
+        val result = uploadDequarantiner.dequarantineFile(fileUpload)
+
+        // Assert
+        verify(fileUploadRepository, never()).save(any())
+        assertFalse(result)
+        assertEquals(originalETag, fileUpload.eTag)
+    }
+
+    @Suppress("ktlint:standard:max-line-length")
```
Travis-Softwire, line 127: Is this left over from when the name of this test was longer?

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/UploadServiceTests.kt
```diff
@@ -0,0 +1,72 @@
...
+    private lateinit var mockUploader: FileUploader
+
+    @Mock
+    private lateinit var mockRepository: FileUploadRepository
+
+    @InjectMocks
+    private lateinit var uploadService: UploadService
+
+    @Test
+    fun `uploadFile uploads the input stream to a file and saves the result as quarantined`() {
```
Travis-Softwire, line 33: Minor, but given this name I'd expect to see an assertion on the repository regarding what's actually been saved?

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/VirusScanProcessingServiceTests.kt
```diff
@@ -1,101 +1,88 @@
...
+        whenever(dequarantiner.dequarantineFile(any())).thenReturn(false)
+        whenever(certificateUploadRepository.findByFileUpload_ObjectKeyAndFileUpload_VersionId(any(), any()))
+            .thenReturn(CertificateUpload(fileUpload, mock(), mock()))
 
         // Act & Assert
-        assertThrows<PrsdbWebException> { virusScanProcessingService.processScan(fileNameInfo, scanResultStatus) }
+        assertThrows<PrsdbWebException> { virusScanProcessingService.processScan(locator, scanResultStatus) }
     }
 
     @EnumSource(ScanResult::class)
```
Travis-Softwire, line 88: Test name doesn't seem quite since we're not asserting on an error being thrown, we're checking that the fileUpload has been deleted?

Travis-Softwire (approved):
A few more thoughts on the tests but all pretty minor. Otherwise LGTM subject to the points we discussed earlier :)

PR #615, opened 2025-08-04
Feat/prsd 1392 change mees exemption loading skipped step data

JasminConterioSW:
## Ticket number

PRSD-1392

## Goal of change

Allow the user to edit their MEES exemption (without needing to go through the earlier EPC steps)

Followed the approach from updating occupancy data on Property Details update.
- Add a duplicate set of EPC steps corresponding to the MEES-only update journey
- Update the step factory so that it can generate steps with either set of EPC steps
- Pre-populate the EPC steps for the MEES-only update journey
- Link to the MEES exemption check page from Property details
- Add a back link from the MEES-only MEES exemption check page to Property Details

I'll add tests in a later PR, to include:
- JourneyDataExtensions tests
- OriginalJourneyDataTests when jumping straight to MEES update
- Integration (Journey) tests
    - including checking the back link on the MEES exemption check page
    - probably updating the current tests to check that the record has been updated

## Checklist

(Tests to follow in a second PR)


src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/journeyExtensions/PropertyComplianceJourneyDataExtensions.kt
```diff
@@ -353,22 +366,26 @@ class PropertyComplianceJourneyDataExtensions : JourneyDataExtensions() {
 
         fun JourneyData.getHasCompletedEpcNotFound() = this.containsKey(PropertyComplianceStepId.EpcNotFound.urlPathSegment)
```
JasminConterioSW, line 359: For `getHasCompletedEpcNotFound()`, `getHasCompletedEpcMissing()`, `getHasCompletedEpcExemptionConfirmation()` etc don't check the "UpdateMees..." copies of steps.

I think this is ok because we only edit the MEES exemption (so we only access this journey) when there is a valid (not expired) EPC there.

PR #620, opened 2025-08-05
PRSD-1410: invalid token page

Kiran-R-K:
## Ticket number

PRSD-1410

## Goal of change

Make sure user can reach Invalid Link error page when using an Invalid Token

I changed the logic in the controller to check for a valid token (without checking for expiry) first and then to have the expiry check happen separately.

Added to tests to catch this bug if it reoccurs.

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
@@ -43,16 +43,15 @@ class RegisterLAUserController(
         // see https://github.com/spring-projects/spring-hateoas/issues/155 for details
         val invitation = invitationService.getInvitationOrNull(token)
 
-        invitation?.let {
-            if (!invitationService.getInvitationHasExpired(it)) {
-                invitationService.storeTokenInSession(token)
-                return "redirect:${LA_USER_REGISTRATION_ROUTE}/${RegisterLaUserStepId.LandingPage.urlPathSegment}"
-            } else {
-                invitationService.deleteInvitation(it)
-            }
+        return if (invitation == null) {
+            "redirect:$LA_USER_REGISTRATION_INVALID_LINK_ROUTE"
+        } else if (invitationService.getInvitationHasExpired(invitation)) {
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLAUserController.kt
```diff
@@ -41,14 +41,15 @@ class RegisterLAUserController(
...
-            invitationService.storeTokenInSession(token)
-            return "redirect:${LA_USER_REGISTRATION_ROUTE}/${RegisterLaUserStepId.LandingPage.urlPathSegment}"
-        }
-
-        val invitation = invitationService.getInvitationFromToken(token)
-        if (invitationService.getInvitationHasExpired(invitation)) {
-            invitationService.deleteInvitation(invitation)
+        val invitation = invitationService.getInvitationOrNull(token)
+
+        invitation?.let {
```
Kiran-R-K, line 46: I'm not sure if this would be clearer as an if statement?

Travis-Softwire: I think this is fine, and it avoids needing a nested if statement

Travis-Softwire (approved):
Once the small refactor we discussed is done this LGTM

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityInvitationServiceTests.kt
```diff
@@ -96,6 +97,27 @@ class LocalAuthorityInvitationServiceTests {
...
+        whenever(mockLaInviteRepository.findByToken(testUuid))
+            .thenReturn(testInvitation)
+
+        val invitation = inviteService.getInvitationOrNull(testUuid.toString())
+
+        assertEquals(invitation, testInvitation)
+    }
+
+    @Test
+    fun `getInvitationOrNull returns null if hee token is not in the database`() {
```
isobel-softwire, line 113: Should this be '...if the token...'

isobel-softwire (approved):
LGTM as well :)

PR #630, opened 2025-08-07
PRSD-NONE: Explicitly determines certificate expiry status

isobel-softwire:
## Ticket number

PRSD-NONE

## Goal of change

Determines certificate expiry status more explicitly 

- Determines cert expiry by checking dates rather than whether expiry confirmation pages have been visited

- This PR only affects the property compliance CYA pages. Something similar will be applied to the incomplete compliance page as part of PRSD-1394

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/cya/GasSafetySummaryRowsFactory.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/cya/GasSafetySummaryRowsFactory.kt
@@ -4,6 +4,7 @@ import kotlinx.datetime.DatePeriod
 import kotlinx.datetime.plus
 import uk.gov.communities.prsdb.webapp.constants.GAS_SAFETY_CERT_VALIDITY_YEARS
 import uk.gov.communities.prsdb.webapp.constants.enums.GasSafetyExemptionReason
+import uk.gov.communities.prsdb.webapp.exceptions.PrsdbWebException
 import uk.gov.communities.prsdb.webapp.forms.JourneyData
 import uk.gov.communities.prsdb.webapp.forms.steps.PropertyComplianceStepId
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/cya/GasSafetySummaryRowsFactory.kt
```diff
@@ -33,18 +35,20 @@ class GasSafetySummaryRowsFactory(
...
-            if (filteredJourneyData.getHasCompletedGasSafetyUploadConfirmation()) {
-                "forms.checkComplianceAnswers.gasSafety.download"
+            if (filteredJourneyData.getHasCompletedGasSafetyExemptionMissing()) {
+                "forms.checkComplianceAnswers.certificate.notAdded"
             } else if (filteredJourneyData.getHasCompletedGasSafetyExemptionConfirmation()) {
                 "forms.checkComplianceAnswers.certificate.notRequired"
-            } else if (filteredJourneyData.getHasCompletedGasSafetyExemptionMissing()) {
-                "forms.checkComplianceAnswers.certificate.notAdded"
-            } else {
+            } else if (filteredJourneyData.getIsGasSafetyCertOutdated() == true) {
```
Travis-Softwire, line 42: Looking at `filteredJourneyData.getIsGasSafetyCertOutdated()`, although it returns a nullable boolean we never use the case when it's null - we either do a not-null assertion on the result or compare it to `true`. Given that, should it just return a non-nullable Boolean instead?

Travis-Softwire (approved):
One small thought from me, but otherwise good to go :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/cya/GasSafetySummaryRowsFactory.kt
isobel-softwire: There seem to be a number of extension methods like this, so I'll deal with them all in a separate PR 👍

PR #639, opened 2025-08-11
PRSD-976: Enforce single confirmation for gas and eicr

AEPR:
## Ticket number
PRSD-976

## Goal of change
Simplify CYA page logic by enforcing that only a single confirmation page can be in the filtered journey data.

This code path can already not be reached normally - filtering should enforce that only one path is possible. This primarliy makes it clearer in the code for anyone working on the CYA page that this is not a risk (and that if it is something else is broken!)

This should be easier to read and means that the logic is not determined by if statement order.

## Checklist


src/test/kotlin/uk/gov/communities/prsdb/webapp/forms/pages/PropertyComplianceCheckAnswersPageTests.kt
```diff
@@ -348,7 +352,9 @@ class PropertyComplianceCheckAnswersPageTests {
     }
 
     @Test
-    fun `the correct summary rows appear when certificate expiry dates have passed since being provided`() {
+    // Commonised set up code mocks functions that are not called in this test
+    @MockitoSettings(strictness = Strictness.LENIENT)
```
isobel-softwire, line 356: Is this still needed if `expectEpcUrl` is set to false?

isobel-softwire (approved):
This LGTM. Might be worth doing something similar for EPCs as well

AEPR:
I'm not going to address EPCs in this PR as the logic there is a little more complicated* and it doesn't impact the download work that's coming up.

\* There isn't a specific confirmation page for "matched EPC which is all good" - you confirm that the EPC is correct and then if there's no problems you move on immediately whereas if it's out of date or not a good enough rating there are separate confirmations. This means that there is a legitimate `else` in the if statements there that mean it doesn't map onto the private enum model quite as smoothly.

PR #641, opened 2025-08-11
PRSD-NONE: Make all properties in proxiable entities open

AEPR:
## Ticket number
PRSD-NONE

## Goal of change
Resolves warning 
```
WARN 23000 --- [prsdb-webapp] [  restartedMain] EntityRepresentationStrategyPojoStandard : HHH000305: Could not create proxy factory for:uk.gov.communities.prsdb.webapp.database.entity.CertificateUpload
```

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/CertificateUpload.kt
```diff
@@ -13,19 +13,19 @@ import uk.gov.communities.prsdb.webapp.constants.enums.FileCategory
 open class CertificateUpload() : ModifiableAuditableEntity() {
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
-    val id: Long = 0
+    open val id: Long = 0
```
isobel-softwire, line 16: Why is `CertificateUpload` an open class?

isobel-softwire (approved):
LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/CertificateUpload.kt
Travis-Softwire: As per slack - it looks like this is because the class is open (it was our only non-abstract open db entity class)

Travis-Softwire (approved):
👍

PR #649, opened 2025-08-12
Feat/prsd 1183 analytics environment variables

JasminConterioSW:
## Ticket number

PRSD-1183

## Goal of change

Update to google analytics, and passing in ids as environment variables

- Edit the cookie deletion so that it works on the deployed environment
- Update cookie consent to only allow `analytics_storage`
- Add some JS tests
- Pass in domain ids as environment variables so we can set them different for different environments if required (e.g. for prod)

## Checklist

  and any related functionality) (done but behaviour may be different on deployed environments!)
- [x] Any special release instructions (e.g. the database will need resetting) have been added as checklist items to a draft PR (merging `main` into `test`) for the next release

### DIFF
--- a/src/main/resources/templates/fragments/layout.html
+++ b/src/main/resources/templates/fragments/layout.html
@@ -40,14 +40,12 @@
 </script>
 <!--/*@thymesVar id="googleAnalyticsMeasurementId" type="String"*/-->
 <script async th:src="'https://www.googletagmanager.com/gtag/js?id='+${googleAnalyticsMeasurementId}"></script>
-<script th:inline="javascript"
-        th:with="measurementIdAsString=${googleAnalyticsMeasurementId}, googleAnalyticsCookieDomainString=${googleAnalyticsCookieDomain}"
->
+<script th:inline="javascript" th:with="measurementIdAsString=${googleAnalyticsMeasurementId}">
     <!-- Google tag (gtag.js) -->
     window.dataLayer = window.dataLayer || [];
     function gtag() { dataLayer.push(arguments); }
     gtag('js', new Date());
-    gtag('config', [[${measurementIdAsString}]], {'cookie_domain': [[${googleAnalyticsCookieDomainString}]]})
+    gtag('config', [[${measurementIdAsString}]])
 </script>
 <script th:inline="javascript">
     // Add environment variables to the window so they can be accessed in js scripts


src/main/js/cookieConsentHandler.js
```diff
@@ -21,19 +21,17 @@ function updateGaConsent(isGranted = false) {
...
 
     if (!isGranted) {
         deleteCookie("_ga")
-        deleteCookie("_ga_PDPW9SQ94W")
+        deleteCookie("_ga_" + window.GOOGLE_ANALYTICS_MEASUREMENT_ID.slice(2))
     }
 }
 
 function deleteCookie(name) {
+    document.cookie = name + '=; expires=Thu, 01 Jan 1970 00:00:00 UTC; path=/; domain=' + window.GOOGLE_COOKIE_DOMAIN + ';';
```
JasminConterioSW, line 34: Deleting cookies only works on integration if the communities.gov.uk domain is specified here, otherwise it just creates (and expires) a new cookie with the domain integration.register-home-to-rent.test.communities.gov.uk. This is set by GA, I don't think we have much control over it!

For tests, it can't find the version with the domain specified, so both are included here.

src/main/resources/templates/fragments/layout.html
```diff
@@ -37,13 +38,19 @@
...
-<script>
+<!--/*@thymesVar id="googleAnalyticsMeasurementId" type="String"*/-->
+<script async th:src="'https://www.googletagmanager.com/gtag/js?id='+${googleAnalyticsMeasurementId}"></script>
+<script th:inline="javascript" th:with="measurementIdAsString=${googleAnalyticsMeasurementId}">
     <!-- Google tag (gtag.js) -->
     window.dataLayer = window.dataLayer || [];
     function gtag() { dataLayer.push(arguments); }
     gtag('js', new Date());
-    gtag('config', 'G-PDPW9SQ94W');
+    gtag('config', [[${measurementIdAsString}]])
```
JasminConterioSW, line 48: `[[${googleAnalyticsCookieDomain}]]` doesn't work here because `G-PDPW9SQ94W` needs to be wrapped in quotes for the config analytic to fire (without it GA doesn't add the cookies)

src/main/resources/templates/fragments/layout.html
```diff
@@ -37,13 +38,21 @@
...
+<script async th:src="'https://www.googletagmanager.com/gtag/js?id='+${googleAnalyticsMeasurementId}"></script>
+<script th:inline="javascript"
+        th:with="measurementIdAsString=${googleAnalyticsMeasurementId}, googleAnalyticsCookieDomainString=${googleAnalyticsCookieDomain}"
+>
     <!-- Google tag (gtag.js) -->
     window.dataLayer = window.dataLayer || [];
     function gtag() { dataLayer.push(arguments); }
     gtag('js', new Date());
-    gtag('config', 'G-PDPW9SQ94W');
+    gtag('config', [[${measurementIdAsString}]], {'cookie_domain': [[${googleAnalyticsCookieDomainString}]]})
```
JasminConterioSW, line 50: `[[${googleAnalyticsCookieDomain}]]` doesn't work here because `G-PDPW9SQ94W` needs to be wrapped in quotes for the config analytic to fire (without it GA doesn't add the cookies)

I'm setting the cookie domain here because we need to know what it is in order to delete the cookie if the user later rejects them (it wasn't deleting the cookie if the domain didn't match the current domain - and the cookie domain by default is just communities.gov.uk).

`googleAnalyticsCookieDomainString` will be set in a environment variable to `communities.gov.uk`

src/test/js/cookieConsentHandler.test.js
```diff
@@ -27,6 +27,10 @@ function setup(onCookiesPage = false) {
     teardown = jsdom(html, {
         url: onCookiesPage ? 'https://example.com/cookies' : 'https://example.com'
     });
+
+    window.GOOGLE_ANALYTICS_MEASUREMENT_ID = 'G-PDPW9SQ94W'
```
Travis-Softwire, line 31: Are we having to use the real (integration) values here because we're actually calling out to GA? If so, does that mean if GA was down for some reason we couldn't deploy our code because the tests couldn't pass?

Travis-Softwire (approved):
A question about whether we are using our 'real' GA account in the tests. If we are it still fine w/ getting it merged for now but we should consider if we can't instead use some mocks/ stubs to isolate our code from GA (and any requirement to make real requests over the internet from our tests)

src/main/resources/templates/fragments/layout.html
JasminConterioSW: I think we should still use the environment variable to specify the domain of the cookie to be deleted (see above), but I can take it out of this config so it just uses the default.

PR #653, opened 2025-08-12
PRSD-1386-beta feedback pages 

Bill-Haigh:
## Ticket number

PRSD-1386

## Goal of change

Creating the feedback pages for the beta including a success page that can be navigated to from both local-authority and landlord routes.

- Added html pages for /feedback and /feedback/success
- created controller and routing
- updated styling

- There are a few elements that feel a bit wonky in terms of my implementation.

## Checklist

image

image


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/controllerAdvice/GlobalModelAttributes.kt
```diff
@@ -11,4 +12,14 @@ class GlobalModelAttributes {
     fun addGlobalModelAttributes(model: Model) {
         model.addAttribute("cookiesUrl", COOKIES_ROUTE)
     }
+
+    @ModelAttribute("feedbackBannerUrl")
+    fun addFeedbackBannerUrl(request: HttpServletRequest): String {
+        val path = request.requestURI
+        return when {
```
Travis-Softwire, line 19: We don't actually need to include `/local-authority` and `/landlord` here at all - we have some code that inserts these into requests based on the host domain, so `_.register-as-a-landlord.whatever.uk/*` will be forwarded to `_.register-as-a-landlord.whatever.uk/landlord/*` and similarly for `search-for-landlord-information`

I think that means you can simplify this to just adding in a constant :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/BetaFeedbackController.kt
```diff
@@ -0,0 +1,108 @@
...
+import uk.gov.communities.prsdb.webapp.constants.FEEDBACK_PATH_SEGMENT
+import uk.gov.communities.prsdb.webapp.constants.LANDLORD_PATH_SEGMENT
+import uk.gov.communities.prsdb.webapp.constants.LOCAL_AUTHORITY_PATH_SEGMENT
+import uk.gov.communities.prsdb.webapp.constants.SUCCESS_PATH_SEGMENT
+import uk.gov.communities.prsdb.webapp.models.viewModels.formModels.BetaFeedbackModel
+
+@PrsdbController
+@RequestMapping
+class BetaFeedbackController {
+    @GetMapping("/${LANDLORD_PATH_SEGMENT}/${FEEDBACK_PATH_SEGMENT}")
```
Travis-Softwire, line 23: I'd suggest delegating both feedback endpoints to a common private method, and the same with the success endopints. That will let us check authorisation on the local authority versions of the endpoints without the code duplication :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/BetaFeedbackController.kt
```diff
@@ -0,0 +1,108 @@
...
+    ): String {
+        val formModel = BetaFeedbackModel()
+        model.addAttribute("betaFeedbackModel", formModel)
+        val referrer = request.getHeader("referer")
+        model.addAttribute("referrerHeader", referrer)
+        model.addAttribute("backUrl", referrer)
+        return "betaBannerFeedback"
+    }
+
+    @GetMapping("/${LANDLORD_PATH_SEGMENT}/${FEEDBACK_PATH_SEGMENT}/${SUCCESS_PATH_SEGMENT}")
```
Travis-Softwire, line 36: And the same with these - they can be delegated to a common private function

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/BetaFeedbackController.kt
```diff
@@ -0,0 +1,108 @@
...
+        if (bindingResult.hasErrors()) {
+            val referrer = request.getHeader("referer")
+            model.addAttribute("referrerHeader", referrer)
+            model.addAttribute("backUrl", referrer)
+            return "betaBannerFeedback"
+        }
+        return "redirect:/${LOCAL_AUTHORITY_PATH_SEGMENT}/${FEEDBACK_PATH_SEGMENT}/${SUCCESS_PATH_SEGMENT}"
+    }
+
+    @RequestMapping("/${LANDLORD_PATH_SEGMENT}/$BETA_BANNER_PATH_SEGMENT")
```
Travis-Softwire, line 102: I don't think you need this endpoint and the one below?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/formModels/BetaFeedbackModel.kt
```diff
@@ -0,0 +1,6 @@
+package uk.gov.communities.prsdb.webapp.models.viewModels.formModels
+
+class BetaFeedbackModel(
```
Travis-Softwire, line 3: Not strictly required for this ticket, but it will be a useful start for the next one

src/main/resources/css/_betaBannerFeedback.scss
```diff
@@ -0,0 +1,9 @@
+@import "govuk/base";
```
Travis-Softwire, line 1: As discussed - we shouldn't need these. The red text is just placeholder text - we should confirm w/ Andrea what should go there. And for text that isn't semantically bold (i.e. a header) we can use a `<strong>` tag in the messages :)

src/main/resources/css/custom.scss
```diff
@@ -1,3 +1,4 @@
 @use "warnings";
 @use "keyDetails";
-@use "dashboard";
```
Travis-Softwire, line 3: As above I don't think we'll need this

src/main/resources/templates/betaBannerFeedback.html
```diff
@@ -0,0 +1,24 @@
...
+                <div id="page-content">
+                    <section>
+                        <div th:replace="~{forms/betaBannerFeedbackForm :: betaBannerFeedbackForm}"></div>
+                    </section>
+                    <section>
+                        <h2 class="govuk-heading-m" th:text="#{betaBannerFeedback.heading.two}">betaBannerFeedback.heading.two</h2>
+                        <h3 class="govuk-heading-s" th:text="#{betaBannerFeedback.heading.three}">betaBannerFeedback.heading.three</h3>
+                        <p class="govuk-body beta-feedback-red" th:text="#{betaBannerFeedback.paragraph.one}">betaBannerFeedback.paragraph.one</p>
+                        <h3 class="govuk-heading-s" th:text="#{betaBannerFeedback.heading.four}">betaBannerFeedback.heading.four</h3>
+                        <p class="govuk-body beta-feedback-red" th:text="#{betaBannerFeedback.paragraph.two}" style="margin-bottom: 0;">betaBannerFeedback.paragraph.two</p>
```
Travis-Softwire, line 17: We shouldn't need the custom inline styling, but I think this is tied together with it being placeholder text. In general it wil be very rare for us to need to do anything different to the standard GDS components styling-wise

src/main/resources/templates/betaBannerFeedback.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
+<html>
+<th:block id="main-content"
```
Travis-Softwire, line 3: I think we should be able to use our question page fragment?

src/main/resources/messages.properties
```diff
@@ -1427,3 +1427,24 @@ generatePasscode.banner.body=The passcode is
 generatePasscode.instructions=Send this passcode in your invitation email to a landlord private beta participant. Please do not generate more passcodes than you need.
 generatePasscode.generateAnother=Generate another passcode
 generatePasscode.returnToDashboard=Return to dashboard
+
+betaBannerFeedback.title=Beta feedback
+betaBannerFeedback.heading=Give feedback on the PRS Database
+betaBannerFeedback.intro=Use this form to let us know if you’ve experienced any issues, or have ideas for how we can improve the service.
```
Travis-Softwire, line 1433: The encoding of the apostrophe's may have gone awry - you probably want it to be `\u2019`

Travis-Softwire (changes requested):
This is a great start :) I've made some suggestions in the comments - quite a few of which we've already discussed, but please do shout if anything doesn't make sense!

src/main/resources/css/_betaBannerFeedback.scss
AEPR: Just to add - on this project we should generally steer away from writing custom CSS of any kind except where strictly necessary because (a) it keeps things a lot simpler and (b) means that we're more closely following the Government Design System.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/BetaFeedbackController.kt
```diff
@@ -0,0 +1,159 @@
...
+    @PreAuthorize("hasAnyRole('LA_USER', 'LA_ADMIN')")
+    @GetMapping("/${LOCAL_AUTHORITY_PATH_SEGMENT}/${FEEDBACK_PATH_SEGMENT}/${SUCCESS_PATH_SEGMENT}")
+    fun localAuthorityFeedbackSuccess(
+        model: Model,
+        request: HttpServletRequest,
+    ): String = renderFeedbackSuccess(model, request)
+
+    // TODO: These are so that the feedback/success pages can be viewed locally, will need to be removed before production
+    // worth noting that you need to manually go to /feedback/success
+    @GetMapping("/${FEEDBACK_PATH_SEGMENT}")
```
Travis-Softwire, line 102: You can include the 'local only' ones in the mapping for the landlord endpoint, i.e. `@GetMapping(value={"/${LANDLORD_PATH_SEGMENT}/${FEEDBACK_PATH_SEGMENT}", "/${FEEDBACK_PATH_SEGMENT}"})`

src/main/resources/templates/betaBannerFeedback.html
```diff
@@ -0,0 +1,17 @@
...
+    <!--/*@thymesVar id="backUrl" type="java.lang.String"*/-->
+    <!--/*@thymesVar id="feedbackUrl" type="java.lang.String"*/-->
+    <!--/*@thymesVar id="referrerHeader" type="java.lang.String"*/-->
+    <th:block th:replace="~{forms/betaBannerFeedbackForm :: betaBannerFeedbackForm}"></th:block>
+    <section>
+        <h2 class="govuk-heading-m" th:text="#{betaBannerFeedback.heading.two}">betaBannerFeedback.heading.two</h2>
+        <h3 class="govuk-heading-s" th:text="#{betaBannerFeedback.heading.three}">betaBannerFeedback.heading.three</h3>
+        <p class="govuk-body" th:text="#{betaBannerFeedback.paragraph.one}">betaBannerFeedback.paragraph.one</p>
+        <h3 class="govuk-heading-s" th:text="#{betaBannerFeedback.heading.four}">betaBannerFeedback.heading.four</h3>
+        <p class="govuk-body" th:text="#{betaBannerFeedback.paragraph.two}" style="margin-bottom: 0;">betaBannerFeedback.paragraph.two</p>
```
Travis-Softwire, line 14: I missed the inline styling here before, but we shouldn't need it - the spacing between paragraphs should be dictated by the GDS design system. If we need no spaceing between the paragraphs then they're probably the same paragraph, possibly with a `<br/>` to force a new line if required.

PR #666, opened 2025-08-15
Feat/prsd 1335 update landlord details confirmation email

AEPR:
## Ticket number
PRSD-1335

## Goal of change
Send a confirmation email when a landlord updates their details

The email is sent for whichever field is updated. If multiple fields are updated the service only sends a confirmation for the first one because the email only allows room for a single field to be described. The journey will only allow one field to actually be updated so this shouldn't currently cause any issues.

## Checklist


Travis-Softwire (approved):
👍

PR #692, opened 2025-08-22
PRSD-1423: Passcode Invalid Page

isobel-softwire:
## Ticket number

PRSD-1423

## Goal of change

Turns passcode used page into invalid passcode page, in line with designs

- Turns passcode used page into passcode invalid page, in line with designs
- Redirects to passcode invalid page in all appropriate cases
- Stops removing passcode redirect URL from session after use so that the redirect can still happen after being sent to the invalid passcode page

N/A

## Checklist

## Screenshots
image


src/main/kotlin/uk/gov/communities/prsdb/webapp/config/interceptors/PasscodeInterceptor.kt
```diff
@@ -79,14 +79,12 @@ class PasscodeInterceptor(
         userId: String,
         submittedPasscode: String,
     ): Boolean {
-        val passcode =
-            passcodeService.findPasscode(submittedPasscode)
-                ?: return redirectToPasscodeEntryAndReturnFalse(request, response)
+        val passcode = passcodeService.findPasscode(submittedPasscode) ?: return redirectToInvalidPasscodeAndReturnFalse(response)
```
Travis-Softwire, line 82: I don't think they should ever actually get a null here, as we check if the passcode is in the DB before we allow it in the session and there isn't currently a way to delete them. But no harm in coding defensively

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PasscodeEntryController.kt
```diff
@@ -68,22 +68,16 @@ class PasscodeEntryController(
 
         // Check for redirect URL in session, otherwise use landlord registration index page
         val redirectUrl = session.getAttribute(PASSCODE_REDIRECT_URL) as? String
-        return if (redirectUrl != null) {
-            // Clear the redirect URL from session after using it
-            session.removeAttribute(PASSCODE_REDIRECT_URL)
```
Travis-Softwire, line 73: To avoid this hanging around, should we instead clear it on any of the successful paths in the interceptor?

Travis-Softwire (approved):
A couple of suggestions, but happy for this to be merged once they've been addressed :)

PR #719, opened 2025-08-28
PRSD-1362: changes requested as per comments

Bill-Haigh:
## Ticket number

PRSD-1362

minor changes to content for landlord start page as requested.


Travis-Softwire:
Super super super minor as it has v. little impact (and definitely don't worry about it for this PR!) but we'd normally use a `fix/` branch for this rather than `hotfix/` which is something slightly different as we'd use that where we're following our hotfix process

Travis-Softwire (approved):
👍

PR #755, opened 2025-09-04
Feat/prsd 1342/user account deletion admin email

Bill-Haigh:
## Ticket number

PRSD-1342

## Goal of change

Send emails to all LA admins for LA area an email when a user in their area is deleted.

 - created email model
 - created template
 - updated LocalAuthorityDataService to fetch all admins and loop through sending emails

There is no facility to send to multiple addressed using notify, hence the loop. Also for now it is decided that this will not be done asynchronously rather this will be addressed in a later ticket.

* ignore the extra bracket this has been fixed but I cannot send any more emails to get an updated screenshot

image

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityDataService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityDataService.kt
@@ -146,10 +146,10 @@ class LocalAuthorityDataService(
             ),
         )
 
-        userDeletionEmailToAdmins(localAuthorityUser)
+        sendUserDeletedEmailsToAdmins(localAuthorityUser)
     }
 
-    private fun userDeletionEmailToAdmins(localAuthorityUser: LocalAuthorityUser) {
+    private fun sendUserDeletedEmailsToAdmins(localAuthorityUser: LocalAuthorityUser) {
         val localAdminsByAuthority =
             localAuthorityUserRepository.findAllByLocalAuthority_IdAndIsManagerTrue(localAuthorityUser.localAuthority.id)
 


src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/emailModels/EmailTemplateModel.kt
```diff
@@ -33,4 +33,5 @@ enum class EmailTemplate {
     UPDATE_EPC_NO_EPC_CONFIRMATION_EMAIL,
     LOCAL_COUNCIL_REGISTRATION_CONFIRMATION_EMAIL,
     LOCAL_AUTHORITY_USER_DELETION_EMAIL,
+    LOCAL_AUTHORITY_USER_DELETION_ADMIN_EMAIL,
```
Kiran-R-K, line 36: I think it's worth re-visiting this name, the general naming convention for these e-mails as there's going to be a few of them. Perhaps something like:  `LOCAL_AUTHORITY_USER_DELETION_INFORM_ADMIN_EMAIL`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityDataService.kt
```diff
@@ -135,6 +137,21 @@ class LocalAuthorityDataService(
...
+            deletionConfirmationSenderAdmin.sendEmail(
+                admin.email,
+                LocalAuthorityUserDeletionAdminEmail(
+                    councilName = localAuthorityUser.localAuthority.name,
+                    email = localAuthorityUser.email,
+                    userName = localAuthorityUser.name,
+                    prsdURL = absoluteUrlProvider.buildLocalAuthorityDashboardUri().toString(),
+                ),
+            )
+        }
```
Kiran-R-K, line 153: I think it's worth pulling this out in to it's own method for ease of reading

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityDataService.kt
```diff
@@ -135,6 +137,21 @@ class LocalAuthorityDataService(
             localAuthorityUserRepository.findByIdOrNull(localAuthorityUserId)
                 ?: throw ResponseStatusException(HttpStatus.NOT_FOUND, "User $localAuthorityUserId does not exist")
 
+        val localAdminsByAuthority =
+            localAuthorityUserRepository.findAllByLocalAuthority_IdAndIsManagerTrue(localAuthorityUser.localAuthority.id)
```
Kiran-R-K, line 141: you could have this variable just store a list of the admins emails by adding `.email` to the end. I think both ways work so up to you.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityDataService.kt
```diff
@@ -135,6 +137,21 @@ class LocalAuthorityDataService(
             localAuthorityUserRepository.findByIdOrNull(localAuthorityUserId)
```
Kiran-R-K, line 137: I'm surprised we don't have some unit tests for this `deleteUser` method in `LocalAuthorityDataServiceTests.kt`. It wasn't part of this ticket to implement the method so I wouldn't expect to see that here but given you're updating the method now would be a could time to add some.
If you want to give it a try `LandlordDeregistrationServiceTests` has some good examples of how to test the repo is being called by the method

Bill-Haigh: have added a few tests but had to give up on adding tests for the sending of emails.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityDataService.kt
```diff
@@ -137,6 +137,18 @@ class LocalAuthorityDataService(
...
+        deletionConfirmationSender.sendEmail(
+            localAuthorityUser.email,
+            LocalAuthorityUserDeletionEmail(
+                councilName = localAuthorityUser.localAuthority.name,
+            ),
+        )
+        localAuthorityUserRepository.deleteById(localAuthorityUserId)
+    }
+
+    private fun emailToAdmins(localAuthorityUser: LocalAuthorityUser) {
```
Kiran-R-K, line 151: I think it's worth giving this a more specific name as there will likely be other similar methods added and we'll want to know which is which

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LocalAuthorityDataService.kt
```diff
@@ -137,6 +137,18 @@ class LocalAuthorityDataService(
...
 
+        emailToAdmins(localAuthorityUser)
+
+        deletionConfirmationSender.sendEmail(
+            localAuthorityUser.email,
+            LocalAuthorityUserDeletionEmail(
+                councilName = localAuthorityUser.localAuthority.name,
+            ),
+        )
+        localAuthorityUserRepository.deleteById(localAuthorityUserId)
```
Kiran-R-K, line 148: I think these should be re-ordered, sorry I didn't catch it earlier. We should have:
- deleting the user from the repo
- then sending the email to the user whose account is being deleted
- then sending the emails to the LC admins

This is because if the actual deletion fails we don't want to have sent an email to the user saying we have deleted them

PR #770, opened 2025-09-11
Feat/prsd 1521 reporting queries

JasminConterioSW:
## Ticket number

PRSD-1521

## Goal of change

Write queries to retrieve KPI information from the prod database

Add queries for

- Counting transactions (LL registration, Property registration, Adding compliance, updates to the above, LC user registration, passcode generation) over the last reporting period (this is for calculating cost per transaction)
- Counting registrations
    - Total so far, and in the last reporting period)
    - Total and per council 
- Calculating transaction times
    - LL registration - time between claiming the passcode and creating a landlord
    - Property registration - time between registering a landlord and registering the first property
    - Add compliance - time between registering a property and adding it's compliance information

These queries aren't run by the webapp but this seemed like a sensible places to put them!

## Checklist

### DIFF
--- a/src/test/resources/databaseQueriesForReports/numberOfRegistrationsAndUpdates.sql
+++ b/src/test/resources/databaseQueriesForReports/numberOfRegistrationsAndUpdates.sql
@@ -1,25 +1,27 @@
+-- A :reference_date of CURRENT_DATE will count records between just before midnight last night and 14 days before that.
+-- A :reference_date of make_date(2025, 09, 15) will count records from 1st September (2025-09-01 00:00:00.000000) - 14th September inclusive (2025-09-015 00:00:00.000000 is NOT included).
 -- Total current registrations, and those in the last 2 weeks
 -- Landlord registrations / updates
 SELECT
     COUNT(*) AS total_landlords,
-    COUNT(*) FILTER (WHERE created_date >= NOW() - INTERVAL '14 DAYS') AS new_landlords_last_2_weeks,
-    COUNT(*) FILTER (WHERE last_modified_date >= NOW() - INTERVAL '14 DAYS') AS updated_landlords_last_2_weeks
+    COUNT(*) FILTER (WHERE created_date >= :reference_date - INTERVAL '14 DAYS' AND created_date < :reference_date) AS new_landlords_last_2_weeks,
+    COUNT(*) FILTER (WHERE last_modified_date >= :reference_date - INTERVAL '14 DAYS' AND last_modified_date < :reference_date) AS updated_landlords_last_2_weeks
 FROM
     landlord;
 
 -- Property registrations / updates
 SELECT
     COUNT(*) AS total_property_ownerships,
-    COUNT(*) FILTER (WHERE created_date >= NOW() - INTERVAL '14 DAYS') AS new_property_ownerships_last_2_weeks,
-    COUNT(*) FILTER (WHERE last_modified_date >= NOW() - INTERVAL '14 DAYS') AS updated_property_ownerships_last_2_weeks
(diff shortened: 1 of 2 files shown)


src/test/resources/databaseQueriesForReports/numberOfRegistrationsAndUpdates.sql
```diff
@@ -0,0 +1,41 @@
+-- Total current registrations, and those in the last 2 weeks
+-- Landlord registrations / updates
+SELECT
+    COUNT(*) AS total_landlords,
+    COUNT(*) FILTER (WHERE created_date >= NOW() - INTERVAL '14 DAYS') AS new_landlords_last_2_weeks,
```
Travis-Softwire, line 5: Hmm, I'm not sure about using 'NOW' here - does this not mean we could accidentally double count/ miss some numbers if we don't run the query at exactly the same time of day every 2 weeks?

src/test/resources/databaseQueriesForReports/timeToRegisterLandlord.sql
```diff
@@ -0,0 +1,20 @@
+-- We count the start of landlord registration as the time when the landlord claims a passcode (assumed to be when their passcode was last updated)
```
Travis-Softwire, line 1: This should work fine, but could also use the created_date for the equivalent one_login_user record as that's created when they claim the passcode (unless they're also an LC user, but that shouldn't happen during pilot)

Travis-Softwire (approved):
A few comments, but overall LGTM :) Main substantive point is we should think about parameterising these to allow us to specify the period regardless of when we run the query.

Also, not as part of this PR, but would be good to either have these run with a script bash script or similar, or if that's going to be challenging, then having a nice ReadMe setting out the process for running them (you may well already have something planned, but mentioning here just in case!)

src/test/resources/databaseQueriesForReports/numberOfRegistrationsAndUpdates.sql
JasminConterioSW: This is a good point, it should probably be something like
`COUNT(*) FILTER (WHERE created_date >= :reference_date - INTERVAL '14 DAYS' AND created_date < :reference_date) AS new_landlords_last_2_weeks`

If `CURRENT_DATE` is passed in as the reference_date, it will not count anything from 00:00 on the current date, but will count everything for 14 days before that

src/test/resources/databaseQueriesForReports/timeToAddCompliance.sql
```diff
@@ -0,0 +1,17 @@
...
+    percentile_cont(0.50) WITHIN GROUP (ORDER BY time_to_add_compliance) AS p50_time_to_add_compliance,
+    percentile_cont(0.95) WITHIN GROUP (ORDER BY time_to_add_compliance) AS p95_time_to_add_compliance,
+    percentile_cont(0.99) WITHIN GROUP (ORDER BY time_to_add_compliance) AS p99_time_to_add_compliance
+FROM (
+     SELECT
+         pc.created_date - po.created_date AS time_to_add_compliance
+     FROM
+         property_ownership po
+             JOIN (
+             SELECT
```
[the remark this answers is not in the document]
JasminConterioSW: I'll be honest I was starting with copilot's suggestion, which joined the result of the query...
I was generally going with only joining the information we needed although maybe it doesn't matter as we never look at that `t` table.

I'll probably leave it as it is for now - I haven't looked into running a script for these yet and the initial version is a very manual "run this script on the database". It feels like selecting minimal information at each stage is probably best when manually poking around the database!

PR #771, opened 2025-09-15
PRSD-1531: Add slack reporting for test and production builds

Kiran-R-K:
## Ticket number

PRSD-1531

## Goal of change

Add slack reporting for test and production builds

Removed the is branch "main" from the re porting on slack action

I have not tested this manually

## Checklist

N/A

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -209,6 +209,7 @@ jobs:
         with:
           payload: |
             {
+              "environment": ${{ toJson(inputs.environment) }},
               "commit": ${{ toJson(github.event.head_commit.message) }},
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-deploy.yml
```diff
@@ -203,7 +203,6 @@ jobs:
     needs: update-ecs-service
     steps:
       - name: Report pipeline status
-        if: always() && github.event_name == 'push' && github.ref == 'refs/heads/main'
```
Travis-Softwire, line 206: I think we still need the `if: always()` or this won't run when the pipeline fails on an earlier step

Travis-Softwire (approved):
✅

PR #778, opened 2025-09-19
PRSD-1553: Remove gas safety engineer row from property record when empty

Kiran-R-K:
## Ticket number

PRSD-1553

## Goal of change

Remove gas safety engineer row from property record when empty

As above and updated test

### After
image

### Before
image

## Checklist


isobel-softwire (approved):
LGTM :)
I think the after screenshot is the same as the before one - I assume the engineer number row doesn't actually appear now?

PR #783, opened 2025-09-23
Feat/prsd 1567/node app for analytics

Bill-Haigh:
## Ticket number

PRSD-1567

## Goal of change

Create a simple JS tool to fetch stats from plausible's stats API.

Created a script to pull stats from plausible's api using queries stored in a json file, and converts and stores the output as csv files in Outputs.

## Checklist

### DIFF
--- a/docs/PlausibleScriptsReadMe.md
+++ b/docs/PlausibleScriptsReadMe.md
@@ -46,7 +46,7 @@
     ```
   - To process a specific input file:
     ```
-    node plausible.mjs --input myfile.json
+    node plausible.mjs --input-file myfile.json
     ```
   - To clear the outputs directory before running:
     ```


scripts/Plausible/Inputs/inputQueries.json
```diff
@@ -0,0 +1,22 @@
+{
+"dwelling_time_per_page": {
+  "site_id": "SITE_ID",
+  "date_range": "5d",
+  "metrics": [
+    "time_on_page"
+  ],
+  "dimensions": [
+    "event:page"
+  ]
```
JasminConterioSW, line 10: It might be worth adding something to check whether we are getting all of the results... it looks like by default the results are paginated with 1000 entries on a page. This can be increased up to 10000, but you could do something like make it return the total number of rows it's expecting and warn you if it's more than the entries per page?
If you add something like 
```
  "include": { "total_rows": true },
  "pagination": { "limit": 10000, "offset": 0 }
```
to the query it gives you "total_rows" in the "meta" bit of the response
image

scripts/Plausible/plausible.mjs
```diff
@@ -0,0 +1,91 @@
+import 'dotenv/config'
+import Papa from 'papaparse'
+import fs from 'fs'
+import path from 'path'
+
+const BASE_URL = process.env.PLAUSIBLE_BASE_URL
+const API_KEY = process.env.PLAUSIBLE_API_KEY
+const SITE_ID = process.env.PLAUSIBLE_SITE_ID
```
JasminConterioSW, line 8: I think BASE_URL and SITE_ID can probably be hardcoded here, I don't think they need to be in the .env file?

API_KEY definitely does though!

scripts/Plausible/plausible.mjs
```diff
@@ -0,0 +1,91 @@
...
+            obj[metric] = row.metrics[i];
+        });
+        dimensions.forEach((dimension, i) => {
+            obj[dimension] = row.dimensions[i];
+        });
+        return obj;
+    });
+}
+
+const INPUT_QUERIES_PATH = path.join('Inputs', 'inputQueries.json')
```
JasminConterioSW, line 55: Is it going to be useful to pass the inputQueries.json filename into this function?

You might end up with a set of queries for getting dwell times, another set of queries for something else which are easier to keep in separate files, so you might want to choose which file to use.

scripts/Plausible/plausible.mjs
```diff
@@ -0,0 +1,91 @@
...
+const inputQueries = JSON.parse(fs.readFileSync(INPUT_QUERIES_PATH, 'utf8'))
+
+function replaceSiteId(obj, siteId) {
```
JasminConterioSW, line 58: Again I think the site_id can be hardcoded (it appears in the code snippet on the webapp pages so that one isn't secret)

.gitignore
```diff
@@ -46,10 +46,12 @@ notify-template-tests.run.xml
 
 ### Node ###
 /node_modules
+scripts/Plausible/node_modules
```
isobel-softwire, line 49: This is super pedantic but we normally start our folder names with a lowercase letter. Would you mind changing `Plausible`, `Plausible/Outputs` and `Plausible/Inputs`?

scripts/Plausible/README.md
isobel-softwire, line 1: Probably worth moving this to our `docs` folder

scripts/Plausible/README.md
```diff
@@ -0,0 +1,36 @@
...
+## Configuration
+- Add your Plausible API key and site ID to a `.env` file in the `scripts` directory:
+  ```
+  PLAUSIBLE_API_KEY=your_api_key
+  PLAUSIBLE_SITE_ID=your_site_id
+  PLAUSIBLE_BASE_URL=plausible_api_base_url
+  ```
+
+## Input Queries
+- Store queries in `inputQueries.json` in `scripts/Plausible`.
```
isobel-softwire, line 16: Should this be `scripts/plausible/inputs`?

scripts/Plausible/plausible.mjs
```diff
@@ -0,0 +1,91 @@
...
+  process.exit(1)
+}
+
+if (!SITE_ID) {
+  console.error('Please provide a Plausible site ID in the PLAUSIBLE_SITE_ID environment variable.')
+  process.exit(1)
+}
+
+async function queryPlausible(query) {
+    const res = await fetch (BASE_URL, {
```
Travis-Softwire, line 23: I expect the results are paginated, in which case we'll want to page through all pages of results and collect them?

Travis-Softwire (commented):
Agree w/ Jasmin on writing the site id into the query json file, and handling pagination in the request code, otherwise LGTM once Jasmin is happy w/ it :)

scripts/plausible/plausible.mjs
```diff
@@ -0,0 +1,137 @@
...
+            obj[dimension] = row.dimensions[i];
+        });
+        return obj;
+    });
+}
+
+const program = new Command();
+program
+  .option('--all', 'Process all input files')
+  .option('--input <filename>', 'Process a specific input file')
```
JasminConterioSW, line 58: I'd maybe make the option name --input-file so it's clear they can't specify a directory.
Maybe there could be an option where the user specifies a different input directory, but I don't think there has to be

JasminConterioSW (approved):
LGTM, thanks this will be really useful!

PR #786, opened 2025-09-25
Chore/prsd none update kpi report db queries

JasminConterioSW:
## Ticket number

PRSD-NONE

## Goal of change

Update database queries for KPI reporting

- Modify queries that count the total registrations so far so that they don't include the current day (current day gets including in the next report as otherwise data might be changing while we are collecting)
- Update the query for property registrations by local council to be grouped by the LC that invited the property's landlord instead of the LC that the address is in
- Update the query for time to register landlord to use the time the one-login user was created instead of the time the passcode was updated (I found some instances where the passcode had been updated after the LL was created)

## Checklist

### DIFF
--- a/src/test/resources/databaseQueriesForReports/timeToRegisterLandlord.sql
+++ b/src/test/resources/databaseQueriesForReports/timeToRegisterLandlord.sql
@@ -1,4 +1,4 @@
--- We count the start of landlord registration as the time when the landlord claims a passcode (assumed to be when their passcode was last updated)
+-- We count the start of landlord registration as the time when the landlord's one-login user account was created.'
 SELECT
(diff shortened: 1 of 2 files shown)


src/test/resources/databaseQueriesForReports/timeToRegisterLandlord.sql
```diff
@@ -5,16 +5,16 @@ SELECT
...
              JOIN (
              SELECT
-                 subject_identifier,
-                 last_modified_date
+                 id,
+                 created_date
              FROM
-                 passcode
-         ) p ON p.subject_identifier = p.subject_identifier
+                 one_login_user
```
Travis-Softwire, line 18: Minor, but should update the commnet at the top of the file to match :)

PR #787, opened 2025-09-25
PRSD-934: Ensure consistent formatting editor config

Kiran-R-K:
## Ticket number

PRSD-934

## Goal of change

Add a `.editorconfig` file for consistent formatting across the repo

As above

I based the config file off the template in the [mhclg-way](https://github.com/communitiesuk/mhclg-way/blob/3ea0660fae611ee80cceb32332081f4dd4605d7a/source/manuals/style-guides/editorconfig) repo. I have adjusted some of the details to better match what we currently have in the codebase.

## Checklist

N/A

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,11 +9,11 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 max_line_length = 140
 
-[*.kt]
+[*.{kt,kts}]
 trim_trailing_whitespace = false
 
 [*.html]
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,18 @@
+root = true
```
Travis-Softwire, line 1: Should we also have line length in here? We've definitely got different values for that set across the team

PR #797, opened 2025-10-08
Feat/prsd 1568/scripts

Bill-Haigh:
## Ticket number

PRSD-1568

## Goal of change

Update the scripts so that the weekly reports can be automated.

- updated the dwell_times query so that we capture data for the dwell times for the journey pages
- created a script to process the csv files further into journeys
- created a script which processes them once more into a format that can be copied and pasted into excel.

## Checklist


scripts/plausible/createMetricCSV.mjs
```diff
@@ -75,25 +66,26 @@ export async function createPageViewCSV() {
...
 export async function createCompletionRateCSV() {
 
+    const metric = 'visitors';
     const llRegCSV = await readCSV(path.resolve('processed_journey_data/outputs/visitors/visitors/landlord_register_as_a_landlord.csv'));
-    const llRegStart = await getVisitorCount(llRegCSV, '/landlord/register-as-a-landlord/start');
-    const llRegConf = await getVisitorCount(llRegCSV, '/landlord/register-as-a-landlord/confirmation');
-    const llRegRate = (llRegConf === 0) ? 0 : ((llRegStart && llRegConf) ? ((llRegConf / llRegStart) * 100).toFixed(2) : null);
+    const llRegStart = await getValueFromCSV(llRegCSV, '/landlord/register-as-a-landlord/start', metric);
+    const llRegConf = await getValueFromCSV(llRegCSV, '/landlord/register-as-a-landlord/confirmation', metric);
+    const llRegRate = (llRegStart === 0 || llRegStart === null) ? null : (llRegConf === 0 || llRegConf === null) ? 0 : ((llRegStart && llRegConf) ? ((llRegConf / llRegStart) * 100).toFixed(2) : null);
```
JasminConterioSW: I don't think you need the check of `(llRegStart && llRegConf)`  at the end of this line as you've already checked that `llRegStart ` and `llRegConf` are not null (same for same for propRegRate, complianceRate, lcUserRate below)

PR #1139, opened 2026-03-25
PDJB-NONE: Fix performance issue causing flaky integration tests

JasminConterioSW:
## Ticket number

PDJB-725

## Goal of change

Fix performance issue causing flaky integration tests

The journey tests in [this PR ](https://github.com/communitiesuk/prsdb-webapp/pull/1132) were getting flaky (2 failed when run all together, one of those passed when run separately, the other passed when run in debug so it paused at the failure point before continuing successfully).

Claude has been digging into it for me, this is the copilot summary:

Root Cause

The checkEpcAnswersStep has 10 OrParents conditions. When computing backUrl, allowingParentSteps evaluates all 10
parents via filter. Each parent's allowsChild() recursively evaluates isStepReachable and outcome on ancestor steps,
traversing the entire journey tree. Without memoization, the same steps were re-evaluated exponentially — 178,000+
evaluations before the 60s timeout killed the request.

Fix

Changed isStepReachable and outcome from computed properties (re-evaluated on every access) to lazy vals (computed
once, cached per step instance). Since step beans are prototype-scoped (fresh per request), the cache is naturally
request-scoped:

 val isStepReachable: Boolean by lazy(LazyThreadSafetyMode.NONE) { parentage.allowsChild() }
 val outcome: TEnum? by lazy(LazyThreadSafetyMode.NONE) { if (isStepReachable) stepConfig.mode(state) else null }

This reduces evaluation from exponential to linear — each step's reachability and outcome are computed at most once
per request.

This came up when developing this PR https://github.com/communitiesuk/prsdb-webapp/pull/1132.
Originally this branch came off that one, I've run the tests and am happy they pass locally.
Now rebased back onto main for PR

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/JourneyStep.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/JourneyStep.kt
@@ -154,8 +154,7 @@ sealed class JourneyStep<out TEnum : Enum<out TEnum>, TFormModel : FormModel, in
     val lifecycleOrchestrator: StepLifecycleOrchestrator
         get() = stepConfig.getStepLifecycleOrchestrator(this)
 
-    val isStepReachable: Boolean
-        get() = parentage.allowsChild()
+    val isStepReachable: Boolean by lazy(LazyThreadSafetyMode.NONE) { parentage.allowsChild() }
 
     val formModelOrNull: TFormModel?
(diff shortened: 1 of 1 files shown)


TomHanmer29 (commented):
Not sure I fully have my head around how this will interact with the full journey structure - if we just compute if a step is reachable once, could it cause issues if we go to a step, then go back and change something? It's _probably_ fine if the tests pass, but might be one @Travis-Softwire should review too :)

PR #1192, opened 2026-04-07
PDJB-241: Remove passcode local council dependency and move endpoint to system-operator

Bill-Haigh:
## Ticket number

PDJB-241

## Goal of change

Decouple passcodes from local councils and move the generate-passcode page under the system-operator URL path.

- Removed the local council foreign key from the `Passcode` entity and dropped the `local_council_id` column via a Flyway migration
- Removed the `LocalCouncilDataService` dependency and `Principal` parameter from `GeneratePasscodeController` — passcode generation no longer needs a council context
- Changed access from `LOCAL_COUNCIL_ADMIN` to `SYSTEM_OPERATOR` role
- Moved the endpoint from `/local-council/generate-passcode` to `/system-operator/generate-passcode`
- Removed the 'Generate passcode' navigation link from the local council dashboard
- Removed the 'Return to dashboard' links from both the generate passcode page and the passcode limit error page
- Simplified `PasscodeService` by removing the `localCouncilId` parameter from all generation methods
- Updated seed data, unit tests, controller tests, integration tests, and page objects to match

The endpoint now falls under the `DefaultSecurityConfig` filter chain (instead of `LocalCouncilSecurityConfig`) which is consistent with other system-operator routes like `ManageLocalCouncilAdminsController`.

## Checklist

### DIFF
--- a/src/main/resources/db/migrations/V1_21_0__remove_local_council_from_passcode.sql
+++ b/src/main/resources/db/migrations/V1_21_0__remove_local_council_from_passcode.sql
@@ -1,0 +1,1 @@
+ALTER TABLE passcode DROP COLUMN local_council_id;
--- a/src/main/resources/db/migrations/V1_21_0__remove_local_council_from_passcode.sql
+++ b/src/main/resources/db/migrations/V1_21_0__remove_local_council_from_passcode.sql
@@ -1,1 +1,2 @@
+ALTER TABLE passcode DROP CONSTRAINT passcode_local_authority_id_fkey;
 ALTER TABLE passcode DROP COLUMN local_council_id;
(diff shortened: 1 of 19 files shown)


src/main/resources/db/migrations/V1_21_0__remove_local_council_from_passcode.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE passcode DROP COLUMN local_council_id;
```
Travis-Softwire: Was this not a fk? Or does dropping the column automatically drop the fk in postgres?

Travis-Softwire (approved):
One Q, but otherwise LGTM

PR #1197, opened 2026-04-09
PDJB-NONE: Profile-gate NftDataSeeder to fix scheduled task startup

Travis-Softwire:
## Ticket number

PDJB-NONE

## Goal of change

Fix scheduled tasks failing to start due to unresolvable `EPC_CERTIFICATE_BASE_URL` placeholder in `NftDataSeeder`.

- Adds `@Profile("nft-data-seeder")` to `NftDataSeeder` so it only loads when that profile is explicitly activated, preventing the `epc.certificate-base-url` property from being resolved during other tasks
- Removes the redundant `@Profile` annotation from `NftDataSeedingTaskApplicationRunner` since `@PrsdbTask("nft-data-seeder")` already gates on the profile via `TaskHasName`
- Removes the blank default from the `@Value` annotation to avoid silent failures when the seeder is actually used
- Updates application tests to remove `NftDataSeeder` from expected bean lists

The previous fix (PR #1194) added a blank default to the `@Value` annotation, but this did not work because Spring resolves the nested `$`{EPC_CERTIFICATE_BASE_URL}` placeholder from `application.yml` first, before the outer default applies.

## Checklist


AEPR (approved):
LGTM!

PR #1198, opened 2026-04-09
PDJB-654: Implement remove electrical cert upload page

AEPR:
## Ticket number

PDJB-654

## Goal of change

Allow users to remove uploaded electrical safety certificates from the property registration journey.

- Implements the `RemoveElectricalCertUploadStepConfig` following the established gas safety remove pattern, using the `areYouSureForm` template with yes/no radios
- Updates the integration test page object to use `FormWithRadios` and `BackLink`
- Adds integration test coverage: uploads two certificates, removes one via the remove page, and verifies only one remains

Neither the gas nor electrical remove steps currently delete the uploaded files from S3 — they only remove the entry from journey state. S3 cleanup for both certificate types is covered in a follow-up PR: #1200.

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/RemoveElectricalCertUploadStepConfig.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/RemoveElectricalCertUploadStepConfig.kt
@@ -3,19 +3,29 @@ package uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps
 import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.JourneyFrameworkComponent
 import uk.gov.communities.prsdb.webapp.journeys.AbstractRequestableStepConfig
 import uk.gov.communities.prsdb.webapp.journeys.JourneyStep.RequestableStep
+import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.states.CertificateUpload
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.states.ElectricalSafetyState
 import uk.gov.communities.prsdb.webapp.journeys.shared.AnyMembers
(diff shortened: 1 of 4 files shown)


TomHanmer29 (approved):
LGTM! Should we have a single page test for the validation error on the remove page or does it not matter because it's using the "are you sure" form?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/RemoveElectricalCertUploadStepConfig.kt
```diff
@@ -3,32 +3,64 @@ package uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps
...
+        return keyToRemove != null && keyToRemove in currentMap.keys
+    }
+
+    override fun afterStepDataIsAdded(state: ElectricalSafetyState) {
+        if (getFormModelFromStateOrNull(state)?.wantsToProceed == false) {
+            return
+        }
+        val currentMap = state.electricalUploadMap.toMutableMap()
+
+        currentMap.remove(collectionKeyParameterService.getParameterOrNull())
```
Travis-Softwire, line 55: If a user did something silly and opened the same remove page in multiple tabs, would they hit an exception here when submitting the second tab? (Appreciate if this is an issue it's probably also an issue for JLs!)

Travis-Softwire (approved):
One Q from me - if it needs action then happy for it to be resolved in a later PR though

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/RemoveElectricalCertUploadStepConfig.kt
AEPR: No, this line is idempotent - it does nothing if the key doesn't exist. They'll actually be returned to the Task list though in secondary tabs because this page is only reachable if the key is in the collection.

PR #1200, opened 2026-04-09
PDJB-654: Delete uploaded files from S3 when removed from journey

AEPR:
## Ticket number

PDJB-654

## Goal of change

Ensure uploaded certificate files are deleted from S3 when a user removes them during property registration, rather than being orphaned.

- Adds a `SafeFileDeleter` interface (with AWS and local implementations) for deleting files from the safe S3 bucket in web server mode
- Adds `deleteUploadedFile` method to `UploadService` that sets the file status to DELETED then deletes the safe bucket copy for scanned files
- Updates both the gas and electrical remove certificate step configs to call `deleteUploadedFile` when a user confirms removal
- Updates `VirusScanProcessingService` to handle the DELETED status — if a file was marked DELETED while still quarantined, the scan processor cleans up the quarantine copy instead of dequarantining it

The design deliberately avoids deleting from the quarantine bucket in web server mode. The `QuarantinedFileDeleter` implementations are `@PrsdbTaskService` beans (not available in web context), and there is a potential race condition with the scan processor. Instead, the scan processor is updated to recognise the DELETED status and handle quarantine cleanup itself.

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/UploadService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/UploadService.kt
@@ -1,5 +1,6 @@
 package uk.gov.communities.prsdb.webapp.services
 
+import jakarta.transaction.Transactional
 import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.PrsdbWebService
 import uk.gov.communities.prsdb.webapp.constants.enums.FileUploadStatus
 import uk.gov.communities.prsdb.webapp.database.entity.FileUpload
@@ -54,6 +55,7 @@ class UploadService(
             null
         }
 
+    @Transactional
     fun deleteUploadedFile(fileUploadId: Long) {
         val fileUpload = getFileUploadById(fileUploadId)
 
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/UploadService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/UploadService.kt
@@ -10,6 +10,7 @@ import java.io.InputStream
 class UploadService(
     private val uploader: FileUploader,
     private val downloader: FileDownloader,
+    private val safeFileDeleter: SafeFileDeleter,
     private val uploadRepository: FileUploadRepository,
 ) {
     fun uploadFile(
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/UploadService.kt
```diff
@@ -52,4 +53,18 @@ class UploadService(
         } else {
             null
         }
+
+    fun deleteUploadedFile(fileUploadId: Long) {
```
Travis-Softwire, line 59: Should this be in a transaction?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/UploadService.kt
```diff
@@ -52,4 +53,18 @@ class UploadService(
...
+        val fileUpload = getFileUploadById(fileUploadId)
+
+        if (fileUpload.status == FileUploadStatus.DELETED) return
+
+        val previousStatus = fileUpload.status
+        fileUpload.status = FileUploadStatus.DELETED
+        uploadRepository.save(fileUpload)
+
+        if (previousStatus == FileUploadStatus.SCANNED) {
+            safeFileDeleter.deleteFile(fileUpload)
```
Travis-Softwire, line 69: If the deletion fails for some reason, what happens?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/UploadService.kt
```diff
@@ -52,4 +53,18 @@ class UploadService(
         } else {
             null
         }
+
+    fun deleteUploadedFile(fileUploadId: Long) {
+        val fileUpload = getFileUploadById(fileUploadId)
+
+        if (fileUpload.status == FileUploadStatus.DELETED) return
+
+        val previousStatus = fileUpload.status
```
Travis-Softwire, line 64: Is there a read/write race condition here if the scanner writes the scanned status between the read and the save?

Travis-Softwire (approved):
A few Qs on the deleter (but suspect at least part of the answer will be that we still need an orphan removal job as a backstop) but overall LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/UploadService.kt
AEPR: Good point - the file could end up dequarrantined in the mean time, then marked as deleted here but not in fact deleted. I think this being transactional resolves that.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/UploadService.kt
AEPR: It silently fails and we have an orphaned file in S3. I've not thought overly hard about that scenario since it will be a relatively small number files. We've discussed this separately and will be covered by a more general improvement to how we handle blob tracking/deletion.

PR #1201, opened 2026-04-09
PDJB-592: Add local council start page with service name interceptor

Bill-Haigh:
## Ticket number

PDJB-592

## Goal of change

Create a public start page for the local council service and display the correct service name ("Check a rental property or landlord") in the header for all local council pages.

- Adds a new public start page at `/local-council/start` with content matching the Figma designs (what the service does, before you start, start now button)
- Adds a `ServiceNameInterceptor` that sets a custom service name for all `/local-council/` routes, displayed in both the authenticated and anonymous page headers
- Updates the anonymous header template to show a GOV.UK service navigation bar when a custom service name is present
- Updates the LC user registration privacy notice form to match revised Figma designs (caption, auto-deletion paragraph)
- Adds the start page route to the LC security config's `permitAll()` list so it's accessible without authentication

- The `ServiceNameInterceptorConfig` is a `WebMvcConfigurer` which means it gets auto-detected by `@WebMvcTest`. It uses a 4-arg `getMessage()` with a default value to avoid `NoSuchMessageException` in test contexts where YAML message files aren't loaded.
- The interceptor skips redirect views to prevent the custom service name leaking into redirect URLs as query parameters.
- `ServiceNameInterceptorConfig` has been added to `ControllerTest`'s `@Import` list so all controller tests have the interceptor active.

## Checklist

image


src/main/kotlin/uk/gov/communities/prsdb/webapp/config/interceptors/ServiceNameInterceptor.kt
```diff
@@ -0,0 +1,25 @@
+package uk.gov.communities.prsdb.webapp.config.interceptors
+
+import jakarta.servlet.http.HttpServletRequest
+import jakarta.servlet.http.HttpServletResponse
+import org.springframework.web.servlet.HandlerInterceptor
+import org.springframework.web.servlet.ModelAndView
+import uk.gov.communities.prsdb.webapp.constants.LOCAL_COUNCIL_PATH_SEGMENT
+
+class ServiceNameInterceptor(
```
Travis-Softwire: Why an interceptor? Feels like this could just be handled by the GlobalModelAttributes class?

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/interceptors/ServiceNameInterceptor.kt
```diff
@@ -0,0 +1,25 @@
...
+    private val localCouncilServiceName: String,
+) : HandlerInterceptor {
+    override fun postHandle(
+        request: HttpServletRequest,
+        response: HttpServletResponse,
+        handler: Any,
+        modelAndView: ModelAndView?,
+    ) {
+        if (request.requestURI.startsWith("/$LOCAL_COUNCIL_PATH_SEGMENT") &&
```
Travis-Softwire: We should also include routes that start with the system operator segment (but as above, that should probably happen in the GlobalModelAttributes class)

src/main/resources/templates/forms/localCouncilPrivacyNoticeForm.html
```diff
@@ -8,17 +8,19 @@
 <!DOCTYPE html>
 <html th:replace="~{fragments/forms/complexQuestionPage :: complexQuestionPage(#{${title}}, ${formModel}, ~{::#form-content}, ~{::#question-content}, ${backUrl}, ${sectionHeaderInfo})}">
     <th:block id="question-content">
+        <span class="govuk-caption-l" th:text="#{localCouncilServiceName}">Check a rental property or landlord</span>
```
Travis-Softwire: Minor but as per the conventions in the repo we usually use the message key as the placeholder, not the message text (and the same for a couple of places further down this file)

src/main/resources/templates/fragments/header/anonymousHeader.html
```diff
@@ -28,4 +28,15 @@
             </a>
         </div>
     </div>
+    <section th:if="${customServiceName} != null" aria-label="Service information"
+             class="govuk-service-navigation"
+             data-module="govuk-service-navigation">
+        <div class="govuk-width-container">
+            <div class="govuk-service-navigation__container">
+                <span class="govuk-service-navigation__service-name">
+                    <span class="govuk-service-navigation__text" th:text="${customServiceName}">Service name</span>
```
Travis-Softwire: Same point here RE the placeholder

src/main/resources/templates/localCouncilStartPage.html
```diff
@@ -0,0 +1,45 @@
...
+<html th:replace="~{fragments/layout :: layout(#{localCouncilStartPage.title}, ~{::main}, false)}">
+<main class="govuk-main-wrapper" id="main-content">
+    <div th:replace="~{fragments/layouts/twoThirdsLayout :: twoThirdsLayout(~{::#page-content})}">
+        <div id="page-content">
+            <h1 class="govuk-heading-xl" th:text="#{localCouncilStartPage.heading}">localCouncilStartPage.heading</h1>
+
+            <p class="govuk-body" th:text="#{localCouncilStartPage.intro.one}">localCouncilStartPage.intro.one</p>
+            <p class="govuk-body" th:text="#{localCouncilStartPage.intro.two}">localCouncilStartPage.intro.two</p>
+
+            <h2 class="govuk-heading-m" th:text="#{localCouncilStartPage.howTheTestingWorks.heading}">How the testing works</h2>
```
Travis-Softwire: And here

Travis-Softwire (changes requested):
I don't think we want to use an interceptor for the service name?

src/main/resources/templates/fragments/header/authenticatedHeader.html
```diff
@@ -78,12 +78,12 @@
             <div class="govuk-service-navigation__container">
 
         <span class="govuk-service-navigation__service-name">
-            <span class="govuk-service-navigation__text" th:text="#{serviceName}">Service name</span>
+            <span class="govuk-service-navigation__text" th:text="${customServiceName} ?: #{serviceName}">Service name</span>
```
Travis-Softwire: Silly question - can't we just use the GlobalModelAttributes to set serviceName for all cases (Landlord, LC, system operator) rather than duplicating?

Travis-Softwire (approved):
One additional point to consider. I think once that's done this should be good to go though

PR #1204, opened 2026-04-09
PDJB-NONE: unify bold text method

TomHanmer29:
## Ticket number

PDJB-NONE

## Goal of change

Unify the way we make text bold in yamls

I noticed we had started using "beforeBold" etc. in the same way we did links on some bold text in property compliance, whereas the rest of the codebase used <strong> tags in the yaml. I believe the <strong> way is cleaner and more readable, and makes the yamls less bloaty so I fixed it.

## Checklist


Bill-Haigh (approved):
LGTM

PR #1206, opened 2026-04-10
PDJB-NONE: Use var convention for revisited pages in integration tests

AEPR:
## Ticket number

PDJB-NONE

## Goal of change

Standardise integration tests to use ar and reassignment when revisiting a page type, instead of creating new al variables with suffixed names.

- When a test revisits the same page type, the first assignment is now ar and subsequent visits reassign the same variable, replacing the old pattern of suffixed names like lookupAddressPageAgain, inalCheckJointLandlordPage, 
ewGeneratePasscodePage, etc.
- Also fixes a pre-existing bug in the "Numbering on page and tables is correct" test where the final assertion incorrectly referenced irstCheckJointLandlordPage instead of inalCheckJointLandlordPage (the refactor naturally corrected this by unifying both into a single checkJointLandlordPage variable)
- Drops misleading irst prefixes from single-visit page variables for consistency with surrounding tests
- 5 files modified, 12 test methods updated

The pre-existing bug on line 807 of PropertyRegistrationSinglePageTests (referencing irstCheckJointLandlordPage instead of inalCheckJointLandlordPage) was silently masked because Playwright locators resolve against the live DOM, so both variables pointed to the same current page state. The refactor fixes this by consolidating into one variable.

## Checklist

### DIFF
--- a/.github/instructions/integration-tests.instructions.md
+++ b/.github/instructions/integration-tests.instructions.md
@@ -75,6 +75,26 @@ val page = navigator.skipToCheckAnswers()
 navigator.navigateToHomePage()
 ```
 
+## Revisiting Pages
+
+When a test revisits the same page type, use `var` and reassign rather than creating new variables with suffixed names:
+
+```kotlin
+// Correct — var + reassignment
+var checkJointLandlordPage = assertPageIs(page, CheckJointLandlordsFormPage::class)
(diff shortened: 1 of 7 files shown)


TomHanmer29 (approved):
LGTM!

PR #1207, opened 2026-04-10
PDJB-714: Update property registration confirmation page with three variants

Bill-Haigh:
## Ticket number

PDJB-714

## Goal of change

Update the property registration confirmation page to match new Figma designs with three variants: occupied with incomplete compliance, unoccupied, and occupied with complete compliance.

- Restructured the confirmation page template to show a green banner with the property address on multiple lines, the PRN in an inset text block, and body paragraphs about keeping the number and using it in adverts
- Added a conditional "What you need to do next" section that appears only when the property is occupied and has incomplete compliance certificates, showing a deadline for uploading them
- Added `PropertyComplianceService` dependency to `RegisterPropertyController` to determine compliance status
- Deleted two unused fragment templates (occupied and unoccupied content) that are now consolidated into the single confirmation template
- Updated integration test page object and assertions to match the new page structure

- The address is rendered using `th:each` over address parts with `th:text` (safe escaping) rather than injecting `<br>` tags via `th:utext`, to avoid XSS risk from user-controlled address data
- Currently EICR and EPC certificates are never provided during registration (only gas safety can be uploaded), so all occupied properties will show the compliance-later variant. The complete-compliance variant will activate when compliance is integrated into registration in a future ticket
- The "What you need to next" heading in the Figma design was a typo — corrected to "What you need to do next" as confirmed with the team

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
@@ -93,16 +93,16 @@ class RegisterPropertyController(
             RegistrationNumberDataModel.fromRegistrationNumber(propertyOwnership.registrationNumber).toString(),
         )
 
-        val hasIncompleteCompliance =
+        val actionRequiredForCompliance =
             if (propertyOwnership.isOccupied) {
                 val compliance = propertyComplianceService.getComplianceForPropertyOrNull(propertyOwnership.id)
                 compliance == null || compliance.isGasSafetyCertMissing || compliance.isEicrMissing || compliance.isEpcMissing
             } else {
                 false
             }
-        model.addAttribute("hasIncompleteCompliance", hasIncompleteCompliance)
+        model.addAttribute("actionRequiredForCompliance", actionRequiredForCompliance)
 
-        if (hasIncompleteCompliance) {
+        if (actionRequiredForCompliance) {
             val completeByDate =
                 CompleteByDateHelper.getIncompletePropertyCompleteByDateFromCreatedDate(propertyOwnership.createdDate)
             val formattedDate =
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
@@ -1,6 +1,7 @@
 package uk.gov.communities.prsdb.webapp.controllers
 
 import jakarta.servlet.http.HttpServletRequest
+import kotlinx.datetime.toJavaLocalDate
 import org.apache.commons.fileupload2.core.FileItemInputIterator
 import org.springframework.http.HttpStatus
 import org.springframework.http.MediaType
@@ -27,6 +28,7 @@ import uk.gov.communities.prsdb.webapp.constants.TASK_LIST_PATH_SEGMENT
 import uk.gov.communities.prsdb.webapp.controllers.LandlordController.Companion.LANDLORD_DASHBOARD_URL
 import uk.gov.communities.prsdb.webapp.controllers.RegisterPropertyController.Companion.PROPERTY_REGISTRATION_ROUTE
 import uk.gov.communities.prsdb.webapp.helpers.CertificateUploadHelper
+import uk.gov.communities.prsdb.webapp.helpers.CompleteByDateHelper
 import uk.gov.communities.prsdb.webapp.helpers.PropertyComplianceJourneyHelper
 import uk.gov.communities.prsdb.webapp.journeys.FormData
 import uk.gov.communities.prsdb.webapp.journeys.JourneyIdProvider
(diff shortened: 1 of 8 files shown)


Bill-Haigh:
### Screenshot: Occupied property with incomplete compliance

confirmation-occupied-incomplete

Note: Screenshot shows the live confirmation page after completing a property registration journey locally.

Bill-Haigh:
> ### Screenshot: Occupied property with incomplete compliance
> confirmation-occupied-incomplete
> 
> Note: Screenshot shows the live confirmation page after completing a property registration journey locally.

nice try copilot but not quite

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
```diff
@@ -81,13 +87,29 @@ class RegisterPropertyController(
...
-        model.addAttribute("singleLineAddress", propertyOwnership.address.singleLineAddress)
+        model.addAttribute("addressParts", propertyOwnership.address.singleLineAddress.split(", "))
         model.addAttribute(
             "prn",
             RegistrationNumberDataModel.fromRegistrationNumber(propertyOwnership.registrationNumber).toString(),
         )
-        model.addAttribute("isOccupied", propertyOwnership.isOccupied)
-        model.addAttribute("propertyComplianceUrl", PropertyComplianceController.getPropertyCompliancePath(propertyOwnership.id))
+
+        val hasIncompleteCompliance =
```
JasminConterioSW: I'm not keen on this variable name... does something like `actionRequiredForCompliance` work?

The compliance can still be complete if the property is not occupied. 
Alternatively, the property can have an incomplete compliance when the property is not occupied (it doesn't matter, but this variable would show hasIncompleteCompliance as false).

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
```diff
@@ -81,13 +87,29 @@ class RegisterPropertyController(
...
+                false
+            }
+        model.addAttribute("hasIncompleteCompliance", hasIncompleteCompliance)
+
+        if (hasIncompleteCompliance) {
+            val completeByDate =
+                CompleteByDateHelper.getIncompletePropertyCompleteByDateFromCreatedDate(propertyOwnership.createdDate)
+            val formattedDate =
+                completeByDate.toJavaLocalDate().format(DateTimeFormatter.ofPattern("d MMMM yyyy", Locale.UK))
+            model.addAttribute("completeByDate", formattedDate)
```
Travis-Softwire, line 110: Why do we need explicit formatting here? Is it because we're doing string interpolation in the template? In general Spring should automatically format datetimes for us, and we should avoid trying to do it ourselves if we can to avoid needing to change it in lots of places if we decided to change the format.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
```diff
@@ -81,13 +87,29 @@ class RegisterPropertyController(
...
             "prn",
             RegistrationNumberDataModel.fromRegistrationNumber(propertyOwnership.registrationNumber).toString(),
         )
-        model.addAttribute("isOccupied", propertyOwnership.isOccupied)
-        model.addAttribute("propertyComplianceUrl", PropertyComplianceController.getPropertyCompliancePath(propertyOwnership.id))
+
+        val hasIncompleteCompliance =
+            if (propertyOwnership.isOccupied) {
+                val compliance = propertyComplianceService.getComplianceForPropertyOrNull(propertyOwnership.id)
+                compliance == null || compliance.isGasSafetyCertMissing || compliance.isEicrMissing || compliance.isEpcMissing
```
JasminConterioSW, line 99: (This bit I would call hasIncompleteCompliance, although you don't need to make it a variable)

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyControllerTests.kt
```diff
@@ -67,7 +79,50 @@ class RegisterPropertyControllerTests(
...
+    fun `getConfirmation returns 200 with correct model attributes for an occupied property with incomplete compliance`() {
+        val propertyRegistrationNumber = 0L
+        val propertyOwnership =
+            createPropertyOwnership(
+                registrationNumber = RegistrationNumber(RegistrationNumberType.PROPERTY, propertyRegistrationNumber),
+                currentNumTenants = 2,
+                currentNumHouseholds = 1,
+                numberOfBedrooms = 1,
+                furnishedStatus = uk.gov.communities.prsdb.webapp.constants.enums.FurnishedStatus.FURNISHED,
+                rentFrequency = uk.gov.communities.prsdb.webapp.constants.enums.RentFrequency.MONTHLY,
```
Travis-Softwire: Can we import rather than using fully qualified names please?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
Travis-Softwire, line 99: Just to check, is this only for missing and not for expired certs?

Travis-Softwire (approved):
A few minor points, but LGTM once those are addressed

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
Bill-Haigh: Yes, it's because the date is passed as a parameter to a message interpolation: `#{...paragraph(${completeByDate})}` which uses `{0}` in the YAML message. Spring's `mvc.format.date` auto-formatting doesn't apply to message parameters — `MessageFormat` would just call `toString()` on the object, giving ISO format (`2026-05-11`) rather than the user-friendly `5 May 2026`. So explicit formatting is needed here to get the right display format.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
Bill-Haigh: yes just for missing

PR #1218, opened 2026-04-14
PDJB-NONE: fix occupied expired low energy routing

TomHanmer29:
## Ticket number

PDJB-NONE

## Goal of change

Fixes the case where a property was occupied, expired and had a low energy rating. 

Previously an occupied expired property where the  tenants moved in before the cert expired would just route straight to done even if the cert was below an E rating. This makes it route through the MEES exemption step first, if it's low energy. We previously had the expired and low energy check as 1 step but this doesn't work in this 1 case. See the diagram to verify:

image

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/EpcAgeCheckStepConfig.kt
```diff
@@ -0,0 +1,27 @@
...
+    }
+}
+
+@JourneyFrameworkComponent
+final class EpcAgeCheckStep(
+    stepConfig: EpcAgeCheckStepConfig,
+) : JourneyStep.InternalStep<EpcAgeCheckMode, EpcState>(stepConfig)
+
+enum class EpcAgeCheckMode {
+    EPC_CURRENT,
```
JasminConterioSW: Maybe call it EPC_NEWER_THAN_10_YEARS

EPC_CURRENT gets confusing when superseded epc's exist (in theory, one could be newer than 10 years but also superseded and therefore not current)

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/EpcEnergyRatingCheckStepConfigTests.kt
```diff
@@ -0,0 +1,59 @@
...
+    @Test
+    fun `mode throws exception when acceptedEpc is null`() {
+        val stepConfig = EpcEnergyRatingCheckStepConfig()
+        whenever(mockState.acceptedEpc).thenReturn(null)
+
+        assertThrows<NotNullFormModelValueIsNullException> { stepConfig.mode(mockState) }
+    }
+
+    @Test
+    fun `mode returns BELOW_THRESHOLD when energy rating is below E`() {
```
JasminConterioSW: The enum values in the test names aren't right in these tests

JasminConterioSW (approved):
A couple of naming points, but other than that LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/EpcAgeCheckStepConfig.kt
TomHanmer29: What about EPC_10_YEARS_OR_NEWER? that way it captures the edge case too

PR #1222, opened 2026-04-14
PDJB-653: Give unique names to file uploads

AEPR:
## Ticket number
PDJB-653

## Goal of change
Multiple file uploads not working when the file names clash.

Ensures the member id is added to the URL when adding another upload.

### DIFF
--- a/src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/CheckElectricalCertUploadsStepConfigTests.kt
+++ b/src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/CheckElectricalCertUploadsStepConfigTests.kt
@@ -9,6 +9,7 @@ import org.junit.jupiter.api.extension.ExtendWith
 import org.mockito.Mock
 import org.mockito.Mockito.mock
 import org.mockito.junit.jupiter.MockitoExtension
+import org.mockito.kotlin.anyOrNull
 import org.mockito.kotlin.whenever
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/CheckElectricalCertUploadsStepConfigTests.kt
```diff
@@ -142,6 +143,7 @@ class CheckElectricalCertUploadsStepConfigTests {
         fun `getStepSpecificContent returns empty upload rows when map is empty`() {
             setupStepMocks(includeRemoveStep = false)
             whenever(mockState.electricalUploadMap).thenReturn(emptyMap())
+            whenever(mockMemberIdService.createParameterPair(anyOrNull())).thenReturn("memberId" to "0")
```
Travis-Softwire, line 146: Now that the add another link accesses nextElectricalUploadMemberId, for all of the other tests (e.g. the one above this one) that call getStepSpecificContent need to mock `createParameterPair(1)` as well? I think it works at the moment because Mokito returns null for unstubbed calls on a mock, but isn't the actual behaviour of the running code.

Travis-Softwire (approved):
Is it worth adding a test for the codepath that returns the random suffix? Appreciate we're not currently using that other than as a defensive fallback, but if we're going to include it then we should test it.

One other comment on mocking in the tests, but the solution itself LGTM

PR #1223, opened 2026-04-14
PDJB-320: Wire up property update confirmation emails to step configs

Bill-Haigh:
## Ticket number

PDJB-320

## Goal of change

Fix property update confirmation emails not being sent after a property is updated.

- The email-sending code was in `PropertyOwnershipService.updatePropertyOwnership()`, which was never called — the actual update flows use individual granular methods (`updateOwnershipType`, `updateLicensing`, etc.) called from journey step configs
- Moved email sending into the 4 step configs that perform property updates: ownership type, licensing, occupancy, and households/tenants — following the established convention where emails are sent from the journey layer (see `PropertyComplianceCyaStepConfig`, `ReasonStepConfig`)
- Removed the dead `updatePropertyOwnership()` method, its helper, the unused `PropertyOwnershipUpdateModel` data class, and associated tests

Each update type now sends its own email (rather than one combined email). This matches the architecture — each update is an independent journey with its own step config.

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/householdsAndTenants/UpdateHouseholdsAndTenantsCyaConfig.kt
```diff
@@ -41,6 +46,24 @@ class UpdateHouseholdsAndTenantsCyaConfig(
...
+    private fun sendUpdateConfirmationEmail(state: UpdateHouseholdsAndTenantsJourneyState) {
+        val propertyOwnership = propertyOwnershipService.getPropertyOwnership(state.propertyId)
+        updateConfirmationEmailService.sendEmail(
+            propertyOwnership.primaryLandlord.email,
+            PropertyUpdateConfirmation(
+                name = propertyOwnership.primaryLandlord.name,
+                multiLineAddress = propertyOwnership.address.toMultiLineAddress(),
+                updatedItems =
+                    listOf(
+                        "The number of households living in this property",
```
TomHanmer29, line 61: Should these live in a yaml somewhere or is it fine to just hardcode them?

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipServiceTests.kt
```diff
@@ -652,224 +636,6 @@ class PropertyOwnershipServiceTests {
...
-
-        // Assert
-        assertEquals(updateModel.ownershipType, propertyOwnership.ownershipType)
-        assertEquals(updateModel.numberOfHouseholds, propertyOwnership.currentNumHouseholds)
-        assertEquals(updateModel.numberOfPeople, propertyOwnership.currentNumTenants)
-        assertEquals(updateModel.licensingType, propertyOwnership.license?.licenseType)
-        assertEquals(updateModel.licenceNumber, propertyOwnership.license?.licenseNumber)
-    }
-
-    @MethodSource("updatesAndConfirmationEmails")
```
TomHanmer29, line 764: We've removed a very large number of email sending tests in this file. Do we have an equivalent for the new email sending process? Ideally we'd want to use a similar data source and test each of the new ones. I.e. 
`verify(emailNotificationService).sendEmail(`

TomHanmer29 (changes requested):
I think it's probably a good idea to reproduce some of these deleted tests for the new methods (unless they already exist, in which case, please say 😆 )

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/householdsAndTenants/UpdateHouseholdsAndTenantsCyaConfig.kt
Bill-Haigh: the existing email models use hardcoded messages but perhaps there is benefit to updating the email models to use yaml files for messages...

TomHanmer29 (approved):

PR #1224, opened 2026-04-14
PDJB-763: Hide One Login header elements for internal access users

Travis-Softwire:
## Ticket number

PDJB-763

## Goal of change

Hide One Login-specific header elements for local council users who authenticate via Internal Access rather than One Login.

- Hides the One Login toggle button and account link for local council users, keeping only the sign-out link
- Ensures the sign-out link remains accessible on mobile when the toggle button is hidden
- Renames the sign-out message key from `oneLoginHeader.logout` to `authenticatedHeader.signOut`
- Hides the service navigation menu when there are no nav links to display
- Simplifies the service navigation toggle button text from `<Service Name> Menu` to `Menu`

The vertical separator on the sign-out link remains visible for LC users. Removing it required custom CSS to override the GOV.UK One Login Service Header styles, which was not worth the complexity for a minor visual issue.

## Checklist

image
image


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/controllerAdvice/GlobalModelAttributes.kt
```diff
@@ -47,7 +47,9 @@ class GlobalModelAttributes(
         model.addAttribute("feedbackBannerUrl", FEEDBACK_URL)
 
         // Authenticated header attributes
+        val uri = request.requestURI
         model.addAttribute("confirmSignOutUrl", "/$CONFIRM_SIGN_OUT_PATH_SEGMENT")
+        model.addAttribute("showOneLoginNav", !uri.startsWith("/$LOCAL_COUNCIL_PATH_SEGMENT"))
```
TomHanmer29: Does this definitely exhaustively cover all pages a local council user could see? I'm thinking cookies (although I think they've been removed now), privacy etc. - anything which may not start in local-council

TomHanmer29 (approved):
LGTM! One tiny question but otherwise all great

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/controllerAdvice/GlobalModelAttributes.kt
Travis-Softwire: Good catch - I'd forgotten about those. Given that, path segments isn't going to be the best approach - will switch to using user roles instead

TomHanmer29 (approved):

PR #1230, opened 2026-04-16
Feat/pdjb 754 update section header component

JasminConterioSW:
## Ticket number

PDJB-754

## Goal of change

Update section header component

* Update section headers to use the "caption" style
* On the property registration journey, include compliance tasks in the "Register you property details" section and include numbering
    * This is for consistency on all pages across the journey, and to make it match the task list that will be in the private beta launch
* Removed unused pageCaption variables
* Removed caption from nonEnglandOrWalesAddressForm (need to check with design if this is ok) - it's the only page in landlord registration with a caption

image

We may or may not want to include checking all the section headers in integration tests - this will be in separate PR so can be dropped if required

I looked for other usages of the caption style - there a couple of other places where it is used, but I think this is ok as they are not section headers (e.g. propertyDetailsView - it's not a journey step, it's just a page)

## Checklist

  mention that here


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationSinglePageTests.kt
```diff
@@ -213,7 +213,7 @@ class PropertyRegistrationSinglePageTests : IntegrationTestWithImmutableData("da
             val licenseNumberPage = assertPageIs(page, HmoMandatoryLicenceFormPagePropertyRegistration::class)
             BaseComponent
                 .assertThat(licenseNumberPage.form.sectionHeader)
-                .containsText("Section 1 of 5 \u2014 Register your property details")
+                .containsText("Section 1 of 2 \u2014 Register your property details")
```
TomHanmer29: Should we be replacing these with the actual character (and same in the above file)? I guess it's fine if they pass, and it does seem like the way we've done it for all the other tests, but we've used the opposite convention for curly apostrophes I think

TomHanmer29 (approved):
LGTM! One little Q

PR #1242, opened 2026-04-17
PDJB-NONE: Rework Address.toMultiLineAddress()

Bill-Haigh:
## Ticket number

PDJB-NONE

## Goal of change

Fix `Address.toMultiLineAddress()` so that multi-line addresses rendered in emails no longer have a spurious newline after the house number.

- Fix the fallback path (which is the only path exercised in production) so it no longer produces a newline after a bare house number. OS Places returns addresses like `1, SAVOY COURT, LONDON, WC2R 0EX` and the previous `replace(", ", "\n")` implementation split on every comma, pushing the house number onto its own line. The new parser merges a leading `\d+[A-Za-z]?` token with the next token (unless that next token is a UK postcode).
- In the structured-component path, move `buildingName` to before the `buildingNumber streetName` line so the output matches Royal Mail PAF conventions.

In production, `Address` entities only ever populate `singleLineAddress` / `uprn` / `townName` / `postcode` / `localCouncil` — the structured component fields are only populated by tests and `NftDataFaker`. So real email sends from `PropertyOwnershipService.getPropertyUpdateConfirmationEmail()` and `JointLandlordInvitationService.sendInvitation()` always hit the fallback path. The component-path ordering fix is for correctness, but the fallback fix is what callers will actually see.

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Address.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Address.kt
@@ -82,7 +82,7 @@ class Address() : ModifiableAuditableEntity() {
         if (hasAddressComponents()) {
             buildMultiLineAddressFromComponents()
         } else {
-            singleLineAddress.replace(", ", "\n")
+            buildMultiLineAddressFromSingleLine(singleLineAddress)
         }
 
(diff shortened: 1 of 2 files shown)


Travis-Softwire:
Let's not do this now - there's a ticket to look at address formatting more generally as I'm not convinced the current parser handles all edge cases anyway (which is why were only using single line address).

PR #1245, opened 2026-04-17
PDJB-682: Note allow pasting requirement in passcode script

Travis-Softwire:
## Ticket number

PDJB-682

## Goal of change

Note the Chrome "allow pasting" requirement in the bulk passcode generation script so users do not get stuck when pasting into the browser console.

- Adds a note to the usage instructions at the top of `scripts/generate_passcodes.js` explaining that if the browser console blocks pasting, the user needs to type "allow pasting" and press Enter before pasting the script.

Follow-up from QA feedback on PDJB-682.


JasminConterioSW (approved):
👍

PR #1246, opened 2026-04-19
PDJB-NONE: Strip query parameters from Plausible analytics events

Travis-Softwire:
## Ticket number

PDJB-NONE

## Goal of change

Stop sending query parameters from page URLs to Plausible analytics.

- Migrates the Plausible tracker from the legacy v1 script (`script.file-downloads.hash.outbound-links.js`) to the v2 site-specific script (`pa-<siteId>.js`), which supports the `transformRequest` init option.
- Adds a `transformRequest` callback that strips the query string from every event URL before it leaves the browser, following the approach in the [Plausible docs](https://plausible.io/docs/stop-tracking-utm-tags#set-transformrequest-option-in-plausibleinit). This removes UTM tags and other params that Plausible would otherwise retain.
- Renames the `plausible.domain-id` config (env var `PLAUSIBLE_ANALYTICS_DOMAIN_ID`) to `plausible.site-id` (env var `PLAUSIBLE_SITE_ID`), since the value semantics change from a Plausible "domain" to the v2 per-site script ID.
- Drops the now-unused `data-domain` attribute and `plausibleDomainId` model attribute (v2 site-specific scripts have the domain baked in).
- Updates `docs/AnalyticsReadMe.md` to describe the query-stripping behaviour and note that optional measurements (outbound links, file downloads, hashed paths) are now configured in the Plausible dashboard rather than via the script filename.

- **Infra change required before merging to any deployed environment**: the env var must be renamed from `PLAUSIBLE_ANALYTICS_DOMAIN_ID` to `PLAUSIBLE_SITE_ID`, and its value must be set to the v2 per-site script ID (e.g. `z_RRGZRChZ_O2RXl2r5-S` for integration). Get this from the Plausible dashboard under Site Settings → Site Installation → Review Installation. Each environment has its own site ID.
- Optional measurements (outbound links, file downloads, hashed page paths) were previously enabled via the script URL filename. They now need enabling per-site in the Plausible dashboard.

## Checklist


docs/AnalyticsReadMe.md
```diff
@@ -8,9 +8,18 @@ Both sets of analytics are added via code snippets on the "layout.html" page.
 
 ## Plausible
 
-Plausible has been set up to track optional events including outbound links, file downloads, 404 error pages and hashed page paths.
-All of those except for hashed page paths add to our billable pageviews.
-Modifying these requires a code change.
+Plausible has been set up to redact query parameters from tracked URLs (via `transformRequest` in
+the inline init script). By default Plausible already strips most query params but retains
+`ref`, `source`, `utm_source`, `utm_medium`, `utm_campaign`, `utm_content` and `utm_term`; our
```
JasminConterioSW, line 13: I was trying to see if this is going to interfere with us collecting the custom events for generating Sankey diagrams, but I'm not obviously seeing where we do this... _do_ we do this? Did it ever make it into the codebase? 

There are some unmerged branches [here](https://github.com/communitiesuk/prsdb-webapp/branches/all?query=1610&lastTab=overview)

JasminConterioSW: (It looks like it would be fine with the approach that's on that branch anyway as we would use the referrer header rather than an url query parameter)

JasminConterioSW (approved):
LGTM :)

docs/AnalyticsReadMe.md
Travis-Softwire: Good spot - will raise a ticket to add the custom event on main

PR #1247, opened 2026-04-20
PDJB-806: Save compliance data

TomHanmer29:
## Ticket number

PDJB-806

## Goal of change

Save the compliance data at the end of the compliance journey

Adds a new internal step "SaveComplianceDataStepConfig" which saves the property compliance info
This is to prepare for a separate screen around compliance data (and why it's not just done in the cya step)
Adds a migration to add eicr_expiry_date given that's the value we're now saving in the journey
Adds TODOs for PDJB-766 to remove eicrIssueDate when we use expiry date in the update journey too
Adds unit tests

Not really sure about saving the registration number in state, but I think we need to if we want the compliance saving in a separate step

## Checklist

  mention that here


src/main/resources/db/migrations/V1_23_0__add_eicr_expiry_date.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE property_compliance ADD COLUMN eicr_expiry_date DATE;
```
Travis-Softwire: Can you add a TODO on the EICR update journey ticket to drop the issue date column as well please?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/dao/NftDataSeederDao.kt
```diff
@@ -120,9 +120,9 @@ class NftDataSeederDao(
             INSERT INTO property_compliance 
             (created_date, last_modified_date, property_ownership_id, gas_safety_upload_id, gas_safety_cert_issue_date, 
              gas_safety_cert_engineer_num, gas_safety_cert_exemption_reason, gas_safety_cert_exemption_other_reason, eicr_upload_id, 
-             eicr_issue_date, eicr_exemption_reason,eicr_exemption_other_reason, epc_url, epc_expiry_date, 
-             tenancy_started_before_epc_expiry, epc_energy_rating, epc_exemption_reason, epc_mees_exemption_reason)
-            VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?,?,?)
+             eicr_issue_date, eicr_exemption_reason, eicr_exemption_other_reason, epc_url, epc_expiry_date,
```
Travis-Softwire: Can we get a TODO here referencing the EICR update journey ticket to remove EICR issue date from here (and anywhere else it's used)?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
```diff
@@ -53,12 +52,15 @@ class PropertyCompliance() : ModifiableAuditableEntity() {
     @JoinTable(name = "electrical_safety_uploads")
     var electricalSafetyFileUploads: MutableList<FileUpload> = mutableListOf()
 
+    // TODO PDJB-766: Remove eicrIssueDate once the compliance update journey uses expiry date instead
```
Travis-Softwire, line 57: Ah, you've quite possibly already done it! :D

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/PropertyRegistrationCyaStepConfig.kt
```diff
@@ -65,59 +65,62 @@ class PropertyRegistrationCyaStepConfig(
...
-                    } else {
-                        null
-                    },
-                baseUserId = SecurityContextHolder.getContext().authentication.name,
-                jointLandlordEmails = jointLandlordEmails,
-                gasSafetyFileUploadIds = state.gasUploadIds,
-                electricalSafetyFileUploadIds = state.electricalUploadIds,
-            )
+            val registrationNumber =
+                propertyRegistrationService.registerProperty(
```
Travis-Softwire: Ah, no - the whole property registration needs to happen in the internal step - as if they're missing compliance info and the property is occupied they'll see a warning screen after this page, at which point they could abort and the property shoudn't be registered

Travis-Softwire (changes requested):
Sorry - crossed wires - it's the whole of the registration that needs to happen in the internal step, not just compliance. That's because CYA won't now be the last step in the journey in some scenarios

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/PropertyRegistrationCyaStepConfig.kt
TomHanmer29: Ahh gotcha - that makes more sense. Will shift it over :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/dao/NftDataSeederDao.kt
```diff
@@ -120,9 +120,9 @@ class NftDataSeederDao(
             INSERT INTO property_compliance 
             (created_date, last_modified_date, property_ownership_id, gas_safety_upload_id, gas_safety_cert_issue_date, 
              gas_safety_cert_engineer_num, gas_safety_cert_exemption_reason, gas_safety_cert_exemption_other_reason, eicr_upload_id, 
-             eicr_issue_date, eicr_exemption_reason,eicr_exemption_other_reason, epc_url, epc_expiry_date, 
-             tenancy_started_before_epc_expiry, epc_energy_rating, epc_exemption_reason, epc_mees_exemption_reason)
-            VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?,?,?)
+             eicr_issue_date, eicr_exemption_reason, eicr_exemption_other_reason, epc_url, epc_expiry_date, -- TODO PDJB-766: Remove eicr_issue_date once the compliance update journey uses expiry date instead
```
Travis-Softwire, line 123: Can we get a TODO here RE removing the issue date as well?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/PropertyRegistrationCyaStepConfig.kt
```diff
@@ -57,81 +46,10 @@ class PropertyRegistrationCyaStepConfig(
...
     override fun resolveNextDestination(
         state: PropertyRegistrationJourneyState,
         defaultDestination: Destination,
-    ): Destination =
-        if (state.isAddressAlreadyRegistered == true) {
-            Destination(state.alreadyRegisteredStep)
-        } else {
-            super.resolveNextDestination(state, defaultDestination)
-        }
+    ): Destination = defaultDestination
```
Travis-Softwire, line 52: I genuinely can't remember, but do you need the override if it just points at the defaultDestination?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfig.kt
```diff
@@ -0,0 +1,154 @@
...
+    override fun afterStepIsReached(state: PropertyRegistrationJourneyState) {
+        val registrationNumberValue: Long
+        try {
+            registrationNumberValue = registerProperty(state)
+        } catch (_: EntityExistsException) {
+            state.isAddressAlreadyRegistered = true
+            return
+        }
+
+        saveRegistrationData(state, registrationNumberValue)
```
Travis-Softwire: Why is this not just part of the registerProperty function?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -98,6 +104,53 @@ class PropertyComplianceService(
...
+                epcExemptionReason != null ||
+                epcMeesExemptionReason != null
+
+        val propertyOwnership =
+            propertyOwnershipRepository.findByRegistrationNumber_Number(registrationNumberValue)
+                ?: throw EntityNotFoundException("Property ownership not found for registration number $registrationNumberValue")
+
+        val compliance = getComplianceForPropertyOrNull(propertyOwnership.id)
+
+        if (compliance == null && !hasAnyData) return
```
Travis-Softwire: This is probably fine in practice, but I think I prefer us creating a compliance record with all null values apart from the registration number vs not creating the record - previously these meant two different things (hasn't completed compliance journey vs having no certs) and there's a risk our logic somewhere might subtly rely on that.

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfigTests.kt
```diff
@@ -0,0 +1,377 @@
...
+                propertyRegistrationService = mockPropertyRegistrationService,
+                propertyComplianceService = mockPropertyComplianceService,
+                epcCertificateUrlProvider = mockEpcCertificateUrlProvider,
+                jointLandlordsStrategy = mockJointLandlordsStrategy,
+            )
+    }
+
+    @AfterEach
+    fun tearDown() {
+        SecurityContextHolder.clearContext()
```
Travis-Softwire, line 79: Out of interest, why do we need to do this?

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceServiceTests.kt
```diff
@@ -793,4 +798,116 @@ class PropertyComplianceServiceTests {
...
         assertFalse(propertyComplianceService.wasPropertyComplianceAddedThisSession(propertyOwnershipId))
     }
+
+    @Nested
+    inner class SaveRegistrationComplianceData {
+        private val mockPropertyOwnership = MockLandlordData.createPropertyOwnership()
+        private val registrationNumberValue = 12345L
+
+        @Test
+        fun `updates existing compliance record when one exists`() {
```
Travis-Softwire: Hmm, are we expecting this function to be used for compliance updates? Appreciate it was for the old compliance update journey but that's because the old update journeys always overwrote the whole record, whereas the new ones should only touch the relevant properties of the record?

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceServiceTests.kt
```diff
@@ -793,4 +798,116 @@ class PropertyComplianceServiceTests {
...
+            assertEquals(epcUrl, saved.epcUrl)
+            assertEquals(epcExpiryDate, saved.epcExpiryDate)
+            assertEquals(epcEnergyRating, saved.epcEnergyRating)
+            assertEquals(true, saved.tenancyStartedBeforeEpcExpiry)
+            assertEquals(epcExemptionReason, saved.epcExemptionReason)
+            assertEquals(meesExemptionReason, saved.epcMeesExemptionReason)
+        }
+
+        @Test
+        fun `does nothing when no data to save and no existing record`() {
```
Travis-Softwire: As above, I don't actually think this is the behaviour we want any more

src/test/kotlin/uk/gov/communities/prsdb/webapp/testHelpers/mockObjects/MockPropertyComplianceData.kt
```diff
@@ -43,6 +44,7 @@ class MockPropertyComplianceData {
             gasSafetyCertExemptionOtherReason = gasSafetyCertExemptionOtherReason,
             eicrUpload = eicrFileUpload,
             eicrIssueDate = eicrIssueDate,
+            eicrExpiryDate = eicrIssueDate?.plusYears(EICR_VALIDITY_YEARS.toLong()),
```
Travis-Softwire: Given we're expecting to remove the issue date, should this be the other way around? Or indeed, since we're expecting it to be null until we remove it should issue date just be null?

Travis-Softwire (approved):
A few points that I think need changing, but if you agree w/ the suggestions then I don't need to see it again. Shout if any of it doesn't make sense though!

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/dao/NftDataSeederDao.kt
TomHanmer29: I think thats what the TODO that's there says? Did you mean something else?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/PropertyRegistrationCyaStepConfig.kt
TomHanmer29: I don't think so - have removed

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfig.kt
TomHanmer29: It should be :) refactoring fail

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfigTests.kt
TomHanmer29: SteupStateForPropertyRegistration calls setMockUser which sets a security context, so I'm just clearing that state between tests

src/test/kotlin/uk/gov/communities/prsdb/webapp/testHelpers/mockObjects/MockPropertyComplianceData.kt
TomHanmer29: Since this is used in some of the old compliance tests as well, I think it's easiest just to mark this bit with the same TODO as the rest of them had - I just missed it here.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/PropertyRegistrationCyaStepConfig.kt
TomHanmer29: Ah hang on - you do need it specifically here. AbstractCheckYourAnswersStepConfig calls state.deleteJourney() (as we expect it to be the last page in a journey) so adding it in here prevents that call specifically for this page

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -98,6 +104,41 @@ class PropertyComplianceService(
     fun getComplianceForPropertyOrNull(propertyOwnershipId: Long): PropertyCompliance? =
         propertyComplianceRepository.findByPropertyOwnership_Id(propertyOwnershipId)
 
+    @Transactional
+    fun saveRegistrationComplianceData(
```
JasminConterioSW, line 97: Just above this there is also createPropertyCompliance which includes the gas and electric uploads, should we be including those uploads here instead? We might not need the createPropertyCompliance that starts on line 91...

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfig.kt
```diff
@@ -0,0 +1,146 @@
...
+                    if (isOccupied) {
+                        state.rentAmount.formModel.rentAmount
+                            .toBigDecimal()
+                    } else {
+                        null
+                    },
+                baseUserId = SecurityContextHolder.getContext().authentication.name,
+                jointLandlordEmails = jointLandlordEmails,
+                gasSafetyFileUploadIds = state.gasUploadIds,
+                electricalSafetyFileUploadIds = state.electricalUploadIds,
```
JasminConterioSW, line 115: These (along with the handling of the file uploads which is currently in `PropertyRegistrationService.registerProperty(...)`) should probably also be passed to `propertyComplianceService.saveRegistrationComplianceData(...)`

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
TomHanmer29: You're completely right - I totally missed this 🙃 . I'll refactor to combine these

JasminConterioSW: I've had another thought, what do we do about properties with no gas supply?
This used to get recorded as an exemption, but we don't ask for those anymore.

Should we just save it as a GasSafetyExemptionReason.NO_GAS_SUPPLY?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/dao/NftDataSeederDao.kt
```diff
@@ -120,9 +120,9 @@ class NftDataSeederDao(
...
+             tenancy_started_before_epc_expiry, epc_energy_rating, epc_exemption_reason, epc_mees_exemption_reason, eicr_expiry_date)
```
JasminConterioSW: You might already be doing this, but does has_gas_safety also need adding here and to NftDataFaker?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfig.kt
```diff
@@ -0,0 +1,150 @@
...
+            )
+        propertyComplianceService.saveRegistrationComplianceData(
+            registrationNumberValue = registrationNumber.number,
+            hasGasSupply = state.hasGasSupplyStep.outcome == YesOrNo.YES,
+            gasSafetyCertIssueDate = state.getGasSafetyCertificateIssueDateIfReachable()?.toJavaLocalDate(),
+            gasSafetyFileUploadIds = state.gasUploadIds,
+            electricalSafetyFileUploadIds = state.electricalUploadIds,
+            eicrExpiryDate = state.getElectricalCertificateExpiryDateIfReachable()?.toJavaLocalDate(),
```
JasminConterioSW: Can you rename this to `electricalSafetyExpiryDate`?

(It used to always be an eicr, but now it could be eicr or ecr)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -46,6 +49,7 @@ class PropertyComplianceService(
         gasSafetyCertExemptionReason: GasSafetyExemptionReason? = null,
         gasSafetyCertExemptionOtherReason: String? = null,
         eicrUploadId: Long? = null,
+        // TODO PDJB-766: Remove eicrIssueDate once the compliance update journey uses expiry date instead
```
JasminConterioSW, line 53: I think you can use the expiry date now... that's what we ask for in the property registration journey.

I don't think the update journey will use this function, it'll always be updating an existing compliance.

If it's easier, keep the issueDate as it's used in the standalone (old) compliance journey, with a TODO to delete it when we remove that (I don't know if we have a ticket for that, but if not you could put one in the backlog!)

JasminConterioSW: Actually, I think this is only used by the old compliance journey, so the whole function can be deleted then

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
JasminConterioSW, line 152: I suspect we won't need this... We could delete it here or Maybe add a TODO (tagging PDJB-764, PDJB-765 and PDJB-766 i.e. the update journeys) to delete it when we are sure we won't need it.

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfigTests.kt
```diff
@@ -0,0 +1,425 @@
...
+        // Act
+        val result = stepConfig.resolveNextDestination(mockState, defaultDestination)
+
+        // Assert
+        verify(mockState).deleteJourney()
+        assertEquals(defaultDestination, result)
+    }
+
+    @Test
+    fun `resolveNextDestination deletes journey and redirects to already registered step when address is already registered`() {
```
JasminConterioSW: Is this right - can it go to a step in the journey if it has deleted the journey? Maybe that should be on the already registered step...

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfig.kt
```diff
@@ -0,0 +1,150 @@
...
+            state.isAddressAlreadyRegistered = true
+            return
+        }
+    }
+
+    override fun resolveNextDestination(
+        state: PropertyRegistrationJourneyState,
+        defaultDestination: Destination,
+    ): Destination {
+        state.deleteJourney()
```
JasminConterioSW: I commented on this on the test - but should it only delete the journey if we are not already registered?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfig.kt
```diff
@@ -0,0 +1,150 @@
...
+                baseUserId = SecurityContextHolder.getContext().authentication.name,
+                jointLandlordEmails = jointLandlordEmails,
+                gasSafetyFileUploadIds = state.gasUploadIds,
+                electricalSafetyFileUploadIds = state.electricalUploadIds,
+            )
+        propertyComplianceService.saveRegistrationComplianceData(
+            registrationNumberValue = registrationNumber.number,
+            hasGasSupply = state.hasGasSupplyStep.outcome == YesOrNo.YES,
+            gasSafetyCertIssueDate = state.getGasSafetyCertificateIssueDateIfReachable()?.toJavaLocalDate(),
+            gasSafetyFileUploadIds = state.gasUploadIds,
```
JasminConterioSW, line 121: If a user goes all the way through and uploads files, but then goes back and says they don't have a gas supply (for example), is this still going to save the uploads in the compliance record and then show as having a gas certificate?

So even though the upload steps are not longer reachable, will it save the data that was previously added?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfig.kt
```diff
@@ -0,0 +1,150 @@
...
+            electricalSafetyFileUploadIds = state.electricalUploadIds,
+            eicrExpiryDate = state.getElectricalCertificateExpiryDateIfReachable()?.toJavaLocalDate(),
+            epcCertificateUrl =
```
JasminConterioSW, line 124: Again if a user accepts an EPC, but then goes back and finishes with a route where the EPC is missing (or exempt), is this still going to save those acceptedEPC details even though those steps should no longer be reachable?

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfigTests.kt
```diff
@@ -0,0 +1,425 @@
...
+
+        whenever(mockState.getLicenceNumberOrNull()).thenReturn(null)
+
+        val mockOwnershipTypeStep = mock<OwnershipTypeStep>()
+        val ownershipTypeFormModel = OwnershipTypeFormModel().apply { ownershipType = OwnershipType.FREEHOLD }
+        whenever(mockState.ownershipTypeStep).thenReturn(mockOwnershipTypeStep)
+        whenever(mockOwnershipTypeStep.formModel).thenReturn(ownershipTypeFormModel)
+
+        whenever(mockState.gasUploadIds).thenReturn(emptyList())
+        whenever(mockState.electricalUploadIds).thenReturn(emptyList())
```
JasminConterioSW: You might not still want these here as they moved out of the registerProperty method

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfigTests.kt
```diff
@@ -0,0 +1,425 @@
...
+            epcExpiryDate = java.time.LocalDate.of(2030, 1, 1),
+            epcEnergyRating = "B",
+            tenancyStartedBeforeEpcExpiry = true,
+            epcExemptionReason = EpcExemptionReason.PROTECTED_ARCHITECTURAL_OR_HISTORICAL_MERIT,
+            epcMeesExemptionReason = MeesExemptionReason.HIGH_COST,
+        )
+    }
+
+    @Test
+    fun `afterStepIsReached passes null for unreachable compliance steps`() {
```
JasminConterioSW: I wonder if this needs a bit more (or more tests for particular cases)

Here, the uploads and acceptedEpc details are already set to null / empty, but does it still work if they are _not_ null and those steps are no longer reachable?

src/test/kotlin/uk/gov/communities/prsdb/webapp/testHelpers/builders/PropertyComplianceBuilder.kt
```diff
@@ -92,11 +92,14 @@ class PropertyComplianceBuilder {
     ): PropertyComplianceBuilder {
         propertyCompliance.eicrFileUpload = fileUpload
         propertyCompliance.eicrIssueDate = issueDate
+        propertyCompliance.eicrExpiryDate = issueDate.plusYears(EICR_VALIDITY_YEARS.toLong())
```
JasminConterioSW: If we're still using this method, the expiryDate should be passed in as that is now what we ask for (you could give it `issueDate.plusYears(EICR_VALIDITY_YEARS.toLong())` as a default?)

JasminConterioSW (changes requested):
Sorry, a few more things to check with the saving (and some other minor comments)

Mostly making sure that only data from the final user journey ends up on the compliance record...

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
TomHanmer29: We have a separate ticket to remove the old compliance journey which is why I've not ripped too much out here - but I can update the TODO with PDJB-812?

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfigTests.kt
TomHanmer29: Yeah, this test doesn't seem correct - it's because copilot saw the nextDestination step redirecting to the already registered page if the address is already registered, and didn't check the order of the flow. It assumed this happened in the cya page but that's an impossible scenario.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfig.kt
TomHanmer29: The already registered step happens when you enter the address so it's always fine to delete here.

JasminConterioSW: It looks like it gets set here too if we get an EntityExistsException when registering the property though?

TomHanmer29: Oooh yes I missed that! I can add in a check before the deleteJourney to be safe. Although saying that, we did previously do this on the cya journey for property registration without any kind of check. Maybe this was wrong though?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
TomHanmer29: I'll tag it with PDJB-812 again - as that should come after all of the above so we don't miss it :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfig.kt
TomHanmer29: No. in GasSafetyState.kt we return an empty list if uploadGasCertStep.isStepReachable is not true for gasUploadIds, and the same for electrical

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfig.kt
TomHanmer29: I'm not convinced we actually have a guard for this in the state in the same way as the others. I think we should though, what do you think (the logic may be more complex though?)

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfigTests.kt
TomHanmer29: I think if we're testing the guards on the uploads / epc stuff separately we don't need to cover this case? (You'd just be testing the guards again)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfig.kt
JasminConterioSW: On `PropertyComplianceCyaStepConfig` (for the standalone compliance journey), I did 

```
val epcDetails =
            if (state.checkMatchedEpcStep.isStepReachable || state.checkAutomatchedEpcStep.isStepReachable) {
                state.acceptedEpc
            } else {
                null
            }
```
Then used the new epcDetails to populate the compliance.

I don't know if we do something similar here or if we there's something we can re-use from the `EpcRegistrationCyaSummaryRowsFactory`/`CheckEpcAnswersStep`?
If we don't show an EPC card on that page, we shouldn't be saving it (presumably!)?

JasminConterioSW: Re-using something from` EpcRegistrationCyaSummaryRowsFactory` is probably best actually, would it make sense to use `isEpcCardShown`? Store it on the state somewhere as `acceptedEpcDetailsShouldBeSaved`?

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfigTests.kt
JasminConterioSW: Yes that should be fine, I think it's currently untested for EPC as it's currently unguarded. Sounds like uploads are covered already :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfig.kt
TomHanmer29: Yep, this is what I implemented but commonising with `EpcRegistrationCyaSummaryRowsFactory` is a great idea. I'll do that now.

TomHanmer29: Having looked into this, I think it's a bit weird to be setting state based off of a condition inside EpcRegistrationCyaSummaryRowsFactory since we use the state to determine this condition in the first place so it all feels a bit circular. I also don't want to have to initialise the factory inside the state in order to access the property - I think it actually makes more sense to set the state to the above snippet and then use that in the factory instead.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/EpcState.kt
```diff
@@ -33,6 +33,14 @@ interface EpcState : JourneyState {
     var updatedEpcRetrievedByCertificateNumber: EpcDataModel?
     var acceptedEpc: EpcDataModel?
 
+    val acceptedEpcIfReachable: EpcDataModel?
+        get() =
+            if (checkUprnMatchedEpcStep.isStepReachable || confirmEpcDetailsRetrievedByCertificateNumberStep.isStepReachable) {
```
JasminConterioSW: The checkSupersededEpcStep is another way to get to acceptedEpc, can you add that one here too?

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/EpcStateTests.kt
```diff
@@ -0,0 +1,106 @@
...
+        assertEquals(epc, state.acceptedEpcIfReachable)
+    }
+
+    @Test
+    fun `acceptedEpcIfReachable returns acceptedEpc when confirmEpcDetailsRetrievedByCertificateNumberStep is reachable`() {
+        val epc = mock<EpcDataModel>()
+        val state = buildTestEpcState(acceptedEpc = epc, uprnStepReachable = false, certStepReachable = true)
+
+        assertEquals(epc, state.acceptedEpcIfReachable)
+    }
```
JasminConterioSW, line 46: There'll be one more for the `checkSupersededEpcStep`

JasminConterioSW (approved):
There's one more step where the user can accept an EPC, but when that's added to acceptedEpcIfReachable this is good to go!

PR #1248, opened 2026-04-20
PDJB-670: fix logic error in epc cya page routing

TomHanmer29:
## Ticket number

PDJB-670

## Goal of change

Fixes a routing error when accepting a cert, then going back and changing the outcome or clicking a change link

when clicking change links, the accepted cert still remained in state. This is good as we want the user to be able to go back down their path with the existing cert. However, this flagged the cert as accepted which meant an EPC card was still shown for pages that it shouldn't have been on. This is now fixed as we use the outcome of a no specific path being non null.

## Checklist


JasminConterioSW (approved):
LGTM :)

PR #1254, opened 2026-04-21
Feat/pdjb 764 gas update journey

JasminConterioSW:
## Ticket number

PDJB-764

## Goal of change

Implement the update gas safety journey

Adds the UpdateGasSafety journey, re-using the existing GasSafetyTask
* Hide the section headers (there's no task list to match these to)
* Hide the "Provide this later" button and don't set HasGasSafetyCert's mode to "Provide this later" to close off that route through the journey
* Save the data with a new step
* Add a change link into the Property Record to access the journey

* Updated PDJB-80 TODOs with more specific ticket numbers
* Add LF line endings to copilot instructions to match .editorconfig

Change button to access the journey (design will be updated in another ticket)
image

Do you have a gas cert page?
Differences from property registration journey version include the page title on the tab, no section headers, no "Provide this later" button
image

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -249,4 +253,41 @@ class PropertyComplianceService(
...
+        gasSafetyCertIssueDate: LocalDate? = null,
+        gasSafetyCertUploadIds: List<Long> = listOf(),
+    ) {
+        val propertyCompliance = getComplianceForProperty(propertyOwnershipId)
+        throwErrorIfLastModifiedDatesConflict(propertyCompliance, initialLastModifiedDate)
+        propertyCompliance.gasSafetyCertIssueDate = gasSafetyCertIssueDate
+        propertyCompliance.gasSafetyCertExemptionReason = if (!hasGasSupply) GasSafetyExemptionReason.NO_GAS_SUPPLY else null
+        val gasUploads = gasSafetyCertUploadIds.map { fileUploadRepository.getReferenceById(it) }
+        propertyCompliance.gasSafetyFileUploads = gasUploads.toMutableList()
+        propertyComplianceRepository.save(propertyCompliance)
```
JasminConterioSW: Actually, we might be able to just call the new saveRegistrationComplianceData function from PDJB-806

JasminConterioSW: Will need to be careful to only update the gas fields though and not override anything else with null

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/UpdateGasSafetyControllerTests.kt
```diff
@@ -0,0 +1,150 @@
...
+import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.HasGasSupplyStep
+import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.UploadGasCertStep
+import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.update.gasSafety.UpdateGasSafetyJourneyFactory
+import uk.gov.communities.prsdb.webapp.services.FileUploadCookieService.Companion.FILE_UPLOAD_COOKIE_NAME
+import uk.gov.communities.prsdb.webapp.services.PropertyOwnershipService
+
+@WebMvcTest(UpdateGasSafetyController::class)
+class UpdateGasSafetyControllerTests(
+    @Autowired webContext: WebApplicationContext,
+) : BasePropertyDetailsUpdateControllerTests(webContext) {
```
JasminConterioSW, line 33: The usual tests of the getUpdateStep and postUpdateStep endpoints are covered in BasePropertyDetailsUpdateControllerTests

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -358,4 +360,45 @@ class PropertyComplianceService(
...
+        val propertyCompliance = getComplianceForProperty(propertyOwnershipId)
+        throwErrorIfLastModifiedDatesConflict(propertyCompliance, initialLastModifiedDate)
+
+        propertyCompliance.apply {
+            populateGasSafetyFields(
+                record = this,
+                hasGasSupply = hasGasSupply,
+                gasSafetyCertIssueDate = gasSafetyCertIssueDate,
+                gasSafetyFileUploadIds = gasSafetyCertUploadIds,
+            )
```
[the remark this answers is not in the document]
JasminConterioSW: gasSafetyFileUpload will be deprecated soon, the compliance summary is due to be updated in PDJB-794/795

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/GasSafetyViewModelFactory.kt
```diff
@@ -26,7 +28,10 @@ class GasSafetyViewModelFactory(
                         propertyCompliance.gasSafetyFileUpload?.let {
                             uploadService.getDownloadUrlOrNull(it, "gas_safety_certificate.${it.extension}")
                         },
```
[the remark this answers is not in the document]
JasminConterioSW: This will be addressed in a later ticket (PDJB-794/PDJB-795)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/HasGasCertStepConfig.kt
```diff
@@ -1,69 +1,66 @@
...
+    override fun mode(state: GasSafetyState) =
+        getFormModelFromStateOrNull(state)?.let {
+            if (it.action == PROVIDE_THIS_LATER_BUTTON_ACTION_NAME) {
+                if (state.allowProvideCertificateLaterRoute) {
+                    HasGasCertMode.PROVIDE_THIS_LATER
+                } else {
+                    // This should never happen as the button to trigger this action should not be shown
+                    // if allowProvideCertificateLaterRoute is false
+                    null
+                }
```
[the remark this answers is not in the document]
JasminConterioSW: Actually, I think the original version of this is correct.
If the action is PROVIDE_THIS_LATER_BUTTON_ACTION_NAME, the mode shouldn't depend on what the value of hasCert is.
We don't want them to be able to select a radio button (to give hasCert a value) then click "Provide this later"

.github/copilot-instructions.md
```diff
@@ -101,6 +101,16 @@ npm run build                    # Build frontend assets only
 
 Integration tests require Docker running (uses testcontainers for PostgreSQL).
 
+## Line Endings
```
Travis-Softwire, line 104: Nice - that's been annoying me!

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/HasGasCertStepConfig.kt
```diff
@@ -1,69 +1,66 @@
...
+
+    override fun mode(state: GasSafetyState) =
+        getFormModelFromStateOrNull(state)?.let {
+            if (it.action == PROVIDE_THIS_LATER_BUTTON_ACTION_NAME) {
+                if (state.allowProvideCertificateLaterRoute) {
+                    HasGasCertMode.PROVIDE_THIS_LATER
+                } else {
+                    // This should never happen as the button to trigger this action should not be shown
+                    // if allowProvideCertificateLaterRoute is false
+                    null
```
Travis-Softwire: I'd be tempted to throw an exception here given it's not a valid code path. What's the effect of returning null here in practice? Will the next step get treated as being unreachable and the user gets redirected?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -187,16 +189,16 @@ class PropertyComplianceService(
 
     private fun updateFileUploadVirusScanningCallbacks(
         propertyOwnershipId: Long,
-        gasSafetyCertUploadIds: List<Long>,
-        electricalSafetyCertUploadIds: List<Long>,
+        gasSafetyCertUploadIds: List<Long>? = null,
```
Travis-Softwire: In what scenario are these null vs being an empty list?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/basePages/CheckGasSafetyAnswersFormBasePage.kt
```diff
@@ -0,0 +1,33 @@
...
+    page: Page,
+    urlSegment: String,
+) : BasePage(page, urlSegment) {
+    val heading = Heading(page.locator("h1"))
+    val sectionHeader = SectionHeader(page.locator("main"))
+    val form = Form(page)
+    val gasSupplySummaryList = GasSupplySummaryList(page)
+    val certSummaryList = CertSummaryList(page)
+
+    class GasSupplySummaryList(
```
Travis-Softwire: These will also be useful for the main CYA page (on of my draft PRs has something v similar) so should probably be spun out into their own class?

src/test/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/GasSafetyViewModelFactoryTests.kt
```diff
@@ -58,7 +63,12 @@ class GasSafetyViewModelBuilderTests {
                         SummaryListRowViewModel(
                             "propertyDetails.complianceInformation.gasSafety.gasSafetyCertificate",
                             "propertyDetails.complianceInformation.gasSafety.downloadCertificate",
-                            // TODO PDJB-80: readd change link
+                            listOf(
```
Travis-Softwire, line 67: For all of these I'd be tempted to tag them with a todo for PDJB-795 (and flag to Tom who's picked that up) because these will be replaced by a summary card and a single change link on the summary card

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -358,4 +360,45 @@ class PropertyComplianceService(
 
         return callbacks.extractFileUpload()
     }
+
+    @Transactional
+    fun updateGasSafety(
```
Travis-Softwire, line 366: Not sure if it makes sense to do it here or as a separate service call from the step (probably the latter) but we also need to handle deleting of the previously uploaded files in s3. We didn't need to worry about that before because we only had one file per cert so we just overwrote it. Let's not try and do it now though - can you raise a bug ticket for them not being deleted and label it with the `Launch_candidate` label?

Travis-Softwire (approved):
A few thoughts, but overall LGTM. We will need to delete the existing files but for right now they can be orphaned and we'll add deletion in a new ticket.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/HasGasCertStepConfig.kt
JasminConterioSW: It ends up throwing an UnrecoverableJourneyStateException when the mode is null - but yes we can explicitly throw that here instead!

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
JasminConterioSW: They were null when those weren't being updated (e.g gasSafetyUpdate would never pass in electricalSafetyCertUploadIds, but you're right they can default to an emptyList

PR #1256, opened 2026-04-22
PDJB-822: Default to https when base URL env vars lack a scheme

Travis-Softwire:
## Ticket number

PDJB-822

## Goal of change

Fix email links being rendered with `http://` instead of `https://` when the configured base URL environment variables omit the URL scheme.

- `AbsoluteUrlProvider` now defensively prepends `https://` to either configured base URL if it does not already include a scheme, so all URIs it builds (and therefore all email links) have an explicit scheme.
- An anchored regex is used to detect existing schemes, so values that already specify `http://`, `https://`, etc. are preserved unchanged.
- Refactored `AbsoluteUrlProvider` to use constructor injection for the `@Value`-bound base URLs so normalisation can happen once, in property initialisers.

The root cause is that the `LANDLORD_BASE_URL` / `LOCAL_COUNCIL_BASE_URL` env vars in deployed environments do not include a URL scheme. GOV.UK Notify then auto-prepends `http://` when it renders scheme-less markdown links. Fixing the env vars would also resolve the bug; this PR is a defence-in-depth change so the application is robust to that configuration shape.

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/AbsoluteUrlProvider.kt
```diff
@@ -16,12 +16,19 @@ import java.net.URI
...
+    @Value("\${base-url.landlord}") landlordBaseUrl: String,
+    @Value("\${base-url.local-council}") localCouncilBaseUrl: String,
+) {
+    private val landlordBaseUrl: String = ensureScheme(landlordBaseUrl)
+    private val localCouncilBaseUrl: String = ensureScheme(localCouncilBaseUrl)
 
-    @Value("\${base-url.local-council}")
-    private lateinit var localCouncilBaseUrl: String
+    private companion object {
+        private val SCHEME_PREFIX_REGEX = Regex("^[a-zA-Z][a-zA-Z0-9+.\\-]*://")
```
TomHanmer29: Do we need the regex to be this loose (i.e. the 0-9 and ._ etc)? And if so, should we add a test for that case?

TomHanmer29 (approved):
LGTM! Just a little question on the choice of regex but I think it's fine :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/AbsoluteUrlProvider.kt
Travis-Softwire: hah - no it doesn't. I think we can be confident that we're not going to be using any protocals other than http/https on the base url :)

PR #1257, opened 2026-04-23
PDJB-780: Separate feature flag configs for test and nft environments

Travis-Softwire:
## Ticket number

PDJB-780

## Goal of change

Replace the single shared feature-flag config across test, nft, and prod with separate per-environment configs so each environment can be tuned independently.

- Adds `application-test.yml` and `application-nft.yml` under `src/main/resources/`, each containing the complete `features.feature-flags` list. Spring picks these up via `SPRING_PROFILES_ACTIVE = "default, <env_name>"` already set by terraform — until now the env-specific profile was a no-op.
- Updates `application.yml` so the default profile represents production: `failover-test-endpoints` is now `false` (was `true`) and the PDJB-780 TODO comment above `subject-identifier-page` has been removed.
- Extends `FeatureFlagConsistencyTests` with nested `TestProfileTests` and `NftProfileTests` mirroring the existing `LocalProfileTests` / `IntegrationProfileTests` pattern, so a missing flag in either new env file is caught at CI rather than at environment boot.

Final flag matrix:

| Flag | Default (prod) | test | nft |
|---|---|---|---|
| failover-test-endpoints | false | true | true |
| joint-landlords | false | false | false |
| subject-identifier-page | true | true | true |

- `subject-identifier-page` is intentionally `true` in all three environments for now. The deleted TODO suggested it should only be enabled in test, but the team's current preference is to leave it on everywhere and disable per environment later. Flagging because the deleted comment line might invite a question.
- Production behaviour change: `failover-test-endpoints` flips from `true` to `false` on the next prod deploy. This is intentional — failover test endpoints should not be live in prod.
- Local bootRun smoke test was skipped because the new `TestProfileTests` / `NftProfileTests` boot a Spring context with the same profiles and exercise the same `FeatureFlagConfig.afterPropertiesSet()` validation that fires at env startup. No other config varies by `test` / `nft` profile in this PR.
- The full gradle `test` suite was not run locally; only `uk.gov.communities.prsdb.webapp.config.featureFlags.*` was run because changes are isolated to feature-flag YAML and the targeted classes cover the consistency invariants. CI will run the full suite.

## Checklist


src/main/resources/application.yml
```diff
@@ -81,13 +81,12 @@ ff4j:
 features:
   feature-flags:
     - name: "failover-test-endpoints"
-      enabled: true
+      enabled: false
```
TomHanmer29, line 84: Not sure I've got full context here but it looks like the flag that's been turned off is not the one that the TODO is under?

TomHanmer29 (approved):
Am approving as I think the change is fine, but it's worth checking the comment I've left first :)

Travis-Softwire:
Yes, sorry, should have made that clear in the PR description. That flag turns on endpoints that force the application to throw 50X exceptions in order to check that Cloudfront fails over to the static maintenance page in s3 - we wanted that on in test but not prod :). The other one was the page to grab people's subject identifier. I'm intending to use that to populate the system operator table pre go-live and then flag it off.

src/main/resources/application.yml
Travis-Softwire: See comment below :)

PR #1258, opened 2026-04-23
PDJB-769: are you sure screen non compliant

TomHanmer29:
## Ticket number

PDJB-769

## Goal of change

Adds an "Are you sure" screen for an occupied property which is non compliant (i.e. doesn't have gas, electrical and EPC).

image

Adds an are you sure page for **occupied** flows with conditional sections depending on what certs the user has added. this includes:
Gas (only shows when has a gas supply and then the user doesn't provide a cert)
Electrical (shows when a user skips or doesn't provide a certificate)
EPC (shows when a user doesn't have a valid EPC)

Currently choosing the "No" option goes back to the main cya page - I've asked Larry for confirmation of this but he's not responded

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/ConfirmMissingComplianceCheckStepConfig.kt
```diff
@@ -0,0 +1,48 @@
...
+import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.JourneyFrameworkComponent
+import uk.gov.communities.prsdb.webapp.journeys.AbstractInternalStepConfig
+import uk.gov.communities.prsdb.webapp.journeys.JourneyStep
+import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.states.CombinedComplianceCheckState
+import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.states.ElectricalSafetyState
+import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.states.EpcState
+import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.states.GasSafetyState
+
+@JourneyFrameworkComponent
+class ConfirmMissingComplianceCheckStepConfig :
```
Travis-Softwire: I think we could probably name this something a bit clearer, e.g. `HasMissingComplianceStep` (same w/ associated classes)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/HasMissingComplianceStepConfig.kt
```diff
@@ -0,0 +1,48 @@
...
+            return isOutdated == null || isOutdated
+        }
+
+        fun isElectricalCertMissingOrExpired(state: ElectricalSafetyState): Boolean {
+            val isOutdated = state.getElectricalCertificateIsOutdated()
+            return isOutdated == null || isOutdated
+        }
+
+        fun isEpcMissing(state: EpcState): Boolean {
+            if (state.acceptedEpcIfReachable != null) return false
```
Travis-Softwire, line 35: Not really part of this PR, but I don't love these <x>IfReachable properties on the state. Using reachability in this way makes it hard for the reader to evaluate when this condition will be triggered, because to do so they need to understand the whole ancestry of the step which might change over time. Maybe one for dev catch-up

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/ConfirmMissingComplianceCheckStepConfig.kt
```diff
@@ -0,0 +1,48 @@
...
+    AbstractInternalStepConfig<ConfirmMissingComplianceCheckResult, CombinedComplianceCheckState>() {
+    override fun mode(state: CombinedComplianceCheckState): ConfirmMissingComplianceCheckResult =
+        if (!state.isOccupied || (!isGasCertMissingOrExpired(state) && !isElectricalCertMissingOrExpired(state) && !isEpcMissing(state))) {
```
Travis-Softwire: Minor, but think you can make this clearer with something like:
```
val anyMissing = isGasCertMissingOrExpired(state) || isElectricalCertMissingOrExpired(state) || isEpcMissing(state)
 return if (state.isOccupied && anyMissing) OCCUPIED_AND_HAS_MISSING_CERTIFICATES else UNOCCUPIED_OR_ALL_CERTIFICATES
 ```

File: src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/PropertyRegistrationJourneyFactory.kt [hunk 4]
```diff
@@ -217,6 +222,17 @@ class PropertyRegistrationJourneyFactory(
             configureStep(journey.checkEpcAnswersStep) {
                 withAdditionalContentProperty { "sectionHeaderInfo" to null }
             }
+            configureStep(journey.confirmMissingComplianceStep) {
+                withAdditionalContentProperty {
+                    "sectionHeaderInfo" to
+                        SectionHeaderViewModel(
```
Travis-Softwire, line 219: I'm surprised we need to supply sectionNumber and totalSections if useNumbering is false? That feels like a bit of a gotcha if true

src/main/resources/messages/registerProperty.yml
```diff
@@ -85,3 +85,38 @@ deleteIncompleteProperties:
       two: 'If you want to register the property, you will need to add any details about the property again.'
     button:
       text: Return to incomplete properties
+submitRegistration: Submit registration
+confirmMissingCompliance:
+  heading: Confirm missing compliance certificates
+  warning: You can be fined or prosecuted for not having these certificates.
+  intro: This property is occupied by tenants, but it's missing compliance certificates. These certificates keep tenants safe and are required by law.
```
Travis-Softwire: Non-curly apostrophe alert

src/main/resources/templates/forms/confirmMissingCompliance.html
```diff
@@ -0,0 +1,100 @@
...
+        </p>
+        <ul class="govuk-list govuk-list--bullet">
+            <li th:text="#{registerProperty.confirmMissingCompliance.epc.bullet.one}">registerProperty.confirmMissingCompliance.epc.bullet.one</li>
+            <li th:text="#{registerProperty.confirmMissingCompliance.epc.bullet.two}">registerProperty.confirmMissingCompliance.epc.bullet.two</li>
+        </ul>
+    </th:block>
+
+    <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
+
+    <div class="govuk-form-group" th:classappend="${#fields.hasErrors('wantsToProceed')} ? 'govuk-form-group--error'">
```
Travis-Softwire: Doesn't this cause problems with the radios fragment also rendering a govuk-form-group? If our existing fragments don't support adding a legend we should probably add that rather than wrapping it

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationSinglePageTests.kt
```diff
@@ -1530,4 +1530,14 @@ class PropertyRegistrationSinglePageTests : IntegrationTestWithImmutableData("da
...
+
+    @Nested
+    inner class ConfirmMissingComplianceStep {
+        @Test
+        fun `Submitting with no option selected returns an error`(page: Page) {
+            val confirmPage = navigator.skipToPropertyRegistrationConfirmMissingCompliancePage()
+            confirmPage.form.submit()
+            assertThat(confirmPage.form.getErrorMessage()).containsText("Select whether you want to submit this registration")
+        }
+    }
```
Travis-Softwire, line 1551: Do we want a test that shows it navigates back to the CYA page? That shouldn't touch the db so I think could live here

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationJourneyTests.kt
```diff
@@ -896,6 +897,15 @@ class PropertyRegistrationJourneyTests : IntegrationTestWithMutableData("data-lo
...
         assertThat(checkAnswersPage.sectionHeader).containsText("Section 2 of 2 — Check and submit your property details")
+
+        // Check Answers - submit to reach Confirm Missing Compliance page
+        checkAnswersPage.form.submit()
+        val confirmMissingCompliancePage = assertPageIs(page, ConfirmMissingComplianceFormPagePropertyRegistration::class)
+
+        // Confirm Missing Compliance - render page
+        assertThat(confirmMissingCompliancePage.heading).containsText("Confirm missing compliance certificates")
+        assertThat(confirmMissingCompliancePage.warning).isVisible()
+        assertThat(confirmMissingCompliancePage.form.sectionHeader).containsText("Submit registration")
```
Travis-Softwire, line 911: Should we check that submitting it does in fact cause the record to be saved?

Travis-Softwire (changes requested):
This looks good, but there are few points that I think probably need addressing first - happy to chat through them if helpful

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/HasMissingComplianceStepConfig.kt
TomHanmer29: Yeah I do see what you mean - if you're happy to keep it in like this for a week or so I'll add it to next week's dev catch up thread, but we can chat about it instead if you'd like it to be changed sooner :)

Travis-Softwire: Yeah let's pick it up at dev catch-up

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/PropertyRegistrationJourneyFactory.kt
TomHanmer29: Yeah, it's a bit of a weird one. I've been digging into this this morning and the current way we set headers is typically via a section header (but we don't want that here as it adds to the total number of sections, and we want sections out of 2) or via the "sectionHeaderInfo" property which uses a sectionHeaderViewModel which currently has all 4 properties required. I don't think we have a case where we set the sectionHeaderInfo outside a section for a single page currently. I think the best way forward could be to add a sectionHeaderOverride to a step in the DSL but may be worth discussing?

Travis-Softwire: Ok, let's leave it as-is for now and add to dev catch-up - it'll probably end up being a QoL ticket

src/main/resources/templates/forms/confirmMissingCompliance.html
TomHanmer29: With regards to the form-group inside a form group thing, it seems like we actually do this in tons of places in the codebase. fieldSet has a govuk-form-group class and every inputGroup also has one. The pattern we use elsewhere is putting stuff like this in a fieldset though - so I will make that change

Travis-Softwire: 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/HasMissingComplianceStepConfig.kt
```diff
@@ -0,0 +1,51 @@
...
+import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.states.ElectricalSafetyState
+import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.states.EpcState
+import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.states.GasSafetyState
+
+@JourneyFrameworkComponent
+class HasMissingComplianceStepConfig : AbstractInternalStepConfig<ConfirmMissingComplianceCheckResult, CombinedComplianceCheckState>() {
+    override fun mode(state: CombinedComplianceCheckState): ConfirmMissingComplianceCheckResult {
+        val anyMissing = isGasCertMissingOrExpired(state) || isElectricalCertMissingOrExpired(state) || isEpcMissing(state)
+        return if (state.isOccupied &&
+            anyMissing
```
Travis-Softwire: Minor, but feels like a very aggressive line break :D

Travis-Softwire (approved):
LGTM!

PR #1260, opened 2026-04-23
Feat/pdjb 764 refactor property services

JasminConterioSW:
## Ticket number

PDJB-764

## Goal of change

Refactor services

Refactors services saving property registration data
* Moves call to create the compliance to the registerProperty function, before confirmation emails sent
* Moves virus scanning callbacks to where we save the compliance
* Refactors the the saveCompliance function to use separate functions for updating gas, electric and epc fields (these can be re-used by the relevant update journey)

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -115,24 +117,98 @@ class PropertyComplianceService(
         val compliance = getComplianceForPropertyOrNull(propertyOwnership.id)
```
TomHanmer29: Not 100% sure I've followed the logic here but do we need this post refactor (can a property that has not been registered have a compliance?)

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceServiceTests.kt
```diff
@@ -840,6 +844,108 @@ class PropertyComplianceServiceTests {
...
+
+            assertThrows<EntityNotFoundException> {
+                propertyComplianceService.saveRegistrationComplianceData(
+                    registrationNumberValue = registrationNumberValue,
+                )
+            }
+        }
+
+        @Test
+        fun `updates existing compliance record when one already exists`() {
```
TomHanmer29: Again, can this actually happen now?

TomHanmer29 (approved):
LGTM! Tiny point about a function but this is much neater now, thanks!

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -115,24 +117,98 @@ class PropertyComplianceService(
...
+        record: PropertyCompliance,
+        hasGasSupply: Boolean?,
+        gasSafetyCertIssueDate: LocalDate?,
+        gasSafetyFileUploadIds: List<Long>,
+    ): PropertyCompliance {
         record.gasSafetyCertExemptionReason = if (hasGasSupply == false) GasSafetyExemptionReason.NO_GAS_SUPPLY else null
         record.hasGasSupply = hasGasSupply
         record.gasSafetyCertIssueDate = gasSafetyCertIssueDate
         record.gasSafetyFileUploads = gasSafetyFileUploadIds.map { fileUploadRepository.getReferenceById(it) }.toMutableList()
+        return record
```
Travis-Softwire: I don't think any of these returned values are used?

Travis-Softwire (approved):
One Q but otherwise LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
JasminConterioSW: Good point, no it can't...

I was thinking we need it for updates but that'll be a different function!

PR #1265, opened 2026-04-25
PDJB-829: Add Plausible Flow custom event for Sankey diagrams

Travis-Softwire:
## Ticket number

PDJB-829

## Goal of change

Adds a Plausible "Flow" custom event so user journeys can be visualised as Sankey diagrams.

- Adds a `PlausibleInterceptor` that populates `plausibleEventReferrer` and `plausibleEventCurrentUrl` model attributes on every successful HTML page response (skipped for redirects, forwards, and null `ModelAndView`).
- Same-origin `Referer` headers are reduced to a path; cross-origin, hostless, malformed, or missing referrers are reported as `external`. Query strings are excluded from both props (matching Plausible's built-in `event:page`).
- For error dispatches, `RequestDispatcher.ERROR_REQUEST_URI` is preferred over `request.requestURI` so the Sankey records the originally requested URL rather than `/error`.
- Adds a CSP-nonce-protected inline script in `layout.html` that fires `plausible('Flow', { props: { referrer, currentUrl } })` per page, guarded against null attributes.

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/PlausibleInterceptorConfig.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/config/PlausibleInterceptorConfig.kt
@@ -1,0 +1,13 @@
+package uk.gov.communities.prsdb.webapp.config
+
+import org.springframework.web.servlet.config.annotation.InterceptorRegistry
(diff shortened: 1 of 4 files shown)


JasminConterioSW (approved):
Looks sensible, I think!

Bill-Haigh (approved):
LGTM

PR #1266, opened 2026-04-27
PDJB-769: fix styling and logic on non compliant occupied screen

TomHanmer29:
## Ticket number

PDJB-769

## Goal of change

Fixes logic and styling for the non compliant occupied check screen

image
Fixes an h1 and makes it an h2 instead
Fixes the case where the EPC was expired but not exempt or low energy but not MEES exempt - this counted as accepted but I'd not explicitly checked for those cases.

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/HasMissingComplianceStepConfig.kt
```diff
@@ -31,9 +31,15 @@ class HasMissingComplianceStepConfig : AbstractInternalStepConfig<ConfirmMissing
             return isOutdated == null || isOutdated
         }
 
-        fun isEpcMissing(state: EpcState): Boolean {
-            if (state.acceptedEpcIfReachable != null) return false
-            return state.epcExemptionStep.formModelIfReachableOrNull?.exemptionReason == null
+        fun isEpcMissingOrExpired(state: EpcState): Boolean {
```
JasminConterioSW: It might be worth calling this `isEpcMissingOrInvalid` as it also covers the case when it has a low energy rating with no exemption

JasminConterioSW (approved):
Might be worth updating the name of isEpcMissingOrExpired, but other than that LGTM :)

PR #1267, opened 2026-04-27
PDJB-795: update compliance tab for private beta

TomHanmer29:
## Ticket number

PDJB-795

## Goal of change

Updates the compliance tab for private beta

image
image

Adds in any missing fields we should display
Adds a skeleton compliance to all properties in the seeder
Updates the screen to use cards, with change links
Removes some unused banner components

See comments - mainly, a bit unsure about displaying the files

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
```diff
@@ -115,14 +115,10 @@ class PropertyCompliance() : ModifiableAuditableEntity() {
...
     val isEpcExpired: Boolean?
-        get() {
-            val isPastExpiryDate = epcExpiryDate?.isBefore(LocalDate.now()) ?: return null
-            return if (!isPastExpiryDate) {
-                false
-            } else {
-                tenancyStartedBeforeEpcExpiry?.not()
-            }
-        }
+        get() = epcExpiryDate?.isBefore(LocalDate.now())
```
TomHanmer29, line 121: I've added this in to ensure the "Did the tenancy start before the EPC expired?" text shows up for expired certs. Existing functionality should be unchanged

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -252,7 +254,8 @@ class PropertyComplianceService(
             propertyCompliance.eicrFileUpload = update.eicrUpdate.fileUploadId?.let { getCertificateFileUpload(it) }
             propertyCompliance.eicrIssueDate = update.eicrUpdate.issueDate
             // TODO PDJB-766: Remove eicrIssueDate and this derived calculation once the compliance update journey uses expiry date
-            propertyCompliance.electricalSafetyExpiryDate = update.eicrUpdate.issueDate?.plusYears(EICR_VALIDITY_YEARS.toLong())
+            propertyCompliance.electricalSafetyExpiryDate =
+                update.eicrUpdate.issueDate?.plusYears(ELECTRICAL_SAFETY_VALIDITY_YEARS.toLong())
```
TomHanmer29: This has just happened as part of the rename, this file will be ripped out soon (which is why it still uses eicrUpdate.issueDate)

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/RichTextValue.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.communities.prsdb.webapp.models.viewModels.summaryModels
+
+data class RichTextValue(
```
TomHanmer29: This is how copilot suggested displaying the big combo lists of links / pending files, I'm not convinced it's the best solution but I can't think of an obvious alternative (I tried looking at the gas CYA page, but I don't think that handles a combination of clickable files and pending things all in the same row)

src/main/resources/templates/propertyDetailsView.html
```diff
@@ -19,14 +19,13 @@
                     </li>
                 </ul>
             </div>
+            <!--/* TODO PDJB-794: Update no-compliance banner contents */-->
```
TomHanmer29, line 22: I've removed change the link from here but kept the banner as is otherwise. I think further changes to it seem out of scope until PDJB-794

src/main/resources/templates/propertyDetailsView.html
```diff
@@ -70,31 +69,12 @@ <h3 class="govuk-heading-s" th:text="#{propertyDetails.landlordDetails.registere
...
 
-                       <h3 class="govuk-heading-s" th:text="#{propertyDetails.complianceInformation.energyPerformance.heading}">propertyDetails.complianceInformation.energyPerformance.heading</h3>
-                       <dl th:replace="~{fragments/summaryList :: summaryList(${complianceDetails.epcSummaryList})}"></dl>
-
-                       <h3 class="govuk-heading-s" th:text="#{propertyDetails.complianceInformation.landlordResponsibilities.heading}">propertyDetails.complianceInformation.landlordResponsibilities.heading</h3>
-                       <p class="govuk-hint" th:text="#{${complianceDetails.landlordResponsibilitiesHintText}}">the hint text for landlord responsibilities</p>
-                       <dl th:replace="~{fragments/summaryList :: summaryList(${complianceDetails.landlordResponsibilitiesSummaryList})}"></dl>
+                       <div th:replace="~{fragments/summaryCard :: summaryCard(${complianceDetails.epcSummaryCard.title}, null, ${complianceDetails.epcSummaryCard.actions}, ${complianceDetails.epcSummaryCard.summaryList})}"></div>
                    </div>
-                    <div class="govuk-grid-row" th:if="${complianceDetails == null}">
```
TomHanmer29, line 86: I've removed the no compliance tab view, since all registered properties now have a skeleton empty compliance (and updated the seed data to match)

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/PropertyComplianceConstants.kt
```diff
@@ -2,6 +2,6 @@ package uk.gov.communities.prsdb.webapp.constants
 
 const val GAS_SAFETY_CERT_VALIDITY_YEARS = 1
 
-const val EICR_VALIDITY_YEARS = 5
+const val ELECTRICAL_SAFETY_VALIDITY_YEARS = 5
```
Travis-Softwire: I think we're dropping this as part of removing the old journey v. soon anyway, but that's fine

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
```diff
@@ -111,21 +111,17 @@ class PropertyCompliance() : ModifiableAuditableEntity() {
...
+
+    val isEpcNonCompliantDueToExpiry: Boolean?
+        get() = if (isEpcExpired == true) tenancyStartedBeforeEpcExpiry != true else isEpcExpired
```
Travis-Softwire: When do you want this to return null? Atm this is equivalent to `isEpcExpired?.and(tenancyStartedBeforeEpcExpiry != true)` but it feels weird to say that if isEpcExpired is null then isEpcNonCompliantDueToExpiry is also null?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/SummaryListRowViewModel.kt
```diff
@@ -26,6 +26,7 @@ data class SummaryListRowViewModel(
 
     private fun getSingleValueMessageKey(value: Any?): Any? =
         when (value) {
+            is RichTextValue -> value
```
Travis-Softwire: Why is the extra case needed? It shouldn't match any of the other cases should it?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/RichTextValue.kt
Travis-Softwire: I agree we do need a  model that's specifically for a line that could be text or could be a link, but it should be more targeted - so optionalParam should be displayName and the name of the model should be something like uploaded file url

src/main/resources/templates/fragments/summaryRow.html
```diff
@@ -10,10 +10,18 @@
            th:text="${#messages.msgOrNull(values)} ?: ${{values}}">
         </a>
         <!--/* If and only if there is a single value (as opposed to a list) we should th:remove the p tag to avoid unnecessary nesting */-->
-        <p th:if="${valueUrl} == null"
-           th:remove="${iteratorStatus.size == 1} ? tag : none"
-           th:each="value, iteratorStatus: ${values}" class="govuk-body"
-           th:text="${#messages.msgOrNull(value)} ?: ${{value}}">row.convertedFieldValue</p>
+        <th:block th:if="${valueUrl} == null" th:each="value, iteratorStatus: ${values}">
+            <!--/* RichTextValue: delegate to dedicated fragment (th:if must be on a wrapper so it runs before th:replace) */-->
+            <th:block th:if="${value.class.simpleName == 'RichTextValue'}">
```
Travis-Softwire: Oh, if we're going to do this based on simpleName then we may as well just have a link view model and then plain text for the non-links?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyCompliance/states/EicrState.kt
```diff
@@ -2,7 +2,7 @@ package uk.gov.communities.prsdb.webapp.journeys.propertyCompliance.states
 
 import kotlinx.datetime.DatePeriod
 import kotlinx.datetime.plus
-import uk.gov.communities.prsdb.webapp.constants.EICR_VALIDITY_YEARS
+import uk.gov.communities.prsdb.webapp.constants.ELECTRICAL_SAFETY_VALIDITY_YEARS
 import uk.gov.communities.prsdb.webapp.helpers.DateTimeHelper
```
Travis-Softwire: Don't forget to also update the names of these files!

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/PropertyComplianceViewModelFactory.kt
```diff
@@ -1,50 +1,67 @@
...
+                title = "propertyDetails.complianceInformation.gasSafety.heading",
+                summaryList = gasSafetyViewModelFactory.fromEntity(propertyCompliance),
+                actions = gasSafetyChangeActions,
+            )
 
-        val landlordResponsibilitiesSummaryList: List<SummaryListRowViewModel> =
-            LandlordResponsibilitiesViewModelBuilder.fromEntity(
-                propertyCompliance,
-                landlordView,
+        val eicrSummaryCard =
```
Travis-Softwire: Minor, but might want to update this variable name as well

Travis-Softwire (approved):
A few minor points, but overall LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/PropertyComplianceConstants.kt
TomHanmer29: I'll actually undo it - since we do just use it in the old journey, I think leaving it as is makes it easier to see what needs removing

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyCompliance/states/EicrState.kt
TomHanmer29: Same here, as these are old journey I think I'll just leave them as is

src/main/resources/templates/fragments/summaryRow.html
TomHanmer29: I'm going to refactor into a url specific one - the main issue with plain text for the non links is that we need to pass in the parameter for the names with (pending) after it so I think it's easier if I just do that in the url fragment.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
TomHanmer29: Yep, there's no harm in making it non nullable - it's only used for ==true cases anyway!

PR #1271, opened 2026-04-27
PDJB-718: Cache occupancy and selected address state for CYA child journeys

Travis-Softwire:
## Ticket number

PDJB-718

## Goal of change

Make `state.isOccupied` and `state.uprn` safe to read from CYA child journeys, where the upstream step instances are uninitialised.

- Adds `cachedOccupied: Boolean?` to `OccupationState` and `cachedSelectedAddress: String?` to `AddressState`, both backed by session-stored delegates on the relevant journey factories.
- Snapshots the values into the cache from `OccupiedStepConfig.afterStepDataIsAdded` and `SelectAddressStepConfig.afterStepDataIsAdded` whenever the user submits those steps.
- Rewrites `PropertyRegistrationJourneyState.isOccupied` and `uprn` to read from the cache first, falling back to the step's submitted form data and lazily backfilling the cache (so journeys submitted before this change continue to work).
- The address cache stores the raw selected address string rather than the resolved UPRN, to distinguish "not yet submitted" (null) from "manual address chosen" (cached non-null but resolves to no UPRN).

This is the third of a stack of four PRs for PDJB-718, and is groundwork for the wiring PR (`feat/PDJB-718-cya-compliance-sub-journeys`).

Without this cache, when the CYA child journey reads `state.isOccupied` or `state.uprn` it throws an `UninitializedPropertyAccessException`, because the upstream step instances (`OccupiedStep`, `SelectAddressStep`) are not part of the child's journey graph and so are never initialised by the framework.

The longer-term fix is for the journey framework itself to allow registering external step dependencies; that is out of scope for PDJB-718.

Stacked on `chore/PDJB-718-extract-compliance-details-tasks`.

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/PropertyRegistrationJourneyFactory.kt
```diff
@@ -490,7 +499,14 @@ class PropertyRegistrationJourney(
...
     override var registrationNumberValue: Long? by delegateProvider.nullableDelegate("registrationNumberValue")
 
-    override val uprn: Long? get() = selectAddressStep.formModelOrNull?.address?.let { getMatchingAddress(it)?.uprn }
+    // Cache reasoning matches isOccupied above. The cached value is the raw selected address string so we can
+    // distinguish "not yet submitted" (null) from "manual address chosen" (cached non-null but resolves to no UPRN).
+    // Cache is populated on step submission (see SelectAddressStepConfig.afterStepDataIsAdded).
+    override val uprn: Long? get() {
+        cachedSelectedAddress?.let { return getMatchingAddress(it)?.uprn }
+        val submittedAddress = selectAddressStep.formModelOrNull?.address ?: return null
+        return getMatchingAddress(submittedAddress)?.uprn
```
TomHanmer29, line 508: How come we have a case for the above which throws `"Cannot use isOccupied until after the occupation step"` but here we just return null? Is it because the address step is the first one in the flow (and/or we're running it through the getMatchingAddress function?)?

TomHanmer29 (approved):
LGTM, one question but not really related to this PR as it was already the case before you made the change :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/PropertyRegistrationJourneyFactory.kt
Travis-Softwire: Good question - it's because you might genuinely not have selected an address, you might have entered a manual address instead (which has no uprn), whereas it should be impossible to arrive on a step that needs isOccupied without having completed the occupied step

PR #1280, opened 2026-04-29
PDJB-806: add file names and EIC/EICR to compliance journey db

TomHanmer29:
## Ticket number

PDJB-806

## Goal of change

Adds a couple of missing fields we should save to the db

Adds file names for use in the pending state
adds whether a cert is EIC/EICR (or NO_CERTIFICATE if the user chooses "none of the above")

I'm currently just saving the full enum for EIC/EICR which includes NO_CERTIFICATE - this could probably be a null instead but it's a bit more ugly to convert it. I can change it if you'd prefer

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
```diff
@@ -63,6 +64,8 @@ class PropertyCompliance() : ModifiableAuditableEntity() {
 
     var electricalSafetyExpiryDate: LocalDate? = null
 
+    var electricalCertType: HasElectricalSafetyCertificate? = null
```
JasminConterioSW: There is also a `CertificateType` enum which gets used elsewhere in the DB (the virus callback table uses it), it might be worth using that? It needs EIC to be added to it

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -142,30 +145,38 @@ class PropertyComplianceService(
...
-        gasSafetyFileUploadIds: List<Long>,
+        gasSafetyFileUploads: Map<Long, String>,
     ) {
         record.gasSafetyCertExemptionReason = if (hasGasSupply == false) GasSafetyExemptionReason.NO_GAS_SUPPLY else null
         record.hasGasSupply = hasGasSupply
         record.gasSafetyCertIssueDate = gasSafetyCertIssueDate
-        record.gasSafetyFileUploads = gasSafetyFileUploadIds.map { fileUploadRepository.getReferenceById(it) }.toMutableList()
+        record.gasSafetyFileUploads =
+            gasSafetyFileUploads.map { (id, fileName) ->
+                fileUploadRepository.getReferenceById(id).apply { this.fileName = fileName }
```
JasminConterioSW: This is updating the filename on the FileUpload record isn't it? I was going to suggest doing that when we create the FileUpload record but maybe that's trickier (it's certainly not so easy to find!)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -142,30 +145,38 @@ class PropertyComplianceService(
...
         updateFileUploadVirusScanningCallbacks(
             propertyOwnershipId = propertyOwnership.id,
-            gasSafetyCertUploadIds = gasSafetyFileUploadIds,
-            electricalSafetyCertUploadIds = electricalSafetyFileUploadIds,
+            gasSafetyCertUploadIds = gasSafetyFileUploads.keys.toList(),
+            electricalSafetyCertUploadIds = electricalSafetyFileUploads.keys.toList(),
         )
     }
 
     private fun populateGasSafetyFields(
```
JasminConterioSW, line 155: The GasSafetyUpdateJourney is merged in and calls this and the updateFileUploadVirusScanningCallbacks function so will also need updating when you next merge in main!

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
TomHanmer29: Yes, it is - since we keep it in compliance state I set it there. But I think it's probably nicer to do it during creation, and I don't think it's much trickier

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
TomHanmer29: Changing this to use Eic as well does work, but it also means we need to handle the EIC case in the virus scan emails now. I've done this, but not convinced it's correct? As an aside, what ARE we doing with the virus scan emails? can they be sent if we have the electrical cert being null now when a user does provide this later or is this not something we need to worry about?

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/PropertyComplianceJourneyHelper.kt
```diff
@@ -35,7 +35,7 @@ class PropertyComplianceJourneyHelper {
             val stepName =
                 when (certificateType) {
                     CertificateType.GasSafetyCert -> GasSafetyCertificateUploadStep.ROUTE_SEGMENT
-                    CertificateType.Eicr -> EicrUploadStep.ROUTE_SEGMENT
+                    CertificateType.Eicr, CertificateType.Eic -> EicrUploadStep.ROUTE_SEGMENT
```
TomHanmer29, line 38: I believe this is only used in the old compliance journey so have just made it work for now

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
JasminConterioSW: I think we don't need to worry about it...

I didn't set this up so may have got it wrong, but I think we save a callback to the  repository for any file that's uploaded when we save the compliance, then this gets used to determine which emails are sent. So it shouldn't matter if there is no electrical cert because it just won't save a callback.

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/CertificateUploadHelper.kt
```diff
@@ -69,6 +69,7 @@ class CertificateUploadHelper(
             uploadFileName,
```
JasminConterioSW: Minor, but should this one have a different name (`uploadFileObjectKey`?) to distinguish it better from `file.name`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/UploadService.kt
```diff
@@ -30,6 +31,7 @@ class UploadService(
                 versionId = uploadResult.versionId,
                 extension = extension,
                 status = FileUploadStatus.QUARANTINED,
+                fileName = fileName,
```
JasminConterioSW, line 34: This should be included in one of the UploadServiceTests

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfig.kt
```diff
@@ -116,6 +116,12 @@ class SavePropertyRegistrationDataStepConfig(
             gasSafetyFileUploadIds = state.gasUploadIds,
             electricalSafetyFileUploadIds = state.electricalUploadIds,
             electricalSafetyExpiryDate = state.getElectricalCertificateExpiryDateIfReachable()?.toJavaLocalDate(),
+            electricalCertType =
```
JasminConterioSW: Might be worth adding a test to check that the EIC version is actually saving EIC

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -191,18 +202,22 @@ class PropertyComplianceService(
...
     ) {
         gasSafetyCertUploadIds.forEach {
             virusScanCallbackService.deleteAllCallbacksForFileUpload(it)
             virusScanCallbackService.saveEmailToMonitoringTeam(propertyOwnershipId, it, CertificateType.GasSafetyCert)
             virusScanCallbackService.saveEmailToOwner(propertyOwnershipId, it, CertificateType.GasSafetyCert)
         }
 
-        // TODO PDJB-765 - do we need to update this to pass CertificateType.Eic when appropriate?
+        if (electricalSafetyCertUploadIds.isNotEmpty()) {
+            requireNotNull(electricalCertType) { "electricalCertType must not be null when electrical safety uploads are present" }
```
JasminConterioSW, line 214: We should test that an error is thrown here

JasminConterioSW (approved):
There are a couple more tests that it's worth adding, but other than that LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/CertificateUploadHelper.kt
TomHanmer29: Yep, agreed :) Nice spot

PR #1288, opened 2026-05-05
PDJB-812: remove old compliance journey files [1 of 2]

TomHanmer29:
## Ticket number

PDJB-812

## Goal of change

Removes all pages that don't rely on the DB

 - Removes the entire old property compliance journey — factory, tasks, step configs, states, controller,
templates, page objects, and integration tests
 - Removes all EICR form models and the PropertyComplianceUpdateModel used by the old journey
 - Adds TODOs to remove or update email classes that were only sent from the old journey (new update journeys have a TODO for PDJB-770)
 - Cleans dead methods and constructor params from PropertyComplianceService
 - Removes dead message keys, constants, and enum values
 - Renames PropertyComplianceJourneyHelper to CertificateFilenameHelper and removes old filename overloads,
adding getUploadStepName() for mapping CertificateType to step names
 - Updates NftDataSeeder to generate filenames in the real format
 - Adds gasSafetyFileUploads to the PropertyCompliance entity constructor

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/CertificateFilenameHelper.kt
```diff
@@ -0,0 +1,32 @@
...
+            stepName: String,
+            memberId: String?,
+        ): String =
+            if (memberId != null) {
+                "certificateUpload.$journeyId.$stepName.$memberId"
+            } else {
+                "certificateUpload.$journeyId.$stepName.${randomSuffix()}"
+            }
+
+        fun getUploadStepName(certificateType: CertificateType): String =
```
TomHanmer29, line 21: This file has been renamed from PropertyComplianceJourneyHelper, and I've tweaked a couple of the bits, but I'm not sure if they're appropriate / functional (for their use in the NftDataSeeder)?

Travis-Softwire: Which bits are only used by the seeder? Ideally we'd keep that in sync with the rest of our changes, but we'll come back to in a separate ticket soon anyway so we can just add a todo (the the ticket is PDJB-239 )

TomHanmer29: getUploadStepName is the only really new bit, which is just used to replace an old function in the seeder that was only used in the compliance journey otherwise. It's for adding file uploads to a batch.

TomHanmer29: I'll put a TODO on just to be safe :)

Travis-Softwire (approved):
LGTM!

PR #1295, opened 2026-05-06
PDJB-766: Implement EPC update journey

JasminConterioSW:
## Ticket number

PDJB-766

## Goal of change

Implement  EPC update journey

- Add EpcUpdateJourney
    - Re-uses the EPC task, but without the "Provide this later"route
- Link to update journey from the compliance tab of the property record
- Added tests.
  - For non-trivial page objects I've added some base pages that the registration and update journeys both inherit from
  -  Note for integration tests I've not covered all possible paths through the EPC journey, but have included 
      - finding/not finding epc by uprn, 
      - epc exemption
      - valid epc with mees exemption
      - expired epc
      - missing epc

The EPC task starts with an internal step, so I've added the StartEpcUpdateStep - this acts an internal step in that it never renders, but it has a route segment so that it can be linked to.

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/epc/StartEpcUpdateStepConfig.kt
```diff
@@ -0,0 +1,31 @@
...
+
+@JourneyFrameworkComponent
+class StartEpcUpdateStepConfig : AbstractRequestableStepConfig<Complete, NoInputFormModel, JourneyState>() {
+    override val formModelClass = NoInputFormModel::class
+
+    override fun getStepLifecycleOrchestrator(journeyStep: JourneyStep<*, *, *>) = RedirectingStepLifecycleOrchestrator(journeyStep)
+
+    override fun getStepSpecificContent(state: JourneyState): Map<String, Any?> = mapOf<String, String>()
+
+    override fun chooseTemplate(state: JourneyState): String = ""
```
JasminConterioSW, line 19: These match AbstractInternalStepConfig as we want this step to behave in a similar way to an internal step, just need to be able to link to it

TomHanmer29: Is it worth making a new step type for an internal step that is linkable to? Or to make it so internal steps can be linked to? I'm not super familiar with what specifics of the journey framework make it so we can't link to an internal step but it could be worth establishing a pattern here, as this feels like it results in some unnecessary duplication.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/epc/UpdateEpcJourneyFactory.kt
```diff
@@ -0,0 +1,153 @@
...
+    private val propertyOwnershipService: PropertyOwnershipService,
+) {
+    final fun createJourneySteps(propertyId: Long): Map<String, StepLifecycleOrchestrator> {
+        val state = stateFactory.getObject()
+
+        if (!state.isStateInitialized) {
+            val propertyOwnership = propertyOwnershipService.getPropertyOwnership(propertyId)
+            val propertyCompliance =
+                propertyOwnership.propertyCompliance
+                    ?: throw PrsdbWebException("Property ownership $propertyId does not have a compliance record")
```
TomHanmer29, line 54: Not really a problem that's come from this ticket, but what should we do about existing data on integration that has been added before compliance was created automatically in registration?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationJourneyTests.kt
```diff
@@ -432,7 +432,6 @@ class PropertyRegistrationJourneyTests : IntegrationTestWithMutableData("data-lo
 
         // Has EPC - render page
         assertThat(hasEpcPage.form.sectionHeader).containsText("Section 1 of 2 — Register your property details")
-        assertThat(hasEpcPage.heading).containsText("Do you have an EPC for this property?")
```
TomHanmer29, line 436: Why have we removed these?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/UpdateEpcJourneyTests.kt
```diff
@@ -0,0 +1,170 @@
...
+        meesExemptionPage.submitExemptionReason(MeesExemptionReason.HIGH_COST)
+
+        val checkAnswersPage = assertPageIs(page, CheckEpcAnswersFormPageUpdateEpc::class, urlArguments)
+        checkAnswersPage.form.submit()
+
+        assertPropertyDetailsUpdated(page, "View EPC (opens in new tab)")
+    }
+
+    @Test
+    fun `A property can have its EPC updated with an expired certificate not found by UPRN`(page: Page) {
```
TomHanmer29: Worth adding test cases explicitly for not being able to find a "provide this later" button?

.secrets.baseline
```diff
@@ -179,7 +179,7 @@
         "filename": "build.gradle.kts",
         "hashed_secret": "54119f2d56461004a7b3d3c428ccae35fd6b53b4",
         "is_verified": false,
-        "line_number": 172,
+        "line_number": 176,
```
TomHanmer29: What are these changes for?

TomHanmer29 (approved):
LGTM! A few questions but nothing major :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/epc/StartEpcUpdateStepConfig.kt
JasminConterioSW: I did wonder about this... It's not something we will usually want to be doing, it's very specifically that the journey started with an internal step so this just leads into it. So I'm not sure if we want it to be a pattern, or something that we specifically think about and decide we need to do...

@AEPR What do think, do we want some kind of NavigableInternalStep?

I might leave it for now but can always add it in later (with suitable documentation!) - I'll add it to dev catchup as well.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationJourneyTests.kt
JasminConterioSW: We decided (maybe after this bit was written) that we may not want to check too much of the copy as it's subject to change! I mainly changed this one as I was making a new base page and didn't think we needed the heading

.secrets.baseline
JasminConterioSW: It's an automated thing, but the linter complains and doesn't let me commit unless I add this...

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/UpdateEpcJourneyTests.kt
JasminConterioSW: The integration tests take a while to spin up, so historically we've been trying to combine them quite a bit where possible (so that check is included in some of the other tests).

Although with that logic maybe I should combine all the EpcUdate tests into one... And similar with gas and electric...

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/epc/UpdateEpcJourneyFactory.kt
JasminConterioSW: It's potentially still useful (it looks like you updated it to add compliance to them, so they should be reasonably consistent) for QAing things like pagination (this is probably why there are so many of them in the first place!)

Maybe they could do with a cleanup though, I'll add that to dev catch up too

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/UpdateEpcJourneyTests.kt
JasminConterioSW: I've gone with combining them (which is the opposite of your comment) as that better fits the pattern of what we've been doing with integration tests. The ones with mutable data are particularly slow as it resets the database between each one, so fewer is better for those!

PR #1296, opened 2026-05-06
PDJB-NONE: Correct CloudWatch metrics property path for Spring Boot 3.x

Travis-Softwire:
## Ticket number

PDJB-NONE

## Goal of change

Fix CloudWatch metrics export, which silently failed to publish anything after #1293 was deployed.

- Moves the CloudWatch registry config from `management.metrics.export.cloudwatch.*` to `management.cloudwatch.metrics.export.*` in both `application.yml` and `application-local.yml`.

### DIFF
--- a/src/main/resources/application-local.yml
+++ b/src/main/resources/application-local.yml
@@ -80,9 +80,9 @@ management:
     web:
       exposure:
         include: metrics,health,info
-  metrics:
-    export:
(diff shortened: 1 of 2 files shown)


TomHanmer29 (approved):
I don't know if this will work but it looks sensible!

PR #1300, opened 2026-05-07
PDJB-588: Fix privacy notice QA defects

Bill-Haigh:
## Ticket number

PDJB-588

## Goal of change

Address three QA defects on the landlord privacy notice page.

- **Section 3.1**: Adds a trailing dot after `3.1` so the numbered paragraphs are punctuated consistently with the rest of the document (section 2 already uses `2.1.`, `2.2.` etc.).
- **UK GDPR Article 9(2)(j) paragraph**: Indents the paragraph so it visually aligns with the bullet list above it, matching the QA mock-up. Implemented by adding the `govuk-!-padding-left-6` GOV.UK Frontend utility class to the existing `<p>` rather than introducing bespoke CSS.
- **Section 10**: Merges the spurious paragraph break between the "...UK General Data Protection Regulation." sentence and "If you are unhappy with the way the Department has acted, you can make a complaint." so the two sentences render in a single paragraph as designed. Done by deleting `paragraph.one` from `privacyNotice.yml` and prepending its text to `paragraph.two.beforeLink`, then removing the corresponding `<p>` from the template.

## Checklist

### DIFF
--- a/src/main/resources/templates/landlordPrivacyNotice.html
+++ b/src/main/resources/templates/landlordPrivacyNotice.html
@@ -106,7 +106,7 @@
                 <li th:utext="#{privacyNotice.section.three.paragraph.two.bullet.one}">privacyNotice.section.three.paragraph.two.bullet.one</li>
                 <li th:utext="#{privacyNotice.section.three.paragraph.two.bullet.two}">privacyNotice.section.three.paragraph.two.bullet.two</li>
             </ul>
-            <p class="govuk-body" th:utext="#{privacyNotice.section.three.paragraph.two.additional}">privacyNotice.section.three.paragraph.two.additional</p>
+            <p class="govuk-body govuk-!-padding-left-6" th:utext="#{privacyNotice.section.three.paragraph.two.additional}">privacyNotice.section.three.paragraph.two.additional</p>
         </section>
         <section>
             <h2 class="govuk-heading-m" th:text="#{privacyNotice.section.four.heading}">privacyNotice.section.four.heading</h2>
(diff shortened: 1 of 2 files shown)


src/main/resources/templates/landlordPrivacyNotice.html
```diff
@@ -106,7 +106,7 @@ <h2 class="govuk-heading-m" th:text="#{privacyNotice.section.three.heading}">pri
                 <li th:utext="#{privacyNotice.section.three.paragraph.two.bullet.one}">privacyNotice.section.three.paragraph.two.bullet.one</li>
                 <li th:utext="#{privacyNotice.section.three.paragraph.two.bullet.two}">privacyNotice.section.three.paragraph.two.bullet.two</li>
             </ul>
-            <p class="govuk-body" th:utext="#{privacyNotice.section.three.paragraph.two.additional}">privacyNotice.section.three.paragraph.two.additional</p>
+            <p class="govuk-body govuk-!-padding-left-6" th:utext="#{privacyNotice.section.three.paragraph.two.additional}">privacyNotice.section.three.paragraph.two.additional</p>
```
TomHanmer29, line 109: I'm not 100% convinced adding padding is the best way to indent the paragraph to match but the design for this is a bit weird anyways. I wonder if you could include it in the ul element above but as a separate paragraph? Not really sure that fits the way ul is supposed to be used though...

TomHanmer29 (approved):
1 comment on the padding, but not sure the best solution for it. Happy for you to merge when addressed though (or you can keep as is if you can't get it to work any other way)

src/main/resources/templates/landlordPrivacyNotice.html
Bill-Haigh: yeah unfortunately that would make screen readers unhappy... I think in this instance that using the gov override is the right approach but definitely shouldn't be making a habit of it

PR #1302, opened 2026-05-07
PDJB-NONE: Include households and tenants in occupancy update email

Bill-Haigh:
## Ticket number

PDJB-NONE

## Goal of change

Fix the property update confirmation email so it lists households and tenants when a landlord transitions a property from unoccupied to occupied.

- The `UpdateOccupancy` journey already wrote new household / tenant counts (and other occupancy-related fields) when a landlord changed a property from unoccupied to occupied, but the confirmation email it sent only ever listed `Whether the property is occupied by tenants`.
- The email now also includes `The number of households living in this property` and `The number of people living in this property` whenever the property's previous state was unoccupied and the landlord has just made it occupied. The wording matches the bullets used by the dedicated `UpdateHouseholdsAndTenants` confirmation email.
- All other transitions (occupied → unoccupied, occupied → occupied, unoccupied → unoccupied) keep the existing single-bullet email behaviour.
- The pre-update `isOccupied` value is now snapshotted onto the journey state at journey-init time (alongside the existing `lastModifiedDate` snapshot), and the CYA submit reads it from there instead of re-deriving it from the database. This keeps the email consistent with what the user saw when they started the journey, and gives us a clear pattern for snapshotting other pre-update fields (rent, bedrooms, etc.) onto the journey state when we want to extend the email to cover more diffs.

The same unoccupied → occupied transition also writes other newly-required occupancy fields (bedrooms, furnished status, rent frequency, rent amount, bills included). Per discussion, the email intentionally lists only the fields that have their own dedicated update-email journey today — i.e. households and tenants — so it stays in step with `UpdateHouseholdsAndTenantsCyaConfig`. None of bedrooms / furnished status / rent / bills currently sends its own confirmation email; when we want them in the email, the new `wasOccupied` snapshot pattern on `UpdateOccupancyJourneyState` is the place to add them.

## Checklist

- [x] Test suite has been run in full locally and is passing — only `UpdateOccupancyCyaConfigTests` and `ktlintCheck` were run; the change is confined to one CYA config and is fully covered by those targeted unit tests, so the full suite was not run
- [x] Branch has been rebased onto main and run locally, with everything working as expected (both for your new feature and any related functionality) — branched from `origin/main` in a fresh worktree; targeted tests + ktlintCheck pass

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/occupancy/UpdateOccupancyCyaConfig.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/occupancy/UpdateOccupancyCyaConfig.kt
@@ -4,7 +4,6 @@ import kotlinx.datetime.Instant
 import kotlinx.datetime.toJavaInstant
 import org.springframework.context.MessageSource
 import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.JourneyFrameworkComponent
-import uk.gov.communities.prsdb.webapp.database.entity.PropertyOwnership
 import uk.gov.communities.prsdb.webapp.exceptions.NotNullFormModelValueIsNullException.Companion.notNullValue
 import uk.gov.communities.prsdb.webapp.journeys.shared.helpers.OccupancyDetailsHelper
 import uk.gov.communities.prsdb.webapp.journeys.shared.stepConfig.AbstractCheckYourAnswersStep
@@ -45,8 +44,6 @@ class UpdateOccupancyCyaConfig(
     override fun afterStepDataIsAdded(state: UpdateOccupancyJourneyState) {
         val isOccupied = isOccupied(state)
         val billsIncludedDataModel = state.getBillsIncludedOrNull()
-        val propertyOwnership = propertyOwnershipService.getPropertyOwnership(state.propertyId)
-        val wasOccupied = propertyOwnership.isOccupied
         propertyOwnershipService.updateOccupancy(
             id = state.propertyId,
             numberOfHouseholds =
@@ -87,18 +84,18 @@ class UpdateOccupancyCyaConfig(
                 },
             initialLastModifiedDate = Instant.parse(state.lastModifiedDate).toJavaInstant(),
         )
-        sendUpdateConfirmationEmail(propertyOwnership, wasOccupied = wasOccupied, isOccupied = isOccupied)
+        sendUpdateConfirmationEmail(state, isOccupied = isOccupied)
     }
 
     private fun sendUpdateConfirmationEmail(
-        propertyOwnership: PropertyOwnership,
-        wasOccupied: Boolean,
+        state: UpdateOccupancyJourneyState,
         isOccupied: Boolean,
     ) {
+        val propertyOwnership = propertyOwnershipService.getPropertyOwnership(state.propertyId)
         val bullets =
             buildList {
                 add("Whether the property is occupied by tenants")
-                if (!wasOccupied && isOccupied) {
+                if (!state.wasOccupied && isOccupied) {
                     add("The number of households living in this property")
                     add("The number of people living in this property")
                 }
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/occupancy/UpdateOccupancyCyaConfig.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/occupancy/UpdateOccupancyCyaConfig.kt
@@ -4,6 +4,7 @@ import kotlinx.datetime.Instant
 import kotlinx.datetime.toJavaInstant
 import org.springframework.context.MessageSource
 import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.JourneyFrameworkComponent
+import uk.gov.communities.prsdb.webapp.database.entity.PropertyOwnership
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/occupancy/UpdateOccupancyCyaConfig.kt
```diff
@@ -84,17 +84,28 @@ class UpdateOccupancyCyaConfig(
...
+    private fun sendUpdateConfirmationEmail(
+        state: UpdateOccupancyJourneyState,
+        isOccupied: Boolean,
+    ) {
         val propertyOwnership = propertyOwnershipService.getPropertyOwnership(state.propertyId)
+        val bullets =
+            buildList {
+                add("Whether the property is occupied by tenants")
+                if (!state.wasOccupied && isOccupied) {
+                    add("The number of households living in this property")
```
Travis-Softwire, line 99: Maybe for another ticket if using this pattern everywhere, but we should consider pulling these from a messages file. Maybe one for dev catch-up?

Travis-Softwire, line 99: You've rightly pointed out in the PR description that there are other fields in the same position - is the plan to do those in another PR? Or is the content for the other relevant update emails not suitable for using here in this context?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/occupancy/UpdateOccupancyCyaConfig.kt
```diff
@@ -84,17 +84,28 @@ class UpdateOccupancyCyaConfig(
...
-    private fun sendUpdateConfirmationEmail(state: UpdateOccupancyJourneyState) {
+    private fun sendUpdateConfirmationEmail(
+        state: UpdateOccupancyJourneyState,
+        isOccupied: Boolean,
+    ) {
         val propertyOwnership = propertyOwnershipService.getPropertyOwnership(state.propertyId)
+        val bullets =
+            buildList {
+                add("Whether the property is occupied by tenants")
+                if (!state.wasOccupied && isOccupied) {
```
Travis-Softwire, line 98: If I update the occupancy state from occupied -> occupied (which I think it lets me do?) do I go straight back to the property record or do I get sent to the households page?

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/UpdateOccupancyCyaConfigTests.kt
```diff
@@ -1,180 +1,239 @@
...
+            )
+        stepConfig.routeSegment = UpdateOccupancyCyaStep.ROUTE_SEGMENT
+        stepConfig.validator = AlwaysTrueValidator()
+        stepConfig.afterStepIsReached(mockState)
+        whenever(mockState.propertyId).thenReturn(propertyId)
+        whenever(mockState.lastModifiedDate).thenReturn(initialLastModifiedDate.toString())
+        whenever(mockState.occupied).thenReturn(mockOccupiedStep)
+        whenever(mockOccupiedStep.formModel).thenReturn(mockOccupancyFormModel)
+        whenever(mockOccupancyFormModel.occupied).thenReturn(true)
+        lenient().`when`(mockState.wasOccupied).thenReturn(false)
```
Travis-Softwire, line 133: What's `lenient` doing here? Not sure I've seen us use that anywhere else so far?

Travis-Softwire (approved):
A few Qs from me - happy to chat through them, but I don't think I'll need to see it again unless something radically changes as a result of those Qs

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/occupancy/UpdateOccupancyCyaConfig.kt
Bill-Haigh: Definitely agree, will add it to the thread

Bill-Haigh: we don't have designs for them, I flagged it when I did the original property update emails so I think there is a ticket somewhere

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/occupancy/UpdateOccupancyCyaConfig.kt
Bill-Haigh: goes to the households page

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/UpdateOccupancyCyaConfigTests.kt
Bill-Haigh: essentially to prevent it from throwing an `UnecessaryStubbingException`

Bill-Haigh: it's used in `NgdAddressLoaderTests`

PR #1304, opened 2026-05-07
PDJB-NONE: Restrict CloudWatch metrics to JVM and process diagnostics

Travis-Softwire:
## Ticket number

PDJB-NONE

## Goal of change

Reduce the volume of metrics published to CloudWatch to keep custom metric costs and namespace clutter under control.

- Adds an additional `MeterFilter` bean that allow-lists only metrics relevant to JVM and process diagnostics
- Allowed prefixes: `jvm.`, `process.`, `system.cpu`, `system.load`
- Drops Tomcat, Hikari, Logback, executor, disk space, file descriptor and Spring Cloud AWS internal SDK metrics from the export

The retained metrics cover everything currently used to diagnose memory and GC behaviour (heap, non-heap pool committed/used, classes loaded/unloaded, threads, direct buffers, GC pauses, CPU). Additional categories can be allow-listed back in if a future investigation needs them.

Expected reduction is roughly 75% of the current published metric count.

The existing `http.server.requests` deny filter is left in place. Both filters apply (Micrometer ANDs `MeterFilter` beans).

## Checklist

- [ ] No special release instructions


TomHanmer29 (approved):
LGTM if you're happy that these are all the types we'll need :)

PR #1310, opened 2026-05-08
PDJB-878: Format numbers correctly on submit

samyou-softwire:
## Ticket number

closes [PDJB-878](https://mhclgdigital.atlassian.net/browse/PDJB-878)

## Goal of change

before saving in the database you could enter valid but odd numbers, eg with trailing zeros. these would be shown to the user on CYA pages and would only be corrected on saving to the database

rent cya page with 00.10 rent amount

this should be fixed to show **0.1**

adds custom setters to impacted fields which call a normalisation function before setting the value

adds tests for the registration and update flows

I think this is the most sensible method vs changing CYA display logic. it means the number will be correct when displayed anywhere which reduces the likelihood of a regression occurring on another field and we keep the best values in long-lived state, though happy to hear otherwise

## Checklist

  mention that here

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/StringExtensions.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/StringExtensions.kt
@@ -1,0 +1,9 @@
+package uk.gov.communities.prsdb.webapp.helpers.extensions
+
+class StringExtensions {
+    companion object {
+        fun String.toNormalizedDecimalString(): String = toBigDecimalOrNull()?.toPlainString() ?: this
+
+        fun String.toNormalizedIntegerString(): String = toIntOrNull()?.toString() ?: this
+    }
+}
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/StringExtensions.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/StringExtensions.kt
@@ -2,7 +2,11 @@ package uk.gov.communities.prsdb.webapp.helpers.extensions
 
 class StringExtensions {
     companion object {
-        fun String.toNormalizedDecimalString(): String = toBigDecimalOrNull()?.toPlainString() ?: this
+        fun String.toNormalizedCurrencyString(): String {
+            val decimal = toBigDecimalOrNull()?.stripTrailingZeros() ?: return this
+            val scaled = if (decimal.scale() < 2) decimal.setScale(2) else decimal
+            return scaled.toPlainString()
+        }
 
         fun String.toNormalizedIntegerString(): String = toIntOrNull()?.toString() ?: this
     }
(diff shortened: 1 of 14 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDetailsUpdateJourneyTests.kt
```diff
@@ -432,7 +456,8 @@ class PropertyDetailsUpdateJourneyTests : IntegrationTestWithMutableData("data-l
                 var propertyDetailsPage = navigator.goToPropertyDetailsLandlordView(occupiedPropertyOwnershipId)
                 // Assert initial rent includes bills status is not Yes
                 assertThat(propertyDetailsPage.propertyDetailsSummaryList.rentIncludesBillsRow.value)
-                    .not().containsText("Yes")
+                    .not()
+                    .containsText("Yes")
```
samyou-softwire, line 460: linter asserts this

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/StringExtensions.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.communities.prsdb.webapp.helpers.extensions
+
+class StringExtensions {
+    companion object {
+        fun String.toNormalizedDecimalString(): String = toBigDecimalOrNull()?.toPlainString() ?: this
```
Travis-Softwire: Out of interest - how does this affect __trailing__ zeros? Is 00.10 -> 0.1 or 0.10? And is 0.100 -> 0.100, 0.10, or 0.1?

Travis-Softwire (approved):
One Q but otherwise LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/StringExtensions.kt
samyou-softwire: good question. I checked and it does not impact trailing zeros. the number is parsed as eg BigDecimal `0.100`. looks like we'd need more code to account for this, 0.1 should probably be displayed as 0.10. worth raising a backlog ticket for this?

Travis-Softwire: I think what I'm getting at is that if we have more string formatting to do on these numbers we should probably do that here, in case it requires replacing this fix with a different one? It should only affect the decimal entries though (which is just this one?)

samyou-softwire: good point; I'll take a look into fixing that

samyou-softwire: turns out this a little finnicky. a question I'm not so sure on in whether "100" should -> "100" or "100.00". this one for the design team? I've pushed some code that does the latter

Travis-Softwire (approved):
Nice - happy with that approach :)

samyou-softwire:
@Travis-Softwire alright to take another quick look at this again? updated a couple tests to reflect the new logic

Travis-Softwire (approved):
Yep, all LGTM

PR #1314, opened 2026-05-11
PDJB-770: compliance update email fixes

TomHanmer29:
## Ticket number

PDJB-770

## Goal of change

Fixes gas and electricity emails not sending, and fixes a small amount of copy

Adds some missing commas, removes an "open in new tab" and fixes emails not sending

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
@@ -197,7 +197,7 @@ class PropertyComplianceService(
             gasSafetyCertUploadIds = gasSafetyCertUploadIds,
         )
 
-        if (gasSafetyCertUploadIds.isNotEmpty()) {
+        if (gasSafetyCertIssueDate != null) {
             sendCertificateUpdateEmail(
                 propertyCompliance,
                 isExpired = propertyCompliance.isGasSafetyCertExpired == true,
@@ -237,7 +237,7 @@ class PropertyComplianceService(
             electricalCertType = electricalCertType,
         )
 
-        if (electricalSafetyCertUploadIds.isNotEmpty()) {
+        if (electricalSafetyExpiryDate != null) {
             val certTypeAbbreviation = if (requireNotNull(electricalCertType) == CertificateType.Eic) "EIC" else "EICR"
             sendCertificateUpdateEmail(
                 propertyCompliance,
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -197,7 +197,7 @@ class PropertyComplianceService(
             gasSafetyCertUploadIds = gasSafetyCertUploadIds,
         )
 
-        if (gasSafetyCertUploadIds.isNotEmpty()) {
+        if (gasSafetyCertIssueDate != null) {
```
samyou-softwire, line 200: to check what was the logic error here? why can't we check if the uploadIds are non-empty to know whether theres any previous gas safety certs?

TomHanmer29: For an expired cert, we won't have any uploadIds as the user won't have uploaded a cert. Using issueDate or expiryDate here ensures that a cert of any kind was added, even if the user didn't end up doing an upload (the flow goes enter date -> upload cert)

samyou-softwire: gotcha 👍

src/main/resources/emails/emailTemplates.json
```diff
@@ -86,29 +86,29 @@
     "bodyLocation": "/emails/BetaFeedbackEmail.md"
   },
   {
-    "test_id": "df943327-ce57-4ccc-990c-a0f8d88293ff",
-    "prod_id": "71e9fede-c824-4f5b-957e-24295a1d3e1b",
+    "test_id": "a596ebce-7bc1-44c7-a3fb-32d09991f635",
+    "prod_id": "7f1d1ab1-4d9a-45ac-b209-0aa83eff2320",
```
samyou-softwire, line 90: will we have to do anything post release to remove the old templates?

samyou-softwire (approved):
LGTM

Travis-Softwire (approved):
👍

src/main/resources/emails/emailTemplates.json
TomHanmer29: Yes, we do. There's already a PR with [special instructions](https://github.com/communitiesuk/prsdb-webapp/pull/1309) to remove. Basically we just carry that along and then delete the old ones when we deploy to prod

PR #1315, opened 2026-05-11
PDJB-879: Fix task list back button from register page

samyou-softwire:
## Ticket number

[PDJB-879](https://mhclgdigital.atlassian.net/browse/PDJB-879)

## Goal of change

Fix an edge case where when navigating from the 'Register a property' flow the task list would have no back button

sets an appropriate back link url using `BackUrlStorageService`. this is used by the template

task list with back url

## Checklist

  mention that here

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
@@ -39,10 +39,10 @@ import uk.gov.communities.prsdb.webapp.journeys.NoSuchJourneyException
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.JointLandlordsPropertyRegistrationStrategy
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.PropertyRegistrationJourneyFactory
 import uk.gov.communities.prsdb.webapp.models.dataModels.RegistrationNumberDataModel
+import uk.gov.communities.prsdb.webapp.services.BackUrlStorageService
 import uk.gov.communities.prsdb.webapp.services.CollectionKeyParameterService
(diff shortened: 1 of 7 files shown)


TomHanmer29 (commented):
Not had a look at how this actually connects up but what's the reason for using this slightly more manual approach instead of doing something like a backStep in the journey framework? Is it because the task list lives outside of the normal step structure? If that's all good then I'll give this an approve :)

samyou-softwire:
> Not had a look at how this actually connects up but what's the reason for using this slightly more manual approach instead of doing something like a backStep in the journey framework? Is it because the task list lives outside of the normal step structure? If that's all good then I'll give this an approve :)

the precedent for other pages is to use a back link override like this, the other example is the `IncompletePropertiesController` which uses the same back link override to make pressing back on the task list go back to incomplete properties (albeit with a lot more mechanism in the way)

I think it makes sense to have this as a separate override, it's not really part of the register property journey

TomHanmer29 (approved):
One other thought I had - how does this interact with the compliance journeys now routing to the task list? Is that back behaviour still preserved? Happy to approve this though, and if there's an issue it can be done in another ticket (but I think it might be fine?)

samyou-softwire:
> One other thought I had - how does this interact with the compliance journeys now routing to the task list? Is that back behaviour still preserved? Happy to approve this though, and if there's an issue it can be done in another ticket (but I think it might be fine?)

don't think we touched that here. the task list doesn't have a back button in any other cases. I can take a look into adding a back button based on where you see it on the registration flow

Travis-Softwire (approved):
I'm fine with this, and suspect this is what we'll need to get it working the way designers would like - although if we do run into any further issues on this then I think for day 1 we just make it static and point it it at the property registration start page.

Travis-Softwire (approved):
LGTM

PR #1316, opened 2026-05-11
Fix/pdjb 882 save cya progress to db

JasminConterioSW:
## Ticket number

PDJB-882

## Goal of change

Save changes from CYA journeys in SavedJourneyState

Refactors FinishCyaJourneyConfig so
* `afterStepIsReached` now
    * Checks whether the journey data has changed while the user was checking answers - moved to here because it needs to be done before the CYA changes are saved in the DB
    * Copies CYA journeyData to the baseJourney (so it happens for every CYA journey, regardless of whether it will be saved to the DB)
 *`saveState` saves the base journey in the DB, not the CYA journey 
    * Note that this function only executes if saving the state is allowed, so for example it won't execute for the landlord registration journey where we should never be saving the journey to the DB
 * `resolveNextDestination` still cleans up, deletes the journey and returns the user to the Cya page as before

Also now saves the address task in the property registration journey - this is important so that if a user resumes their journey, changes the address and does nothing else, the updated address is still saved in the SavedJourneyState and updates on the incomplete property page

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/FinishCyaJourneyConfig.kt
```diff
@@ -12,23 +13,31 @@ import uk.gov.communities.prsdb.webapp.journeys.shared.states.CheckYourAnswersJo
...
-            state.deleteJourney()
         } else {
-            state.deleteJourney()
             throw CyaDataHasChangedException("Journey data has changed since the user started checking their answers.")
         }
+    }
+
+    override fun saveState(state: CheckYourAnswersJourneyState): SavedJourneyState {
+        return state.getBaseJourneyState().save()
+        // return super.saveState(state)
```
TomHanmer29: We should probably get rid of this comment

TomHanmer29 (approved):
LGTM! One tiny comment

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/FinishCyaJourneyConfig.kt
JasminConterioSW: Oops!

PR #1317, opened 2026-05-11
PDJB-851: Show CYA certificate filenames as download links once scanned

Bill-Haigh:
## Ticket number

PDJB-851

## Goal of change

Fix the property registration Check Your Answers pages so uploaded gas/electrical safety certificates appear as download links once virus scanning completes, instead of always being shown as plain filename text.

- On the Gas Safety and Electrical Safety registration CYA pages, uploaded certificate rows now render as a download link once the file is scanned, mirroring the behaviour already used on the property compliance journey.
- Files still being virus-scanned show the filename followed by "(Pending virus scan)" rather than appearing as a plain (non-functional) link.
- Files that have been deleted (e.g. flagged as infected) are filtered out of the CYA row instead of leaving a stale entry.
- The same fix is applied to the property compliance journey's compliance details summary, because it builds its summary via the same two factory classes (via ComplianceDetailsHelper).

The fix touches both the registration CYA pages and the compliance details summary used after registration, because both call into the same row factories. If you would prefer the compliance-journey change split into a separate PR, happy to do that.

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/ElectricalSafetyRegistrationCyaSummaryRowsFactory.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/ElectricalSafetyRegistrationCyaSummaryRowsFactory.kt
@@ -1,6 +1,5 @@
 package uk.gov.communities.prsdb.webapp.journeys.propertyRegistration
 
-import uk.gov.communities.prsdb.webapp.constants.enums.FileUploadStatus
 import uk.gov.communities.prsdb.webapp.constants.enums.HasElectricalSafetyCertificate
 import uk.gov.communities.prsdb.webapp.journeys.Destination
 import uk.gov.communities.prsdb.webapp.journeys.JourneyStep
@@ -8,10 +7,8 @@ import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.states.Elec
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.ElectricalSafetyScenario
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.HasElectricalCertMode
 import uk.gov.communities.prsdb.webapp.models.viewModels.summaryModels.SummaryListRowViewModel
-import uk.gov.communities.prsdb.webapp.models.viewModels.summaryModels.UploadedFileUrl
+import uk.gov.communities.prsdb.webapp.models.viewModels.summaryModels.propertyComplianceViewModels.toUploadedFileUrls
 import uk.gov.communities.prsdb.webapp.services.UploadService
-
-private const val VIRUS_SCAN_PENDING_WITH_NAME_KEY = "propertyCompliance.uploadedFile.virusScanPendingWithName"
 
 class ElectricalSafetyRegistrationCyaSummaryRowsFactory(
     private val state: ElectricalSafetyState,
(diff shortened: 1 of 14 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/ElectricalSafetyRegistrationCyaSummaryRowsFactory.kt
```diff
@@ -33,16 +39,37 @@ class ElectricalSafetyRegistrationCyaSummaryRowsFactory(
...
+                            )
+
+                        FileUploadStatus.QUARANTINED ->
+                            UploadedFileUrl(
+                                messageKey = VIRUS_SCAN_PENDING_WITH_NAME_KEY,
+                                displayName = upload.fileName,
+                                url = null,
+                            )
+
+                        FileUploadStatus.DELETED -> null
```
JasminConterioSW: There's a `ComplianceUploadRowHelper `with something very similar to this, you might be able to either just use that or commonise some of it? Here and for the gas version

TomHanmer29: Yep, I definitely think they should be commonised! Was about to leave the same point :) Pretty sure the logic could be extracted and put into a shared place if you can't just directly reuse it

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/ElectricalSafetyRegistrationCyaSummaryRowsFactoryTests.kt
```diff
@@ -8,15 +8,21 @@ import org.junit.jupiter.api.Test
...
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.states.ElectricalSafetyState
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.CheckElectricalCertUploadsStep
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.ElectricalCertExpiryDateStep
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.HasElectricalCertMode
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.HasElectricalCertStep
+import uk.gov.communities.prsdb.webapp.models.viewModels.summaryModels.UploadedFileUrl
+import uk.gov.communities.prsdb.webapp.services.UploadService
 
 @ExtendWith(MockitoExtension::class)
 class ElectricalSafetyRegistrationCyaSummaryRowsFactoryTests {
```
TomHanmer29, line 28: There's a lot of duplication between the tests for this and the gas one. I wonder if we can commonise them more than they already are, given they're sharing a helper method now too (perhaps by restricting more of the tests to just checking `.toUploadedFileUrls`).

TomHanmer29 (commented):
Have you checked that this displays properly for multiple file names? I know in the compliance tab, I had to make an HTML change to add in a box that would show multiple links stacked on top of each other, and I'm not sure we've made a similar change / reused that component here (unless it already does it in the summary list, which it might!!). I'm happy to approve (other than the tests) if you can grab a screenshot for me :)

Bill-Haigh:
> Have you checked that this displays properly for multiple file names? I know in the compliance tab, I had to make an HTML change to add in a box that would show multiple links stacked on top of each other, and I'm not sure we've made a similar change / reused that component here (unless it already does it in the summary list, which it might!!). I'm happy to approve (other than the tests) if you can grab a screenshot for me :)

screenshots:

image

image

image

TomHanmer29 (approved):

PR #1321, opened 2026-05-12
PDJB-884: update page back link fixes

TomHanmer29:
## Ticket number

PDJB-884

## Goal of change

Adds back links to property details update journey pages that were missing them.

Adds some missing back links for update pages that did not have them.

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/states/CheckYourAnswersJourneyState.kt
```diff
@@ -76,6 +76,10 @@ interface CheckYourAnswersJourneyState : JourneyState {
                 initialStep()
                 backDestination { journey.returnToCyaPageDestination }
                 nextStep { journey.finishCyaStep }
+                conditionallyConfigure(
+                    condition = { true },
+                    configuration = { backDestinationIfNotSet { journey.returnToCyaPageDestination } },
+                )
```
AEPR: I think this means that all steps within the journey without explicit back links will instead go back to the CYA page.

Is that right?

TomHanmer29: Yes, that's the intention. When a user presses a change link, and then gets taken to the change page but goes back, I think we always want to go back to the cya page again. See @Travis-Softwire comments (with ?) on [PDJB-884](https://mhclgdigital.atlassian.net/browse/PDJB-884)

JasminConterioSW: Is that just for the step that you navigate too, or for other steps in the task?

So for something like the address task in Landlord Registration, if they got a couple of steps through the task and clicked back would they go to the CYA page instead of the earlier step in the task? 

I think we don't want them to go further back in the journey than the entry point, but probably do want them to be able to go back up to that entry point?

JasminConterioSW: (I've not looked at it properly though)

Travis-Softwire: I'm not sure this is quite what we want. I think what we want is to set the back link to the Cya page IFF the current step is the one that the journey is checking, i.e. `state.checkingAnswerFor == step.routeSegement`. So if I check the answer for e.g. are bills included, change from no -> yes, and go forwrad to which bills are included, the back link should go back to bills included, and clicking the back link __again__ should take me back to the CYA page, whereas if I click on the change link for which bills are included, the back link should take me straight back to the CYA page. Shout if that doesn't make sense!

Travis-Softwire (changes requested):

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/updateLicensing/UpdatePropertyLicensingJourneyFactory.kt
```diff
@@ -84,7 +86,7 @@ class UpdateLicensingJourneyFactory(
             step(journey.cyaStep) {
                 routeSegment(UpdateLicensingCyaStep.ROUTE_SEGMENT)
                 parents { journey.licensingTask.isComplete() }
-                nextUrl { LANDLORD_PROPERTY_DETAILS_ROUTE }
+                nextUrl { propertyDetailsRoute }
```
TomHanmer29, line 89: this seems like the standard convention, not sure why this page didn't use it (they resolve to the same thing)

TomHanmer29:
Chatted with @Travis-Softwire , we've decided to shelve the additional back link fixes for going back through mutli-journey steps (probably best achieved via adding some config to a task that lets you apply a back link to all it's children, so we can do that to the CYA tasks), and just use this PR for fixing the missing back links!

Travis-Softwire (approved):

PR #1323, opened 2026-05-12
PDJB-891: Fix 500 when changing bills included from rent includes bills CYA page

Bill-Haigh:
## Ticket number

PDJB-891

## Goal of change

Fix a 500 error when clicking the "Change" link next to "Which bills are included" on the rent-includes-bills update Check Your Answers page.

- The `finishCyaStep` in `UpdateRentIncludesBillsJourneyFactory.checkYourAnswersJourneyMap` previously gated its parents on `rentIncludesBillsTask.isComplete()`. `Task.isComplete()` reads a `lateinit` `subJourneyBuilder` that is only initialised when `Task.makeSubJourney()` runs (via `checkAnswerTask`). The `BillsIncludedStep` branch of the CYA journey map uses `checkAnswerStep` instead, so the task was never initialised and the parents predicate threw `UninitializedPropertyAccessException` → 500.
- Changed that step to `initialStep()`, matching the pattern used in `UpdateOccupancyJourneyFactory.checkYourAnswersJourneyMap` (the other update journey that mixes `checkAnswerTask` and `checkAnswerStep` branches).
- Added a regression integration test that drives the bug repro (CYA → Change "Which bills are included" → update selection → submit) and asserts the new value is persisted.

- This is a one-line behavioural fix. The other three update journey factories that still use `parents { task.isComplete() }` (`UpdatePropertyLicensingJourneyFactory`, `UpdateRentFrequencyAndAmountJourneyFactory`, `UpdateHouseholdsAndTenantsJourneyFactory`) only have `checkAnswerTask` branches, so they don't hit the lateinit and don't need changing.
- Possible follow-up (not in scope here): defensively guarding `Task.isComplete()` to return `false` when `subJourneyBuilder` is uninitialised would prevent any future occurrence of this class of bug.

## Checklist

- [x] Test suite has been run in full locally and is passing (RentIncludesBills nested class — 3/3, plus local smoke test against bootRun)

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/rentIncludesBills/UpdateRentIncludesBillsJourneyFactory.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/rentIncludesBills/UpdateRentIncludesBillsJourneyFactory.kt
@@ -100,7 +100,7 @@ class UpdateRentIncludesBillsJourneyFactory(
                 else -> throw IllegalStateException("Unknown step being checked: $checkingAnswersFor")
             }
             step(journey.finishCyaStep) {
-                parents { journey.rentIncludesBillsTask.isComplete() }
+                initialStep()
                 nextDestination { Destination.Nowhere() }
             }
             configureStep(journey.rentIncludesBills) {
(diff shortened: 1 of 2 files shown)


JasminConterioSW (approved):
It seems odd to make it an initial step, but looks like we do that in a few places already so let's go with it for now

PR #1324, opened 2026-05-12
PDJB-887: Restrict system operator pages to link to other system operator pages

samyou-softwire:
## Ticket number

[PDJB-887](https://mhclgdigital.atlassian.net/browse/PDJB-887)

## Goal of change

makes it so links in /system-operatorr still direct to /system-operator and not the local authority pages

add needed handlers to ManageLocalAuthorityCouncilUsersController

make needed changes on templates to use these methods

all the routes existed already, just weren't linked to

the dashboard button wasn't touched, it's waiting for the system operator dashboard to be created

user list system operator view

## Checklist

  mention that here

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalCouncilUsersController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalCouncilUsersController.kt
@@ -432,28 +432,57 @@ class ManageLocalCouncilUsersController(
         }
     }
 
+    private fun isSystemOperatorRequest(request: HttpServletRequest): Boolean =
+        request.requestURI.startsWith("/$SYSTEM_OPERATOR_PATH_SEGMENT/")
+
     private fun getRequestPathPrefix(request: HttpServletRequest) =
-        if (request.requestURI.startsWith("/$SYSTEM_OPERATOR_PATH_SEGMENT/")) {
+        if (isSystemOperatorRequest(request)) {
             SYSTEM_OPERATOR_PATH_SEGMENT
         } else {
             LOCAL_COUNCIL_PATH_SEGMENT
         }
 
-    private fun getManageUsersRoute(
+    private fun getSystemOperatorOrLocalCouncilRoute(
+        request: HttpServletRequest,
+        systemManagerRoute: String,
+        localCouncilRoute: String,
+    ): String =
+        if (isSystemOperatorRequest(request)) {
+            systemManagerRoute
+        } else {
+            localCouncilRoute
+        }
+
+    fun getManageUsersRoute(
         localCouncilId: Int,
         request: HttpServletRequest,
-    ): String = "/${getRequestPathPrefix(request)}/$localCouncilId/$MANAGE_USERS_PATH_SEGMENT"
+    ): String =
+        getSystemOperatorOrLocalCouncilRoute(
+            request,
+            getSystemOperatorManageUsersRoute(localCouncilId),
+            getLocalCouncilManageUsersRoute(localCouncilId),
+        )
 
     private fun getInviteNewUserRoute(
         localCouncilId: Int,
         request: HttpServletRequest,
-    ): String = "/${getRequestPathPrefix(request)}/$localCouncilId/$INVITE_NEW_USER_PATH_SEGMENT"
+    ): String =
+        getSystemOperatorOrLocalCouncilRoute(
+            request,
+            getSystemOperatorInviteNewUserRoute(localCouncilId),
+            getLocalCouncilInviteNewUserRoute(localCouncilId),
+        )
 
     private fun getDeleteUserRoute(
         localCouncilId: Int,
         localCouncilUserId: Long,
         request: HttpServletRequest,
-    ): String = "/${getRequestPathPrefix(request)}/$localCouncilId/$DELETE_USER_PATH_SEGMENT/$localCouncilUserId"
+    ): String =
+        getSystemOperatorOrLocalCouncilRoute(
+            request,
+            getSystemOperatorDeleteUserRoute(localCouncilId, localCouncilUserId),
+            getLocalCouncilDeleteUserRoute(localCouncilId, localCouncilUserId),
+        )
 
     private fun getEditUserBaseUrl(
         localCouncilId: Int,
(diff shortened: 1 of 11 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalCouncilUsersControllerTests.kt
```diff
@@ -460,27 +576,27 @@ class ManageLocalCouncilUsersControllerTests(
...
         @WithMockUser(roles = ["SYSTEM_OPERATOR"])
         fun `updateUserAccessLevel updates the given user's access level when called by a system operator`() {
             setupLocalCouncilForSystemOperator(NON_ADMIN_LOCAL_COUNCIL_ID)
 
-            postUpdateUserAccessLevelAndAssertSuccess()
+            postUpdateUserAccessLevelAndAssertSuccess(getSystemOperatorManageUsersRoute(DEFAULT_LOCAL_COUNCIL_ID))
         }
 
-        private fun postUpdateUserAccessLevelAndAssertSuccess(localCouncilId: Int = DEFAULT_LOCAL_COUNCIL_ID) {
+        private fun postUpdateUserAccessLevelAndAssertSuccess(expectedRedirectUrl: String) {
```
samyou-softwire: tweaked this function as the `localCouncilId` arg didnt seem to be used

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalCouncilUsersController.kt
```diff
@@ -428,6 +432,57 @@ class ManageLocalCouncilUsersController(
...
+        localCouncilUserId: Long,
+        request: HttpServletRequest,
+    ): String =
+        if (request.isUserInRole(ROLE_SYSTEM_OPERATOR)) {
+            getSystemOperatorDeleteUserRoute(localCouncilId, localCouncilUserId)
+        } else {
+            getLocalCouncilDeleteUserRoute(localCouncilId, localCouncilUserId)
+        }
+
+    private fun getEditUserBaseUrl(
```
samyou-softwire, line 463: this is not like the other functions, the current template for edit users adds the ID as part of the template. in theory this could be moved to the view model, but this seemed overkill so I've left it as is. see src/main/resources/templates/manageLocalCouncilUsers.html line 32

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalCouncilUsersController.kt
```diff
@@ -428,6 +432,57 @@ class ManageLocalCouncilUsersController(
         }
     }
 
+    private fun getManageUsersRoute(
+        localCouncilId: Int,
+        request: HttpServletRequest,
+    ): String =
+        if (request.isUserInRole(ROLE_SYSTEM_OPERATOR)) {
```
JasminConterioSW: I'm not sure about this, because one user can have multiple roles so they could be both a SYSTEM_OPERATOR and a LOCAL_COUNCIL_ADMIN.

Instead of checking the user role, can you check the request URI to see if it contains `SYSTEM_OPERATOR_PATH_SEGMENT`?

samyou-softwire: ah good point, I'll update that

out of interest what would it mean for a user to be both a system operator and a local council admin? can system operators not access everything?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalCouncilUsersController.kt
```diff
@@ -428,6 +432,39 @@ class ManageLocalCouncilUsersController(
...
     }
 
+    private fun getRequestPathPrefix(request: HttpServletRequest) =
+        if (request.requestURI.startsWith("/$SYSTEM_OPERATOR_PATH_SEGMENT/")) {
+            SYSTEM_OPERATOR_PATH_SEGMENT
+        } else {
+            LOCAL_COUNCIL_PATH_SEGMENT
+        }
+
+    private fun getManageUsersRoute(
```
Travis-Softwire: It feels quite odd to have both these and the companion object functions below (which AFAICT are only used in the tests?). Is there no way to commonise them?

Travis-Softwire (changes requested):

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalCouncilUsersController.kt
JasminConterioSW: I don't think they can access the Local Council dashboard or the search pages.

Of course they can just invite themselves to be an LC admin if they want to look at those pages!

Travis-Softwire: Tbf, you can't __currently__ be both a system operators (which requires a OL account) and a LC admin (which requires an internal access account). But this used to be possible, and likely will be possible again in the future..

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalCouncilUsersController.kt
samyou-softwire: I've done some commonisation so the route structures are only defined once. I think I could do more, but not sure if we prefer having each URL defined explicitly as a const (leading to two consts per URL) or if we're fine to have a method that generates each route dynamically

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalCouncilUsersController.kt
```diff
@@ -428,6 +432,68 @@ class ManageLocalCouncilUsersController(
...
+        request.requestURI.startsWith("/$SYSTEM_OPERATOR_PATH_SEGMENT/")
+
+    private fun getRequestPathPrefix(request: HttpServletRequest) =
+        if (isSystemOperatorRequest(request)) {
+            SYSTEM_OPERATOR_PATH_SEGMENT
+        } else {
+            LOCAL_COUNCIL_PATH_SEGMENT
+        }
+
+    private fun getSystemOperatorOrLocalCouncilRoute(
```
Travis-Softwire: I don't think we need this extra level of indirection? I think you can call `isSystemOperatorRequest(request)` once in each controller method, and the just pass in a boolean. I think that keeps the signatures of the helpers a bit cleaner, as the request is passed down quite a long way just to evaluate the type of the user. It also probably means that you could just combine the system operator and LC versions of the `get<role><route type>Route()` functions since they are only used here and in the relevant tests.

Travis-Softwire (approved):
One further suggestion - LGTM once that's address though

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/ManageLocalCouncilUsersController.kt
samyou-softwire: so I've
- removed redundancy in the companion object methods
- made an enum for the type of view (felt more appropriate than a boolean)
- there are `getManageUsersRoute` that take in a request object and just get the request type from it to pass to a function of the same name. think this is fine for keeping the call sites a little cleaner so it only needs to call 1 function

sound good?

Travis-Softwire (approved):

PR #1325, opened 2026-05-13
Fix/pdjb 793 add pagination to incomplete properties

JasminConterioSW:
## Ticket number

PDJB-793

## Goal of change

Add pagination to the incomplete properties page

* Retrieves pages of incomplete properties from the database
* Adds pagination to the incomplete properties page (10 items per page)
* Refactors the property registration journey so that saving the incomplete property has its own step (moved from propertyType, now happens after the address is added)

image
image

This follows the approach we have used in other pagination components, see for example `index(...)` in `ManageLocalCouncilUsersController`

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/PropertyTypeStepConfig.kt
```diff
@@ -54,13 +50,6 @@ class PropertyTypeStepConfig(
     override fun chooseTemplate(state: JourneyState): String = "forms/propertyTypeForm"
 
     override fun mode(state: JourneyState) = getFormModelFromStateOrNull(state)?.let { Complete.COMPLETE }
-
-    override fun afterSaveState(
-        state: JourneyState,
-        saveStateId: SavedJourneyState,
-    ) {
-        incompletePropertyForLandlordService.addIncompletePropertyToLandlord(saveStateId)
```
JasminConterioSW, line 62: This shouldn't really be linked to `PropertyType`, that was just the first step where we were saving (which it no longer is)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/LandlordIncompletePropertiesRepository.kt
```diff
@@ -14,4 +16,10 @@ interface LandlordIncompletePropertiesRepository : JpaRepository<LandlordIncompl
 
     @Suppress("ktlint:standard:function-naming")
     fun countBySavedJourneyState_CreatedDateBefore(cutoffDate: Instant): Long
+
+    @Suppress("ktlint:standard:function-naming")
```
TomHanmer29, line 20: Do you know why we have weirdly named functions in here? Just curious!

TomHanmer29 (approved):
LGTM! 1 question but nothing that needs changing :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/LandlordIncompletePropertiesRepository.kt
JasminConterioSW: Spring JPA has a bunch of query methods you can construct so if you define in the interface, you don't have to explicitly write out what the function does.

So thinks like findBy... it knows what to do with

The underscores here tell it to look at the table referred to, so in `findByLandlord_BaseUser_Id`:
* LandlordIncompleteProperties has a landlord field(which is a Landlord entity)
    * Landlord has a BaseUser field (which is a PrsdbUser entity)
        * PrsdbUser has an Id - which is one we actually want to search by and match to the principalName
        
There's a bit more here https://docs.spring.io/spring-data/jpa/reference/repositories/query-methods-details.html

PR #1329, opened 2026-05-13
PDJB-869: user feedback surveys

TomHanmer29:
## Ticket number

PDJB-869

## Goal of change

Adds the feedback page insert as per the designs

image

image

No links have been added yet as we do not currently have them. I've added TODOS but we have no ticket number for this yet

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
@@ -202,7 +202,7 @@ class RegisterPropertyController(
                 "?$CONTEXT_ID_URL_PARAMETER={contextId}"
 
         // TODO: Replace with real survey URL
-        const val PROPERTY_REGISTRATION_SURVEY_URL = "https://example.com/property-registration-survey"
+        const val PROPERTY_REGISTRATION_SURVEY_URL = "#"
 
         fun getResumePropertyRegistrationPath(journeyId: String): String =
             UriTemplate(RESUME_PROPERTY_REGISTRATION_JOURNEY_ROUTE).expand(journeyId).toASCIIString()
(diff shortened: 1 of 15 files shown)


src/main/resources/templates/registerPropertyConfirmation.html
```diff
@@ -46,11 +46,23 @@
...
+                    ${propertyRegistrationSurveyUrl})}">
+                </div>
+            </th:block>
+
+            <p class="govuk-body" th:if="${!isFirstProperty}">
                 <a class="govuk-link"
                    th:href="@{${landlordDashboardUrl}}"
                    th:text="#{registerProperty.confirmation.goToDashboard}">registerProperty.confirmation.goToDashboard</a>
             </p>
+            <th:block th:if="${isFirstProperty}">
```
samyou-softwire: this seems a bit odd to me but checking ticket comments I see that it's on purpose to have this link/button conditional logic

samyou-softwire: I'm not super familiar with thymeleaf, though I wonder is there a reason we wrap this in a block? can we apply the th:if class to the a tag directly?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
```diff
@@ -120,6 +119,15 @@ class RegisterPropertyController(
             model.addAttribute("completeByDate", formattedDate)
         }
 
+        val propertyCount = propertyOwnershipService.getPropertyCountForLandlord(principal.name)
+        if (propertyCount == 0L) {
```
samyou-softwire, line 123: what would happen if this threw?

src/main/resources/templates/registerPropertyConfirmation.html
TomHanmer29: Yep, I think it's a bit odd, but confirmed with Larry

TomHanmer29: This tends to just be the pattern we're using in the codebase I think. I think it makes the conditionals look a bit neater, but you can apply the if to the tag directly in thymeleaf

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
TomHanmer29: It should just take you to the generic error page

samyou-softwire (approved):

PR #1331, opened 2026-05-14
PDJB-875: Log more details for virus scans

AEPR:
## Ticket number
PDJB-875

## Goal of change
Logs the target file and scan result when a scan finishes. We might be getting duplicate processing so this will help diagnose.


samyou-softwire (approved):
to check, what would an example log line look like?

AEPR:
image
This is an example of the success message, it will be the same but with the string text indicated.

> Processing virus scan result NoThreats for UploadedFileLocator(objectKey=noObjectSet, eTag=test-etag, versionId=33)

PR #1334, opened 2026-05-14
Not a PR, notes on seeding NFT

JasminConterioSW:
## Ticket number

PDJB-NONE

## Goal of change

Seed a few hundred records for pen testing, including a load in a particular LC

* We no longer have a certificate_uploads table, so seeded that is removed
* File uploads (gas and electric certificates) should be there but may not be correctly implemented - I removed these for this seeded but they should be implemented properly
* Some of the landlord data specifies the ids, but doesn't account for there already being data in the db. Added some constants which let the user add an offset to the ids if there are already records in the db (these might want renaming actually so it's the highest id in the db)
* PREFERRED_PROPERTY_ADDRESS_IDS lets you add ids of particular addresses you want to use - I used this to add addresses in a particular LC (once those ran out, it continued to select random address to make the rest of the property ownership records)

Running it:
* Try again the local DB using local-nft-seeder
* To run again nft, remove the "local" profile and set RDS_URL to point to where you have the nft db running
* A few environment variables then need adding to the local .env file (I added 8 - it will tell you what's missing when you try to run)

### DIFF
--- a/.run/local-nft-seeder.run.xml
+++ b/.run/local-nft-seeder.run.xml
@@ -18,7 +18,6 @@
     <method v="2">
       <option name="RunConfigurationTask" enabled="true" run_configuration_name="local_resources_up" run_configuration_type="docker-deploy" />
       <option name="Make" enabled="true" />
-      <option name="RunConfigurationTask" enabled="true" run_configuration_name="flywayClean-local-except-address" run_configuration_type="GradleRunConfiguration" />
     </method>
   </configuration>
 </component>
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NftDataSeeder.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NftDataSeeder.kt
@@ -40,7 +40,7 @@ class NftDataSeeder(
 
     private val registrationNumberGenerator = RegistrationNumberGenerator()
     private val landlordAddressGenerator = AddressGenerator()
-    private val propertyOwnershipAddressGenerator = AddressGenerator(restrictToAvailable = true)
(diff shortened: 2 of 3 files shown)


.run/local-nft-seeder.run.xml
```diff
@@ -18,7 +18,6 @@
     <method v="2">
       <option name="RunConfigurationTask" enabled="true" run_configuration_name="local_resources_up" run_configuration_type="docker-deploy" />
       <option name="Make" enabled="true" />
-      <option name="RunConfigurationTask" enabled="true" run_configuration_name="flywayClean-local-except-address" run_configuration_type="GradleRunConfiguration" />
```
JasminConterioSW, line 21: Make sure it doesn't try to wipe the database before running the task!

Might need to remove` flywayClean-local-except-address` from the run configuration locally from the "Before launch" section

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NftDataSeeder.kt
```diff
@@ -231,32 +230,47 @@ class NftDataSeeder(
...
+
+            // Flush any remaining batched items if we ran out of landlords before reaching NUM_OF_PROPERTIES
+            if (newPropertyRegistrations > 0 && newPropertyRegistrations % BATCH_SIZE != 0) {
+                registrationNumberStmt.executeBatch()
+                licenceStmt.executeBatch()
+                propertyOwnershipStmt.executeBatch()
+                propertyComplianceStmt.executeBatch()
+                reminderEmailSentStmt.executeBatch()
+                savedJourneyStateStmt.executeBatch()
+                incompletePropertyStmt.executeBatch()
```
JasminConterioSW, line 270: This was copilot, it may or may not be working....

It won't necessarily add all the properties if it runs out of landlords first (it just adds properties when it's adding landlords)

PR #1351, opened 2026-05-19
PDJB-904: align notify and app email validation

TomHanmer29:
## Ticket number

PDJB-904

## Goal of change

Aligns the email validation rules between notify and PRSDB

Implements the notify rules used [here](https://github.com/alphagov/notifications-utils/blob/995faf0d925f7e95ecac6ecec383b1d162b2eceb/notifications_utils/recipient_validation/email_address.py)
Also adds comprehensive tests covering all the cases

I've not included any updates to the case where notify still errors - this prevents a property being registered, which I think is fine.

There's also a weird test that started erroring as part of my changes. I removed it as I think the test is testing an assertion which is actually not true - the first element returned when there are multiple violations is not always from the constraint earlier in the declaration order - the test displayed 2 violations, but the first one was the email error. The issue is that previously, the email validation didn't exclude null or blank emails, so there was only 1 violation anyway.

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/EmailConstraintValidator.kt
```diff
@@ -1,9 +1,48 @@
...
+        private const val VALID_LOCAL_CHARS = """a-zA-Z0-9.!#$%&'*+/=?^_`{|}~\-"""
+        private val EMAIL_REGEX = Regex("^[$VALID_LOCAL_CHARS]+@([^.@][^@\\s]+)$")
+        private val HOSTNAME_PART_REGEX = Regex("^(xn|[a-z0-9]+)(-?-[a-z0-9]+)*$", RegexOption.IGNORE_CASE)
+        private val TLD_PART_REGEX = Regex("^([a-z]{2,63}|xn--([a-z0-9]+-)*[a-z0-9]+)$", RegexOption.IGNORE_CASE)
+        private const val MAX_EMAIL_LENGTH = 320
+        private const val MAX_HOSTNAME_LENGTH = 253
+        private const val MAX_HOSTNAME_PART_LENGTH = 63
+    }
+
+    override fun isValid(value: Any?): Boolean {
```
samyou-softwire, line 19: could we provide a link to the code that this re-implements? https://github.com/alphagov/notifications-utils/blob/995faf0d925f7e95ecac6ecec383b1d162b2eceb/notifications_utils/recipient_validation/email_address.py or similar

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/EmailConstraintValidator.kt
```diff
@@ -1,9 +1,48 @@
...
+        val email = (value as? CharSequence)?.toString() ?: return false
```
samyou-softwire, line 20: the python file contains a method `strip_and_remove_obscure_whitespace`. do we implement that here?

src/test/kotlin/uk/gov/communities/prsdb/webapp/validation/IsValidPrioritisedValidatorTest.kt
```diff
@@ -135,18 +135,6 @@ class IsValidPrioritisedValidatorTest {
             assertTrue(violations.isEmpty())
         }
 
-        @Test
-        fun `first violation comes from constraints earlier in declaration order, even via composed annotations`() {
```
samyou-softwire, line 139: if we're confident (cc @JasminConterioSW) that this is what the case should be could we update the test to reflect rather than removing?

TomHanmer29: I've removed it here as I think the assertion is not true, so there's no behaviour to test in the first place, it's not that the behaviour is wrong. I'm not sure what I can test here instead as it's just not a thing as far as I can tell

JasminConterioSW: I _suspect_ it's (almost) what it was _meant_ to do, but both the validator and the test weren't quite right in such a way that the errors cancelled out...

So it should probably broadly be that if there are multiple problems with an input, the first failing constraint that was declared is the one that shows up on the violation (so is the one that the error message displays)

This a fix that copilot suggested (with some prodding)
https://github.com/communitiesuk/prsdb-webapp/tree/fix/PDJB-none-is-valid-prioritised
(Just running all the tests with it to see if it breaks anything else)

src/test/kotlin/uk/gov/communities/prsdb/webapp/validation/EmailConstraintValidatorTests.kt
```diff
@@ -0,0 +1,101 @@
...
+                Named.of("single char local part", "a@example.com"),
+                Named.of("numeric local part", "123@example.com"),
+                Named.of("long TLD", "user@example.museum"),
+                Named.of("two letter TLD", "user@example.uk"),
+            )
+
+        @JvmStatic
+        fun provideInvalidEmails() =
+            arrayOf(
+                Named.of("null value", null),
```
Travis-Softwire, line 32: Should we have one with just whitespace?

Travis-Softwire (commented):
One minor comment from me. I'm happy with it once Sammy & Jasmin are!

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/EmailConstraintValidator.kt
TomHanmer29: They actually implement that separately and call it on all emails during sending etc. If we want something like that I think it should be a new ticket, and we should apply it to all email addresses we send off too :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/validation/IsValidPrioritisedValidatorTest.kt
TomHanmer29: Have cherry picked your changes!

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/EmailConstraintValidator.kt
samyou-softwire: sure; suppose this way our service would reject emails that notify might accept which is safer. alright to raise that ticket?

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/IsValidPrioritisedValidator.kt
```diff
@@ -1,168 +1,171 @@
-package uk.gov.communities.prsdb.webapp.validation
```
samyou-softwire, line 1: could we keep the same line endings?

samyou-softwire (approved):
LGTM, just looking for that follow on ticket to be raised

src/test/kotlin/uk/gov/communities/prsdb/webapp/validation/IsValidPrioritisedValidatorTest.kt
JasminConterioSW: I think there was also an update to IsValidPrioritisedValidatorTest that made it actually check what it said it was checking but I can't see that file in the diff?

TomHanmer29: Yeah you're right, I don't know where that went - I'll add it in. Sorry!

src/main/kotlin/uk/gov/communities/prsdb/webapp/validation/EmailConstraintValidator.kt
TomHanmer29: [PDJB-933](https://mhclgdigital.atlassian.net/browse/PDJB-933)

PR #1354, opened 2026-05-19
PDJB-906: Fix no registered property details back links

samyou-softwire:
## Ticket number

[PDJB-906](https://mhclgdigital.atlassian.net/browse/PDJB-906)

## Goal of change

add back url overrides to the links on the registered properties page when there are no registered properties

## Description of main change(s)

adds back link service to the controller
allows for back links to include a fragment

no registered properties view

## Anything you'd like to highlight to the reviewer?

see comments

## Checklist

Delete any that are not applicable, and add explanation below for any that are applicable but haven't been done

- [x] Screenshots of any UI changes have been added
- [x] A new journey integration test has been added for any new journeys
- [x] Test suite has been run in full locally and is passing
- [x] Branch has been rebased onto main and run locally, with everything working as expected (both for your new feature
  and any related functionality)
- [x] TODO comments referencing this JIRA ticket have been searched for and removed - if a future PR will address them,
  mention that here
- [x] QA instructions have been added to the ticket (particularly if this is the last PR required to complete the ticket)

PR #1355, opened 2026-05-19
PDJB-886: Preserve registered properties tab on back navigation

Bill-Haigh:
## Ticket number

PDJB-886

## Goal of change

When a landlord (or local council user) clicks a property from the `Registered properties` tab of the landlord details page and presses the back link on the property details page, return them to the `Registered properties` tab instead of the default `Personal details` tab. Likewise, when clicking the landlord link from the `Landlord details` tab on the property details page, return them to the `Landlord details` tab rather than the default `Property details` tab.

- The `BackUrlStorageService.storeCurrentUrlReturningKey()` `fragment` parameter (added in #1354) is now also used by `PropertyOwnershipService.getRegisteredPropertiesForLandlord{User}` (passing `REGISTERED_PROPERTIES_FRAGMENT`) and by `PropertyDetailsController` for the landlord-details link on the property details page (passing the new `LANDLORD_DETAILS_FRAGMENT`).
- A `LANDLORD_DETAILS_FRAGMENT = "landlord-details"` constant has been added alongside the existing tab fragment constants.
- Behaviour is preserved on back navigation because the stored back URL now carries the active tab fragment, so on return govuk-tabs reopens the correct tab.

Root cause was that URL fragments are not transmitted to the server in HTTP requests, so the stored back URL omitted the active-tab fragment. Fixing at the server-side back-URL storage layer keeps the existing back-link mechanism authoritative and avoids any client-side changes.

This PR was originally scoped to the registered-properties tab only; it has been expanded during review to also cover the property details page's landlord-details tab, which exhibited the same bug. The compliance-information tab is not affected because no outbound link from that tab uses `storeCurrentUrlReturningKey`.

## Checklist


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordDetailTests.kt
```diff
@@ -50,7 +50,8 @@ class LandlordDetailTests : IntegrationTestWithImmutableData("data-local.sql") {
                 )
 
             propertyDetailsView.backLink.clickAndWait()
-            assertPageIs(page, LandlordDetailsPage::class)
+            val detailsPageAfterBack = assertPageIs(page, LandlordDetailsPage::class)
+            assertEquals("registered-properties", detailsPageAfterBack.tabs.activeTabPanelId)
```
samyou-softwire: I think we should probably be checking against the constant rather than the magic string, as the test in `PropertyDetailsTests` does below

I notice I didn't do this in #1354. which I think is wrong too! if you change here please also update the other tests in `LandlordDetailsTests`

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
```diff
@@ -46,7 +47,7 @@ class PropertyDetailsController(
 
         val landlordDetailsUrl =
             LandlordDetailsController.LANDLORD_DETAILS_FOR_LANDLORD_ROUTE
-                .overrideBackLinkForUrl(backLinkStorageService.storeCurrentUrlReturningKey())
+                .overrideBackLinkForUrl(backLinkStorageService.storeCurrentUrlReturningKey(LANDLORD_DETAILS_FRAGMENT))
```
samyou-softwire, line 50: good catch, this is the back button from this link which needs to send back to having this tab open?

property landlords view

could this be documented on the ticket? (say, in QA notes)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
samyou-softwire: I worry that this could become a footgun - if you were to use the registered properties method in some other context you'd need to make sure you sending a different fragment otherwise you might implicitly send a junk fragment to the user

could the fragment be an optional parameter, which is passed by the caller LandlordDetailsControlller?

samyou-softwire (approved):
LGTM

PR #1358, opened 2026-05-19
PDJB-792: Add feature flag for compliance actions page

samyou-softwire:
## Ticket number

[PDJB-792](https://mhclgdigital.atlassian.net/browse/PDJB-792)

## Goal of change

adds feature flag and skeleton new view for the compliance actions page

- adds the feature flag that will toggle if we use the new redesign
- adds a release as this is the first feature flag in the release
- adds a copy of the compliance actions page that we can update independent of the old compliance actions page
- adds a test to verify we get the correct page back

making this a sep PR so we can get this release registered in the code early so other people can use this

may also split the compliance actions page redesign into smaller tickets that can be worked on in parallel

the actual redesign is TODO. both templates should be identical

## Checklist

- [ ] Screenshots of any UI changes have been added (no UI change)
  mention that here (there is one, but this'll be addressed when we progress with any redesign)
- [ ] QA instructions have been added to the ticket (particularly if this is the last PR required to complete the ticket) (no QA thus far)


src/main/resources/application-local.yml
```diff
@@ -103,7 +103,15 @@ features:
...
-  releases: []
+    - name: "use-compliance-actions-page-redesign"
+      enabled: false
+      expiry-date: "2026-12-31"
+      release: "private-beta-release-2"
+  releases:
+    - name: "private-beta-release-2"
+      enabled: true
+      strategy-config:
+        release-date: "2025-12-31"
```
samyou-softwire: these dates being TODO

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/FeatureFlagNames.kt
```diff
@@ -14,9 +14,12 @@ const val JOINT_LANDLORDS = "joint-landlords"
 
 const val SUBJECT_IDENTIFIER_PAGE = "subject-identifier-page"
 
+const val USE_COMPLIANCE_ACTIONS_PAGE_REDESIGN = "use-compliance-actions-page-redesign"
```
TomHanmer29: nit: does this need the "use" at the start?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordController.kt
```diff
@@ -98,7 +101,11 @@ class LandlordController(
         )
         model.addAttribute("backUrl", LANDLORD_DASHBOARD_URL)
 
-        return "complianceActions"
+        return if (featureFlagManager.checkFeature(USE_COMPLIANCE_ACTIONS_PAGE_REDESIGN)) {
```
TomHanmer29: There may be a "nicer" way we can do this, see [former examples](https://github.com/communitiesuk/prsdb-webapp/pull/1153/changes#diff-8ec374423a5d2d273b398c5a46e58ef732ee231441d61d5e4a6cd5ff376e07d5)
particularly, 
`@GetMapping(FEATURED_FLAGGED_TEMPLATE_TEST_URL_SEGMENT)
    fun featureFlaggedTemplateTest(): String = exampleFeatureFlaggedService.getTemplateName()`

src/main/resources/application-local.yml
TomHanmer29: I think this is fine, assuming we want integration off by default and local on by default? Maybe if int is off local should be off too so other people don't see a broken looking page?

TomHanmer29 (commented):
The thing about the if is the only real comment, the rest is all very minor :) Looks good!

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordController.kt
samyou-softwire: I think I'll leave this not as an extra service for now. I may need it for [PDJB-928](https://mhclgdigital.atlassian.net/browse/PDJB-928) so if it seems appropriate I'll rework then

src/main/resources/application-local.yml
samyou-softwire: good point, I'll turn this off locally. integration and all other environments should be off too

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/FeatureFlagNames.kt
samyou-softwire: I suppose not 👍, have removed

samyou-softwire:
I've also renamed the 'old' and 'new' templates to make the new one specific to this redesign

TomHanmer29 (approved):
LGTM!

PR #1361, opened 2026-05-20
PDJB-912: Add back link to privacy notice pages

Bill-Haigh:
## Ticket number

PDJB-912

## Goal of change

Add a GOV.UK back link to the standalone privacy notice pages so users can easily return to where they came from, without showing a back link when the page is opened in a new tab.

- Adds a GOV.UK back link to the standalone landlord, local council and unauthenticated privacy notice pages
- The back link is only rendered when the user reached the page via a same-tab link (footer or dashboard link); it is hidden when the page is opened in a new tab (e.g. from the registration journey) or accessed directly
- Reuses the existing `BackLinkInterceptor` / `BackUrlStorageService` mechanism already used by the cookies page: the link URL carries a `withBackUrl=<key>` query param, the interceptor populates the `backUrl` model attribute, and the existing `backLink` fragment renders the link only when that attribute is set

The in-journey privacy notice form pages (registration journeys) are intentionally out of scope — those journeys link to the standalone page with `target="_blank"` (opens in new tab), and the requirement is explicitly "back link only when not opening in a new tab".

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/controllerAdvice/GlobalModelAttributes.kt
```diff
@@ -55,7 +55,10 @@ class GlobalModelAttributes(
 
         // Footer attributes
         model.addAttribute("prsdbEmail", PRSD_EMAIL)
-        model.addAttribute("privacyUrl", "/$PRIVACY_NOTICE_PATH_SEGMENT")
+        model.addAttribute(
+            "privacyUrl",
+            "/$PRIVACY_NOTICE_PATH_SEGMENT".overrideBackLinkForUrl(backUrlStorageService.storeCurrentUrlReturningKey()),
```
samyou-softwire, line 60: it's not part of this ticket, but it feels like this code suggests that the global privacy notice link used in the footer will always send to /privacy-notice (the landlord privacy notice) even if clicked from other views (local council pages)

alas I don't seem to have LC access on integration right now so I can't verify this but I suspect this may be a bug. are you able to check and if so raise a bug ticket?

samyou-softwire (approved):
LGTM, one comment on some out of scope behaviour & the ticket is looking for QA notes.

I notice the run tests pipeline never reported its status. are you able to rerun that pipeline?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/controllerAdvice/GlobalModelAttributes.kt
Bill-Haigh: it takes you to the local council privacy notice so no issue there (there is a url-rewriter that inserts landlord or local-council depending

PR #1362, opened 2026-05-20
PDJB-925: Fix EPC unchangeable in specific case

TomHanmer29:
## Ticket number

PDJB-925

## Goal of change

Fixes the EPC being unremoveable in a very specific case

If the user has an automatched EPC, rejects it, adds a manual EPC, goes back and tries to say they have no EPC, you'd still get the manually entered one. This fixes that by altering logic to remove isStepReachable and use outcomes instead.

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/EpcState.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/EpcState.kt
@@ -24,6 +24,8 @@ import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.Prope
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.ProvideEpcLaterStep
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.StartEpcStep
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.tasks.EpcDetailsTask
+import uk.gov.communities.prsdb.webapp.journeys.shared.Complete
+import uk.gov.communities.prsdb.webapp.journeys.shared.YesOrNo
 import uk.gov.communities.prsdb.webapp.models.dataModels.EpcDataModel
 
 interface EpcState : JourneyState {
@@ -38,10 +40,11 @@ interface EpcState : JourneyState {
     var acceptedEpc: EpcDataModel?
     val allowProvideCertificateLaterRoute: Boolean
 
-    val acceptedEpcIfReachable: EpcDataModel?
+    val acceptedEpcIfStillAccepted: EpcDataModel?
         get() =
-            if (checkUprnMatchedEpcStep.isStepReachable || confirmEpcDetailsRetrievedByCertificateNumberStep.isStepReachable ||
-                checkSupersededEpcStep.isStepReachable
+            if (checkUprnMatchedEpcStep.outcome == YesOrNo.YES ||
+                confirmEpcDetailsRetrievedByCertificateNumberStep.outcome == YesOrNo.YES ||
+                checkSupersededEpcStep.outcome == Complete.COMPLETE
             ) {
                 acceptedEpc
             } else {
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/EpcState.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/EpcState.kt
@@ -47,6 +47,7 @@ interface EpcState : JourneyState {
                 checkSupersededEpcStep.outcome == Complete.COMPLETE
             ) {
                 acceptedEpc
+                    ?: throw PrsdbWebException("acceptedEpc is null despite a confirm step having a positive outcome")
             } else {
                 null
             }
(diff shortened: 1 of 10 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/EpcStateTests.kt
```diff
@@ -28,54 +28,71 @@ import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.Prope
...
+        val state = buildTestEpcState(acceptedEpc = epc)
 
-        assertNull(state.acceptedEpcIfReachable)
+        assertNull(state.acceptedEpcIfStillAccepted)
     }
 
     @Test
-    fun `acceptedEpcIfReachable returns null when acceptedEpc is null and a confirm step is reachable`() {
-        val state = buildTestEpcState(acceptedEpc = null, uprnStepReachable = true, certStepReachable = false)
+    fun `acceptedEpcIfStillAccepted returns null when acceptedEpc is null and a confirm step has positive outcome`() {
```
JasminConterioSW: Is this behaviour correct? If a confirm step has a positive outcome but acceptedEpc is null, should that throw an error instead?

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/EpcStateTests.kt
```diff
@@ -28,54 +28,71 @@ import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.Prope
...
+        val state = buildTestEpcState(acceptedEpc = null, uprnStepOutcome = YesOrNo.YES)
 
-        assertNull(state.acceptedEpcIfReachable)
+        assertNull(state.acceptedEpcIfStillAccepted)
+    }
+
+    @Test
+    fun `acceptedEpcIfStillAccepted returns null when checkUprnMatchedEpcStep has outcome NO`() {
```
JasminConterioSW: If checkUprnMatchedEpcStep is NO the should still be able to get to confirmEpcDetailsRetrievedByCertificateNumberStep, which can be YES, so the name of this test sounds wrong...

I think the actual behaviour is covered by `acceptedEpcIfStillAccepted returns null when acceptedEpc is null and a confirm step has positive outcome` though?

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/EpcStateTests.kt
```diff
@@ -28,54 +28,71 @@ import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.Prope
...
+        val epc = mock<EpcDataModel>()
+        val state = buildTestEpcState(acceptedEpc = epc, uprnStepOutcome = YesOrNo.NO)
+
+        assertNull(state.acceptedEpcIfStillAccepted)
+    }
+
+    @Test
+    fun `acceptedEpcIfStillAccepted returns null when confirmEpcDetailsRetrievedByCertificateNumberStep has outcome NO`() {
+        val epc = mock<EpcDataModel>()
```
JasminConterioSW: Again I _think_ this title is incorrect, because they could get back to the checkUprnMatchedEpcStep and accept the EPC there. But we shouldn't need a separate test

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/EpcState.kt
```diff
@@ -38,10 +40,11 @@ interface EpcState : JourneyState {
...
-    val acceptedEpcIfReachable: EpcDataModel?
+    val acceptedEpcIfStillAccepted: EpcDataModel?
         get() =
-            if (checkUprnMatchedEpcStep.isStepReachable || confirmEpcDetailsRetrievedByCertificateNumberStep.isStepReachable ||
-                checkSupersededEpcStep.isStepReachable
+            if (checkUprnMatchedEpcStep.outcome == YesOrNo.YES ||
+                confirmEpcDetailsRetrievedByCertificateNumberStep.outcome == YesOrNo.YES ||
+                checkSupersededEpcStep.outcome == Complete.COMPLETE
             ) {
                 acceptedEpc
```
JasminConterioSW, line 49: There's a comment on the test, but it might be worth throwing an error here if the users has confirmed some EPC details but the acceptedEpc is null as I don't think that should happen

JasminConterioSW (commented):
This looks right to me, just a question on the behaviour when acceptedEpc is null but one of the confirm pages has a positive outcome, and a couple of tests which can come out

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/EpcStateTests.kt
TomHanmer29: Good point - I'll update this behaviour so that it throws if we have a positive outcome but no acceptedEpc :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/EpcStateTests.kt
TomHanmer29: I'll delete these 2 - I agree it's covered by the other case :)

JasminConterioSW (approved):
LGTM :)

PR #1363, opened 2026-05-20
PDJB-922: Fix gas showing as non compliant when no gas supply is present

TomHanmer29:
## Ticket number

PDJB-922

## Goal of change

Gas was showing as non compliant when no gas supply is present. This fixes that

Adds a new NOT_REQUIRED compliance state.
Also handles errors better when the site attempts to display an ADDED state.

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/converters/MessageKeyConverter.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/converters/MessageKeyConverter.kt
@@ -158,8 +158,13 @@ class MessageKeyConverter {
 
         private fun convertComplianceCertStatus(complianceCertStatus: ComplianceCertStatus): String =
             when (complianceCertStatus) {
+                ComplianceCertStatus.NOT_REQUIRED -> throw IllegalStateException(
+                    "NOT_REQUIRED is a valid compliance state and should not be converted to a message key",
+                )
+                ComplianceCertStatus.ADDED -> throw IllegalStateException(
+                    "ADDED is a valid compliance state and should not be converted to a message key",
+                )
                 ComplianceCertStatus.NOT_STARTED -> "complianceActions.status.notStarted"
-                ComplianceCertStatus.ADDED -> "complianceActions.status.added"
                 ComplianceCertStatus.NOT_ADDED -> "complianceActions.status.notAdded"
                 ComplianceCertStatus.EXPIRED -> "complianceActions.status.expired"
             }
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/converters/MessageKeyConverter.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/converters/MessageKeyConverter.kt
@@ -159,10 +159,13 @@ class MessageKeyConverter {
         private fun convertComplianceCertStatus(complianceCertStatus: ComplianceCertStatus): String =
             when (complianceCertStatus) {
                 ComplianceCertStatus.NOT_STARTED -> "complianceActions.status.notStarted"
+
                 ComplianceCertStatus.NOT_ADDED -> "complianceActions.status.notAdded"
+
                 ComplianceCertStatus.EXPIRED -> "complianceActions.status.expired"
+
                 else -> throw IllegalStateException(
-                    "${complianceCertStatus.name} is a valid compliance state and should not be converted to a message key",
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/converters/MessageKeyConverter.kt
```diff
@@ -158,8 +158,13 @@ class MessageKeyConverter {
 
         private fun convertComplianceCertStatus(complianceCertStatus: ComplianceCertStatus): String =
             when (complianceCertStatus) {
+                ComplianceCertStatus.NOT_REQUIRED -> throw IllegalStateException(
+                    "NOT_REQUIRED is a valid compliance state and should not be converted to a message key",
+                )
+                ComplianceCertStatus.ADDED -> throw IllegalStateException(
+                    "ADDED is a valid compliance state and should not be converted to a message key",
+                )
```
samyou-softwire: would it be more useful to have an `_` escape clause for this `when` to reject any not know statuses rather than this predefined list?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ComplianceStatusDataModel.kt
```diff
@@ -15,7 +15,8 @@ data class ComplianceStatusDataModel(
     val isOccupied: Boolean,
 ) {
     fun shouldShowCert(status: ComplianceCertStatus): Boolean =
-        status == ComplianceCertStatus.EXPIRED || (isOccupied && status != ComplianceCertStatus.ADDED)
+        status == ComplianceCertStatus.EXPIRED ||
```
samyou-softwire, line 18: I've seen in some language it's more natural to do something like

`isOccupied & ![added, notrequired].contains(status)`

to make it a little clearer that we're checking status is not one of these two states

src/test/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ComplianceStatusDataModelTests.kt
```diff
@@ -100,6 +100,23 @@ class ComplianceStatusDataModelTests {
         assertFalse(dataModel.shouldShowOnComplianceActionsPage)
     }
 
+    @Test
+    fun `fromPropertyCompliance returns ADDED gas safety status when property has no gas supply`() {
```
samyou-softwire: I think this test name needs changing (ADDED?)

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/converters/MessageKeyConverter.kt
TomHanmer29: Yes! That's a great idea, I was looking for how to do this in kotlin originally but forgot. I'll add it now :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ComplianceStatusDataModel.kt
TomHanmer29: Sure, happy to swap to this :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/converters/MessageKeyConverter.kt
```diff
@@ -159,9 +159,11 @@ class MessageKeyConverter {
         private fun convertComplianceCertStatus(complianceCertStatus: ComplianceCertStatus): String =
             when (complianceCertStatus) {
                 ComplianceCertStatus.NOT_STARTED -> "complianceActions.status.notStarted"
-                ComplianceCertStatus.ADDED -> "complianceActions.status.added"
                 ComplianceCertStatus.NOT_ADDED -> "complianceActions.status.notAdded"
                 ComplianceCertStatus.EXPIRED -> "complianceActions.status.expired"
+                else -> throw IllegalStateException(
+                    "${complianceCertStatus.name} is a valid compliance state and should not be converted to a message key",
```
samyou-softwire: I do wonder a little on this error message, if I were reading this and was told there was an error due to it being a 'valid compliance state' it may be a little confusing what the intent is here (error are for invalid data). maybe it could be more general 'xyz could not be converted'? though happy to keep this one if there's no clear replacement

samyou-softwire (approved):
LGTM, one final thought on the error message but happy to go ahead either way

PR #1364, opened 2026-05-20
Fix/pdjb 919 link change epc cya to start of task

JasminConterioSW:
## Ticket number

PDJB-919 (and PDJB-918)

## Goal of change

Point the EPC change link to the start page of the EPC task rather than specifying the hasEpc page

* Changes the link to go to the start page
* This is now reachable for properties where an EPC is found by uprn as well as those where it is not, so the change link appears for those properties too (this fixes PDJB-918)
* The change EPC link will now do the lookup by uprn before deciding which page the user sees instead of always taking them to "Do you have an EPC"
    * This could be desirable e.g. if the user has also changed the property address from one where the epc was not found by uprn to an address where the epc is found by uprn

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/EpcRegistrationCyaSummaryRowsFactory.kt
```diff
@@ -108,7 +108,7 @@ class EpcRegistrationCyaSummaryRowsFactory(
         if (!isEpcCardShown()) return null
         val epc = state.acceptedEpc ?: throw IllegalStateException("An EPC should be present when showing EPC card")
         val epcUrl = epcCertificateUrlProvider.getEpcCertificateUrl(epc.certificateNumber)
-        val changeUrl = destinationProvider(state.hasEpcStep).toUrlStringOrNull()
+        val changeUrl = destinationProvider(state.startEpcStep).toUrlStringOrNull()
```
samyou-softwire, line 111: to confirm, this section confused me as I was expecting us to remove an if condition on this change link

is the point that the hasEpcStep was unreachable in cases where it was auto matched and so the check answers page is smart enough to hide the change link in those cases? but now we link to a step that always appears so it'll always show? it's a little surprising it worked this way but suppose now it's less surprising

samyou-softwire (approved):
LGTM, just wanted to check how exactly we solved [PDJB-918](https://mhclgdigital.atlassian.net/browse/PDJB-918)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/EpcRegistrationCyaSummaryRowsFactory.kt
Travis-Softwire: I assume this is because previously this evalutated to null in those cases, and we hide the change link if it's destination is null

Travis-Softwire (approved):
Just a respones to Sammy's comment, LGTM!

PR #1370, opened 2026-05-26
PDJB-584: Normalize line endings to LF and add CI check

Travis-Softwire:
## Ticket number

PDJB-584

## Goal of change

Normalize all repository line endings to LF and prevent new CRLF endings from being committed.

- Adds `.gitattributes` with `* text=auto eol=lf` as the default, plus explicit text/binary rules for the file types present in the repo. Git will now auto-normalize text files to LF on commit on every platform.
- Renormalizes the 139 existing files that had CRLF (134) or mixed (5) line endings — purely whitespace changes (`git diff --ignore-all-space --stat` shows non-whitespace changes only in the two intentional files).
- Adds a `Check line endings` step to `build-and-test.yml` that fails the CI build and lists offending paths if any file in the index has CRLF or mixed endings. Runs before the JDK setup so it fails fast.

- The `.bat` rule maps to LF (not the Windows convention of CRLF) because `.editorconfig` already mandates LF for all files and the only existing `.bat` file (`gradlew.bat`) is already LF in the index and works correctly.
- The CI check uses `awk -F'\t'` rather than the default space delimiter because, with `.gitattributes` in place, the `attr/` column in `git ls-files --eol` output contains spaces (e.g. `attr/text=auto eol=lf`).
- Verified locally: `./gradlew ktlintCheck` passes, `npm test` passes, and the check script correctly reports 0 bad files after renormalization (and 139 before, as a sanity check).

## Checklist


JasminConterioSW (approved):
LGTM, thanks for the tip about the "ignore whitespace" setting...

PR #1371, opened 2026-05-26
Feat/pdjb 260 jl invite accept reject skeleton journey

JasminConterioSW:
## Ticket number

PDJB-260

## Goal of change

Set up the controller and journey for a JL accepting / rejecting their invitation

* Switch the joint landlords flag on in integration
* Setup new controller
    * Endpoints use a similar pattern to RegisterLocalCouncilUserController but validating the token will be done in the journey
    * Until the user accepts the invitation, they do not need to be logged in with One Login (or to be a landlord). They will need to log in / register once they accept the invitation.
* Set up new journey factory with some placeholder places
* update the url added to landlord invitation emails in AbsoluteUrlProvider

There will be another PR actually implementing the main page for this ticket

## Checklist

- TODO comments referencing this JIRA ticket have been searched for and removed  - this will be done in a later PR


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/AcceptOrRejectJointLandlordInvitationController.kt
```diff
@@ -0,0 +1,104 @@
...
+class AcceptOrRejectJointLandlordInvitationController(
+    private val journeyFactory: AcceptOrRejectJointLandlordInvitationJourneyFactory,
+    private val invitationService: JointLandlordInvitationService,
+) {
+    @GetMapping
+    @AvailableWhenFeatureEnabled(JOINT_LANDLORDS)
+    fun startJourney(
+        @RequestParam(value = TOKEN, required = true) token: String,
+    ): ModelAndView {
+        val journeyId = journeyFactory.initializeJourneyState(token)
```
samyou-softwire: why do we pass in the token here? it doesn't seem to be used in `initializeJourneyState`

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/AcceptOrRejectJointLandlordInvitationController.kt
```diff
@@ -0,0 +1,104 @@
...
+    @PreAuthorize("hasRole('LANDLORD')")
+    @GetMapping("/property-joined-$CONFIRMATION_PATH_SEGMENT")
+    @AvailableWhenFeatureEnabled(JOINT_LANDLORDS)
+    fun getConfirmation(model: Model): ModelAndView {
+        model.addAttribute("title", "TODO: PDJB-265 - Property joined confirmation")
+
+        return ModelAndView("placeholder")
+    }
+
+    private fun initializeAndRedirect(
```
samyou-softwire, line 88: can we use this in the `startJourney` endpoint?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/AcceptOrRejectJointLandlordInvitationController.kt
```diff
@@ -0,0 +1,104 @@
...
+            val journeyMap = journeyFactory.createJourneySteps(token)
+            journeyMap[stepRouteSegment]?.postStepModelAndView(formData)
+                ?: throw ResponseStatusException(HttpStatus.NOT_FOUND, "Step not found")
+        } catch (_: NoSuchJourneyException) {
+            initializeAndRedirect(token, stepRouteSegment)
+        }
+    }
+
+    @PreAuthorize("hasRole('LANDLORD')")
+    @GetMapping("/property-joined-$CONFIRMATION_PATH_SEGMENT")
```
samyou-softwire: could this be a const?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        if (state.invitationToken != token) {
+            throw PrsdbWebException(
+                "Journey state token ${state.invitationToken} does not match provided token $token",
+            )
+        }
+
+        return journey(state) {
+            unreachableStepStep { journey.validateTokenStep }
```
samyou-softwire, line 39: to check; what is this doing? it looks like there's a similar step below this one

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
```diff
@@ -0,0 +1,91 @@
...
+                parents { journey.validateTokenStep.hasOutcome(TokenValidationResult.VALID) }
+                // TODO PDJB-260
+                //   Accept and not a landlord -> LL registration journey
+                //   Reject -> rejection confirmation screen PDJB-261
+                nextUrl { JOINT_LANDLORD_INVITATION_ACCEPTED_CONFIRMATION_ROUTE }
+            }
+            step(journey.inviteUnavailableStep) {
+                routeSegment(InviteUnavailableStep.ROUTE_SEGMENT)
+                parents { journey.validateTokenStep.hasOutcome(TokenValidationResult.INVALID) }
+                nextUrl { JOINT_LANDLORD_INVITATION_ACCEPTED_CONFIRMATION_ROUTE }
```
samyou-softwire, line 62: is this the case? seems an odd next step if the invite is unavailable

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationService.kt
```diff
@@ -43,4 +46,10 @@ class JointLandlordInvitationService(
         invitationRepository.save(JointLandlordInvitation(token, email, propertyOwnership))
         return token.toString()
     }
+
+    fun storeTokenInSession(token: String) = session.setAttribute(JOINT_LANDLORD_INVITATION_TOKEN, token)
```
samyou-softwire, line 50: this does confuse me a little; why do we have both this token stored in the session state and also the journey state? is it enough to verify the token once when starting the journey?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/AcceptOrRejectJointLandlordInvitationController.kt
JasminConterioSW: Actually `initializeJourneyState `probably should be using it as the seed...

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
Travis-Softwire: This is telling the journey framework what step to go to if the user tries to access a step that's unreachable (hence unreachable step step)

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/security/LandlordSecurityConfig.kt
```diff
@@ -62,6 +66,20 @@ class LandlordSecurityConfig(
                     .permitAll()
                     .requestMatchers("${BetaFeedbackController.LANDLORD_FEEDBACK_URL}/**")
                     .permitAll()
+                    .requestMatchers(AcceptOrRejectJointLandlordInvitationController.ACCEPT_OR_REJECT_JOINT_LANDLORD_INVITATION_ROUTE)
```
Travis-Softwire, line 66: Could this just be `..ACCEPT_OR_REJECT_JOINT_LANDLORD_INVITATION_ROUTE/**`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
Travis-Softwire: Assume this is just missing a TODO to correct the routing once the relevant steps have been implemented?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/steps/ValidateTokenStepConfig.kt
```diff
@@ -0,0 +1,39 @@
...
+import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.JourneyFrameworkComponent
+import uk.gov.communities.prsdb.webapp.database.repository.JointLandlordInvitationRepository
+import uk.gov.communities.prsdb.webapp.journeys.AbstractRequestableStepConfig
+import uk.gov.communities.prsdb.webapp.journeys.JourneyStep
+import uk.gov.communities.prsdb.webapp.journeys.StepLifecycleOrchestrator.RedirectingStepLifecycleOrchestrator
+import uk.gov.communities.prsdb.webapp.journeys.acceptOrRejectJointLandlordInvitation.AcceptOrRejectJointLandlordInvitationJourneyState
+import uk.gov.communities.prsdb.webapp.models.requestModels.formModels.NoInputFormModel
+
+@JourneyFrameworkComponent
+class ValidateTokenStepConfig(
```
Travis-Softwire, line 12: If all the interesting stuff is happening in `mode` - does this need to be visitable? Or could it be an internal step which is added as an AND parent for each visitiable step?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationService.kt
Travis-Softwire: We do need to verify the token each time, or at least again at the end of the journey (the pattern is that we consume the token at the end of the journey, so this prevents the same token being reused multiple times by starting several journeys in parallel).

However I think I agree that we don't need to store the token in state - the controller can grab it from session and use it as the initialisation seed, and after that I don't think we need it until the end when we consume the token, when again we should be able to inject the relevant service and pull it directly from session there?

Travis-Softwire (approved):
A few additional thoughts, but subject to those LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationService.kt
JasminConterioSW: We need to retrieve the property address and inviting landlord name for the accept / reject page, which we'll want to do via the token.

The `joint-landlord-invitation` already includes the property ownership which we can use for the address, but I think we'll need to add the inviting landlord to that entity too (in another PR)

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/security/LandlordSecurityConfig.kt
JasminConterioSW: There are some pages that the user only gets to after they have logged in ("Confirm that you're a joint landlord..." and the confirmation screen) so we're going to end up with a mixture either way.

I'll change it for now though, there are actually quite a few that are not behind the One Login wall...

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
JasminConterioSW: Yes, that'll be covered in another ticket - I'll add a TODO

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationService.kt
JasminConterioSW: Also we'll need it for the validateTokenStep!

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/steps/ValidateTokenStepConfig.kt
JasminConterioSW: The issue is similar to the one we had for the Epc Update journey - it's the first step in the journey so we need to be able to navigate to it.

Although we could add another start step that we can navigate to but then takes us straight to the validation step (as there is nothing in the validation step that requires the navigation)

JasminConterioSW: Ah, you're thinking we never have to actually get to the step, the mode will get evaluated when the child step's parentage is evaluated...

JasminConterioSW: I've added a comment to PDJB-266 to look into this there as that's where the validation step will be implemented so we can check what works

PR #1374, opened 2026-05-26
PDJB-928: Part 1: Update certificate status rows

samyou-softwire:
## Ticket number

## Goal of change

add the new copy for what shows on the status row for an epc 

updates behaviour in `ComplianceStatusDataModel` to provide the following behaviour, agreed on the ticket

new behaviour

happy to hear other ways the conditions could be structured to achieve this

adds structure to allow for a date to be put into the compliance row

adds tests, including integration tests for all 20 EPC tests to make sure we catch them all!

<details><summary>screenshots</summary>
example compliance page
</details>

see comments

## Checklist

  mention that here
- [ ] QA instructions have been added to the ticket (particularly if this is the last PR required to complete the ticket) (to be done on next ticket)


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
```diff
@@ -97,7 +105,10 @@ class PropertyCompliance() : ModifiableAuditableEntity() {
         }
 
     val isEpcMissing: Boolean
-        get() = (epcUrl == null && !hasEpcExemption) || (isEpcRatingLow == true && epcMeesExemptionReason == null)
+        get() =
+            (epcUrl == null && !hasEpcExemption) ||
+                (isEpcRatingLow == true && epcMeesExemptionReason == null) ||
+                (isEpcExpired == true && tenancyStartedBeforeEpcExpiry == false)
```
samyou-softwire: this technically is a breaking change, but previously expired would take precedence over no valid cert. so, on the old page this third condition will never be used as it'd already be considered expired by this point

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
```diff
@@ -86,6 +86,14 @@ class PropertyCompliance() : ModifiableAuditableEntity() {
     val isEpcNonCompliantDueToExpiry: Boolean
         get() = isEpcExpired == true && tenancyStartedBeforeEpcExpiry != true
 
+    val isEpcExpiredAndWouldBeCompliant: Boolean
```
samyou-softwire: this is the new logic, we only show 'expired' if it would've been valid; otherwise it is just 'no valid cert'

src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/converters/MessageKeyConverterTests.kt
```diff
@@ -81,12 +81,6 @@ class MessageKeyConverterTests {
         assertMessageKeyResolves(MessageKeyConverter.convert(value))
     }
 
-    @ParameterizedTest
-    @EnumSource(ComplianceCertStatus::class, names = ["NOT_REQUIRED", "ADDED"], mode = EnumSource.Mode.EXCLUDE)
-    fun `convert returns a resolvable message key for every ComplianceCertStatus`(value: ComplianceCertStatus) {
-        assertMessageKeyResolves(MessageKeyConverter.convert(value))
```
samyou-softwire, line 87: this is no longer true; the controllers need to do some extra work to get the key

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordDashboardTests.kt
```diff
@@ -150,7 +150,7 @@ class LandlordDashboardTests : IntegrationTestWithImmutableData("data-local.sql"
                 fun `the notification banner loads with correct message for multiple compliance actions`(page: Page) {
                     val dashboard = navigator.goToLandlordDashboard()
                     assertThat(dashboard.notificationBanner.content.heading)
-                        .containsText("You have 3 properties awaiting compliance information: View compliance actions")
+                        .containsText("You have 5 properties awaiting compliance information: View compliance actions")
```
samyou-softwire, line 153: the .sql file this is based on now has 2 extra compliance actions that were added to test other cases

src/test/kotlin/uk/gov/communities/prsdb/webapp/testHelpers/builders/PropertyComplianceBuilder.kt
```diff
@@ -205,6 +205,7 @@ class PropertyComplianceBuilder {
                 .withExpiredElectricalSafety()
                 .withElectricalCertType()
                 .withExpiredEpc()
+                .withTenancyStartedBeforeEpcExpiry()
```
samyou-softwire: for an occupied property, we now only show expired if the tenancy started before epc expiry

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ComplianceStatusDataModel.kt
```diff
@@ -72,13 +88,24 @@ data class ComplianceStatusDataModel(
...
                 }
 
-        private val PropertyCompliance.epcStatus: ComplianceCertStatus
-            get() =
+        private fun getEpcStatus(
+            propertyCompliance: PropertyCompliance,
+            useMay26Redesign: Boolean,
+        ): ComplianceCertStatus =
+            if (useMay26Redesign) {
+                when {
```
Travis-Softwire, line 88: Rather than deeply passing in this flag (which will be annoying to unpick) we could just add an extra property to the resulting data model so that it can include both old and new epcStatus, and then choose between them in the builder (which you already have old and new versions of)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordController.kt
```diff
@@ -51,10 +51,11 @@ class LandlordController(
         val landlord =
             landlordService.retrieveLandlordByBaseUserId(principal.name)
                 ?: throw PrsdbWebException("User ${principal.name} is not registered as a landlord")
+        val useMay2026Redesign = featureFlagManager.checkFeature(COMPLIANCE_ACTIONS_PAGE_MAY26_REDESIGN)
 
         val numberOfComplianceActions =
             propertyOwnershipService.getNumberOfIncompleteCompliancesForLandlord(principal.name) +
-                propertyComplianceService.getNumberOfNonCompliantPropertiesForLandlord(principal.name)
+                propertyComplianceService.getNumberOfNonCompliantPropertiesForLandlord(principal.name, useMay2026Redesign)
```
Travis-Softwire: I think we can avoid adding this parameter, see my other comment

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ComplianceStatusDataModel.kt
```diff
@@ -72,13 +88,24 @@ data class ComplianceStatusDataModel(
...
+                    propertyCompliance.epcProvideLater == true -> ComplianceCertStatus.PROVIDE_LATER
+                    propertyCompliance.isEpcExpiredAndWouldBeCompliant == true -> ComplianceCertStatus.EXPIRED
```
Travis-Softwire: `isEpcExpiredAndWouldBeCompliant` isn't nullable so you don't need the `== true` here

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ComplianceStatusDataModel.kt
```diff
@@ -48,9 +57,16 @@ data class ComplianceStatusDataModel(
                         ).toString(),
                 gasSafetyStatus = propertyCompliance.gasSafetyStatus,
                 eicrStatus = propertyCompliance.eicrStatus,
-                epcStatus = propertyCompliance.epcStatus,
+                epcStatus = getEpcStatus(propertyCompliance, useMay26Redesign),
                 isComplete = true,
                 isOccupied = propertyCompliance.propertyOwnership.isOccupied,
+                provideLaterDeadline =
+                    propertyCompliance.propertyOwnership.lastOccupiedDate?.plusDays(
+                        PROVIDE_LATER_DEADLINE_DAYS.toLong(),
```
Travis-Softwire, line 60: Silly question, but if they selected provide later more than 28 days ago, what do we display?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/ComplianceActionViewModelBuilder.kt
```diff
@@ -26,19 +30,19 @@ class ComplianceActionViewModelBuilderOld {
                     if (dataModel.shouldShowCert(dataModel.gasSafetyStatus)) {
                         addRow(
                             "complianceActions.summaryRow.old.gasSafety",
-                            MessageKeyConverter.convert(dataModel.gasSafetyStatus),
+                            "${MessageKeyConverter.convert(dataModel.gasSafetyStatus)}.old",
```
Travis-Softwire, line 34: While I don't love this - recognise that it's probably the least-worst option for the moment until we pick up that ticket to handle it at the message key level!

src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/converters/MessageKeyConverterTests.kt
Travis-Softwire: Will it be true once we remove the feature flag? Or is just not true at all any more? If the former, can we disable it with a TODO instead of deleting?

Travis-Softwire (changes requested):
A few thoughts from me - happy to chat about any of them if need be!

src/test/kotlin/uk/gov/communities/prsdb/webapp/helpers/converters/MessageKeyConverterTests.kt
samyou-softwire: ah yup. since we've no ticket to reference a TODO against I've added a test to verify that these keys don't resolve and added some instructions on what to do if they ever do (which should happen when somebody strips the feature flag out)

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ComplianceStatusDataModel.kt
samyou-softwire: we've had this confirmed after 28 days nothing happens, at least on the site

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ComplianceStatusDataModel.kt
samyou-softwire: ah yup have removed this

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ComplianceStatusDataModel.kt
samyou-softwire: I've reworked this to be an extra property. did require a couple extra caller methods too that sit between the controller and this dataModel but not too bad

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ComplianceStatusDataModel.kt
Travis-Softwire: So does that mean the message will tell the user they need to provide it by <date in the past>? I don't think we need to resolve it as part of this PR but can we triple check designers realise & are happy w/ that?

Travis-Softwire (approved):
One non-blocking Q, but all LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ComplianceStatusDataModel.kt
samyou-softwire: the messaging is Provide this later (before <28 days after submission>), so after 28 days pass it'll start telling you when you should have submitted by. I'm not 100% I've answered your question but I'm happy that it was confirmed nothing should change in the wording after 28 days passes

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
```diff
@@ -96,8 +96,9 @@ class PropertyCompliance() : ModifiableAuditableEntity() {
             }
         }
 
-    val isEpcMissing: Boolean
-        get() = (epcUrl == null && !hasEpcExemption) || (isEpcRatingLow == true && epcMeesExemptionReason == null)
+    val isEpcNotValid: Boolean
+        get() =
+            (epcUrl == null && !hasEpcExemption) || (isEpcRatingLow == true)
```
samyou-softwire: isEpcRatingLow already checks if theres a mees exemption

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
```diff
@@ -104,11 +96,9 @@ class PropertyCompliance() : ModifiableAuditableEntity() {
             }
         }
 
-    val isEpcMissing: Boolean
+    val isEpcNotValid: Boolean
```
TomHanmer29: I was thinking about this and I really hate this wording. I think it's confusing and it makes it really hard to understand the logic (because there's 2 or 3 definitions of "valid" we're working with, legally valid, "valid" in terms of do we show the "valid" badge on the tab and "not valid" in terms of do we show a compliance action). I think isEpcMissing is a little better, but maybe something else could be even more good - we want to convey that they attempted to provide an Epc but the one they gave is not useble but not because it's expired... which is quite hard to do in a short form way. The logic here looks good though.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ComplianceStatusDataModel.kt
```diff
@@ -82,16 +81,19 @@ data class ComplianceStatusDataModel(
                 when {
                     isEpcNonCompliantDueToExpiry -> ComplianceCertStatus.EXPIRED
                     epcProvideLater == true -> ComplianceCertStatus.PROVIDE_LATER
-                    isEpcMissing -> ComplianceCertStatus.NOT_ADDED
+                    isEpcNotValid -> ComplianceCertStatus.NOT_ADDED
```
TomHanmer29: I wonder if it makes sense for this to be called NOT_ADDED any more - similar with my other naming comment.

TomHanmer29 (commented):
I'm happy with the logic here, but I really don't like how everything is named. Feel free to disagree, but I think it makes it really confusing when it's read all together

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ComplianceStatusDataModel.kt
samyou-softwire: have renamed to suit

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
samyou-softwire: as discussed have gone with 'HAS_FAULT'

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/ComplianceActionViewModelBuilder.kt
```diff
@@ -84,26 +89,82 @@ class ComplianceActionViewModelBuilderMay26Redesign {
...
+            }
+        }
+
+        private fun getCertStatusValueParam(
+            status: ComplianceCertStatus,
+            provideLaterDeadline: LocalDate?,
+            expiryDate: LocalDate?,
+        ): String? =
+            when (status) {
+                ComplianceCertStatus.PROVIDE_LATER -> provideLaterDeadline?.format(DATE_FORMATTER)
```
Travis-Softwire: Minor - but in Tom's new GasSafetyViewModelService & Electrical equivalent we have a fallback in case provideLaterDeadline is null. I think that's just defense in depth but good to be consistent

src/test/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyComplianceTests.kt
```diff
@@ -109,28 +109,28 @@ class PropertyComplianceTests {
     ) {
         assertEquals(expectedIsXMissing, propertyCompliance.isGasSafetyCertMissing)
         assertEquals(expectedIsXMissing, propertyCompliance.isElectricalSafetyMissing)
-        assertEquals(expectedIsXMissing, propertyCompliance.isEpcMissing)
+        assertEquals(expectedIsXMissing, propertyCompliance.epcHasFaults)
```
Travis-Softwire, line 112: Does the name of this test + the arg need to be updated as well?

src/test/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyComplianceTests.kt
```diff
@@ -109,28 +109,28 @@ class PropertyComplianceTests {
...
     }
 
     @Test
-    fun `isEpcMissing returns true when EPC rating is low and there's no MEES exemption`() {
+    fun `isEpcNotValid returns true when EPC rating is low and there's no MEES exemption`() {
```
Travis-Softwire: The assertion is on epcHasFaults rather than isEpcNotValid?

src/test/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyComplianceTests.kt
```diff
@@ -109,28 +109,28 @@ class PropertyComplianceTests {
...
                 .withEpc()
                 .withLowEpcRating()
                 .build()
-        assertTrue(propertyCompliance.isEpcMissing)
+        assertTrue(propertyCompliance.epcHasFaults)
     }
 
     @Test
-    fun `isEpcMissing returns false when EPC rating is low and there is a MEES exemption`() {
+    fun `isEpcNotValid returns false when EPC rating is low and there is a MEES exemption`() {
```
Travis-Softwire: Same here

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
```diff
@@ -110,7 +110,7 @@ class RegisterPropertyController(
         val actionRequiredForCompliance =
             if (propertyOwnership.isOccupied) {
                 val compliance = propertyComplianceService.getComplianceForPropertyOrNull(propertyOwnership.id)
-                compliance == null || compliance.isGasSafetyCertMissing || compliance.isElectricalSafetyMissing || compliance.isEpcMissing
+                compliance == null || compliance.isGasSafetyCertMissing || compliance.isElectricalSafetyMissing || compliance.epcHasFaults
```
Travis-Softwire, line 113: For low rated EPCs with no MEES exemption that have expired before the tenancy started, doesn't this mean we'd no longer show the 28 day warning on the confirmation page? Is that what we wanted?

Travis-Softwire (changes requested):
One substantive comment (sorry!) and some minor ones. I think the substantive one does need working through - shout if I'm wrong about it and I can come back and approve

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
samyou-softwire: I don't think is a situation we can disambiguate - if the EPC expired before tenancy started we do not ask the MEES exemption question. so here it falls that the EPC is expired and was expired when the property started so it will count as needing compliance action

happy to chat further about the logic here. I did double check and submitting an EPC with those properties showed the 28 days action copy
confirmation with 28 days later action

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/ComplianceActionViewModelBuilder.kt
samyou-softwire: I spoke with tom and we've decided to change it to explicitly throw here. provide later is only shown for occupied properties and so the provide later deadline (which is based on last occupied date) really should be set

src/test/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyComplianceTests.kt
samyou-softwire: they do - I've updated this test name to account for that the epc has slightly different naming as well as the rest of this file

src/test/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyComplianceTests.kt
samyou-softwire: see above

src/test/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyComplianceTests.kt
samyou-softwire: see above

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterPropertyController.kt
Travis-Softwire: Ah fair - if we don't ask the MEES exemption question in that scenario then this is fine :)

Travis-Softwire (approved):

PR #1377, opened 2026-05-27
PDJB-792: Enable private beta release 2 on integration

samyou-softwire:
quick PR to allow for this feature to be QA tested


Travis-Softwire (approved):

samyou-softwire:
Jasmin pointed out that there's not much point having the strategy config if we've not decided yet, so I've left it as enabled: false/true for all except integration

JasminConterioSW (approved):

PR #1378, opened 2026-05-27
PDJB-794: Add compliance tab redesign feature flag, banner and skeleton

TomHanmer29:
## Ticket number

PDJB-794

## Goal of change

image
image
image
image
image
image
image

Add the feature flag infrastructure and notification banner/skeleton changes for the compliance tab redesign.

- Adds `PROPERTY_COMPLIANCE_TAB_MAY26_REDESIGN` feature flag to the `private-beta-release-2` release, initially disabled
- Adds flag constant to `FeatureFlagNames.kt` and all YAML profile configs
- Adds notification banner view model service interface with `@PrsdbFlip` annotation
- Adds legacy and redesign implementations for notification banner logic
- Adds all-valid inset text logic behind the flag
- Updates the compliance tab skeleton/structure

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/NotificationBannerViewModelServiceRedesign.kt
```diff
@@ -0,0 +1,61 @@
...
+import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.PrsdbWebService
+import uk.gov.communities.prsdb.webapp.constants.COMPLIANCE_INFO_FRAGMENT
+import uk.gov.communities.prsdb.webapp.database.entity.PropertyCompliance
+
+@PrsdbWebService("notificationBannerViewModelServiceRedesign")
+class NotificationBannerViewModelFactory : NotificationBannerViewModelService {
+    override fun getNotificationMessageKeys(
+        propertyCompliance: PropertyCompliance,
+    ): List<PropertyComplianceViewModel.PropertyComplianceNotificationMessage> {
+        val isGasExpired = propertyCompliance.isGasSafetyCertExpired == true
```
TomHanmer29: This could be commonised with [Sammy's PR](https://github.com/communitiesuk/prsdb-webapp/pull/1374) in some ways, but I think it's fine for now. I can always do a refactor in the final PR of the chain

samyou-softwire: agree; I think the exact logic of what we consider 'expired' and what we consider 'missing' could do with a closer inspection, though this file shouldn't change much as `PropertyCompliance.kt` should handle that

Travis-Softwire: Yes, shouldn't  the compliance status data model should encapsulate this?

Travis-Softwire: And if that doesn't quite fit the bill then I think we want a data model that sits between the DB entity and our business logic to encapsulate all of this

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/TagValue.kt
```diff
@@ -0,0 +1,6 @@
+package uk.gov.communities.prsdb.webapp.models.viewModels.summaryModels
+
+data class TagValue(
```
Travis-Softwire, line 6: This would be handy to use elsewhere as well!

src/main/resources/templates/fragments/summaryRow.html
```diff
@@ -11,13 +11,18 @@
         </a>
         <!--/* If and only if there is a single value (as opposed to a list) we should th:remove the p tag to avoid unnecessary nesting */-->
         <th:block th:if="${valueUrl} == null" th:each="value, iteratorStatus: ${values}">
+            <!--/* TagValue: render as a GOV.UK tag */-->
```
Travis-Softwire, line 18: Worth having a small tag fragment? And again using it more widely

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/NotificationBannerViewModelFactory.kt
```diff
@@ -0,0 +1,67 @@
+package uk.gov.communities.prsdb.webapp.models.viewModels.summaryModels.propertyComplianceViewModels
+
+import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.PrsdbWebService
+import uk.gov.communities.prsdb.webapp.constants.COMPLIANCE_INFO_FRAGMENT
+import uk.gov.communities.prsdb.webapp.database.entity.PropertyCompliance
+
+@PrsdbWebService("notificationBannerViewModelServiceRedesign")
+class NotificationBannerViewModelFactory : NotificationBannerViewModelService {
```
Travis-Softwire: This isn't really a factory? Feels like an odd name for an implementation of a service

Travis-Softwire (approved):
There's a fair bit here I think could be commonised with Sammy's work & more widely. Happy for that to be picked up in a later PR though. Other than that just one minor point on naming

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/TagValue.kt
TomHanmer29: I agree, I think I'll do the tag changes in this PR since they don't seem too large

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/enums/ComplianceCertStatus.kt
```diff
@@ -7,4 +7,10 @@ enum class ComplianceCertStatus {
     HAS_FAULTS,
     PROVIDE_LATER,
     EXPIRED,
+    ;
+
+    companion object {
+        val FAULTY_STATUSES = listOf(HAS_FAULTS, PROVIDE_LATER)
```
samyou-softwire: I wonder if this is quite descriptive. HAS_FAULTS is the faulty status, the point being these statuses suggest an action will be needed

maybe clearest would be two lists `NEEDS_COMPLIANCE_IF_OCCUPIED_STATUSES` and `NEEDS_COMPLIANCE_ALWAYS_STATUSES` which should fit into `shouldShowCert` below

or, if these status lists only really make sense in the context of the compliance status page, maybe it could be moved to that class? I'm not sure if we'd want to use these specific lists anywhere else

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ComplianceStatusDataModel.kt
```diff
@@ -10,23 +10,23 @@ data class ComplianceStatusDataModel(
...
     val provideLaterDeadline: LocalDate? = null,
     val gasSafetyExpiryDate: LocalDate? = null,
-    val eicrExpiryDate: LocalDate? = null,
+    val electricalSafetyExpiryDate: LocalDate? = null,
     val epcExpiryDate: LocalDate? = null,
 ) {
     fun shouldShowCert(status: ComplianceCertStatus): Boolean =
         status == ComplianceCertStatus.EXPIRED ||
-            (isOccupied && !listOf(ComplianceCertStatus.ADDED, ComplianceCertStatus.NOT_REQUIRED).contains(status))
+            (isOccupied && status !in ComplianceCertStatus.VALID_STATUSES)
```
samyou-softwire: I wonder why we use the negative contains here. would it make more sense to list the statuses that do need a compliance action?

samyou-softwire (approved):
happy with the logic here. I think we should be careful with our method/constant names to be very clear the context we expect them to be used in to avoid us tying logic together that need not be

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/enums/ComplianceCertStatus.kt
TomHanmer29: I might move them, but if I do I'll do it in the later branches :) I reuse one of these in the gas one

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ComplianceStatusDataModel.kt
TomHanmer29: Have cleaned this up a bit. Previously, we just used the negative as there were a ton of statuses and it was easier to just use the 2 it wasn't, but now it's a bit neater.

samyou-softwire (approved):

PR #1381, opened 2026-05-27
PDJB-794: EPC compliance tab redesign

TomHanmer29:
## Ticket number

PDJB-794

## Goal of change

Redesign the EPC section of the compliance tab with certificate status, expiry logic, low energy rating warnings and MEES exemption handling.

image
Provide later (occupied)

image
Provide later (unoccupied)

image
Valid

image
Expired but in date when tenancy began

image
Expired but in date, low energy but exempt

image
EPC not required (exempt)

image
EPC not given

image
Below E, no exemption, occupied

image
Expired occupied

image
BECOMES expired (occupied)

image
BECOMES expired (occupied) - low energy and not exempt

image
BECOMES expired (occupied) - low energy and exempt

image
BECOMES expired (unoccupied) - low energy

image
BECOMES expired (unoccupied)

- Adds EPC view model service interface with `@PrsdbFlip` for flag switching
- Adds redesign implementation with certificate status tags, expiry dates and supplementary sections
- Adds legacy implementation that preserves existing behaviour when flag is off
- Adds low energy rating inset text and MEES exemption row logic
- Adds naturally expired EPC inset text with tenancy start date context

## Checklist

- [ ] Branch has been rebased onto main and run locally, with everything working as expected


src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/EpcViewModelFactory.kt
```diff
@@ -0,0 +1,225 @@
...
+import uk.gov.communities.prsdb.webapp.models.viewModels.summaryModels.EpcExpiredInsetViewModel
+import uk.gov.communities.prsdb.webapp.models.viewModels.summaryModels.SummaryCardSupplementarySection
+import uk.gov.communities.prsdb.webapp.models.viewModels.summaryModels.SummaryListRowViewModel
+import uk.gov.communities.prsdb.webapp.models.viewModels.summaryModels.TagValue
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
+import java.util.Locale
+
+@PrsdbWebService("epcViewModelServiceRedesign")
+class EpcViewModelFactory(
```
TomHanmer29, line 23: This may want a bit of a heavy refactor for neatness when I rebase [Sammy's code](https://github.com/communitiesuk/prsdb-webapp/pull/1374) in, but can be looked at now too

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/EpcViewModelFactory.kt
```diff
@@ -0,0 +1,223 @@
...
+    override fun getInsetTextKey(propertyCompliance: PropertyCompliance): String? {
+        val status = ComplianceStatusDataModel.fromPropertyCompliance(propertyCompliance).epcStatus
+
+        return when {
+            status == ComplianceCertStatus.NOT_ADDED &&
+                propertyCompliance.propertyOwnership.isOccupied -> {
+                "propertyCompliance.epcTask.checkEpcAnswers.occupiedNoEpcInset"
+            }
+
+            propertyCompliance.isEpcExpired == true &&
```
Travis-Softwire: I'm possibly missing something obvious here, but we have a bunch of explicit `== true` conditions - what other truthy value can these fields have? I get `== false` to distinguish from `null` but why do we need the same for `true`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/EpcViewModelFactory.kt
```diff
@@ -0,0 +1,223 @@
...
+class EpcViewModelFactory(
+    private val messageSource: MessageSource,
+) : EpcViewModelService {
+    override fun getInsetTextKey(propertyCompliance: PropertyCompliance): String? {
+        val status = ComplianceStatusDataModel.fromPropertyCompliance(propertyCompliance).epcStatus
+
+        return when {
+            status == ComplianceCertStatus.NOT_ADDED &&
+                propertyCompliance.propertyOwnership.isOccupied -> {
+                "propertyCompliance.epcTask.checkEpcAnswers.occupiedNoEpcInset"
```
Travis-Softwire: We shouldn't be using a CYA message key here - if it's definitely always going to be the same message then let's commonise it into a suitable message file/ create one, if not then we should duplicate the message

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/EpcViewModelFactory.kt
```diff
@@ -0,0 +1,223 @@
...
+                null
+            }
+        }
+    }
+
+    override fun getEpcExpiredInsetViewModel(propertyCompliance: PropertyCompliance): EpcExpiredInsetViewModel? {
+        if (propertyCompliance.isEpcExpired != true) return null
+        if (propertyCompliance.tenancyStartedBeforeEpcExpiry != null) return null
+        if (!propertyCompliance.propertyOwnership.isOccupied) return null
+        if (propertyCompliance.isEpcRatingLow == true && propertyCompliance.epcMeesExemptionReason == null) return null
```
Travis-Softwire: Given the discussions on slack, can we triple check the logic here is consistent with what we're saying elsewhere? Conscious that if you click on the link from the compliance actions page it takes you here, so we don't want them to contradict each other

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/EpcViewModelFactory.kt
Travis-Softwire: Yes, I think we want to have the data model be the place where we can easily query the state of the certs and keep this to just mapping those states to the view model

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/PropertyComplianceViewModelFactory.kt
```diff
@@ -71,13 +76,32 @@ class PropertyComplianceViewModelFactory(
                 insetTextKey = electricalSafetyViewModelService.getInsetTextKey(propertyCompliance),
             )
 
+        val epcCertificateUrl =
+            propertyCompliance.epcUrl
```
Travis-Softwire: Wait, what? What is `epcUrl` if it's not the certificate URL?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/EpcViewModelFactory.kt
```diff
@@ -0,0 +1,223 @@
...
+                        key = "propertyDetails.complianceInformation.energyPerformance.registeredExemption",
+                        value = MessageKeyConverter.convert(propertyCompliance.epcMeesExemptionReason!!),
+                    )
+                }
+            }.toList()
+
+    private fun getMissingCertValue(
+        status: ComplianceCertStatus,
+        propertyCompliance: PropertyCompliance,
+    ): Any {
```
Travis-Softwire: Why Any instead of String for the return type?

Travis-Softwire (approved):
A few minor points - I think the bigger ones probably needs to be covered by the refactor though as you say. Assuming the refactor is in a separate PR I'm happy I don't need to see this one again

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/EpcViewModelFactory.kt
TomHanmer29: It's actually a kotlin requirement as far as I can tell - the linter enforces a strict == true check for a nullable boolean when used in an if

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/PropertyComplianceViewModelFactory.kt
TomHanmer29: You're right, this is crazy! I think copilot misunderstood how the storing worked... I should have probably caught this 😅

TomHanmer29:
I'm going to get this merged for demo, and then finish refactoring in https://github.com/communitiesuk/prsdb-webapp/pull/1382

PR #1382, opened 2026-05-27
PDJB-794: Compliance tab feature flag integration tests and refactor

TomHanmer29:
## Ticket number

PDJB-794

## Goal of change

Add integration tests verifying the compliance tab feature flag correctly toggles between legacy and redesigned views.
Refactors EPC to store the logic in a more proper place.

- Adds `PropertyComplianceTabFeatureFlagTests` with symmetric flag-on/flag-off test coverage
- Tests notification banner visibility, gas/electrical/EPC certificate status row presence
- Adds typed page object components (`GasSafetyCard`, `ElectricalSafetyCard`, `EpcCard`) with named summary list rows and inset text locators

## Checklist

- [ ] Branch has been rebased onto main and run locally, with everything working as expected


src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/PropertyComplianceEpcExtensions.kt
```diff
@@ -0,0 +1,34 @@
...
+import uk.gov.communities.prsdb.webapp.constants.EPC_ACCEPTABLE_RATING_RANGE
+import uk.gov.communities.prsdb.webapp.database.entity.PropertyCompliance
+
+val PropertyCompliance.isEpcEnergyRatingLow: Boolean
+    get() = epcEnergyRating?.uppercase()?.let { it !in EPC_ACCEPTABLE_RATING_RANGE } ?: false
+
+val PropertyCompliance.isEpcNonExpiredButLowRating: Boolean
+    get() = isEpcExpired != true && isEpcRatingLow == true
+
+val PropertyCompliance.shouldShowCouncilWillSeeEpcInset: Boolean
```
samyou-softwire: are these methods very reusable? do they make sense in the context of the compliance tab view? if not, I think they should be part of that view model

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/PropertyComplianceEpcExtensions.kt
```diff
@@ -0,0 +1,34 @@
...
+                    (isEpcExpired == true && tenancyStartedBeforeEpcExpiry == false) ||
+                    (isEpcRatingLow == true)
+            )
+
+val PropertyCompliance.shouldShowEpcExpiredNaturallyInset: Boolean
+    get() =
+        isEpcExpired == true &&
+            tenancyStartedBeforeEpcExpiry == null &&
+            propertyOwnership.isOccupied &&
+            isEpcRatingLow != true
```
samyou-softwire: it does feel a bit off that we have all of these conditions. they're not particularly easy to reason with and I have a suspicion some of it may be duplicated.

could we decompose these into smaller methods? then, it should be more obvious what exactly we're doing here and it should achieve the goal better of letting us query compliance easier. ideally at the level here (methods specific to the compliance tab) they'll be the && of a couple well named methods which is much easier to read

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/PropertyComplianceEpcExtensions.kt
samyou-softwire, line 1: in any case PropertyCompliance is our class; I don't particular see a need to define this as an extension class?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/EpcViewModelFactory.kt
```diff
@@ -2,89 +2,78 @@ package uk.gov.communities.prsdb.webapp.models.viewModels.summaryModels.property
...
-            (!propertyCompliance.hasEpcUrl && !propertyCompliance.hasEpcExemption && propertyCompliance.epcProvideLater != true) ||
-                (propertyCompliance.isEpcExpired == true && propertyCompliance.tenancyStartedBeforeEpcExpiry == false) ||
-                (propertyCompliance.isEpcRatingLow == true)
+    override fun getStatus(propertyCompliance: PropertyCompliance): ComplianceCertStatus =
+        ComplianceStatusDataModel.fromPropertyCompliance(propertyCompliance).epcStatusMay2026Redesign
 
-        return if (shouldShowInset) OCCUPIED_NO_EPC_INSET_KEY else null
-    }
+    override fun getInsetTextKey(propertyCompliance: PropertyCompliance): String? =
+        if (propertyCompliance.shouldShowCouncilWillSeeEpcInset) occupiedNoCertInsetKey else null
```
samyou-softwire, line 30: the base class does this based off of the status. why can't we do that here? if there is an exception or two to why we can't base this off the status alone, would it make more sense to rewrite the logic as

shows the inset text if:
- status is one that shows the inset
- AND some other condition

rather than re-expressing from base principles

samyou-softwire (changes requested):
like the idea behind these changes, though I feel we should be moving to a stage where we never need write another 'is this valid' check again - I think the functions are too specific right now to allow for this

samyou-softwire:
not part of this PR but I think also worth a look: `isEpcValidDespiteExpiry` in `ComplianceStatusDataModel`. could this be part of the `PropertyCompliance` class?

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/PropertyComplianceEpcExtensions.kt
TomHanmer29: Really, these are specific to the compliance tab - they're the extra methods that I needed on top of the existing cert stuff to add in the extra few rows. They are only used in the compliance tab. I wonder if this is the best place for them to live though, maybe renaming this file to something like `PropertyComplianceTabLogicEpcExtensions` could be enough?

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/PropertyComplianceEpcExtensions.kt
TomHanmer29: I just felt like the class was getting a bit bloated, and these ones specifically are only used in the EPC tab, but it's true that they could just be in the class as is.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/EpcViewModelFactory.kt
TomHanmer29: So the logic for this breaks down like so. Using statuses, we show the council will see insert for EPC if:
`isOccupied && (HAS_FAULTS || (EXPIRED && (tenancyStartedBeforeExpiryDate == false || isEpcRatingLow == true)))` which feels a bit gross. I would like to make stuff a bit more common though.

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/PropertyComplianceEpcExtensions.kt
TomHanmer29: Have just done this, lmk what you think.

TomHanmer29:
> not part of this PR but I think also worth a look: isEpcValidDespiteExpiry in ComplianceStatusDataModel. could this be part of the PropertyCompliance class?

I've moved this into the extension methods, as I think it was definitely in the wrong place. But we may want to move these further, since it is a bit weird having them in this little side file 😆

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/extensions/PropertyComplianceEpcExtensions.kt
samyou-softwire: could they go with the view model they are used in? you can define private extensions there

TomHanmer29: Sure, I've moved them in! It still feels like a bit of a weird place to have them but it's better than them dangling in a weird helper I think.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/ComplianceViewModelFactoryBase.kt
samyou-softwire, line 1: minor - this called a `ServiceBase` & the implementors are `Factory`. should they be the same format?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/ComplianceViewModelServiceBase.kt
```diff
@@ -0,0 +1,62 @@
...
+            occupiedNoCertInsetKey
+        } else {
+            null
+        }
+
+    protected fun getMissingCertValue(
+        status: ComplianceCertStatus,
+        propertyCompliance: PropertyCompliance,
+    ): Any {
+        val isOccupied = propertyCompliance.propertyOwnership.isOccupied
```
samyou-softwire: seems a bit odd to define this const that's only used once

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/ComplianceViewModelServiceBase.kt
```diff
@@ -0,0 +1,62 @@
...
+
+        return when {
+            !isOccupied -> provideLaterUnoccupiedKey
+            status == ComplianceCertStatus.PROVIDE_LATER ->
+                getProvideLaterWithDeadlineText(propertyCompliance.propertyOwnership.lastOccupiedDate)
+            else -> missingCertOccupiedValue
+        }
+    }
+
+    protected fun getProvideLaterWithDeadlineText(lastOccupiedDate: LocalDate?): String {
```
samyou-softwire: why is this fun protected? I can't see any other usages

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/EpcViewModelFactory.kt
```diff
@@ -153,3 +148,44 @@ class EpcViewModelFactory(
                 }
             }.toList()
 }
+
+private val PropertyCompliance.isEpcEnergyRatingLow: Boolean
```
samyou-softwire: `PropertyCompliance` contains a method `isEpcRatingLow`. what's different about this method?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/EpcViewModelFactory.kt
```diff
@@ -153,3 +148,44 @@ class EpcViewModelFactory(
...
+    get() = epcEnergyRating?.uppercase()?.let { it !in EPC_ACCEPTABLE_RATING_RANGE } ?: false
+
+private val PropertyCompliance.isEpcNonExpiredButLowRating: Boolean
+    get() = isEpcExpired != true && isEpcRatingLow == true
+
+private val PropertyCompliance.hasNoEpcAndNotProvidingLater: Boolean
+    get() = !hasEpcUrl && !hasEpcExemption && epcProvideLater != true
+
+private val PropertyCompliance.isEpcExpiredAfterTenancyStart: Boolean
```
samyou-softwire: I think this is a useful method that can go into `PropertyCompliance`. ditto for `didEpcBecomeExpired` and `isEpcValidDespiteExpiry`

samyou-softwire (approved):
happier with this now. I haven't done a full deep dive on the logic though this was checked in the three smaller PRs. some last qs about methods that we have put as private extensions but I think can be promoted to public methods in PropertyCompliance

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/EpcViewModelFactory.kt
TomHanmer29: Nothing, I didn't find isEpcRatingLow the first time 😄

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
```diff
@@ -104,6 +104,15 @@ class PropertyCompliance() : ModifiableAuditableEntity() {
     val hasEpcUrl: Boolean
         get() = epcUrl != null
 
+    val isEpcExpiredAfterTenancyStart: Boolean
+        get() = isEpcExpired == true && tenancyStartedBeforeEpcExpiry == false
+
+    val didEpcBecomeExpired: Boolean
+        get() = isEpcExpired == true && tenancyStartedBeforeEpcExpiry == null
```
samyou-softwire, line 111: not something for now, though I wonder for how much longer there will be a meaningful difference between this method and the one above

PR #1384, opened 2026-05-27
PDJB-928: Part 2: EPC insert

samyou-softwire:
## Ticket number

## Goal of change

add the conditional EPC insert
compliance action with EPC insert

adds an insert at the bottom of the summary card if the EPC is expired and occupied

adds capability to summary card to have a footer element

are the conditions for showing the insert fine? we don't need to actually verify the EPC would be valid, since if it were not we'd be showing the no valid EPC copy

is there a better way to have the outer page provide this kind of conditional element to the summary card? wanted something where we could place arbitrary elements with any sequence of insets, links, normal copy etc

## Checklist

  mention that here


src/main/resources/templates/fragments/summaryCard.html
```diff
@@ -1,4 +1,4 @@
-<div th:fragment="summaryCard(title, cardNumber, actions, summaryList)" class="govuk-summary-card">
+<div th:fragment="summaryCard(title, cardNumber, actions, summaryList, additionalContent)" class="govuk-summary-card">
```
samyou-softwire: there's a couple changes through the rest of this PR to add a default for this; it seems thymeleaf isn't that great at allowing optional params? so maybe there's nothing more to be done here

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/ComplianceActionViewModelBuilder.kt
```diff
@@ -165,6 +167,20 @@ class ComplianceActionViewModelBuilderMay26Redesign {
                 else -> null
             }
 
+        private fun getInsetViewModel(dataModel: ComplianceStatusDataModel): ComplianceActionInsetViewModel? {
+            val epcExpiryDate = dataModel.epcExpiryDate
+            return if (dataModel.epcStatus == ComplianceCertStatus.EXPIRED && dataModel.isOccupied &&
+                epcExpiryDate != null
```
Travis-Softwire: What's the case where we can have a status of expired with a null expiry date?

src/main/resources/templates/fragments/summaryCard.html
Travis-Softwire: An alternative would be to wrap the version with the extra content with a new fragment that has passes in `~{}` to `additionalContent` and insert that wrapper fragment everywhere else instead of this 'inner' version

Travis-Softwire (approved):
One suggestion in response to your Q, but LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/ComplianceActionViewModelBuilder.kt
samyou-softwire: this was one that was more to appease the compiler since the expiry date is marked nullable

are we happy to say, mark the `epcExpiryDate` usage as `epcExpiryDate!!`?

src/main/resources/templates/fragments/summaryCard.html
samyou-softwire: have reworked the fragments to work like this

src/main/resources/templates/fragments/complianceActionsInset.html
```diff
@@ -0,0 +1,8 @@
+<div th:fragment="epcExpiredInset(insetViewModel)" th:if="${insetViewModel != null}" class="govuk-inset-text">
+    <span th:utext="#{complianceActions.inset.epcExpired.beforeLink(${insetViewModel.expiryDate})}"></span>
+    <span th:text="' '"></span>
```
Travis-Softwire: I don't think I've seen us handle extra spaces this way before - I think we just add them inline into the relevant message? So in this case at the end of `..epcExpired.beforeLink`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/ComplianceActionViewModelBuilder.kt
Travis-Softwire: Ah ok, in that case we should make the behaviour we want explicit - i.e. if we want to return null if epcExpiryDate is null when we hit this function then having `val epcExpirydate = dataModel.epcExpiryDate ?: return null` makes that behaviour clear and should satisfy the compiler

Travis-Softwire: Or if we want to throw an error because it should be impossible, then move the variable creation inside the return if block and swap the return for a throw statement with a helpful error message :)

Travis-Softwire (approved):
Two minor points, good to go once those are addressed, I don't need to see it again :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/ComplianceActionViewModelBuilder.kt
samyou-softwire: have gone for the latter

src/main/resources/templates/fragments/complianceActionsInset.html
samyou-softwire: think this is outright unneeded, a space will be added automatically after the span
image

PR #1385, opened 2026-05-27
PDJB-792: Move to a single feature flag for the entire compliance redesign

samyou-softwire:
## Ticket number

[PDJB-792](https://mhclgdigital.atlassian.net/browse/PDJB-792)

## Goal of change

rename the compliance feature flag to one that encompasses the entire redesign, for #1374, #1384 and #1378 

renames to just the compliance may2026 redesign

specifies the whole year, may26 was a bit ambiguous. there's some other may26 references we could change but I'll leave to just the feature flag for now

no

## Checklist


TomHanmer29 (approved):
Change LGTM. Since this is in the release that's currently already turned on on local/int, should we toggle it off again, or does it not really matter?

samyou-softwire:
I think it's fine to leave this on so we can test

PR #1375, opened 2026-05-26
PDJB-703: address QA markups on local council name audit

Bill-Haigh:
## Ticket number

PDJB-703

## Goal of change

Address QA markups raised against the original PDJB-703 (local council name audit) PR.

## Description of main change(s)

Seven small content/structure fixes on local council pages flagged by QA:

- **Search for a landlord** and **Search for a property**: removes the redundant `Tell us the landlord's/property's contact details` hint above the search field, and updates the bottom details/reveal panel heading and body to the agreed copy (`If you cannot find a landlord/property`).
- **Invite a new local council user**: adds a back link to the manage-users page. (Controller now sets `backLinkPath`; template restructured to use the `~{::body/content()}` layout pattern so the back link is included in the rendered layout, matching `deleteLocalCouncilUser.html`.)
- **Cancel local council user invitation** warning page: adds a `govuk-section-break` between the user-details section and the body paragraphs to match the spacing on `deleteLocalCouncilUser.html`.
- **Cancel invitation success** page: removes the redundant `What happens next` `<h2>` so the paragraph flows directly under the confirmation banner.
- **Delete local council user success** page: moves the parameterised `You've removed X's account from Y` line out of the banner body and into the banner heading (passing `null` as the body, matching the pattern used in `deregisterPropertyConfirmation.html` / `deleteIncompletePropertyConfirmation.html`).
- **Register local council user invalid-link** page: updates the heading to `This link is not valid`.

Integration test updates:
- `LocalCouncilUserRegistrationJourneyTests` and `LocalCouncilUserRegistrationSinglePageTests`: invalid-link heading assertions updated.
- `InviteNewLocalCouncilUserPage` page object exposes the back-link component.
- New single-page test asserts the new back link is visible.

## Anything you'd like to highlight to the reviewer?

- The largest structural change is in `inviteLocalCouncilUser.html`: it was previously using `<html th:replace="layout(... ~{::main} ...)">`, which would not have included a sibling back link. I restructured it to the `<body th:replace="layout(... ~{::body/content()} ...)">` pattern used by `deleteLocalCouncilUser.html` so that the back link is part of the rendered layout. Same end result, no other behaviour change.
- The delete-user success banner now uses the existing fragment with a `null` body — already an established usage in two other templates, so no fragment change was needed.

## Checklist

- [x] Screenshots of any UI changes have been added _(QA verified against Figma; smoke-tested locally on port 8081)_
[...]

PR #1386, opened 2026-05-28
PDJB-703: Fix broken back link on invite-new-user page

Bill-Haigh:
## Ticket number

PDJB-703

## Goal of change

Fix the broken back link on the invite-new-user page that 404s when clicked, and tighten the integration test so this kind of regression is caught in future.

- Change the back link path on the invite-new-user page from `../manage-users` to `manage-users` so the browser resolves it to `/local-council/{id}/manage-users` instead of `/local-council/manage-users`
- Update the single page integration test to actually click the back link and assert it lands on the manage users page (the previous assertion only checked the link was visible, which did not catch the bug)

The `../` prefix works on the cancel-invitation and delete-user pages because their URLs have an extra `{id}` segment for the browser to consume when resolving `..`, but the invite-new-user URL is only one segment deep below `{localCouncilId}`, so `..` ended up dropping the council id.

Spotted while testing on the integration environment after #1375 was merged.

## Checklist


samyou-softwire (approved):
LGTM

can there be some QA notes for this on the ticket?

PR #1392, opened 2026-05-29
PDJB-268: Add joint landlord invitation expiry scheduled task and notification email

Bill-Haigh:
## Ticket number

PDJB-268

## Goal of change

Expire pending joint landlord invitations after 28 days and notify the inviting landlord by email.

- Adds a scheduled task that finds joint landlord invitations older than the configured lifetime (28 days), emails the primary landlord about each one, and deletes the invitation
- Adds a new "joint landlord invitation expired" Notify email template, view model, and markdown body, with the expiry-days value derived from `JOINT_LANDLORD_INVITATION_LIFETIME_IN_HOURS` rather than hardcoded
- Adds a repository query to fetch invitations created before a given cutoff
- Adds a helper on `AbsoluteUrlProvider` to build the absolute URL to the landlord-facing property details page (linked from the expiry email)
- Wraps the expiry service behind the `JOINT_LANDLORDS` feature flag (no-op when the flag is off)
- Registers the service impls with `@PrsdbTaskService` and extends `PrsdbTaskApplicationTests` so the beans are exercised in task-runner mode

- Expiry emails are currently only sent to the primary landlord. Once PDJB-260 lands, accepted joint landlords should also be notified — there's a TODO in the service flagging this.

## Checklist

- [x] New email templates have been added to `/src/main/resources/emails/emailTemplates.json`


src/main/kotlin/uk/gov/communities/prsdb/webapp/application/DeleteIncompletePropertiesTaskApplicationRunner.kt
```diff
@@ -16,22 +16,13 @@ class DeleteIncompletePropertiesTaskApplicationRunner(
     override fun run(args: ApplicationArguments) {
         println("Executing delete incomplete properties scheduled task")
 
-        // Separating into its own method to allow this to be tested without "exitProcess" being called
```
samyou-softwire, line 19: this seemed like an antipattern to me, the test we arranged for this for only verifies one method call will call another method, which didn't feel like a very useful test

simplified this file and the new regular job that's based off this file and removed the tests

src/main/kotlin/uk/gov/communities/prsdb/webapp/application/ExpireJointLandlordInvitationsTaskApplicationRunner.kt
```diff
@@ -0,0 +1,27 @@
...
+
+@PrsdbScheduledTask("expire-joint-landlord-invitations-scheduled-task")
+class ExpireJointLandlordInvitationsTaskApplicationRunner(
+    private val context: ApplicationContext,
+    private val jointLandlordInvitationExpiryService: JointLandlordInvitationExpiryService,
+) : ApplicationRunner {
+    override fun run(args: ApplicationArguments?) {
+        println("Executing expire joint landlord invitations scheduled task")
+
+        jointLandlordInvitationExpiryService.expirePendingInvitations()
```
samyou-softwire: do we want any additional logging here?

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/InvitationLifetimeConstants.kt
```diff
@@ -1,4 +1,4 @@
 package uk.gov.communities.prsdb.webapp.constants
 
 const val LOCAL_COUNCIL_INVITATION_LIFETIME_IN_HOURS: Int = 48
-const val JOINT_LANDLORD_INVITATION_LIFETIME_IN_HOURS: Int = 672 // 28 days
```
samyou-softwire: we only ever treat this as a number of days so seemed odd to have it in hours

src/main/kotlin/uk/gov/communities/prsdb/webapp/application/DeleteIncompletePropertiesTaskApplicationRunner.kt
JasminConterioSW: I would leave that test there to check that the scheduled task is calling the service function, unless that is tested elsewhere?

JasminConterioSW: (Also add the equivalent to the new task runner)
There might be a better way of doing it, but the functionality should still be tested

samyou-softwire: mm for me it seems unclear when this test would fail. I trust that this service that does only this will succeed. if there was some logic in the way it may be different. if this is out testing approach on PRS I'm happy to follow suit

JasminConterioSW: It would fail if someone changed `deleteIncompletePropertiesTaskLogic `to not call that service method, e.g. if they hardcoded `numberOfRecordsDeleted` or set to `numberOfRecordsDeleted = someOtherService.returnANumber()`

src/main/kotlin/uk/gov/communities/prsdb/webapp/application/ExpireJointLandlordInvitationsTaskApplicationRunner.kt
JasminConterioSW: Yeah might be helpful for debugging purposes, maybe something like "{n} invitations deleted"? I don't think we need lots (although we've not had to do much debugging of scheduled tasks yet so may end up wanting more later!)

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/InvitationLifetimeConstants.kt
JasminConterioSW: Agree, I suspect it started as hours to copy the local council version but I don't think we need to be so granular for 28 days

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitation.kt
```diff
@@ -33,6 +33,14 @@ class JointLandlordInvitation(
     lateinit var invitingLandlord: Landlord
         private set
 
+    @Column(nullable = false)
+    var expired: Boolean = false
```
JasminConterioSW: I don't think this should be added to the database, I think we should calculate it from the createdDate when we need to check it.

I'm looking at adding 
```
    fun getInvitationHasExpired(invitation: JointLandlordInvitation): Boolean {
        val dateTimeHelper = DateTimeHelper()

        val expiresOnDate =
            DateTimeHelper
                .getDateInUK(invitation.createdDate.toKotlinInstant())
                .plus(DatePeriod(days = JOINT_LANDLORD_INVITATION_LIFETIME_IN_DAYS))

        return dateTimeHelper.getCurrentDateInUK() > expiresOnDate
```
to JointLandlordInvitationService in my current ticket, maybe something like that would work here?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationExpiryService.kt
```diff
@@ -0,0 +1,84 @@
...
+        expiredInvitations.forEach { invitation ->
+            try {
+                sendExpiryEmailsForInvitation(invitation)
+                invitation.markAsExpired()
+                invitationRepository.save(invitation)
+            } catch (ex: PersistentEmailSendException) {
+                printFailureMessage(ex, invitation)
+            } catch (ex: TransientEmailSentException) {
+                printFailureMessage(ex, invitation)
+            }
```
JasminConterioSW: I don't think we need to mark invitations as expired in the db, we should generally just check if they were created before the cutoff.

Although... I'm guessing we want to only send them one email to tell them that the invitation has expired? We did something similar with the incomplete properties reminder task - we added a new ReminderEmailSent table to record in the db which incomplete properties we had sent a reminder email about, and checked to make sure we didn't send another.

We might not need this to be as complicated though, maybe add an `invitationExpiredEmailSent `field to the invitation entity?

JasminConterioSW (changes requested):

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitation.kt
samyou-softwire: mm I see, the reason why I was thinking to do it this way is that it means that the 'your invite is expired' email and the expiration happening occur at the same time. though thinking again the email is more of a formality so agree let's 'expire' immediately and send the email later

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationExpiryService.kt
samyou-softwire: I'll rework the flag to work more like this

src/main/kotlin/uk/gov/communities/prsdb/webapp/application/ExpireJointLandlordInvitationsTaskApplicationRunner.kt
samyou-softwire: yeah, I think if I was debugging this task it'd be useful to get some reference to which landlords it expired. we can put the invitation ID in the logs as is it non-PII

src/main/kotlin/uk/gov/communities/prsdb/webapp/application/DeleteIncompletePropertiesTaskApplicationRunner.kt
samyou-softwire: sure - I've put it back

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitation.kt
samyou-softwire: I'm thinking of having this method as a prop on JointLandlordInvitation itself as it's specific to the invitation model itself

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitation.kt
```diff
@@ -33,6 +33,14 @@ class JointLandlordInvitation(
     lateinit var invitingLandlord: Landlord
         private set
 
+    @Column(nullable = false)
+    var invitationExpiredEmailSent: Boolean = false
```
JasminConterioSW, line 46: It might be useful to have this as a date that the email was sent (that's what we save in ReminderEmailSent) in case this is useful for tracking (it could be null if nothing has been sent).

But I don't think anyone has asked for this so it's probably fine!

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationExpiryService.kt
```diff
@@ -0,0 +1,89 @@
...
+                    invitedEmail = invitation.invitedEmail,
+                    propertyAddress = propertyAddress,
+                    propertyRecordUri = propertyRecordUri,
+                    expiryDays = JOINT_LANDLORD_INVITATION_LIFETIME_IN_DAYS,
+                ),
+            )
+        }
+    }
+
+    // TODO PDJB-260: include accepted joint landlords once that data model exists.
```
JasminConterioSW: This is most likely to be PDJB-264 or Alex's PDJB-432 ticket

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitation.kt
JasminConterioSW: Adding it to JointLandlordService might be more consistent with what we do elsewhere (it's what I'm likely to do here https://github.com/communitiesuk/prsdb-webapp/pull/1409 if you don't beat me to it)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationExpiryService.kt
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+@PrsdbTaskService("joint-landlord-invitation-expiry-flag-on")
+class JointLandlordInvitationExpiryServiceImplFlagOn(
+    private val invitationRepository: JointLandlordInvitationRepository,
+    private val expiryEmailNotificationService: EmailNotificationService<JointLandlordInvitationExpiryEmail>,
+    private val absoluteUrlProvider: AbsoluteUrlProvider,
+) : JointLandlordInvitationExpiryService {
+    override fun expirePendingInvitations(): List<Long> {
```
JasminConterioSW: This one will probably want to be renamed - expiring isn't something we do to an invitation, it just happens when they get old.

It's more like sendExpiredInvitationEmails (or you could include something saying that we record that too)

JasminConterioSW (commented):
I've not looked at everything, just a few comments as I'm off tomorrow!

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitation.kt
samyou-softwire: see #1401, it's now a merged method on `JointLandlordInvitation` as well as one for grabbing the expiry date

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationExpiryService.kt
samyou-softwire: agree yeah, I think some renaming will be needed here across the files

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationExpiryService.kt
samyou-softwire: ah good point, have swapped to [PDJB-432](https://mhclgdigital.atlassian.net/browse/PDJB-432)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitation.kt
samyou-softwire: I'm not too fussed - notify will also be logging this as well as our own logs

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitation.kt
```diff
@@ -39,6 +39,14 @@ class JointLandlordInvitation(
...
```
Travis-Softwire, line 46: I think this is the first actually mutable field on the entity, which means we should probably be extending `ModifiableAuditableEntity` instead of just `AuditableEntity`

src/main/kotlin/uk/gov/communities/prsdb/webapp/application/JointLandlordInvitationExpiryEmailTaskApplicationRunner.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import org.springframework.boot.ApplicationArguments
+import org.springframework.boot.ApplicationRunner
+import org.springframework.boot.SpringApplication
+import org.springframework.context.ApplicationContext
+import uk.gov.communities.prsdb.webapp.annotations.taskAnnotations.PrsdbScheduledTask
+import uk.gov.communities.prsdb.webapp.services.JointLandlordInvitationExpiryEmailService
+import kotlin.system.exitProcess
+
+@PrsdbScheduledTask("jl-invitation-expiry-email-scheduled-task")
```
Travis-Softwire, line 11: Haven't cross checked myself, but just confirming that the first part needs to match the name in the infra PR

Travis-Softwire (approved):
Two small points, but otherwise LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitation.kt
JasminConterioSW: Thanks, I'll use that on mine too then instead of the service method

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/JointLandlordInvitationRepository.kt
```diff
@@ -2,8 +2,11 @@ package uk.gov.communities.prsdb.webapp.database.repository
 
 import org.springframework.data.jpa.repository.JpaRepository
 import uk.gov.communities.prsdb.webapp.database.entity.JointLandlordInvitation
+import java.time.Instant
 import java.util.UUID
 
 interface JointLandlordInvitationRepository : JpaRepository<JointLandlordInvitation, Long> {
     fun findByToken(token: UUID): JointLandlordInvitation?
+
+    fun findAllByInvitationExpiredEmailSentFalseAndCreatedDateBefore(cutoff: Instant): List<JointLandlordInvitation>
```
JasminConterioSW: This one can be updated now to use isExpired rather than the CreatedDateBefore(cutoff)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationExpiryEmailService.kt
```diff
@@ -0,0 +1,92 @@
...
+class JointLandlordInvitationExpiryEmailServiceImplFlagOn(
+    private val invitationRepository: JointLandlordInvitationRepository,
+    private val expiryEmailNotificationService: EmailNotificationService<JointLandlordInvitationExpiryEmail>,
+    private val absoluteUrlProvider: AbsoluteUrlProvider,
+) : JointLandlordInvitationExpiryEmailService {
+    override fun sendExpiryEmailsForExpiredInvitations(): List<Long> {
+        val cutoff = Instant.now().minus(JOINT_LANDLORD_INVITATION_LIFETIME_IN_DAYS.toLong(), ChronoUnit.DAYS)
+        val expiredInvitations =
+            invitationRepository
+                .findAllByInvitationExpiredEmailSentFalseAndCreatedDateBefore(cutoff)
```
JasminConterioSW: I don't think you need to calculate the cutoff again here, just use the isExpired on the invitation

JasminConterioSW (approved):
One comment about where the expiry is checked (it's happening in 2 places at the moment) but other than that LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/application/JointLandlordInvitationExpiryEmailTaskApplicationRunner.kt
samyou-softwire: infra PR is `jl-invitation-expiry-email` which should match

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitation.kt
samyou-softwire: did make this change, though ended up finding that it looks like the auditing automations currently don't work on scheduled tasks

the `AuditingConfig` class was defined as `@PrsdWebConfiguration` so would only run on web and so wouldn't work for the scheduled runner. I've changed to `@Configuration` though wary of whether this was done on purpose in the first place. I'll check in with Alex who made this change

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/JointLandlordInvitationRepository.kt
samyou-softwire: I think there is some merit on keeping this check - reduces the number of invitations returned by the query. we run the more costly `isExpired` check later to make sure we're aligned with the source of truth

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationExpiryEmailService.kt
samyou-softwire: see above

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/JointLandlordInvitationRepository.kt
samyou-softwire: had a chat and agreed it's not worth the optimisation to have this split across two places, so have made it check the flag only. the number of unexpired invites shouldn't grow too much

samyou-softwire:
want to confirm the change to `AuditingConfig.kt` before merging

samyou-softwire:
have agreed in standup that this should be fine to merge, we'll monitor whether there are any issues having this annotation turned on for non-web configurations

PR #1393, opened 2026-05-29
PDJB-928: Document how we deal with compliance

samyou-softwire:
## Ticket number

## Goal of change

document what we do in every case around compliance and offer explanations for surprising changes

I've been knee-deep in compliance for the last week which can make me a poor metric for what is surprising here. please mention if there's any aspects of these docs that nee further explanation

### DIFF
--- a/docs/ComplianceReadMe.md
+++ b/docs/ComplianceReadMe.md
@@ -6,12 +6,10 @@ This document will break down, for the 3 certificate types, what properties the 
 # Certificate state
 
 Once the compliance journey has been completed, the certificate can be in one of 4 states.
-- `ADDED`: Certificate is valid to be used for the property
-- `PROVIDE_LATER`: The user has opted to provide their certificate later, by pressing the 'Provide later' button at the start of the flow
-- `EXPIRED`: The certificate was valid but is now expired
-- `HAS_FAULTS`: The certificate has an issue and cannot be used for this property, regardless of if it is expired or not.
-
-If a certificate is both expired and has faults, 'HAS_FAULTS' takes precedence.
+- `ADDED`: Certificate is valid to be used for the property.
+- `PROVIDE_LATER`: The user has opted to provide their certificate later, by pressing the 'Provide later' button at the start of the flow.
+- `EXPIRED`: The certificate was valid but is now expired.
+- `HAS_FAULTS`: The certificate has an issue and cannot be used for this property. This takes precedence over expired.
 
 The possible states for a certificate are:
 
--- a/docs/ComplianceReadMe.md
+++ b/docs/ComplianceReadMe.md
@@ -5,7 +5,7 @@ This document will break down, for the 3 certificate types, what properties the 
 
 # Certificate state
 
-Once the compliance journey has been completed, the certificate can be in one of 4 states.
+Once the compliance steps has been completed, the certificate can be in one of 4 states.
 - `ADDED`: Certificate is valid to be used for the property.
 - `PROVIDE_LATER`: The user has opted to provide their certificate later, by pressing the 'Provide later' button at the start of the flow.
 - `EXPIRED`: The certificate was valid but is now expired.
--- a/docs/ComplianceReadMe.md
+++ b/docs/ComplianceReadMe.md
@@ -15,13 +15,15 @@ The possible states for a certificate are:
 
 ## Gas
 
-| State                                                    | If occupied   | If unoccupied |
-|----------------------------------------------------------|---------------|---------------|
-| Has gas, provide it later                                | PROVIDE_LATER | PROVIDE_LATER |
-| Has gas, has certificate, not expired                    | VALID         | VALID         |
-| Has gas, has certificate, it's expired                   | EXPIRED       | EXPIRED       |
-| Has gas, no certificate                                  | HAS_FAULTS    | HAS_FAULTS    |
-| No gas                                                   | VALID         | VALID         |
+| State                                                    | If occupied   | If unoccupied    |
+|----------------------------------------------------------|---------------|------------------|
+| Has gas, provide it later                                | PROVIDE_LATER | PROVIDE_LATER    |
+| Has gas, has certificate, not expired                    | VALID         | VALID            |
+| Has gas, has certificate, it's expired                   | EXPIRED       | EXPIRED          |
+| Has gas, no certificate                                  | HAS_FAULTS    | PROVIDE_LATER[1] |
+| No gas                                                   | VALID         | VALID            |
+
+[1]: If a cert has faults but the property is unoccupied, we treat as just provide later
 
 ## Electric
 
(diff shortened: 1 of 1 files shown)


docs/ComplianceReadMe.md
```diff
@@ -0,0 +1,65 @@
...
+The compliance flow on PRSDB has a number of branching paths.
+This document will break down, for the 3 certificate types, what properties the certificate can have, and how we combine these properties together to display on the site.
+
+# Certificate state
+
+Once the compliance journey has been completed, the certificate can be in one of 4 states.
+- `ADDED`: Certificate is valid to be used for the property
+- `PROVIDE_LATER`: The user has opted to provide their certificate later, by pressing the 'Provide later' button at the start of the flow
+- `EXPIRED`: The certificate was valid but is now expired
+- `HAS_FAULTS`: The certificate has an issue and cannot be used for this property, regardless of if it is expired or not.
```
TomHanmer29: Maybe something like "this does not take into account if the certificate is expired or not" could make this a little clearer?

TomHanmer29 (commented):
Seems nice and clear to me, thanks! I'll hold off approving as I think someone who has not spent a week looking at this should make sure it makes sense too

docs/ComplianceReadMe.md
samyou-softwire: this is kind of the case, it doesn't take expired certificate into account since it takes precedence. I've moved up the below wording that says this explicitly rather than saying it twice (with this time being in uncertain terms)

TomHanmer29: I think it might also be worth clarifying exactly what we mean by "has an issue and cannot be used". Could just be giving some examples (e.g. EPC low energy, certificate not given)

docs/ComplianceReadMe.md
```diff
@@ -0,0 +1,63 @@
+# Compliance Certifcates on PRSDB
+
+The compliance flow on PRSDB has a number of branching paths.
+This document will break down, for the 3 certificate types, what properties the certificate can have, and how we combine these properties together to display on the site.
+
+# Certificate state
+
+Once the compliance journey has been completed, the certificate can be in one of 4 states.
```
Travis-Softwire: Minor - but we don't have a compliance journey any more :)

docs/ComplianceReadMe.md
```diff
@@ -0,0 +1,63 @@
...
+The possible states for a certificate are:
+
+## Gas
+
+| State                                                    | If occupied   | If unoccupied |
+|----------------------------------------------------------|---------------|---------------|
+| Has gas, provide it later                                | PROVIDE_LATER | PROVIDE_LATER |
+| Has gas, has certificate, not expired                    | VALID         | VALID         |
+| Has gas, has certificate, it's expired                   | EXPIRED       | EXPIRED       |
+| Has gas, no certificate                                  | HAS_FAULTS    | HAS_FAULTS    |
```
Travis-Softwire: Why is this Has faults if the property is unoccupied?

docs/ComplianceReadMe.md
samyou-softwire: sounds good have given those examples. I think those are the only two reasons

docs/ComplianceReadMe.md
```diff
@@ -0,0 +1,63 @@
...
+|-----------------------------------------------------------------------------|---------------|---------------|
+| Provide it later                                                            | PROVIDE_LATER | PROVIDE_LATER |
+| Has EPC, not expired, high rating                                           | VALID         | VALID         |
+| Has EPC, not expired, low rating, has exemption                             | VALID         | VALID         |
+| Has EPC, not expired, low rating, no exemption                              | HAS_FAULTS    | HAS_FAULTS    |
+| Has EPC, expired, EPC in date when tenancy began, high rating               | EXPIRED[2]    | EXPIRED[1]    |
+| Has EPC, expired, EPC in date when tenancy began, low rating, has exemption | EXPIRED       | EXPIRED[1]    |
+| Has EPC, expired, EPC in date when tenancy began, low rating, no exemption  | HAS_FAULTS    | EXPIRED[1][2] |
+| Has EPC, expired, EPC not in date when tenancy began                        | EXPIRED       | EXPIRED[1]    |
+| No EPC, it is required                                                      | HAS_FAULTS    | HAS_FAULTS    |
```
Travis-Softwire: When is an EPC required when the property is unoccupied?

docs/ComplianceReadMe.md
```diff
@@ -0,0 +1,63 @@
...
+
+## EPC
+
+| State                                                                       | If occupied   | If unoccupied |
+|-----------------------------------------------------------------------------|---------------|---------------|
+| Provide it later                                                            | PROVIDE_LATER | PROVIDE_LATER |
+| Has EPC, not expired, high rating                                           | VALID         | VALID         |
+| Has EPC, not expired, low rating, has exemption                             | VALID         | VALID         |
+| Has EPC, not expired, low rating, no exemption                              | HAS_FAULTS    | HAS_FAULTS    |
+| Has EPC, expired, EPC in date when tenancy began, high rating               | EXPIRED[2]    | EXPIRED[1]    |
```
Travis-Softwire, line 47: I don't think the EPC in date when tenancy began part makes sense here for unoccupied properties? Same with the next 3 rows

Travis-Softwire, line 47: I'm v. surprised by this row and the one below. This is still a valid EPC and there is no action for the landlord to take if the property is still occupied...

Travis-Softwire (changes requested):
A few coments from me. I wonder if we should just drop the if unoccupied column since as far as I can see the only row where there's a difference can't happen in practice because we don't ask one of the questions?

docs/ComplianceReadMe.md
samyou-softwire: ah yup, they're just steps

docs/ComplianceReadMe.md
samyou-softwire: I had to check this one with Tom since for the compliance actions page this doesn't matter (has faults & unoccupied is no action). on the compliance tab we treat this as provide later. I'll update this table and our code to reflect that

docs/ComplianceReadMe.md
samyou-softwire: see above, this'll become PROVIDE_LATER

docs/ComplianceReadMe.md
samyou-softwire: yeah it's a bit of a limitation with the table, we 'treat' this and the next 3 rows the same as eachother as just 'Has EPC, expired'. I've clarified this further

samyou-softwire: as discussed on slack, the reason we do this is that we can't definitely know that we'll show this warning again when the landlord re-lets, so we default to showing it anyway. [PDJB-979](https://mhclgdigital.atlassian.net/browse/PDJB-979) should improve this

samyou-softwire:
as discussed on slack, there's still some questions on what we want to do with the expired but not expired when tenancy began states so we'll leave this documentation unfinished till we can get that ironed out

PR #1395, opened 2026-05-29
Feat/pdjb 260 handle auth and registration for accepted invitation

JasminConterioSW:
## Ticket number

PDJB-260

## Goal of change

If the user accepts the invitation, make sure they are logged in and integrate with the landlord registration journey if the user is not already a landlord

* Fix the local-auth profile for local testing
* Modified how the tokens are stored so they are tied to a journeyId
* Update getJourneyStep and postJourneyData endpoints so they can't initialise a new journey - we only do this from the startJourney step
* Add an authenticated endpoint after the acceptance step to make the user log in and check if they are a landlord
* If a user is not a landlord, integrate with the landlord registration flow
    * Check for a session attribute in the landlord registration journey to decide where the user should be sent after landlord CYA
    * Modify the back link on the first step to take the user back to the accept joint landlord invitation flow
        * The back link takes them to the actual journey they were in, so keeps their answer. 
        * Now also set up so that when they click "Continue" again, they re-enter the same landlord registration journey.

* Storing invitation tokens linked to a journeyId
    * Unlike local council user registration (the other place we use invitation tokens), a user could have invitations to multiple properties and start the journeys for them at the same time (in different tabs)
    * I changed the way the tokens are stored to include the journey id so that multiple tokens can be stored in one session
* Updating the getJourneyStep and postJourneyData endpoints so they can't initialize a journey
    * This is mainly so I don't need the token to initialise the journey again from here - there could be more than one in the session and I wanted to use the journeyId to decide which one to use!
* Integration with landlord registration - I originally planned to put most of landlord registration in a task with an exit step, then call that task from the joint landlord flow but the cya page (which we need) deletes the journey so loses the remaining steps. Handing between journeys is messy though, so may be worth revisiting this

## Checklist

  mention that here


src/test/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationServiceTests.kt
```diff
@@ -129,22 +129,114 @@ class JointLandlordInvitationServiceTests {
...
+        assertEquals(pairs, result)
+    }
+
+    @Test
+    fun `addJourneyIdInvitationTokenPairToSession adds a pair when session is empty`() {
+        // Act
+        invitationService.addJourneyIdInvitationTokenPairToSession("journey-1", "token-1")
+
+        // Assert
+        val stored = invitationService.getJourneyIdInvitationTokenPairsFromSession()
```
JasminConterioSW, line 159: These tests might want updating so they aren't relying on other service methods

JasminConterioSW:
Passing between the two journeys is getting a bit messy, I'm going to go back to adding a Landlord Registration task into the joint landlord journey instead

PR #1398, opened 2026-06-01
Feat/pdjb 260 handle auth and landlord registration during JL acceptance

JasminConterioSW:
## Ticket number

PDJB-260

## Goal of change

If the user accepts the invitation, make sure they are logged in and integrate with the landlord registration journey if the user is not already a landlord

- Fix the local-auth profile for local testing
- Modified how the tokens are stored so they are tied to a journeyId (allows multiple journeys)
- Take the token out of the journeyState
- Update getJourneyStep and postJourneyData endpoints so they can't initialise a new journey - we only do this from the startJourney step
- Add an authenticated endpoint after the acceptance step to make the user log in and check if they are a landlord
- If a user is not a landlord, integrate with the landlord registration flow
    - Refactor landlord registration journey
        - Put the bulk of the journey into a task with a new state
        - Move deleting the journey out of the CYA step and into it's own step (so it's not in the task but is in the full Landlord Registration Journey)
    - Add the landlord registration task to this journey and include it for non-landlord users

* Storing invitation tokens linked to a journeyId
    * Unlike local council user registration (the other place we use invitation tokens), a user could have invitations to multiple properties and start the journeys for them at the same time (in different tabs)
    * I changed the way the tokens are stored to include the journey id so that multiple tokens can be stored in one session
* Taking the token out of the journeyState
    * I had been getting this from the session and passing it into createJourneySteps so it could be initialised
    * But this caused problems if the user got into a cya journey during landlord registration as there was no token tied to the cya journey id
    * We can retrieve the token from the session when required
* Updating the getJourneyStep and postJourneyData endpoints so they can't initialize a journey
    * This is mainly so I don't need the token to initialise the journey again from here - there could be more than one in the session and I wanted to use the journeyId to decide which one to use!
* Integration with landlord registration - I[ looked at pointing the user to the actual landlord](https://github.com/communitiesuk/prsdb-webapp/pull/1395) registration journey, but passing them between journeys gets messy (making sure to pass to the correct journeyId)

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/config/security/LandlordSecurityConfig.kt
```diff
@@ -65,10 +66,13 @@ class LandlordSecurityConfig(
...
                     .permitAll()
+                    .requestMatchers(
+                        AcceptOrRejectJointLandlordInvitationController.ACCEPT_OR_REJECT_JOINT_LANDLORD_INVITATION_ROUTE +
+                            "/${CheckUserRoleStep.ROUTE_SEGMENT}",
+                    ).authenticated()
                     .requestMatchers(
                         "${AcceptOrRejectJointLandlordInvitationController.ACCEPT_OR_REJECT_JOINT_LANDLORD_INVITATION_ROUTE}/**",
-                    )
-                    .permitAll()
+                    ).permitAll()
```
Travis-Softwire, line 87: Just to double check that we're happy for someone unauthenticated to have permissions for any of the other endpoints (incl those after the check user role endpoint)? If the step isn't reachable then it __should__ be fine as long as none of the earlier steps implicitly assume the user will be logged in

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/AcceptOrRejectJointLandlordInvitationController.kt
```diff
@@ -18,46 +18,43 @@ import uk.gov.communities.prsdb.webapp.constants.JOINT_LANDLORD_INVITATION_PATH_
...
     ): ModelAndView {
-        val token = invitationService.getTokenFromSession() ?: throw(PrsdbWebException("Token not found in session"))
-        return try {
-            val journeyMap = journeyFactory.createJourneySteps(token)
-            journeyMap[stepRouteSegment]?.getStepModelAndView()
-                ?: throw ResponseStatusException(HttpStatus.NOT_FOUND, "Step not found")
-        } catch (_: NoSuchJourneyException) {
-            initializeAndRedirect(token, stepRouteSegment)
-        }
+        val journeyMap = journeyFactory.createJourneySteps()
```
Travis-Softwire: What happens if a user tries to navigate straight here now? Are we handling that gracefully?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
```diff
@@ -1,43 +1,99 @@
...
+        checkingAnswersFor: String,
+    ): Map<String, StepLifecycleOrchestrator> =
+        journey(state) {
+            unreachableStepDestination { journey.returnToCyaPageDestination }
+            configure {
+                withAdditionalContentProperty { "title" to "registerAsALandlord.title" }
+            }
+            configureFirst { backDestination { journey.returnToCyaPageDestination } }
+            when (checkingAnswersFor) {
+                NameStep.ROUTE_SEGMENT -> checkAnswerStep(journey.nameStep, NameStep.ROUTE_SEGMENT)
```
Travis-Softwire: This does feel less than ideal, as if we change part of the LL registration journey we now need to remember to change it here

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
```diff
@@ -1,43 +1,99 @@
...
+                EmailStep.ROUTE_SEGMENT -> checkAnswerStep(journey.emailStep, EmailStep.ROUTE_SEGMENT)
+                PhoneNumberStep.ROUTE_SEGMENT -> checkAnswerStep(journey.phoneNumberStep, PhoneNumberStep.ROUTE_SEGMENT)
+                CountryOfResidenceStep.ROUTE_SEGMENT ->
+                    checkAnswerStep(
+                        journey.countryOfResidenceStep,
+                        CountryOfResidenceStep.ROUTE_SEGMENT,
+                    )
+                LookupAddressStep.ROUTE_SEGMENT -> checkAnswerTask(journey.addressTask)
+            }
+            step(journey.finishCyaStep) {
```
Travis-Softwire: Should this be in the CYA journey?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
```diff
@@ -57,38 +113,97 @@ class AcceptOrRejectJointLandlordInvitationJourneyFactory(
                 parents { journey.validateTokenStep.hasOutcome(TokenValidationResult.VALID) }
                 nextUrl { mode ->
                     when (mode) {
-                        // TODO PDJB-260 - if they submit yes we will need to check if the user is a registered latndlord and if not they need to register
-                        //   The existing landlord registration journey covers most of this but can't be used as-is as we need different exit points here
-                        YesOrNo.YES -> JOINT_LANDLORD_INVITATION_ACCEPTED_CONFIRMATION_ROUTE
+                        YesOrNo.YES -> {
```
Travis-Softwire: Should the rejected invitation not just be part of the journey so we can use `nextStep` instead of `nextUrl`?

src/main/resources/application-local.yml
```diff
@@ -185,17 +185,24 @@ spring:
...
           internal-access:
-            client-id: ${INTERNAL_ACCESS_CLIENT_ID}
+            client-id: 902e8a06-4086-423d-8e67-3838c4e52d7f
             client-secret: ${INTERNAL_ACCESS_CLIENT_SECRET}
             client-authentication-method: client_secret_basic
             authorization-grant-type: authorization_code
             scope: openid,email,profile
-            redirect-uri: "{baseUrl}/local-council/login/oauth2/code/internal-access"
+            redirect-uri: http://localhost:${server.port}/local-council/login/oauth2/code/internal-access
+          epc-register:
```
Travis-Softwire, line 194: This should probably be behind a different profile rather than tying it to auth

Travis-Softwire (approved):
I think this is mostly fine. I don't love that duplicating the LL registration journey means that we have to expliciltly duplicate the CYA journey structure. I suspect the answer there is to tweak the journey framework to allow reusable CYA journey maps rather than trying to do something clever here via the actual LL registration journey

src/main/resources/application-local.yml
JasminConterioSW: This should already just be on the local-auth profile (we normally run with local-no-auth which logs in the Mock User automatically)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
JasminConterioSW: We've usually put confirmation pages outside of the journey, presumably so we can delete the journey on the way out (I should add a TODO for that...) and they can still view / refresh the confirmation page

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/AcceptOrRejectJointLandlordInvitationController.kt
JasminConterioSW: Hmm. No, they just get the error page.

Also currently if they try clicking the link again, they get an error because we use the token to seed the journey, and that journey already exists so it doesn't like trying to initialise it.

How about something like:
* If we get a `NoSuchJourneyException` we catch it and thrown 400: Bad request (on the basis that they shouldn't be trying to go straight here - we need the invitation token for this journey which we would have if they access the journey by clicking the link)
* I change the seed to include the token and timestamp (instead of just the token) so that clicking the link again can start a new journey (similar to what we do for landlord registration and maybe others that use principal as the seed)
* That way, if they go straight here
    * They get "Bad request"
    * But they can click on their invitation link again and get a journey

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/security/LandlordSecurityConfig.kt
JasminConterioSW: Actually now that we have all the landlord registration journey steps in here as well we probably should switch to specifying particular routes that have` permitAll()`

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
JasminConterioSW: I just copied it from landlord registration...

But yes, the `checkAnswerStep(journey.nameStep, NameStep.ROUTE_SEGMENT)` (or `checkAnswerTask(journey.addressTask)` functions set that as the `nextStep` once the answers have been updated

`FinishCyaJourneyConfig` does the cleanup like checking the user hasn't tried to do cya journeys in parallel, copying the new answers across to the base journey, and deleting the cya journey

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
JasminConterioSW: Argh yes that's not ideal...

I'll have a look and see if I can find a quick fix, but if not should I make a new ticket to look into making CYA journey maps re-usable?

JasminConterioSW: Maybe we can get it on the task... 🤔

JasminConterioSW: Yes, the journey map can go in the companion object on LandlordRegistrationTask, then both journeys can use it from there :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/AcceptOrRejectJointLandlordInvitationController.kt
JasminConterioSW: Oh actually "Bad request" still shows them the "Sorry there is a problem with the service" error page

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
Travis-Softwire: Ok, fair enough. I guess this is just an annoying case where it's a step or a URL. In that case can we use `nextDestination` instead, instead of coercing a step into a URL? As you can do `Destination.ExternalUrl(<url)` and `Destination(<step>)` or something like that

src/main/resources/application-local.yml
Travis-Softwire: Sorry, yes I meant putting it under it's own profile like we have with notify, so we can turn using the non-mocked epc register on and off independently of turning non-mocked auth on and off

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
Travis-Softwire: Ah ofc, we need it there so we can reference it from other steps - makes sense, thanks!

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
Travis-Softwire: Ah, nice! Is this the first time we're doing that? If so can we add the Q to dev catch-up (but we should go ahead and merge this anyway in the mean time - mostly a Q on whether we still need a ticket to do it a different way)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/AcceptOrRejectJointLandlordInvitationController.kt
Travis-Softwire: I suspect we need a slightly more sophisticated initialisation check instead, along the lines of if not initialised, is the token in the session, if it is initialise, if not redirect to the bad token page?

Travis-Softwire: Or if we don't want to initialise, redirect (preserving the token query param?) to the start page

JasminConterioSW: Hmm at the moment we've only include the token in the query parameter for the start step, rather than for every step, so we don't have access to it here. I'm storing that in the session with the journeyId (to allow us to have multiple journey with multiple tokens), but then that's not accessible if we don't have a valid journeyId

JasminConterioSW: Also the bad token page is currently a step in the journey

JasminConterioSW: After some discussion, we're going with:

If we can't initialise a journey

Redirect the user to the start page (with no token)

_Currently_
When the reach the start page, it will complain about the missing token, throw bad request and show them a generic error page telling them to try again later

_In a new ticket [PDJB-1037](https://mhclgdigital.atlassian.net/browse/PDJB-1037)_
We can catch the missing token error and show them an error page telling them to use their invitation link again

src/main/resources/application-local.yml
JasminConterioSW: I don't think we have a proper mocked version, if I miss this out it complains at me if I try to go to an EPC page

image

I'll leave it for now so we have a working version

PR #1401, opened 2026-06-03
PDJB-NONE: add isExpired accessor for JL invitation

samyou-softwire:
to be used for [PDJB-300](https://mhclgdigital.atlassian.net/browse/PDJB-300) and other tickets dealing with expired invitations

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitation.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitation.kt
@@ -7,6 +7,11 @@ import jakarta.persistence.GenerationType
 import jakarta.persistence.Id
 import jakarta.persistence.JoinColumn
 import jakarta.persistence.ManyToOne
+import kotlinx.datetime.DatePeriod
+import kotlinx.datetime.plus
+import kotlinx.datetime.toKotlinInstant
+import uk.gov.communities.prsdb.webapp.constants.JOINT_LANDLORD_INVITATION_LIFETIME_IN_DAYS
+import uk.gov.communities.prsdb.webapp.helpers.DateTimeHelper
 import java.util.UUID
 
 @Entity
@@ -44,6 +49,18 @@ class JointLandlordInvitation(
         this.registeredOwnership = registeredPropertyId
         this.invitingLandlord = invitingLandlord
     }
+
+    val isExpired: Boolean
+        get() {
+            val dateTimeHelper = DateTimeHelper()
+
+            val expiresOnDate =
+                DateTimeHelper
+                    .getDateInUK(createdDate.toKotlinInstant())
+                    .plus(DatePeriod(days = JOINT_LANDLORD_INVITATION_LIFETIME_IN_DAYS))
+
+            return dateTimeHelper.getCurrentDateInUK() > expiresOnDate
+        }
 
     constructor(
         id: Long,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitation.kt
```diff
@@ -45,6 +50,18 @@ class JointLandlordInvitation(
...
+    val isExpired: Boolean
+        get() {
+            val dateTimeHelper = DateTimeHelper()
+
+            val expiresOnDate =
+                DateTimeHelper
+                    .getDateInUK(createdDate.toKotlinInstant())
+                    .plus(DatePeriod(days = JOINT_LANDLORD_INVITATION_LIFETIME_IN_DAYS))
+
+            return dateTimeHelper.getCurrentDateInUK() > expiresOnDate
```
TomHanmer29: Just to check, is it expired ON the expiresOnDate or after the expiresOnDate? i.e. does it make sense to call this expiresOnDate and then use a >

TomHanmer29 (approved):
Just want to clarify one bit of logic/naming but otherwise seems sensible!

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitation.kt
samyou-softwire: good point, I'll double check what we want here with product and update accordingly

TomHanmer29 (changes requested):
Just read Rowan's comments on https://github.com/communitiesuk/prsdb-webapp/pull/1402 and agree we should probably add some unit tests with edge cases for this :)

samyou-softwire:
agree - product has clarified exactly how they want expiration to work so I'll get on this

samyou-softwire:
added some tests @TomHanmer29, based on this from product (if the invitation expiry date was 06/06)

> "I think it should expire at midnight on 06/06 so that’d be the last day it’s valid"

src/test/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitationTests.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import org.junit.jupiter.api.Assertions.assertFalse
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Test
+import uk.gov.communities.prsdb.webapp.constants.JOINT_LANDLORD_INVITATION_LIFETIME_IN_DAYS
+import uk.gov.communities.prsdb.webapp.testHelpers.mockObjects.MockJointLandlordData
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+
+class JointLandlordInvitationTests {
```
TomHanmer29, line 11: Worth having a test specifically at midnight to ensure this is behaving as expected? Or is that overkill?

TomHanmer29 (approved):
Tiny comment on tests, but happy that this aligns with the product decision.

src/test/kotlin/uk/gov/communities/prsdb/webapp/database/entity/JointLandlordInvitationTests.kt
samyou-softwire: this would be a bit tricky to implement yeah, since we'd have to mock out the current time to test it expiring over the time boundary. I'll leave as it

PR #1403, opened 2026-06-03
PDJB-NONE: Make worktree creation optional

samyou-softwire:
instructs AI to ask before creating the worktree if you want to use an existing worktree

adds a script to keep .env files etc synced across arbitrary worktrees

### DIFF
--- a/.github/skills/development-workflow/phases/phase-1-setup.md
+++ b/.github/skills/development-workflow/phases/phase-1-setup.md
@@ -11,6 +11,6 @@
      Change the working directory to the worktree.
    - **Current workspace** — stay in the current working directory. Create or
      switch to the branch using `git checkout -b <branch>` (or
-     `git checkout <branch>` if it already exists). If gitignored config
-     files are missing (e.g. `.env`), offer to copy them from a sibling
-     worktree using the `copy-config-files` script.
+     `git checkout <branch>` if it already exists). If the worktree is
+     not the default one, offer to copy gitignored config files from a
+     sibling worktree using the `copy-config-files` script.
--- a/.github/skills/development-workflow/phases/phase-1-setup.md
+++ b/.github/skills/development-workflow/phases/phase-1-setup.md
@@ -4,7 +4,13 @@
    If there is no ticket, use `PDJB-NONE`.
 2. Use the `branch-and-commit-naming` skill to determine the branch name for
    the first PR.
-3. Use the `using-git-worktrees` skill to create a worktree for the branch.
-4. Launch IntelliJ in the worktree folder using the command recorded during
-   preflight (e.g. `idea.cmd <worktree-path>`).
-5. Change the working directory to the worktree.
+3. Ask the user whether to create a new worktree or use the current workspace:
+   - **New worktree** — use the `using-git-worktrees` skill to create a
+     worktree for the branch. Launch IntelliJ in the worktree folder using
+     the command recorded during preflight (e.g. `idea.cmd <worktree-path>`).
+     Change the working directory to the worktree.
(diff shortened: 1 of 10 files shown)


.github/skills/development-workflow/phases/phase-1-setup.md
```diff
@@ -4,7 +4,13 @@
...
-3. Use the `using-git-worktrees` skill to create a worktree for the branch.
-4. Launch IntelliJ in the worktree folder using the command recorded during
-   preflight (e.g. `idea.cmd <worktree-path>`).
-5. Change the working directory to the worktree.
+3. Ask the user whether to create a new worktree or use the current workspace:
+   - **New worktree** — use the `using-git-worktrees` skill to create a
+     worktree for the branch. Launch IntelliJ in the worktree folder using
+     the command recorded during preflight (e.g. `idea.cmd <worktree-path>`).
+     Change the working directory to the worktree.
+   - **Current workspace** — stay in the current working directory. Create or
```
Travis-Softwire, line 12: I think this should be asking if you're not on the default worktree, offer to synchronize any gitgnored files to the latest version from the default worktree using `copy-config-files`, or similar - as the most likely scenario if you're using long-lived worktrees is that anything gitignored will have gone stale

scripts/git-worktrees/copy-config-files.ps1
```diff
@@ -0,0 +1,115 @@
...
+                    $skip = $true
+                    break
+                }
+            }
+            if ($skip) { continue }
+
+            $source = Join-Path $SourcePath $file
+            $dest = Join-Path $DestinationPath $file
+
+            # Skip if destination already has the file
```
Travis-Softwire: As above - I don't think this is the behaviour you want, I think you want to overwrite with the version from the source worktree to keep everything up to date

Travis-Softwire (changes requested):
I've left some thoughts RE syncing vs. just copying missing files, see what you think

.github/skills/development-workflow/phases/phase-1-setup.md
samyou-softwire: makes sense, can have it offer. not sure how often I'd actually want to do this; my .env for the workspace already has the ports set up etc so if it were overwritten then it'd break

scripts/git-worktrees/copy-config-files.ps1
samyou-softwire: good catch thanks, have removed this check

TomHanmer29:
Ignore the above mention, typed the wrong ticket 😆

.github/skills/development-workflow/phases/phase-1-setup.md
Travis-Softwire: Oh, good point - the copy script should also run the port setup part of the worktree creation script to make sure we maintain port segregation even if the long-lived worktrees change.

Travis-Softwire, line 12: Oh - we should also make sure the rest of the workflow skills can handle being in a long-lived worktree. I think that might just be Phase 7

Travis-Softwire (approved):
Two further small comments, but LGTM, I don't need to see it again :)

.github/skills/development-workflow/phases/phase-1-setup.md
samyou-softwire: have reviewed this and fixed some references

samyou-softwire: have updated the script to handle this

PR #1406, opened 2026-06-04
PDJB-306: Add marked joint landlord flag

samyou-softwire:
## Ticket number

[PDJB-306](https://mhclgdigital.atlassian.net/browse/PDJB-306) (and others in the future)

## Goal of change

adds a database column that tracks whether the user indicated this is a JL property

right now filled in from the outcome of this question on the registration journey

invite joint landlords question

this will be used later on for designs where the landlord needs to specifically state later on that the property has no joint landlords

adds database column

adds hooks into the journey flow to populate it

view of database

## For reviewer

nothing here except the population of the field is feature flagged. I think this is fine as the default 'false' will be true for any registrations made before we turn on JL

## Checklist

  mention that here
- [ ] QA instructions have been added to the ticket (particularly if this is the last PR required to complete the ticket) (nothing to QA yet, this is an internal change)


TomHanmer29 (approved):
Looks sensible to me! I think false by default is a good state for this to be in

PR #1407, opened 2026-06-04
PDJB-1032: Add metrics dashboard page with date range filtering

Bill-Haigh:
## Ticket number

PDJB-1032

## Goal of change

Add a system operator metrics dashboard page that can be filtered by a date range.

- Adds a `/system-operator/metrics` page, restricted to the `SYSTEM_OPERATOR` role, rendering metrics as a summary list
- Adds a date range filter form with custom prioritised validation for the from/to dates
- Adds a `ReportingPeriod` data model that converts a UK-local date range into an inclusive instant range, capping the end at "now" when the end date is today

This is the page skeleton — the metric rows are currently empty pending the data wiring in a follow-up.

## Screenshots

image

image

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/annotations/webAnnotations/MetricsDateRangeValidationAnnotations.kt
Bill-Haigh, line 1: this is a lot of duplication of `DateValidationAnnotations` but in the instance of the metrics there are two date fields in the form that need validation so we have this duplication. 

There is likely a way around this but given that this is relevant to the metrics only I am unsure if I leave as is or if I look for an alternative.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ReportingPeriod.kt
```diff
@@ -0,0 +1,41 @@
...
+        ): ReportingPeriod {
+            val now = clock.instant()
+            val todayInUk = now.atZone(UK_ZONE).toLocalDate()
+
+            val start = from.atStartOfDay(UK_ZONE).toInstant()
+            val end =
+                if (to == todayInUk) {
+                    now
+                } else {
+                    to.atTime(LocalTime.of(23, 59, 59)).atZone(UK_ZONE).toInstant()
```
Travis-Softwire: To avoid a 1 second gap I think we can use LocalTime.MAX here

Travis-Softwire (commented):
Will let Jasmin approve, but overall LGTM subject to one minor point (I couldn't immediately see a good way to consolidate the annotations so I'm fine with that duplication). Minor but the PR description claims this includes the summary list but atm it doesn't - which is fine, just a discrepency in the description

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ReportingPeriod.kt
```diff
@@ -0,0 +1,41 @@
...
+         * If the [to] date is the current date in the UK, the end is the current instant instead of
+         * end-of-day, so the period only covers up to "now".
+         */
+        fun fromDateRange(
+            from: LocalDate,
+            to: LocalDate,
+            clock: Clock = Clock.systemDefaultZone(),
+        ): ReportingPeriod {
+            val now = clock.instant()
+            val todayInUk = now.atZone(UK_ZONE).toLocalDate()
```
JasminConterioSW: You can use the existing `DateTimeHelper().getCurrentDateInUk()` for this one.

And maybe add something to `DateTimeHelper `to get the end of the day?

src/main/resources/templates/fragments/forms/dateInputScoped.html
```diff
@@ -0,0 +1,49 @@
+<div th:fragment="dateInputScoped(fieldName, dayField, monthField, yearField, fieldSetHeading, fieldSetHint)" class="govuk-form-group"
```
JasminConterioSW, line 1: Might be worth seeing if some of this can be commonised with `dateInput.html`

src/test/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ReportingPeriodTests.kt
```diff
@@ -0,0 +1,91 @@
...
+                to = LocalDate.of(2025, 3, 5),
+                clock = clock,
+            )
+
+        assertEquals(LocalDate.of(2025, 3, 5).atStartOfDay(ukZone).toInstant(), period.start)
+        assertEquals(ZonedDateTime.of(2025, 3, 5, 23, 59, 59, 0, ukZone).toInstant(), period.end)
+    }
+
+    @Test
+    fun `start is computed correctly across the spring DST boundary`() {
```
JasminConterioSW, line 77: Worth adding a test for the clocks going back as well? (I'm nitpicking and probably wouldn't have thought of it if there wasn't a test for the clocks going forward)

JasminConterioSW (approved):
A couple of suggestions where we might be able to re-use some of the existing code, but generally LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/ReportingPeriod.kt
Bill-Haigh: at the moment this is using `java.time` so I would have to refactor to use `kotlinx.datetime` which I can do if you think it's worthwhile

PR #1408, opened 2026-06-04
PDJB-302: Remove (hide) expired invites

samyou-softwire:
## Ticket number

[PDJB-302](https://mhclgdigital.atlassian.net/browse/PDJB-302)

## Goal of change

allow for expired invites to be removed from the expired invites list

invites list with removed flash visible

implemented this as a column on an invite - as I understand we should never permanently remove invites, for auditing reasons

adds a notification box based on a flash

adds a new status property to an invitation to make it clear how to use the flags and make it natural to exclude deleted invites

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
```diff
@@ -108,6 +110,19 @@ class PropertyDetailsController(
...
     }
 
+    @PreAuthorize("hasRole('LANDLORD')")
+    @GetMapping(REMOVE_EXPIRED_INVITE_ROUTE)
+    fun removeExpiredInvite(
+        @PathVariable propertyOwnershipId: Long,
+        @PathVariable invitationId: Long,
+        redirectAttributes: RedirectAttributes,
+    ): String {
+        val baseUserId = SecurityContextHolder.getContext().authentication.name
```
samyou-softwire, line 121: copilot wanted this a POST but I didn't really see the point of wrapping the link in a form to make that work

Travis-Softwire: Sorry, but I agree with Copilot here - although it is really an issue with the design (we've explained that state changes = button, and that GDS really doesn't like buttons styled as links...). As it stands though if a user clicks on that link twice in quick succession (before the first response has arrived) they'll get a 404 page, because browsers don't protect against accidental repeated GET requests. The security side of things is less important because hiding an already expired invite is low stakes - but just having the code smell of GETs that make state changes as a precedent risks others copying it in higher stakes cases where we want the protection of a CSRF token.

My suggestion would be that we merge as-is but we a) create a ticket to add a confirmation page, b) put a bit TODO & health warning here about it being a GET when it should be a POST, just in case that change never materialises, c) ask the designers for a confirmation page (and maybe suggest this should be a 'delete' or 'cancel' rather than a 'hide' action to make clear that it's destructive).

Travis-Softwire: Tbf have spotted this won't trigger a 404, we'll just set hidden = true again even though it was already true. The wider point still stands though

Travis-Softwire (approved):
See comment RE GET vs POST. As per that, I think we just need new ticket + TODO & health warning for now so am approving this subject to that, but do reach out if you disagree

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
samyou-softwire: I've raised [PDJB-1060](https://mhclgdigital.atlassian.net/browse/PDJB-1060) and added a TODO for it here

I wasn't sure where to place this ticket? I'd want to put it into the JL epic though I don't this is necessary for MVP

PR #1417, opened 2026-06-09
PDJB-1033: Add database performance metrics to the metrics dashboard

Bill-Haigh:
## Ticket number

PDJB-1033

## Goal of change

Populate the system-operator metrics dashboard with four database-performance metrics computed over the user-selected reporting period.

- Adds a `MetricsService` that computes four metrics for a `ReportingPeriod`:
  - **Landlord registrations** — landlords created within the period.
  - **Properties** — property ownerships created within the period.
  - **Landlords with a property** — distinct landlords who had at least one property by the **end** of the period.
  - **Average time to first property** — mean gap between a landlord's registration and their first-ever property, for landlords whose registration and first property both fall in the period (`No data` when none qualify).
- Wires the service into `MetricsController` so the four rows render on a valid date-range submission (and no rows on an invalid submission).
- Adds the supporting repository query methods and message keys for the row labels and the humanised average (days/hours/minutes).

- Metrics 3 and 4 count **all** landlords on a property (joint landlords included) via `po.landlords` — `primaryLandlord` is no longer a persistent attribute after the recent `landlords` many-to-many refactor, so it can't be used in JPQL. The current seed has no joint landlords, so values are unaffected.
- The integration test asserts exact values (33 / 36 / 3 / 124 days) derived from `data-local.sql`.

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
@@ -48,7 +48,6 @@ interface PropertyOwnershipRepository :
     @Query(
         "SELECT l.createdDate, MIN(po.createdDate) FROM PropertyOwnership po " +
             "JOIN po.landlords l " +
-            "WHERE l.createdDate BETWEEN :start AND :end " +
             "GROUP BY l.id, l.createdDate " +
             "HAVING MIN(po.createdDate) BETWEEN :start AND :end",
     )
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
@@ -1,7 +1,10 @@
 package uk.gov.communities.prsdb.webapp.database.repository
 
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Query
+import org.springframework.data.repository.query.Param
 import uk.gov.communities.prsdb.webapp.database.entity.PropertyOwnership
+import java.time.Instant
 
 // The underscore tells JPA to access fields relating to the referenced table
 @Suppress("ktlint:standard:function-naming")
@@ -26,4 +29,30 @@ interface PropertyOwnershipRepository :
     ): Boolean
 
     fun countByLandlords_BaseUser_Id(userId: String): Long
+
+    fun countByCreatedDateBetween(
+        start: Instant,
+        end: Instant,
+    ): Long
+
+    @Query(
+        "SELECT COUNT(DISTINCT l.id) FROM PropertyOwnership po " +
+            "JOIN po.landlords l " +
+            "WHERE po.createdDate <= :end",
+    )
+    fun countDistinctLandlordsWithPropertyCreatedOnOrBefore(
+        @Param("end") end: Instant,
+    ): Long
+
+    @Query(
+        "SELECT l.createdDate, MIN(po.createdDate) FROM PropertyOwnership po " +
+            "JOIN po.landlords l " +
+            "WHERE l.createdDate BETWEEN :start AND :end " +
+            "GROUP BY l.id, l.createdDate " +
+            "HAVING MIN(po.createdDate) BETWEEN :start AND :end",
+    )
+    fun findLandlordAndFirstPropertyCreatedDates(
+        @Param("start") start: Instant,
+        @Param("end") end: Instant,
+    ): List<Array<Instant>>
 }
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
@@ -38,9 +38,10 @@ interface PropertyOwnershipRepository :
     @Query(
         "SELECT COUNT(DISTINCT l.id) FROM PropertyOwnership po " +
             "JOIN po.landlords l " +
-            "WHERE po.createdDate <= :end",
+            "WHERE po.createdDate BETWEEN :start AND :end",
     )
-    fun countDistinctLandlordsWithPropertyCreatedOnOrBefore(
+    fun countDistinctLandlordsWithPropertyCreatedBetween(
+        @Param("start") start: Instant,
         @Param("end") end: Instant,
     ): Long
 
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/MetricsDataModel.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/MetricsDataModel.kt
@@ -7,5 +7,7 @@ data class MetricsDataModel(
     val numberOfVerifiedLandlords: Long,
     val numberOfProperties: Long,
     val numberOfLandlordsWithAProperty: Long,
-    val averageTimeToFirstProperty: Duration?,
+    val medianTimeToFirstProperty: Duration?,
+    val p90TimeToFirstProperty: Duration?,
+    val p95TimeToFirstProperty: Duration?,
 )
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/MetricsDataModel.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/MetricsDataModel.kt
@@ -4,6 +4,7 @@ import java.time.Duration
 
 data class MetricsDataModel(
     val numberOfLandlordRegistrations: Long,
+    val numberOfVerifiedLandlords: Long,
     val numberOfProperties: Long,
     val numberOfLandlordsWithAProperty: Long,
     val averageTimeToFirstProperty: Duration?,
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/MetricsDataModel.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/MetricsDataModel.kt
@@ -1,0 +1,10 @@
+package uk.gov.communities.prsdb.webapp.models.dataModels
+
+import java.time.Duration
+
+data class MetricsDataModel(
+    val numberOfLandlordRegistrations: Long,
+    val numberOfProperties: Long,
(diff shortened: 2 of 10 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/MetricsDataModel.kt
```diff
@@ -0,0 +1,10 @@
+package uk.gov.communities.prsdb.webapp.models.dataModels
+
+import java.time.Duration
+
+data class MetricsDataModel(
+    val numberOfLandlordRegistrations: Long,
+    val numberOfProperties: Long,
+    val numberOfLandlordsWithAProperty: Long,
+    val averageTimeToFirstProperty: Duration?,
```
JasminConterioSW: I think Lewis also wants to know how many landlords were verified by one-login

(Basically analytics requirements are a bit in flux...)

JasminConterioSW: For this one, last time we reported p50, p90 and p95.
I think out of those p50 (median) is probably the most useful as the other numbers get into days - but they might want one of those to get an idea?

You can query those p numbers straight from the db, there's an example in the file below.

Queries we used before are in the codebase ([this ](https://github.com/communitiesuk/prsdb-webapp/blob/chore/PDJB-NONE-kpi-database-queries/src/test/resources/databaseQueriesForReports/privateBetaInitialKpiQueries.sql)is what I used last week to get numbers, although it'll be outdated with the database changes)

We'll want median rather than mean though as mean will be too affected by people coming back another day

src/main/resources/messages/metrics.yml
```diff
@@ -12,3 +12,16 @@ refreshButton: Refresh
 dateRange:
   error:
     toBeforeFrom: The ‘to’ date must be the same as or after the ‘from’ date
+rows:
+  landlordRegistrations: Number of registrations (landlords)
+  properties: Number of properties
+  landlordsWithProperty: Number of landlords with at least 1 property
```
JasminConterioSW, line 19: We might need to make clear which metrics only take into account the "To" date (unless we make sure everything takes that into account?)

Otherwise this looks a little odd

Image

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/MetricsControllerTests.kt
```diff
@@ -125,4 +144,56 @@ class MetricsControllerTests(
                 }
             }
     }
+
+    @Test
+    @WithMockUser(roles = ["SYSTEM_OPERATOR"])
+    fun `submitMetrics populates four metric rows for a valid date range`() {
```
JasminConterioSW: We're testing more here in the controller than we usually do, but I think that's ok (we're doing some stuff in the controller which often fits into a journey framework so maybe that's not surprising!)

(Not anything that needs an action, just a comment that it's a bit different from other controller tests)

JasminConterioSW (changes requested):
Looks pretty good!

The time between landlord and first property registration would I think be better as a median (it can be queried direct from the database).
Also might be worth adding a line on how many verified landlords we have (Lewis was asking about this last week so we might as well add it)

There's a thread on private beta reporting [here](https://teams.microsoft.com/l/message/19:FYwH4lYraM0QKE33xG2XZMOlLqezbT6lOevTpztA5Vo1@thread.skype/1779794154894?tenantId=bf346810-9c7d-43de-a872-24a2ef3995a8&groupId=b8b6dd1f-a94c-4cb4-8ca7-4772310f9ac6&parentMessageId=1779794154894&teamName=Private%20Rented%20Sector&channelName=PRS%20Database%20Service%20Team%20(Digital%20and%20Operations)&ngc=true&allowXTenantAccess=true) which might be worth keeping an eye on

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/MetricsService.kt
```diff
@@ -0,0 +1,40 @@
...
+            numberOfLandlordRegistrations =
+                landlordRepository.countByCreatedDateBetween(period.start, period.end),
+            numberOfProperties =
+                propertyOwnershipRepository.countByCreatedDateBetween(period.start, period.end),
+            numberOfLandlordsWithAProperty =
+                propertyOwnershipRepository.countDistinctLandlordsWithPropertyCreatedOnOrBefore(period.end),
+            averageTimeToFirstProperty = getAverageTimeToFirstProperty(period),
+        )
+
+    private fun getAverageTimeToFirstProperty(period: ReportingPeriod): Duration? {
```
Travis-Softwire: Will this break once we have JLs? JLs could register __after__ the property was registered...

Travis-Softwire (commented):
One additional Q in addition to Jasmin's but will let her approve

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/MetricsService.kt
JasminConterioSW: Ooh that's a point.

Once we have joint landlords we no longer know who did the property registration do we? Will this metric become a bit meaningless?

Bill-Haigh: Good catch - are we / should we be tracking who registers the property? If so then this metric would still be helpful.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/MetricsDataModel.kt
Bill-Haigh: added this

Bill-Haigh: have added all three medians for now - can always reduce them if they decide they don't want all of them

src/main/resources/messages/metrics.yml
Bill-Haigh: have bound all of the metrics to the to and from dates for simplicity

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/MetricsService.kt
JasminConterioSW: It sounds like they will want this metric but we will need to change how we calculate it when joint landlords is on.

We don't currently track who registers a property, Alex is looking into auditing to see which landlord was linked to the property ownership first.

I think leave this in for now (you could modify the text to say it doesn't account for joint landlords) but add another ticket to update this metric when we know how we want to track joint landlords.

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyOwnershipRepository.kt
```diff
@@ -26,4 +29,31 @@ interface PropertyOwnershipRepository :
...
+        @Param("start") start: Instant,
+        @Param("end") end: Instant,
+    ): Long
+
+    @Query(
+        "SELECT l.createdDate, MIN(po.createdDate) FROM PropertyOwnership po " +
+            "JOIN po.landlords l " +
+            "WHERE l.createdDate BETWEEN :start AND :end " +
+            "GROUP BY l.id, l.createdDate " +
+            "HAVING MIN(po.createdDate) BETWEEN :start AND :end",
```
JasminConterioSW, line 52: Hmm this will only count when both the landlord and property ownerships were created in the date range...

That could skew the result particularly for something like p90 / p95 as it could exclude cases where the landlord took a long time to register their first property.

Does it work if we remove the filter on the landlord created date, so we are looking at properties registered in the date range, and measuring how long it took the landlord to register those (if they are the landlord's first property)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/MetricsService.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    private fun getTimesToFirstProperty(period: ReportingPeriod): List<Duration> =
+        propertyOwnershipRepository
+            .findLandlordAndFirstPropertyCreatedDates(period.start, period.end)
+            .map { (landlordCreatedDate, firstPropertyCreatedDate) ->
+                Duration.between(landlordCreatedDate, firstPropertyCreatedDate)
+            }
+
+    private fun percentile(
```
JasminConterioSW, line 44: Is there an in-built function for calculating this? It seems surprising to need to write our own function for it.

When getting the metrics before, I've queried percentile_cont(0.5) direct from the database. Then again this sometimes gave a weirdly rendered result like 1 day 28 hours 68 minutes so maybe your version is better!

If it's awkward this is probably fine, especially as this is one where we might need to change the database query for joint landlords anyway

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/MetricsServiceTests.kt
```diff
@@ -0,0 +1,115 @@
...
+        val metrics = metricsService.getMetrics(period)
+
+        // rank = fraction * (n - 1) = fraction * 10
+        assertEquals(Duration.ofDays(60), metrics.medianTimeToFirstProperty) // rank 5 -> 60
+        assertEquals(Duration.ofDays(100), metrics.p90TimeToFirstProperty) // rank 9 -> 100
+        assertEquals(Duration.ofDays(105), metrics.p95TimeToFirstProperty) // rank 9.5 -> midway 100..110
+    }
+
+    @Test
+    fun `getMetrics computes sub-day percentiles`() {
```
JasminConterioSW: Minor but can we make this check with some more realistic / less round numbers?

Our real data for the first couple of weeks showed

* median - 27 minutes
* p90 - 1 day 4 hours 2 min
* p95 - 2 days 12 hours 38 min

We don't need to it to be that, but maybe something with a combination of minutes / hours / days

JasminConterioSW (commented):
Can you take another look at the filtering for findLandlordAndFirstPropertyCreatedDates? Thinking about it filtering by created date on both landlord and property ownership might skew the results...

There are a couple of other minor comments to check but mostly good!

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/MetricsService.kt
Bill-Haigh: In essence there isn't a particularly clean way of doing this with percentile_cont without having to convert to seconds and then back again, I think i'll leave it as is for now as it may well change anyway

JasminConterioSW (approved):
LGTM now, thanks :)

PR #1420, opened 2026-06-09
PDJB-304: Standalone joint landlord journey fixes

TomHanmer29:
## Ticket number

PDJB-304

## Goal of change

image
Fix some validation rules and notify sending on the standalone JL flow.

Adds validation for each of the sent invites so you can't send another to a landlord you've invited already.
Adds validation for existing JLs on the property record so you can't send another to a landlord that's already on the record.
Also fixes back button behaviour using a new startStep, similar to how we did it for EPC

Worth giving this a check to ensure we actually want this logic

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/inviteJointLandlord/InviteJointLandlordJourneyFactory.kt
```diff
@@ -49,8 +55,27 @@ class InviteJointLandlordJourneyFactory(
...
+                routeSegment(StartInviteJointLandlordStep.ROUTE_SEGMENT)
                 initialStep()
+                nextStep { journey.hasAnyJointLandlordsInvitedStep }
+            }
+            step(journey.hasAnyJointLandlordsInvitedStep) {
+                parents { journey.startInviteJointLandlordStep.isComplete() }
+                nextStep { mode ->
+                    when (mode) {
+                        AnyLandlordsInvited.NO_LANDLORDS -> journey.inviteJointLandlordsTask.firstStep
+                        AnyLandlordsInvited.SOME_LANDLORDS -> journey.checkJointLandlordsStep
```
JasminConterioSW: Is this step reachable if the user hasn't completed `inviteJointLandlordStep`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/InviteJointLandlordPropertyRegistrationState.kt
```diff
@@ -9,6 +9,9 @@ interface InviteJointLandlordPropertyRegistrationState : SharedInviteJointLandlo
     val hasAnyJointLandlordsInvitedStep: HasAnyJointLandlordsInvitedStep
     val hasJointLandlordsStep: HasJointLandlordsStep
     val inviteJointLandlordsTask: SharedInviteJointLandlordsTask
```
JasminConterioSW: Minor and it's not changed in this PR, but I'd probably just call it `InviteJointLandlordsTasks` without the `Shared` as tasks are inherently shareable. Same with the state.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/inviteJointLandlord/InviteJointLandlordJourneyFactory.kt
JasminConterioSW: I think we don't really want to navigate to the middle of a task, it might be better to either split the task up or add more into it to make it more standalone...

Here I'd probably add more into it and include a version of the `hasAnyJointLandlordsInvitedStep` in the task

JasminConterioSW (changes requested):
I'm not sure about navigating into the middle of a task as a pattern (a step in the middle of a task might not be reachable), can you move things around a bit so the task is more standalone?

It should be a bit of a black box - the journey gives it inputs and take the outputs but doesn't know what's going on in the middle

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/inviteJointLandlord/InviteJointLandlordJourneyFactory.kt
TomHanmer29: It's a bit of a tricky one - the issue is we want to share some stuff between the current task in property reg and the new task that is standalone, the problem is that the standalone one doesn't have the question that asks you if you have joint landlords, so it makes it hard to just add the hasAnyJointLandlordsInvitedStep to the task. It does work currently but I agree it's an antipattern so I'll try and remove it

TomHanmer29: I've separated the tasks back out, but kept the state shared. I can't think of a good way to commonise them if we need the checkJointLandlordsStep to be an entry point 🤔 The only thing I hadn't tried is finding a way to conditionally make it so the hasJointLandlordsStep only shows in property registration, but this feels a bit janky

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/InviteJointLandlordPropertyRegistrationState.kt
```diff
@@ -9,6 +9,9 @@ interface InviteJointLandlordPropertyRegistrationState : SharedInviteJointLandlo
...
+
+    override val checkJointLandlordsBackUrl: String?
+        get() = hasAnyJointLandlordsInvitedStep.backUrl
```
Travis-Softwire: Is this level of indirection required? What's the actual back logic for the step? Can we not just write it straight into the body of `backUrl {...}`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/inviteJointLandlord/InviteJointLandlordJourneyFactory.kt
Travis-Softwire: Don't we already have this situation for the 'add another' pattern for file uploads? I think there the first step in the task is an internal step that checks which step to route to based on whether any <x> have been added yet

Travis-Softwire (changes requested):
It feels like we're making some of the routing a bit overcomplicated here when we should (I think) have patterns we can work from. Happy to chat about it if you disagree or if I've missed something though

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/InviteJointLandlordPropertyRegistrationState.kt
TomHanmer29: Nope, removed this now :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/inviteJointLandlord/InviteJointLandlordJourneyFactory.kt
TomHanmer29: All sorted now, we've gone with having the hasAnyJointLandlordsInvitedStep both inside the task (to choose between inviteJointLandlordsStep and checkJointLandlordsStep, and also reused in property reg to decide whether to show the hasJointLandlordsStep

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationService.kt
```diff
@@ -41,6 +41,12 @@ class JointLandlordInvitationService(
         return Pair(pending, expired)
     }
 
+    fun getExistingInvitedEmails(ownershipId: Long): List<String> =
+        invitationRepository
+            .findByRegisteredOwnershipId(ownershipId)
+            .filter { it.status == JointLandlordInvitationStatus.PENDING || it.status == JointLandlordInvitationStatus.EXPIRED }
```
JasminConterioSW: I don't know what we're going to end up with with status's, but should this just be `it.status != JointLandlordInvitationStatus.HIDDEN`?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/InviteJointLandlordJourneyTests.kt
```diff
@@ -67,4 +67,18 @@ class InviteJointLandlordJourneyTests : IntegrationTestWithMutableData("data-loc
         assertThat(confirmationPage.goBackToPropertyRecordLink).containsText("Go back to the property record")
         assertThat(page.locator("main")).containsText("The joint landlords you’ve invited have 28 days to join the property")
     }
+
+    @Test
+    fun `Submitting an email of an existing landlord on the property shows an error`(page: Page) {
+        val firstStepUrl =
```
JasminConterioSW: This bit should be a navigator method

JasminConterioSW (approved):
A couple of minor points but other than that LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationService.kt
TomHanmer29: Yep, I think this sounds sensible :) I'll update

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/tasks/JointLandlordsPropertyRegistrationTask.kt
```diff
@@ -13,14 +13,14 @@ import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.HasJo
...
-class JointLandlordsTask : Task<InviteJointLandlordPropertyRegistrationState>() {
+class JointLandlordsPropertyRegistrationTask : Task<InviteJointLandlordPropertyRegistrationState>() {
     override fun makeSubJourney(state: InviteJointLandlordPropertyRegistrationState) =
         subJourney(state) {
             taskStatus {
                 when {
                     exitStep.isStepReachable -> TaskStatus.COMPLETED
                     journey.hasJointLandlordsStep.outcome != null -> TaskStatus.IN_PROGRESS
-                    journey.checkJointLandlordsStep.outcome != null -> TaskStatus.IN_PROGRESS
+                    journey.inviteJointLandlordsTask.exitStep.isStepReachable -> TaskStatus.IN_PROGRESS
```
Travis-Softwire: What's the scenario where we need this one but neither the case above or below would catch it?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/inviteJointLandlord/InviteJointLandlordJourneyFactory.kt
```diff
@@ -49,8 +52,13 @@ class InviteJointLandlordJourneyFactory(
             configure {
                 withAdditionalContentProperty { "title" to "inviteJointLandlord.title" }
             }
-            task(journey.inviteJointLandlordsTask) {
+            step(journey.startInviteJointLandlordStep) {
```
Travis-Softwire, line 56: Can we add a TODO here to switch to addressable tasks if/when they're implemented? Referencing PDJB-896

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/InviteJointLandlordJourneyTests.kt
```diff
@@ -34,10 +34,10 @@ class InviteJointLandlordJourneyTests : IntegrationTestWithMutableData("data-loc
 
     @Test
     fun `Landlord can complete the standalone invite joint landlord journey`(page: Page) {
-        val firstStepUrl =
-            "http://localhost:$port${InviteJointLandlordController.getInviteJointLandlordRoute(propertyOwnershipId)}/" +
-                InviteJointLandlordStep.INVITE_FIRST_ROUTE_SEGMENT
-        page.navigate(firstStepUrl)
+        navigator.navigate(
```
Travis-Softwire: Should be a navigator.goToX function

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/InviteJointLandlordJourneyTests.kt
```diff
@@ -67,4 +67,18 @@ class InviteJointLandlordJourneyTests : IntegrationTestWithMutableData("data-loc
         assertThat(confirmationPage.goBackToPropertyRecordLink).containsText("Go back to the property record")
         assertThat(page.locator("main")).containsText("The joint landlords you’ve invited have 28 days to join the property")
     }
+
+    @Test
+    fun `Submitting an email of an existing landlord on the property shows an error`(page: Page) {
+        navigator.navigate(
```
Travis-Softwire: Same here

src/test/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/InviteJointLandlordsFormModelTests.kt
```diff
@@ -62,4 +62,26 @@ class InviteJointLandlordsFormModelTests {
 
         assertTrue(formModel.isEmailNotAlreadyInvited())
     }
+
+    @Test
+    fun `a user cannot invite an email that is already a joint landlord on the property`() {
```
Travis-Softwire, line 67: Is a specific requirement?

Travis-Softwire (approved):
Happy w/ the new approach. A few minor Qs but overall LGTM once those are resolved

src/test/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/InviteJointLandlordsFormModelTests.kt
TomHanmer29: Yes, this is a specific requirement in the designs.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/tasks/JointLandlordsPropertyRegistrationTask.kt
TomHanmer29: You're right, copilot thought it was something to do with the state if you add landlords and then remove them all but it had not gone through the journey properly. I can remove it.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/InviteJointLandlordJourneyTests.kt
TomHanmer29: Whoops, messed this up the first time 🙃

PR #1425, opened 2026-06-09
PDJB-794: fix QA points

TomHanmer29:
## Ticket number

PDJB-794

## Goal of change

QA fixes for the EPC tab

image
image

Fixes some copy on the expired EPC page
Prevents tenancy started row from showing when a property becomes unoccupied

I've not included logic changes here (as to flipping "tenancy started before expiry to null etc.) as I feel this is a legal / product decision. Let me know if you disagree @Travis-Softwire / @AEPR 

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
@@ -263,6 +263,9 @@ class PropertyOwnershipService(
         if (!wasOccupied && propertyOwnership.isOccupied) {
             propertyOwnership.lastOccupiedDate = LocalDate.now()
         }
+        if (wasOccupied && !propertyOwnership.isOccupied) {
+            propertyOwnership.propertyCompliance?.tenancyStartedBeforeEpcExpiry = null
+        }
         propertyOwnershipRepository.save(propertyOwnership)
     }
 
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
@@ -263,7 +263,7 @@ class PropertyOwnershipService(
         if (!wasOccupied && propertyOwnership.isOccupied) {
             propertyOwnership.lastOccupiedDate = LocalDate.now()
         }
-        if (wasOccupied && !propertyOwnership.isOccupied) {
+        if (!propertyOwnership.isOccupied) {
             propertyOwnership.propertyCompliance?.tenancyStartedBeforeEpcExpiry = null
         }
         propertyOwnershipRepository.save(propertyOwnership)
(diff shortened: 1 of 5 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/EpcViewModelFactoryTests.kt
```diff
@@ -152,6 +152,15 @@ class EpcViewModelFactoryTests {
             PropertyComplianceBuilder.createWithInDateCertsAndLowEpcRating(
                 propertyIsOccupied = true,
             )
+        private val expiredWithTenancyBeforeExpiryUnoccupied =
```
JasminConterioSW, line 155: So this comes up when

* The property starts as occupied
* They have an expired EPC but the EPC was in date when the tenancy began
* They later changed the status to unoccupied

Is that right?

JasminConterioSW (commented):
I think we should reset tenancy started before expiry to null if a property becomes unoccupied.

We could get into a bit of a tangle if the property becomes occupied again and we haven't made them re-answer that question - it'll show up as "true" still even though it isn't.

src/test/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/EpcViewModelFactoryTests.kt
TomHanmer29: yes

TomHanmer29:
> I think we should reset tenancy started before expiry to null if a property becomes unoccupied.
> 
> We could get into a bit of a tangle if the property becomes occupied again and we haven't made them re-answer that question - it'll show up as "true" still even though it isn't.

Yes, I do agree with this in principle. We had a big discussion about this when originally doing the ticket around automatically updating the state of this value when the occupancy state changes but I think decided that we shouldn't assume anything unless we go through product/legal with that decision - but I do agree it could lead to incorrect data being shown. @Travis-Softwire any more thoughts / updates on this one?

Travis-Softwire:
I think we're fine to reset this one. There are some that we shouldn't reset (e.g. number of households) but in this case we know for sure that this answer will be invalid when the property is next marked as occupied unless it was a mistake (in which case they'll need to complete it again anyway)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -263,6 +263,9 @@ class PropertyOwnershipService(
         if (!wasOccupied && propertyOwnership.isOccupied) {
             propertyOwnership.lastOccupiedDate = LocalDate.now()
         }
+        if (wasOccupied && !propertyOwnership.isOccupied) {
```
Travis-Softwire: Minor, but do we need the `wasOccupied` element of this check?

Travis-Softwire (approved):
One small point, but LGTM 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
TomHanmer29: Yes, that's true - we should just never be saving this when the property is unoccupied, I'll remove the wasOccupied condition

PR #1426, opened 2026-06-10
Feat/pdjb 432 extend joint landlord

AEPR:
## Ticket number
PDJB-432

## Goal of change
Updates the NFT seeder to work with the current database and prevents orphaning of ownerships.

The NFT seeder just creates landlordship meberships for each ownership a landlord would have previously had. This means there are no joint landlords being generated currently.
The application code now check when deleting a landlord how many other landlords are attached to their landlordships. Where it is 0 it is deleted, where it is >0 that landlord is removed.

## Checklist

- [ ] TODO comments referencing this JIRA ticket have been searched for and removed - currently we have no email templated for handling deleting landlords with multiple properties some of which have other landlords. I've left a TODO with a 432 reference, but will update to another JIRA number at some point.
- [X] `NftDataSeeder` has been updated to reflect any changes to the database schema


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Landlord.kt
```diff
@@ -70,6 +70,8 @@ class Landlord() : ModifiableAuditableEntity() {
     @ManyToMany(mappedBy = "landlords")
     private lateinit var propertyOwnerships: MutableSet<PropertyOwnership>
 
+    val landlordships: Set<PropertyOwnership> get() = propertyOwnerships
```
samyou-softwire, line 73: what is this for?

AEPR: The type it returns is a non-mutable set because the relationship is "owned" by the PropertyOwnership entity. If you modify the mutableSet "propertyOwnership" it doesn't change the database which is why I made the underlying collection private.

Travis-Softwire (approved):
LGTM (and the LL deregistration logic is actually included in a later ticket, but makes sense to do here!)

PR #1427, opened 2026-06-10
PDJB-NONE: Add some more invitations to local seed data

samyou-softwire:
under 1 PRSDB Square, EG1 2AA

view of pending invitations


AEPR (approved):
Seems very sensible.

PR #1430, opened 2026-06-10
PDJB-432: Fix joint landlord view property permissions

samyou-softwire:
## Ticket number

[PDJB-432](https://mhclgdigital.atlassian.net/browse/PDJB-432)

## Goal of change

fixes it so landlords can view and edit any properties they're associated with

updates the relevant methods in PropertyOwnershipService

this means that we can test flows properly on integration that require another landlord to be added to the property. right now you can only test this if you are the lowest ID landlord as that's the primaryLandlord fallback

## Checklist


Travis-Softwire:
Can you flag on PDJB-1069 that this has been actioned here? Ta

Travis-Softwire (approved):

PR #1431, opened 2026-06-10
PDJB-NONE: Add message instructions

samyou-softwire:
copilot struggles a lot with adding new messages, it keeps setting the placeholder to what it sees in the design. make it explicit to not do that


Bill-Haigh (approved):
LGTM

PR #1433, opened 2026-06-11
Feat/pdjb 1056 add landlord to property ownership

JasminConterioSW:
## Ticket number

PDJB-1056

## Goal of change

Add the joint landlord to the property ownership

* Modified the landlord details shown on the confirmYouAreALandlord... step so this is the landlord who just completed landlord registration (saved in an internal step after completing the registration)
* Get the details of the logged in landlord and save these to the property ownership record
    * (Note those two sets of landlord details should be the same unless the user logged out and then logged in again as a different landlord user)

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyOwnership.kt
```diff
@@ -150,6 +150,10 @@ class PropertyOwnership() : ModifiableAuditableEntity() {
         this.markedJointLandlord = markedJointLandlord
     }
 
+    fun addLandlord(landlord: Landlord) {
```
Travis-Softwire: In two minds whether we need this vs just `propertyOwnership.landlords.add(landlord)`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/steps/ConfirmYouAreALandlordForThisPropertyStepConfig.kt
```diff
@@ -2,40 +2,34 @@ package uk.gov.communities.prsdb.webapp.journeys.acceptOrRejectJointLandlordInvi
...
-                val landlord =
-                    landlordService.retrieveLandlordByBaseUserId(baseUserId)
-                        ?: throw PrsdbWebException(
-                            "Landlord record not found for user with baseUserId $baseUserId after they completed landlord registration",
-                        )
-                RegistrationNumberDataModel.fromRegistrationNumber(landlord.registrationNumber).toString()
-            } else {
-                null
-            }
+        val registrationNumber = state.registeredLandlordRegistrationNumber
```
Travis-Softwire, line 32: Rather than doing the landlord look-up twice (once in the internal step, once again below in afterStepDataIsAdded) could we not do the look-up either here or in an init block and store the landlord as a field on the class?

Travis-Softwire (approved):
Couple of minor points but overall LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/steps/ConfirmYouAreALandlordForThisPropertyStepConfig.kt
JasminConterioSW: I was thinking the user could log out and log in again as a different landlord after the landlord registration step, but actually logging out would just clear the session anyway so I'll simplify this a bit!

JasminConterioSW: Actually it's looking fiddlier than I thought so I might leave it as is...

* I don't think storing it on the class will persist between requests, we are looking up to get the registration number on the GET request and then looking up to get the whole landlord on the POST request
* Copilot says `We can't store a JPA Landlord entity directly — it's not @Serializable and has Hibernate proxies/lazy associations` so caching the whole landlord in the journey state the first time we look it up won't work

PR #1436, opened 2026-06-11
PDJB-265: Implement the confirmation page

JasminConterioSW:
## Ticket number

PDJB-265

## Goal of change

Add confirmation page and emails

* Implement the confirmation page, storing required attributes from the journey such as the property address in the session
* Add emails going to existing landlords and the new joint landlord confirming that the invitation has been accepted

Confirmation page
image

Email to new landlord
image

Email to existing landlords
image

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/AcceptOrRejectJointLandlordInvitationController.kt
```diff
@@ -82,9 +82,18 @@ class AcceptOrRejectJointLandlordInvitationController(
     @GetMapping("/$PROPERTY_JOINED_CONFIRMATION_PATH_SEGMENT")
     @AvailableWhenFeatureEnabled(JOINT_LANDLORDS)
     fun getConfirmation(model: Model): ModelAndView {
-        model.addAttribute("title", "TODO: PDJB-265 - Property joined confirmation")
+        val propertyAddress =
```
Travis-Softwire: Feels a bit weird to store them in one call but get them in two. If we're not using them anywhere else we could store and retreive them in something simple like a pair?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationService.kt
```diff
@@ -249,4 +251,19 @@ class JointLandlordInvitationService(
...
+        session.setAttribute(ACCEPTED_JOINT_LANDLORD_PROPERTY_ADDRESS, address)
+        session.setAttribute(ACCEPTED_JOINT_LANDLORD_PROPERTY_OWNERSHIP_ID, propertyOwnershipId)
+    }
+
+    fun getLastAcceptedPropertyAddressFromSession(): String? = session.getAttribute(ACCEPTED_JOINT_LANDLORD_PROPERTY_ADDRESS) as? String
+
+    fun getLastAcceptedPropertyOwnershipIdFromSession(): Long? =
+        session.getAttribute(
+            ACCEPTED_JOINT_LANDLORD_PROPERTY_OWNERSHIP_ID,
+        ) as? Long
```
Travis-Softwire: Copilot is flagging that this is brittle because it relies on how the serializer serializes numbers. It's recommending `?toString()?.toLong()` instead which we seem to use in a few other places.

src/test/kotlin/uk/gov/communities/prsdb/webapp/testHelpers/mockObjects/MockLandlordData.kt
```diff
@@ -133,6 +134,7 @@ class MockLandlordData {
 
             ReflectionTestUtils.setField(propertyOwnership, "id", id)
             ReflectionTestUtils.setField(propertyOwnership, "createdDate", createdDate)
+            ReflectionTestUtils.setField(propertyOwnership, "landlords", landlords ?: mutableSetOf(primaryLandlord))
```
Travis-Softwire: I don't think landlords is nullable?

Travis-Softwire (approved):
Two super minor points, LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationService.kt
JasminConterioSW: This is now combined with a function getting `Pair<String, Long>`

Local copilot claims this will be ok...

```
Not quite the same problem. The original concern with as? Long was likely that session serialization or intermediate code could store the value as an Int (e.g., from form data), making as? Long silently return null.
                                                                              
With as? Pair<String, Long>, due to JVM type erasure, the runtime only checks whether the attribute is a Pair — the generic parameters <String, Long> aren't verified at runtime (hence the @Suppress("UNCHECKED_CAST")). So the cast won't fail on type mismatch within the pair. 

Since we control both the store and retrieve (we always write Pair(String, Long) explicitly in storeLastAcceptedPropertyInSession), the types are guaranteed to be correct within the same session. The brittleness risk is effectively eliminated because there's no separate code path that could store a different type under the same key
```

PR #1438, opened 2026-06-11
PDJB-306: Part 2: Add switch to JL flow

samyou-softwire:
## Ticket number

[PDJB-306](https://mhclgdigital.atlassian.net/browse/PDJB-306)

## Goal of change

Add the switch to JL flow without breaking the two other JL adding flows

<details><summary>screenshots</summary>
property details with no JL
start of JL add flow
finish JL flow with now in JL mode
</details>

updates invite JL journey to potentially include the has JL step

the copy between the has JL page on prop registration and the invite joint landlord flow is the same (thankfully)

finishing this new journey sets the property JL flag if appropriate

see comments

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/inviteJointLandlord/CompleteInviteJointLandlordStepConfig.kt
```diff
@@ -35,6 +36,24 @@ class CompleteInviteJointLandlordStepConfig(
...
 
 @JourneyFrameworkComponent
-class CompleteInviteJointLandlordStep(
-    stepConfig: CompleteInviteJointLandlordStepConfig,
+class CompleteInviteJointLandlordYesJointStep(
+    stepConfig: CompleteInviteJointLandlordYesJointStepConfig,
+) : JourneyStep.InternalStep<Complete, InviteJointLandlordJourneyState>(stepConfig)
+
+@JourneyFrameworkComponent
+class CompleteInviteJointLandlordNoJointStepConfig : AbstractInternalStepConfig<Complete, InviteJointLandlordJourneyState>() {
```
samyou-softwire: these are in the same file as they're very similar, just the sort of 'outcome handler' steps for if you pick yes or pick no to having JL

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/inviteJointLandlord/InviteJointLandlordJourneyFactory.kt
```diff
@@ -62,22 +68,26 @@ class InviteJointLandlordJourneyFactory(
                 parents { journey.startInviteJointLandlordStep.isComplete() }
                 backUrl { propertyDetailsLandlordTab }
                 nextDestination { _ ->
-                    if (journey.invitedJointLandlords.isEmpty()) {
-                        Destination.ExternalUrl(propertyDetailsLandlordTab)
-                    } else {
+                    if (journey.hasJointLandlordsStep.outcome == YesOrNo.YES || state.propertyMarkedAsJointLandlord) {
```
samyou-softwire: the second condition is that for on the add more JL flow we won't ask the hasJointLandlordsStep question (we assume yes). happy if there's a better way of phrasing this condition

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/inviteJointLandlord/InviteJointLandlordsTask.kt
```diff
@@ -25,43 +47,43 @@ class InviteJointLandlordsTask(
             }
             step(journey.inviteJointLandlordStep) {
                 routeSegment(InviteJointLandlordStep.INVITE_FIRST_ROUTE_SEGMENT)
-                parents { hasAnyJointLandlordsInvitedStep.hasOutcome(AnyLandlordsInvited.NO_LANDLORDS) }
```
samyou-softwire, line 28: I don't know if there's a meaningful difference between getting steps from the journey object or having them wired in, in any case I made it consistent

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/inviteJointLandlord/InviteJointLandlordsTask.kt
```diff
@@ -6,16 +6,38 @@ import uk.gov.communities.prsdb.webapp.journeys.Task
...
 
 @JourneyFrameworkComponent
-class InviteJointLandlordsTask(
-    private val hasAnyJointLandlordsInvitedStep: HasAnyJointLandlordsInvitedStep,
-) : Task<InviteJointLandlordState>() {
+class InviteJointLandlordsTask : Task<InviteJointLandlordState>() {
     override fun makeSubJourney(state: InviteJointLandlordState) =
         subJourney(state) {
-            step(hasAnyJointLandlordsInvitedStep) {
+            step(journey.isMarkedAsJointLandlordStep) {
```
samyou-softwire: the important logic here, needs to handle the using saying no JL, skipping that question if we already have JL, and skipping the email question initially if we already have JL entered on the flow to make the back buttons etc work

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/inviteJointLandlord/IsMarkedAsJointLandlordStepConfig.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.JourneyFrameworkComponent
+import uk.gov.communities.prsdb.webapp.journeys.AbstractInternalStepConfig
+import uk.gov.communities.prsdb.webapp.journeys.JourneyStep
+import uk.gov.communities.prsdb.webapp.journeys.shared.YesOrNo
+import uk.gov.communities.prsdb.webapp.journeys.shared.states.InviteJointLandlordState
+
+@JourneyFrameworkComponent
+class IsMarkedAsJointLandlordStepConfig : AbstractInternalStepConfig<YesOrNo, InviteJointLandlordState>() {
+    override fun mode(state: InviteJointLandlordState): YesOrNo = if (state.propertyMarkedAsJointLandlord) YesOrNo.YES else YesOrNo.NO
```
samyou-softwire: this has to be in the state since I can't check the property registration directly here (it may not exist)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -236,6 +236,13 @@ class PropertyOwnershipService(
         propertyOwnershipRepository.save(propertyOwnership)
     }
 
+    @Transactional
+    fun markAsJointLandlord(id: Long) {
```
samyou-softwire: funny this didn't already exist; suppose since we only set jl to true on registration before

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/inviteJointLandlord/CompleteInviteJointLandlordStepConfig.kt
```diff
@@ -35,6 +36,24 @@ class CompleteInviteJointLandlordStepConfig(
 }
 
 @JourneyFrameworkComponent
-class CompleteInviteJointLandlordStep(
-    stepConfig: CompleteInviteJointLandlordStepConfig,
+class CompleteInviteJointLandlordYesJointStep(
```
Travis-Softwire: I definitely don't like having a step that is named after the outcome for reaching it. The steps should be named for what they themselves do. If that's not possible (as with the 'No' case here) then we've probably got the journey architecture wrong

Travis-Softwire:
I think we've massively overcomplicated this - I think we just want:
- the inner task, which doesn't have the hasJointLandlord step in it
- the full task, which does have that step
- two versions of the invite JL journey, either behind different urls, OR with a simple condition in the controller that picks between the two versions of the journey based on whether the property ownership is marked as being a JL one

Travis-Softwire (changes requested):
I think we need to rethink this - this is a lot of complexity and I don't think it's necessary

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/inviteJointLandlord/IsMarkedAsJointLandlordStepConfig.kt
TomHanmer29: This should probably become logic in the controller picking which route to use

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/inviteJointLandlord/InviteJointLandlordsTask.kt
TomHanmer29: I think it feels really messy to have to include this in the task that's also used in property registration, I think we should be able to leave this file pretty much as is

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/inviteJointLandlord/InviteJointLandlordJourneyFactory.kt
```diff
@@ -62,22 +68,26 @@ class InviteJointLandlordJourneyFactory(
...
             }
-            step(journey.completeInviteJointLandlordStep) {
+            step(journey.completeInviteJointLandlordYesJointStep) {
                 parents { journey.checkInvitationsStep.isComplete() }
                 nextUrl { confirmationRoute }
             }
+            step(journey.completeInviteJointLandlordNoJointStep) {
+                parents { journey.inviteJointLandlordsTask.isComplete() }
+                nextUrl { propertyDetailsLandlordTab }
+            }
```
TomHanmer29: We should just be able to copy the structure used in JointLandlordsPropertyRegistrationTask.kt here with an exitStep I think? It might be a bit different since in one case we route into the task, and in the other we want to route back to the dashboard, but we shouldn't need 2 steps to decide this

TomHanmer29 (changes requested):
Agree with Travis here, I think it's much simpler (even if it's a small amount of duplication) to have 2 versions of the task, 1 with has joint landlords and 1 without. I don't really have a preference as to whether we have 2 different routes or if we choose which version to use in the controller.

samyou-softwire:
I'll rework these files and journeys now. the overarching point being that the journey framework works best by compositing smaller tasks into journeys, keeping journey specifics to the journey file and making the tasks as limited in scope as possible

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/inviteJointLandlord/CompleteInviteJointLandlordStepConfig.kt
```diff
@@ -16,13 +16,16 @@ class CompleteInviteJointLandlordStepConfig(
...
 
     override fun afterStepIsReached(state: InviteJointLandlordJourneyState) {
-        val propertyOwnership = propertyOwnershipService.getPropertyOwnership(state.propertyId)
-        // TODO PDJB-1069 - do not use primary landlord when it is not needed
-        jointLandlordInvitationService.sendInvitationEmails(
-            jointLandlordEmails = state.invitedJointLandlords,
-            propertyOwnership = propertyOwnership,
-            invitingLandlord = propertyOwnership.primaryLandlord,
-        )
+        if (state.invitedJointLandlords.isNotEmpty()) {
```
samyou-softwire, line 23: specific override needed as it's now necessary to not run `propertyOwnershipService.markAsJointLandlord(state.propertyId)` if you didn't add any landlords

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/inviteJointLandlord/InviteJointLandlordJourneyFactory.kt
```diff
@@ -56,30 +72,64 @@ class InviteJointLandlordJourneyFactory(
...
+                        journey.confirmAndInviteJointLandlordsTask.isComplete()
+                    }
+                }
                 nextStep { journey.completeInviteJointLandlordStep }
             }
             step(journey.completeInviteJointLandlordStep) {
-                parents { journey.checkInvitationsStep.isComplete() }
-                nextUrl { confirmationRoute }
+                parents {
+                    if (propertyMarkedAsJointLandlord) {
```
samyou-softwire: could also split this into two separate journeys if the branching here is too finnicky

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/inviteJointLandlord/InviteJointLandlordsTask.kt
samyou-softwire: aside from the `journey.` changes it now should be

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/states/InviteJointLandlordPropertyRegistrationState.kt
```diff
@@ -1,15 +1,8 @@
...
-import uk.gov.communities.prsdb.webapp.journeys.shared.inviteJointLandlord.InviteJointLandlordsTask
-import uk.gov.communities.prsdb.webapp.journeys.shared.states.InviteJointLandlordState
+import uk.gov.communities.prsdb.webapp.journeys.shared.states.ConfirmAndInviteJointLandlordState
 
-interface InviteJointLandlordPropertyRegistrationState : InviteJointLandlordState {
-    val hasAnyJointLandlordsInvitedStep: HasAnyJointLandlordsInvitedStep
-    val hasJointLandlordsStep: HasJointLandlordsStep
-    val inviteJointLandlordsTask: InviteJointLandlordsTask
-}
+interface InviteJointLandlordPropertyRegistrationState : ConfirmAndInviteJointLandlordState
```
TomHanmer29: Do we need to redeclare this state?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/inviteJointLandlord/InviteJointLandlordJourneyFactory.kt
TomHanmer29: I definitely think this should be 2 journeys, wrapping DSL logic in if/elses isn't a pattern we have anywhere else and just makes it harder to read imo. We specifically tried to avoid this when adding the joint landlords flag in the first place so shouldn't add it here.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/inviteJointLandlord/InviteJointLandlordsTask.kt
TomHanmer29: That's true, I don't think there's a real difference. I think usually we tend to go the other way and not use the journey. but there's quite a few of them in the codebase

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/tasks/JointLandlordsPropertyRegistrationTask.kt
```diff
@@ -25,47 +20,12 @@ class JointLandlordsPropertyRegistrationTask : Task<InviteJointLandlordPropertyR
...
-                }
-                nextDestination { _ ->
-                    if (journey.invitedJointLandlords.isEmpty()) {
-                        Destination(journey.hasJointLandlordsStep)
-                    } else {
-                        Destination(exitStep)
-                    }
-                }
+            // this is split out into its own task to be common with the 'Add a JL' button on property details
+            task(journey.confirmAndInviteJointLandlordsTask) {
```
TomHanmer29: Why do we need this wrapped in another task here? is it just for the taskStatus section? Could we not put that inside ConfirmAndInviteJointLandlordsTask? I just think it feels a little odd to have a task which is pretty much just 1 other task

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/inviteJointLandlord/CompleteInviteJointLandlordStepConfig.kt
TomHanmer29: Would it be simpler to just extract out hasJointLandlords, and make the "No" option in the standalone journey route straight back to the property record, and the no option in the registration journey skip to the next section? That way we could remove the ConfirmAndInviteJointLandlord task, put these steps in the JointLandlordsPropertyRegistrationTask/InviteJointLandlordsTasks and avoid having too many nested tasks, which I think introduces quite a lot of complexity.

TomHanmer29 (changes requested):
I think there's still some pretty big changes I'd make here I'm afraid. But let me know if you or Travis disagree, I'm open to other suggestions :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/inviteJointLandlord/CompleteInviteJointLandlordStepConfig.kt
TomHanmer29: It leaves you with 3 journeys which each use 1 task, rather than 2 journeys using a task nested 2-3 layers deep, which I think is neater overall

samyou-softwire: I'll refactor to this and see how it looks

TomHanmer29 (approved):
Thanks for sticking with this, I think this is looking good now :) Still worth Travis having a look too though!

Travis-Softwire (approved):
Yep, happy with this as well, thanks for persevering!

PR #1440, opened 2026-06-11
PDJB-269: JL invite deletion job

TomHanmer29:
## Ticket number

PDJB-269

## Goal of change

Deletes invites that have been expired for 7 days or more

Includes hidden invites

## Checklist

  mention that here

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationDeletionService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationDeletionService.kt
@@ -36,7 +36,7 @@ class JointLandlordInvitationDeletionServiceImplFlagOn(
         val invitationsToDelete =
             invitationRepository
                 .findAllByCreatedDateBefore(cutoffDate)
-                .filter { it.status != JointLandlordInvitationStatus.PENDING }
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationDeletionService.kt
```diff
@@ -0,0 +1,46 @@
...
+    @Transactional
+    override fun deleteExpiredInvitations(): List<Long> {
+        val totalGracePeriodInDays =
+            (JOINT_LANDLORD_INVITATION_LIFETIME_IN_DAYS + JOINT_LANDLORD_INVITATION_DELETION_GRACE_PERIOD_IN_DAYS).toLong()
+        val cutoffDate = Instant.now().minus(totalGracePeriodInDays, ChronoUnit.DAYS)
+
+        val invitationsToDelete =
+            invitationRepository
+                .findAllByCreatedDateBefore(cutoffDate)
+                .filter { it.status in listOf(JointLandlordInvitationStatus.EXPIRED, JointLandlordInvitationStatus.HIDDEN) }
```
TomHanmer29, line 39: We don't technically need this because status is calculated based on date anyway, but I figured it's better to have it defensively in case we do change how we calculate these.

Travis-Softwire (approved):
LGTM

PR #1441, opened 2026-06-11
PDJB-NONE: Document updated release process

samyou-softwire:
this should match what was discussed in dev catchups

### DIFF
--- a/docs/FeatureFlagsReadMe.md
+++ b/docs/FeatureFlagsReadMe.md
@@ -21,9 +21,10 @@ added to check that strategies could be combined)
 
 ## Feature flag name
 
-Each feature flag should equate to an epic on Jira.
+Each feature flag should relate to a single ticket on Jira. This will either be the task/story ticket that relates to the feature flag, or the epic if this feature flag relates to a series of tickets.
+
 We should include the ticket number in the feature flag name.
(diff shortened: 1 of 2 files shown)


docs/FeatureFlagsReadMe.md
```diff
@@ -19,6 +19,12 @@ override any strategy config on individual flags in that release.
 Currently, we are set up to add ReleaseDateFlipStrategy (feature disabled until the release date), and a boolean flipping strategy (mainly
 added to check that strategies could be combined)
 
+## Feature flag name
+
+Each feature flag should equate to an epic on Jira.
```
Travis-Softwire: I think it's less specific than that - bigger features with multiple tickets behind the same flag will be an epic - smaller features or changes might be behind their own feature flag and so that flag should reference the relevant task/story ticket

Travis-Softwire (approved):
One nuance on the ticket<->flag link, happy for you to finessse the wording without me needing to see it again though

docs/FeatureFlagsReadMe.md
samyou-softwire: ah good point, have reworded to account for this

samyou-softwire (commented):
PDJB-NONE: Account for single ticket feature flags

edit: that isn't the commit dialog

PR #1445, opened 2026-06-12
PDJB-1035: Add Plausible analytics journey completion rate metrics

Bill-Haigh:
## Ticket number

PDJB-1035

## Goal of change

Surface journey completion rates (landlord registration, property registration, local council user registration) on the system operator metrics dashboard, sourced from Plausible analytics.

- Adds a Plausible analytics integration (`PlausibleClient` + `PlausibleApiConfig`) that queries the Plausible Stats API for per-page visitor counts.
- Adds three completion-rate rows to the metrics dashboard: confirmation-page visitors as a percentage of start-page visitors for each registration journey.
- Rates degrade gracefully to "No data" on API failure (now logged at `warn`) and are capped at 100%.
- Adds a local stub (`MockPlausibleController`, `@Profile("local")`) so local dev exercises the real client/parsing path without an outbound call or API key.

- This branch was rebased onto main after PDJB-1033 merged. PDJB-1033 evolved during its review (single `averageTimeToFirstProperty` row → median/p90/p95 percentile rows), so the rebase keeps main's percentile rows and layers the new completion-rate rows on top — the metrics list is now 10 rows.

## Checklist

  mention that here

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
@@ -26,18 +26,18 @@ class PlausibleMetricsService(
     fun getCompletionRates(period: ReportingPeriod): JourneyCompletionRatesDataModel =
         try {
             val response = plausibleClient.query(buildQuery(period))
-            val visitorsByPage =
+            val pageViewsByPage =
                 response.results
                     .filter { it.dimensions.isNotEmpty() && it.metrics.isNotEmpty() }
                     .associate { it.dimensions.first() to it.metrics.first() }
             JourneyCompletionRatesDataModel(
                 landlordRegistration =
-                    completionRate(visitorsByPage, LANDLORD_REGISTRATION_START_PAGE_ROUTE, LANDLORD_REGISTRATION_CONFIRMATION_ROUTE),
+                    completionRate(pageViewsByPage, LANDLORD_REGISTRATION_START_PAGE_ROUTE, LANDLORD_REGISTRATION_CONFIRMATION_ROUTE),
                 propertyRegistration =
-                    completionRate(visitorsByPage, PROPERTY_REGISTRATION_ROUTE, PROPERTY_REGISTRATION_CONFIRMATION_ROUTE),
+                    completionRate(pageViewsByPage, PROPERTY_REGISTRATION_ROUTE, PROPERTY_REGISTRATION_CONFIRMATION_ROUTE),
                 localCouncilUserRegistration =
                     completionRate(
-                        visitorsByPage,
+                        pageViewsByPage,
                         LOCAL_COUNCIL_USER_REGISTRATION_PRIVACY_NOTICE_ROUTE,
                         LOCAL_COUNCIL_USER_REGISTRATION_CONFIRMATION_ROUTE,
                     ),
@@ -51,21 +51,21 @@ class PlausibleMetricsService(
         PlausibleQuery(
             siteId = siteId,
             dateRange = listOf(period.start.toUkDate(), period.end.toUkDate()),
-            metrics = listOf("visitors"),
+            metrics = listOf("pageviews"),
             dimensions = listOf("event:page"),
             filters = listOf(listOf("is", "event:page", ALL_PAGES)),
         )
 
     private fun completionRate(
-        visitorsByPage: Map<String, Double>,
+        pageViewsByPage: Map<String, Double>,
         startPage: String,
         confirmationPage: String,
     ): Double? {
-        val startVisitors = visitorsByPage[startPage] ?: return null
-        if (startVisitors == 0.0) return null
-        val confirmationVisitors = visitorsByPage[confirmationPage] ?: 0.0
+        val startPageViews = pageViewsByPage[startPage] ?: return null
+        if (startPageViews == 0.0) return null
+        val confirmationPageViews = pageViewsByPage[confirmationPage] ?: 0.0
         return BigDecimal
-            .valueOf(confirmationVisitors / startVisitors * 100)
+            .valueOf(confirmationPageViews / startPageViews * 100)
             .setScale(2, RoundingMode.HALF_UP)
             .toDouble()
             .coerceAtMost(100.0)
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
@@ -49,6 +49,8 @@ class PlausibleMetricsService(
             JourneyCompletionRatesDataModel(null, null, null)
         }
 
+    // TODO PDJB-1055: Add pagination handling for Plausible queries. This query is currently bounded to a fixed set of
+    //  pages, but pagination will need to be addressed if a query ever returns more than 10,000 entries from Plausible.
     private fun buildQuery(period: ReportingPeriod): PlausibleQuery =
         PlausibleQuery(
             siteId = siteId,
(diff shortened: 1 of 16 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
```diff
@@ -0,0 +1,95 @@
...
+                        LOCAL_COUNCIL_USER_REGISTRATION_PRIVACY_NOTICE_ROUTE,
+                        LOCAL_COUNCIL_USER_REGISTRATION_CONFIRMATION_ROUTE,
+                    ),
+            )
+        } catch (e: Exception) {
+            println("Failed to fetch journey completion rates from Plausible: ${e.message}")
+            JourneyCompletionRatesDataModel(null, null, null)
+        }
+
+    private fun buildQuery(period: ReportingPeriod): PlausibleQuery =
```
Travis-Softwire, line 54: There's a ticket in the backlog to handle pagination of Plausible API queries - assuming that's non-trivial to add, can we put a TODO somewhere referencing it?

Travis-Softwire (commented):
One request, but LGTM. Will let Jasmin approve though. 

Reminder to NOT merge this until you've raised and merged the infra PR to create the new env variable, otherwise you'll break deployments to integration

src/main/resources/application.yml
```diff
@@ -81,6 +81,8 @@ base-url:
 
 plausible:
   site-id: ${PLAUSIBLE_SITE_ID}
+  base-url: ${PLAUSIBLE_BASE_URL:https://plausible.io}
+  api-key: ${PLAUSIBLE_API_KEY:}
```
JasminConterioSW: Does the colon mean that it _won't_ break if `PLAUSIBLE_API_KEY` is not set, unless someone tries to use the dashboard page?

I _think_ this is ok, but it won't be so immediately obvious that we're missing the environment variable until someone tries to use the page.

src/main/resources/application.yml
```diff
@@ -81,6 +81,8 @@ base-url:
 
 plausible:
   site-id: ${PLAUSIBLE_SITE_ID}
+  base-url: ${PLAUSIBLE_BASE_URL:https://plausible.io}
```
JasminConterioSW: I don't think we need the PLAUSIBLE_BASE_URL environment variable here as we're not expecting it to change between deployed environments, can it just be set to `https://plausible.io`

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
```diff
@@ -0,0 +1,95 @@
...
+        PlausibleQuery(
+            siteId = siteId,
+            dateRange = listOf(period.start.toUkDate(), period.end.toUkDate()),
+            metrics = listOf("visitors"),
+            dimensions = listOf("event:page"),
+            filters = listOf(listOf("is", "event:page", ALL_PAGES)),
+        )
+
+    private fun completionRate(
+        visitorsByPage: Map<String, Double>,
```
JasminConterioSW: For completion rate, I think we want to calculate this based on `pageViews` rather `visitors`, @Travis-Softwire does that sound right?

Either would work for landlord registration but we would want pageViews to capture multiple property registrations by the same landlord.

pageViews results could get skewed by users returning to / refreshing a page.
visitors is inaccurate when users complete a journey over multiple days (they could do this for property registration although not landlord registration)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
JasminConterioSW, line 54: Can we rename this to specify which metric we're querying (visitors or pageViews)?

JasminConterioSW (commented):
Can we check whether we want to use visitor numbers or pageViews to calculate completion rates? I was assuming pageViews would work better for property registration (as one visitor could complete multiple journeys) and we should use the same for the different journeys to be consistent.

Not for this PR, but do we want to try and look at update journeys by querying the flow event e.g. for returns to propertyDetails from updateEpc CYA? This might not be this ticket, it also might not be how we want to do it - there _is_ no CYA for single page update journeys so we might need a different way of measuring anyway

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
Travis-Softwire: Yep, agree for property registration we want pageViews. We can live with the odd refresh on that page I think.

JasminConterioSW: Do we want pageViews for both? Or do we use visitors for landlord registration?

JasminConterioSW:
> Can we check whether we want to use visitor numbers or pageViews to calculate completion rates? I was assuming pageViews would work better for property registration (as one visitor could complete multiple journeys) and we should use the same for the different journeys to be consistent.
> 
> Not for this PR, but do we want to try and look at update journeys by querying the flow event e.g. for returns to propertyDetails from updateEpc CYA? This might not be this ticket, it also might not be how we want to do it - there _is_ no CYA for single page update journeys so we might need a different way of measuring anyway

Oh hang on, we don't need update journeys here, that's for counting transactions (for cost per transaction), sorry!

src/main/resources/application.yml
Travis-Softwire: Copilot often believes this is the case, but from memory it doesn't actually work in practice...

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
Travis-Softwire: Good Q - I think visitors for landlord registration since they can't revisit in a new session and should only hit it once per user

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
```diff
@@ -0,0 +1,105 @@
...
+        confirmationPage: String,
+    ): Double? {
+        val startCount = countsByPage[startPage] ?: return null
+        if (startCount == 0.0) return null
+        val confirmationCount = countsByPage[confirmationPage] ?: 0.0
+        return BigDecimal
+            .valueOf(confirmationCount / startCount * 100)
+            .setScale(2, RoundingMode.HALF_UP)
+            .toDouble()
+            .coerceAtMost(100.0)
```
JasminConterioSW: I'd probably leave out the` .coerceAtMost(100.0)` - a value over 100% raises questions but I think that's ok, we can answer the question if it comes up (mostly I expect it'll be "the user refreshed the confirmation page so the pageView was double counted"). It shows to take the numbers with a bit of a pinch of salt I think!

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    private fun Instant.toUkDate(): String = DateTimeFormatter.ISO_LOCAL_DATE.format(this.atZone(UK_ZONE).toLocalDate())
+
+    companion object {
+        private val UK_ZONE = ZoneId.of("Europe/London")
+
+        // Visitors are used for one-off-per-user journeys (landlord and local council registration); page views are
+        // used for property registration, where a single landlord may register multiple properties.
```
JasminConterioSW, line 83: Can we get this comment (or similar) on the actual page, just so they are clear without asking us which metric was used for the calculation?

JasminConterioSW (approved):
A couple of minor comments but then LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
Bill-Haigh: done, removed.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
Bill-Haigh: have added an inset at the bottom explaining.

PR #1448, opened 2026-06-15
PDJB-1075: Add Notify allowlist error page for non-production environments

Travis-Softwire:
## Ticket number

PDJB-1075

## Goal of change

Show a specific error page in non-production environments when an email fails to send because the recipient is not on the GOV.UK Notify allowlist, instead of the generic service-problem page.

- Detects the GOV.UK Notify "can't send to this recipient" allowlist failure and, only when running against allowlisted (non-production) Notify, shows a dedicated error page explaining that the recipient is not on the Notify allowlist.
- Production behaviour is unchanged: gated on the existing `notify.use-production-notify` config, so allowlist-shaped errors still fall through to the generic error handling in production.
- The new page is modelled on the standard error page (no designs, as it is not user-facing).

- Detection is gated on `notify.use-production-notify` (the exact signal — the allowlist only exists on non-production Notify) and matches the Notify message fragment "send to this recipient", which covers both the team-only-key and trial-mode messages.
- `NotifyAllowlistException` deliberately subclasses `PersistentEmailSendException` so the existing task-context `catch (PersistentEmailSendException)` sites (incomplete-properties reminder and JL invite expiry task) keep working unchanged, while the web `@ExceptionHandler` catches the more specific subclass.
- Verified end-to-end with a local smoke test against **real Notify** (`use-notify` profile): inviting a local council user with a non-allowlisted address redirects to the new page and renders correctly.

## Checklist

- [x] TODO comments referencing this JIRA ticket have been searched for and removed (none found)
- [x] Test suite has been run in full locally and is passing — targeted unit, controller and integration tests for the affected classes pass; the full suite was not run as the change is isolated to the email error path
- [x] QA instructions have been added to the ticket — to be added (smoke tested locally against real Notify by triggering a send to a non-allowlisted recipient)

image


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationService.kt
```diff
@@ -104,4 +120,8 @@ class NotifyEmailNotificationService<EmailModel : EmailTemplateModel>(
         @SerialName("Exception")
         EXCEPTION,
     }
+
+    companion object {
+        private const val NOTIFY_ALLOWLIST_MESSAGE_FRAGMENT = "send to this recipient"
```
samyou-softwire: is there something more specific we could use here? "can't send to this recipient", nervous that this checks for text that suggests we could send to this recipient

samyou-softwire (approved):

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/NotifyEmailNotificationService.kt
Travis-Softwire: Good point, have updated to "can't send to this recipient" and added some handling for the curly apostrophe :)

PR #1449, opened 2026-06-15
PDJB-317: confirm property deregistration page

TomHanmer29:
## Ticket number

PDJB-317

## Goal of change

Adds the confirm property deletion page

image
Adds the confirm property deletion page

This hasn't touched any of the existing email sending logic, but it has moved it to the confirm page (as it replaced the reason page)

## Checklist

  mention that here

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyDeregistration/stepConfig/ConfirmStepConfig.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyDeregistration/stepConfig/ConfirmStepConfig.kt
@@ -7,7 +7,6 @@ import uk.gov.communities.prsdb.webapp.journeys.Destination
 import uk.gov.communities.prsdb.webapp.journeys.JourneyStep.RequestableStep
 import uk.gov.communities.prsdb.webapp.journeys.propertyDeregistration.PropertyDeregistrationJourneyState
 import uk.gov.communities.prsdb.webapp.journeys.shared.Complete
-import uk.gov.communities.prsdb.webapp.models.dataModels.RegistrationNumberDataModel
 import uk.gov.communities.prsdb.webapp.models.requestModels.formModels.NoInputFormModel
 import uk.gov.communities.prsdb.webapp.models.viewModels.emailModels.PropertyDeregistrationConfirmationEmail
 import uk.gov.communities.prsdb.webapp.services.EmailNotificationService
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyDeregistration/stepConfig/ConfirmStepConfig.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    override fun chooseTemplate(state: PropertyDeregistrationJourneyState) = "forms/confirmPropertyDeregistrationForm"
+
+    override fun mode(state: PropertyDeregistrationJourneyState): Complete? = getFormModelFromStateOrNull(state)?.let { Complete.COMPLETE }
+
+    override fun afterStepDataIsAdded(state: PropertyDeregistrationJourneyState) {
+        val propertyOwnership = propertyOwnershipService.getPropertyOwnership(state.propertyOwnershipId)
+
+        // TODO PDJB-319 - do not use primary landlord
+        val primaryLandlordEmailAddress = propertyOwnership.primaryLandlord.email
```
Travis-Softwire: How big an extra change would it be to upgrade this to use the landlords list instead? Appreciate this is just a refactor but I'd like to avoid new references to primaryLandlord if we can

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyDeregistration/stepConfig/ConfirmStepConfig.kt
```diff
@@ -0,0 +1,77 @@
...
+        val propertyRegistrationNumber = propertyOwnership.registrationNumber
+        val propertyAddress = propertyOwnership.address.singleLineAddress
+
+        propertyDeregistrationService.deregisterProperty(state.propertyOwnershipId)
```
Travis-Softwire: Looks like there are a few db actions here - should they all be in a service layer method inside a transaction?

Travis-Softwire (approved):
2 Qs, otherwise LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyDeregistration/stepConfig/ConfirmStepConfig.kt
TomHanmer29: Yes, ideally. This is all lifted straight from the previous reasonStep which didn't do it but I think it's safer to do it :) I'll make the change

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyDeregistration/stepConfig/ConfirmStepConfig.kt
TomHanmer29: Sure, I can use the landlord list. The main issue here is that the email isn't being changed in this ticket so there'll be a bit of time it won't make sense to the JLs but if it's flagged off anyway it shouldn't matter :)

PR #1450, opened 2026-06-16
PDJB-1033: Add metrics test cohorts to local and integration seed data

Bill-Haigh:
## Ticket number

PDJB-1033

## Goal of change

Make the System Operator metrics dashboard testable by seeding enough landlord/property data to exercise the median/p90/p95 "time to first property" percentiles.

- Adds a **deterministic cohort** (101 landlords, "time to first property" of 0–100 days) to `data-local.sql` and `data-integration.sql`, giving known, exact percentiles of median/p90/p95 = 50/90/95 days for easy verification.
- Adds a **realistic randomised cohort** (~150 landlords with varied registration dates, verification status, and 0–3 properties each) to produce a natural spread on the dashboard.
- Both cohorts use fixed high IDs plus `ON CONFLICT DO NOTHING`, so they are idempotent under `spring.sql.init` (`mode: always` re-runs the script on every startup).

- Written as plain set-based `INSERT … SELECT generate_series(...)` rather than a procedural `DO` block, because Spring's SQL init script runner splits on `;` and does not understand PostgreSQL dollar-quoting (a `DO` block would be shredded and silently skipped).
- The deterministic cohort offsets are added as **absolute seconds** (`make_interval(secs => (i-1)*86400)`), not a `days` interval. In Postgres, `timestamptz + interval 'N days'` is DST-aware under the app's Europe/London JDBC session, so offsets crossing the 2030 spring-forward landed an hour short and floored p90/p95 to 89/94. Absolute seconds keep the known values exact regardless of session timezone.

## Checklist


src/main/resources/data-integration.sql
```diff
@@ -392,3 +392,216 @@ VALUES ('PRSD22', current_date, null, 'urn:fdc:gov.uk:2022:mGHDySEVfCsvfvc6lVWf6
...
+-- duration exact regardless of session timezone, so the known values stay 50/90/95.
+-- so "time to first property" takes the values 0, 1, 2, ..., 100 days. The
+-- Metrics service computes percentiles as rank = fraction * (n - 1) with linear
+-- interpolation, so for n = 101 this gives exactly:
+--   median = 50 days, p90 = 90 days, p95 = 95 days, average = 50 days.
+--
+-- HOW TO VERIFY: open System Operator > Metrics and submit the date range
+--   From: 1 / 1 / 2030    To: 31 / 12 / 2030
+-- Expected: 101 landlord registrations, 101 verified, 101 properties,
+--   101 landlords with a property, median/p90/p95 = 50/90/95 days.
```
JasminConterioSW: Most of this comment doesn't need to be in the sql file I think, some might want to go on the ticket at it looks more like qa instructions!

src/main/resources/data-integration.sql
```diff
@@ -392,3 +392,216 @@ VALUES ('PRSD22', current_date, null, 'urn:fdc:gov.uk:2022:mGHDySEVfCsvfvc6lVWf6
...
+-- =============================================================================
+INSERT INTO prsdb_user (id, created_date)
+SELECT 'metrics-test-user-' || i, TIMESTAMPTZ '2030-01-01 09:00:00+00'
+FROM generate_series(1, 101) AS s(i)
+ON CONFLICT DO NOTHING;
+
+INSERT INTO address (id, created_date, single_line_address, local_council_id, postcode, building_number)
+SELECT 9000000 + i, TIMESTAMPTZ '2030-01-01 09:00:00+00',
```
JasminConterioSW: Can you check the ids are still ok when you need to add a new record?

The other seeded data has for example `SELECT setval(pg_get_serial_sequence('property_compliance', 'id'), (SELECT MAX(id) FROM property_compliance));` after the statement, then any new records are created with an id higher than any used so far. That setval is before these seeding statements so presumably new records will be added in with id ~50 (or whatever it is)

I think we can use a similar approach here, start the ids at 1000 (still way above what we're likely to manually seed), then run the `setval` commands after all the inserts?

src/main/resources/data-integration.sql
```diff
@@ -392,3 +392,216 @@ VALUES ('PRSD22', current_date, null, 'urn:fdc:gov.uk:2022:mGHDySEVfCsvfvc6lVWf6
...
+-- NOTE: the offset is added as absolute SECONDS, not as a `days` interval. In
+-- Postgres, `timestamptz + interval 'N days'` is DST-aware under a session whose
+-- timezone observes DST (the app's JDBC connection uses Europe/London): an offset
+-- that crosses the spring-forward boundary (2030-03-31) lands 1 hour short, so
+-- toDays() floors p90/p95 to 89/94. Using absolute seconds (24h * offset) makes the
+-- duration exact regardless of session timezone, so the known values stay 50/90/95.
+-- so "time to first property" takes the values 0, 1, 2, ..., 100 days. The
+-- Metrics service computes percentiles as rank = fraction * (n - 1) with linear
+-- interpolation, so for n = 101 this gives exactly:
+--   median = 50 days, p90 = 90 days, p95 = 95 days, average = 50 days.
```
JasminConterioSW: I think it will be more convincing that it works as expected if you can make it come out as

Median = 22 minutes
p90 = 1 day, x hours, y minutes
p95 = 2 days, x hours, y minutes

src/main/resources/data-integration.sql
```diff
@@ -392,3 +392,216 @@ VALUES ('PRSD22', current_date, null, 'urn:fdc:gov.uk:2022:mGHDySEVfCsvfvc6lVWf6
...
+--
+-- HOW TO VERIFY: open System Operator > Metrics and submit the date range
+--   From: 1 / 1 / 2030    To: 31 / 12 / 2030
+-- Expected: 101 landlord registrations, 101 verified, 101 properties,
```
JasminConterioSW: Maybe worth making a few of the landlords unverified, and a few of them without properties?

It's more convincing if the numbers aren't all the same

src/main/resources/data-integration.sql
JasminConterioSW: Ah, I see you're covering this with the random one.

src/main/resources/data-integration.sql
```diff
@@ -392,3 +392,216 @@ VALUES ('PRSD22', current_date, null, 'urn:fdc:gov.uk:2022:mGHDySEVfCsvfvc6lVWf6
...
+FROM generate_series(1, 101) AS s(i)
+ON CONFLICT DO NOTHING;
+
+INSERT INTO landlordship_members (landlord_id, landlordship_id)
+SELECT 9000000 + i, 9001000 + i
+FROM generate_series(1, 101) AS s(i)
+ON CONFLICT DO NOTHING;
+
+-- =============================================================================
+-- Metrics test cohort 2: realistic / randomised data
```
JasminConterioSW: I don't know how much we need a fully randomised set, it feels like it's quite complicated and isn't necessarily giving the answers we want (I'm getting a lot of times where the median is in days when I give it a try on local)

Maybe the one set giving 50 days, 90 days and 95 days as you have, and another giving more realistic values - but again where you can say ("query between 1/1/2028 and 1/12/2028 to get realistic values")

src/main/resources/data-integration.sql
Bill-Haigh: Copilot suggested leaving the id's high to avoid collisions with AddressBase/NGD since they add addresses with ids spanning 1 ... 7.4 mil

I am inclined to leave it as is unless you have any suggestions?

Travis-Softwire: Why do we need to insert additional addresses at all?

JasminConterioSW: Ah yes I think you'll need them for local but not integration

src/main/resources/data-integration.sql
```diff
@@ -392,3 +392,149 @@ VALUES ('PRSD22', current_date, null, 'urn:fdc:gov.uk:2022:mGHDySEVfCsvfvc6lVWf6
...
+-- =============================================================================
+-- Metrics test cohort 2: deterministic "realistic" data (System Operator > Metrics)
+-- =============================================================================
+-- A second deterministic cohort giving lifelike but reproducible percentiles (the
+-- previous version used random() and so changed on every load). 120 landlords register
+-- across 2028; the first 100 each own one property and the last 20 own none, and ~60%
+-- are verified, so the totals are not all identical. Time to first property is engineered
+-- (sub-hour for ~half, hours for ~a third, days for the rest) so the dashboard shows
+-- median = 22 minutes, p90 = 1 day, p95 = 2 days. The duration formatter shows a single
+-- unit rounded down, so e.g. 1 day 6 hours displays as "1 day". Fixed ids in a high,
```
JasminConterioSW: **The duration formatter shows a single
-- unit rounded down, so e.g. 1 day 6 hours displays as "1 day"**

I don't think I'd twigged this, is that what we want?
Will it show for 1 day 23 hours as 1 day as well?

It might be useful to show at least 2 units

src/main/resources/data-integration.sql
JasminConterioSW: So I think:

- Seed addresses for local
- Don't seed addresses for integration

But everything seeded can have normal ids, then run the `setval ` commands after all the inserts (matching the existing pattern)

JasminConterioSW: (Addresses with ids between 1000 and 2000 all seem to exist on the integration database if you want to use those)

JasminConterioSW (approved):
LGTM once the tests are passing :)

PR #1456, opened 2026-06-16
PDJB-304: Standalone JL flow fixups

TomHanmer29:
## Ticket number

PDJB-304

## Goal of change

QA markups for PDJB-304

Adds a db check to prevent simultaneous journeys saving the same email twice
Makes sure emails aren't sent if the DB update fails

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationService.kt
```diff
@@ -57,9 +57,16 @@ class JointLandlordInvitationService(
...
-            val token = createInvitationToken(email, propertyOwnership, invitingLandlord)
-            val invitationUri = absoluteUrlProvider.buildJointLandlordInvitationUri(token)
+        // Re-check against the current state of the database when finishing the journey. The form-level checks happen
+        // when an email is submitted, so without this a concurrent journey could invite the same email twice.
+        val alreadyInvitedEmails = getExistingInvitedEmails(propertyOwnership.id)
+        val existingLandlordEmails = propertyOwnership.landlords.map { it.email }
+        val emailsToInvite =
+            jointLandlordEmails.filter { it !in alreadyInvitedEmails && it !in existingLandlordEmails }
+
+        emailsToInvite.forEach { email ->
```
TomHanmer29, line 67: This fixes the race condition issue, but also makes it so if the email doesn't send then we don't store the token in the DB. Instead of having a token in the DB when an email didn't send, we now could have an email sending but no token in the DB which would be a user error. I'm not sure which is safer so open to changing this further.

Travis-Softwire: Shouldn't the sequence be:
- Save token
- Try to send email
- If email failed to send, delete the token
?

Travis-Softwire: I'm also not convinced that duplicate emails are that bad in the first place - and in fact there could be cases where say a married couple are both JLs with a 3rd person but they share an email address?

Travis-Softwire (approved):
I've suggested a way of resolving the question you've asked, but happy to chat about it. Subject to that I'm fine with this.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationService.kt
TomHanmer29: Ahh yeah that seems like a very sensible way of going about it. I'll refactor this 👍

PR #1457, opened 2026-06-17
PDJB-268/PDJB-269: Apply transaction correctly on scheduled jobs

samyou-softwire:
## Ticket number

[PDJB-268](https://mhclgdigital.atlassian.net/browse/PDJB-268) [PDJB-269](https://mhclgdigital.atlassian.net/browse/PDJB-269)

## Goal of change

fixes the two feature flagged regular jobs

@Transactional cannot be on the same class as the FF4J flip, this seems to break it and makes it so FF4J can no longer find the bean class. moves it to the application task runner instead, which is equivalent as the task runner just calls the service method

as this entire job is transactional if any email fails or if there is any fail the entire job will break. [PDJB-964](https://mhclgdigital.atlassian.net/browse/PDJB-964) will make sure this alerts us

## Checklist


TomHanmer29 (approved):
LGTM!

PR #1459, opened 2026-06-17
[PR 1 of 2] PDJB-318: deregister property confirmation page

TomHanmer29:
## Ticket number

PDJB-318

## Goal of change

Adds the deregister property confirm page

image

Adds the confirm page.
Doesn't add any emails

## Checklist

  mention that here

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/SessionAttributes.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/SessionAttributes.kt
@@ -6,7 +6,7 @@ const val LOCAL_COUNCIL_USER_ID = "local-council-user-id"
 
 const val PROPERTY_REGISTRATION_NUMBER = "propertyRegistrationNumber"
 
-const val PROPERTIES_DEREGISTERED_THIS_SESSION = "propertiesDeregisteredThisSession"
+const val PROPERTIES_DEREGISTERED_THIS_SESSION = "propertiesDeregisteredThisSessionWithAddresses"
 
 const val LANDLORD_HAD_ACTIVE_PROPERTIES = "landlordHadActiveProperties"
 
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/SessionAttributes.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/SessionAttributes.kt
@@ -6,7 +6,7 @@ const val LOCAL_COUNCIL_USER_ID = "local-council-user-id"
 
 const val PROPERTY_REGISTRATION_NUMBER = "propertyRegistrationNumber"
 
-const val PROPERTIES_DEREGISTERED_THIS_SESSION = "propertiesDeregisteredThisSessionWithAddresses"
+const val PROPERTIES_DEREGISTERED_THIS_SESSION_WITH_ADDRESSES = "propertiesDeregisteredThisSessionWithAddresses"
 
 const val LANDLORD_HAD_ACTIVE_PROPERTIES = "landlordHadActiveProperties"
 
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/SessionAttributes.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/SessionAttributes.kt
@@ -6,7 +6,7 @@ const val LOCAL_COUNCIL_USER_ID = "local-council-user-id"
 
 const val PROPERTY_REGISTRATION_NUMBER = "propertyRegistrationNumber"
 
-const val PROPERTY_DEREGISTERED_THIS_SESSION = "propertyDeregisteredThisSession"
+const val PROPERTIES_DEREGISTERED_THIS_SESSION_WITH_ADDRESSES = "propertiesDeregisteredThisSessionWithAddresses"
 
 const val LANDLORD_HAD_ACTIVE_PROPERTIES = "landlordHadActiveProperties"
 
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/SessionAttributes.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/SessionAttributes.kt
@@ -6,7 +6,7 @@ const val LOCAL_COUNCIL_USER_ID = "local-council-user-id"
 
 const val PROPERTY_REGISTRATION_NUMBER = "propertyRegistrationNumber"
 
-const val PROPERTIES_DEREGISTERED_THIS_SESSION_WITH_ADDRESSES = "propertiesDeregisteredThisSessionWithAddresses"
+const val PROPERTY_DEREGISTERED_THIS_SESSION = "propertyDeregisteredThisSession"
 
 const val LANDLORD_HAD_ACTIVE_PROPERTIES = "landlordHadActiveProperties"
 
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyController.kt
@@ -106,7 +106,7 @@ class DeregisterPropertyController(
         model.addAttribute("landlordDashboardUrl", LANDLORD_DASHBOARD_URL)
 
         return if (featureFlagManager.checkFeature(JOINT_LANDLORDS)) {
-            model.addAttribute("address", propertyDeregistrationService.getDeregisteredPropertyAddress())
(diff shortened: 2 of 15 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/SessionAttributes.kt
```diff
@@ -6,7 +6,7 @@ const val LOCAL_COUNCIL_USER_ID = "local-council-user-id"
 
 const val PROPERTY_REGISTRATION_NUMBER = "propertyRegistrationNumber"
 
-const val PROPERTIES_DEREGISTERED_THIS_SESSION = "propertiesDeregisteredThisSession"
+const val PROPERTIES_DEREGISTERED_THIS_SESSION = "propertiesDeregisteredThisSessionWithAddresses"
```
samyou-softwire: should the constant name update too?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyDeregistrationService.kt
```diff
@@ -25,14 +25,20 @@ class PropertyDeregistrationService(
...
+
+    fun getDeregisteredPropertyOwnershipIdsFromSession(): MutableList<Long> = getDeregisteredPropertiesFromSession().keys.toMutableList()
+
+    fun getDeregisteredPropertyAddress(propertyOwnershipId: Long): String? = getDeregisteredPropertiesFromSession()[propertyOwnershipId]
 
     @Suppress("UNCHECKED_CAST")
-    fun getDeregisteredPropertyOwnershipIdsFromSession(): MutableList<Long> =
-        session.getAttribute(PROPERTIES_DEREGISTERED_THIS_SESSION) as MutableList<Long>?
-            ?: mutableListOf()
+    private fun getDeregisteredPropertiesFromSession(): MutableMap<Long, String> =
```
samyou-softwire: does this need to be a map/list? I don't think we need to store more than one

src/main/resources/templates/deregisterPropertyConfirmationJune26Redesign.html
samyou-softwire, line 1: I think for page redesigns (eg compliance actions page redesign) we opted for `xyzold` and `xyzJune26Redesign` in case we need to track more than one old version

src/main/resources/templates/deregisterPropertyConfirmation.html
```diff
@@ -1,21 +1,32 @@
 <!DOCTYPE html>
 <html th:replace="~{fragments/layout :: layout(#{deregisterProperty.title}, ~{::main}, false)}">
-<main class="govuk-main-wrapper" id="main-content">
-    <div id="page-contents"
-         th:replace="~{fragments/layouts/twoThirdsLayout :: twoThirdsLayout(~{::#page-contents/content()})}">
-        <div id="confirmation-banner"
-             th:replace="~{fragments/banners/confirmationPageBanner :: confirmationPageBanner(#{deregisterProperty.confirmation.banner.heading}, null)}">
+    <main class="govuk-main-wrapper" id="main-content">
+        <div id="page-contents" th:replace="~{fragments/layouts/twoThirdsLayout :: twoThirdsLayout(~{::#page-contents/content()})}">
+
```
samyou-softwire: dont think this space is needed

src/main/resources/templates/deregisterPropertyConfirmation.html
```diff
@@ -1,21 +1,32 @@
...
+            <p class="govuk-body" th:text="#{deregisterProperty.confirmation.body.two}">
+                deregisterProperty.confirmation.body.two
+            </p>
+
+            <p class="govuk-body">
+                <a class="govuk-link"
+                   th:href="@{${landlordDashboardUrl}}"
+                   th:text="#{common.confirmationPage.goToDashboard}">common.confirmationPage.goToDashboard</a>
+            </p>
+
```
samyou-softwire: or this

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDeregistrationJourneyTests.kt
samyou-softwire, line 1: do we want to keep some integration tests around that verify we still get the old view & it all works if the feature flag is off?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyDeregistration/stepConfig/ReasonStepConfig.kt
```diff
@@ -41,7 +41,7 @@ class ReasonStepConfig(
         val propertyAddress = propertyOwnership.address.singleLineAddress
 
         propertyDeregistrationService.deregisterProperty(state.propertyOwnershipId)
-        propertyDeregistrationService.addDeregisteredPropertyOwnershipIdToSession(state.propertyOwnershipId)
+        propertyDeregistrationService.addDeregisteredPropertyOwnershipIdToSession(state.propertyOwnershipId, propertyAddress)
```
samyou-softwire: something seems off here; the reason step only appears on the old journey I think so why does the address get stored in our session?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyController.kt
```diff
@@ -105,7 +105,13 @@ class DeregisterPropertyController(
 
         model.addAttribute("landlordDashboardUrl", LANDLORD_DASHBOARD_URL)
 
-        return "deregisterPropertyConfirmation"
+        return if (featureFlagManager.checkFeature(JOINT_LANDLORDS)) {
+            model.addAttribute("address", propertyDeregistrationService.getDeregisteredPropertyAddress(propertyOwnershipId))
```
samyou-softwire, line 109: I do wonder why we store the address in the session state at all. at this point we have access to the property registration ID and we have confirmed it's the one in the session. couldn't we retrieve the address when we render the page?

samyou-softwire (changes requested):
looks good, only two I want to check in are the last two comments

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyController.kt
TomHanmer29: At the point we show the confirm page, the property is deleted so I don't think we can?

samyou-softwire: ah good point, ok agree then session is necessary

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDeregistrationJourneyTests.kt
TomHanmer29: This file actually already has integration tests for both new and old, it just updates some contents for the new ones. Previously the tests used the same page for new and old

samyou-softwire (approved):
LGTM, to check my understanding so I'm happy this won't break anything, previously we stored the prop ID in your session only for the purpose of deciding whether to show you the confirmation page or not. now, we store the ID and the address as we check the presence of the ID and if so show the relevant address?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyController.kt
```diff
@@ -128,10 +134,10 @@ class DeregisterPropertyController(
             .getIsPrimaryLandlord(propertyOwnershipId, principal.name)
 
     private fun checkPropertyHasBeenDeregisteredInThisSession(propertyOwnershipId: Long) {
-        if (propertyOwnershipId !in propertyDeregistrationService.getDeregisteredPropertyOwnershipIdsFromSession()) {
+        if (propertyOwnershipId != propertyDeregistrationService.getDeregisteredPropertyOwnershipIdFromSession()) {
```
Travis-Softwire: I think I'm missing something - why do we now think we can only deregister one property in a session?

Travis-Softwire (commented):
I'm fine with the substantive change, but confused why we're now limiting to one deregistration per session? I think subject that it's fine but want to understand that before approving

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterPropertyController.kt
samyou-softwire: to me it felt like a lot of mechanism for not much gain so I requested it here; it's what we did on the cancel invitation flow. happy to leave as is though if we feel it's worthwhile

PR #1461, opened 2026-06-17
PDJB-268/PDJB-269: Move the transaction away from exitProcess

samyou-softwire:
## Ticket number

[PDJB-268](https://mhclgdigital.atlassian.net/browse/PDJB-268) [PDJB-269](https://mhclgdigital.atlassian.net/browse/PDJB-269)

## Goal of change

fix it so the transaction block does not contain exitProcess, since exiting the process reverts the transaction

refactors all tasks to follow this pattern

also updates tests as we now have a reason to split the task logic away from the runner so no more private method accessing

## Checklist


TomHanmer29 (approved):
Think I'm happy with this, so long as it all seems to run locally :)

PR #1462, opened 2026-06-17
PDJB-1035: Add Plausible domain-id config for Stats API metrics

Bill-Haigh:
## Ticket number

PDJB-1035

## Goal of change

Fix the journey completion-rate metrics returning no data on deployed environments by separating the two Plausible identifiers that were incorrectly sharing a single config value.

- Adds a second Plausible config value so the browser tracking snippet and the Stats API use the correct identifier each:
  - `plausible.site-id` (`PLAUSIBLE_SITE_ID`) — unchanged; the `pa-<id>` installation id used to build the browser tracking snippet URL.
  - `plausible.domain-id` (`PLAUSIBLE_DOMAIN_ID`) — new; the site **domain**, used by the Stats API to fetch completion-rate metrics.
- Updates `PlausibleMetricsService` to query the Stats API with `domain-id`; `GlobalModelAttributes` continues to build the snippet from `site-id`.
- Updates `AnalyticsReadMe.md` to document the two distinct values and their per-environment meaning.

A matching infra change is required: every deployed environment keeps its existing `PLAUSIBLE_SITE_ID` (the `pa-` id) and must additionally set a new `PLAUSIBLE_DOMAIN_ID` (the domain, e.g. `register-home-to-rent.communities.gov.uk` for non-prod). Without `PLAUSIBLE_DOMAIN_ID` the app will fail to start, so infra must be updated before this deploys.

## Checklist

- [x] Branch is based on the latest `main`

Note: this is a config split with an existing unit test updated to match; the full suite has not yet been run locally.

### DIFF
--- a/src/main/resources/application-local.yml
+++ b/src/main/resources/application-local.yml
@@ -88,6 +88,7 @@ management:
 
 plausible:
   site-id: local-no-analytics
+  domain-id: local-no-analytics
   base-url: http://localhost:${server.port}/local/plausible
   api-key: local-no-analytics
 
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -205,7 +205,7 @@
         "filename": "src\\main\\resources\\application-local.yml",
         "hashed_secret": "ce8c74ce64b55dc94183540aac1aee68fd6d27ad",
         "is_verified": false,
-        "line_number": 135,
+        "line_number": 136,
         "is_secret": false
       }
     ],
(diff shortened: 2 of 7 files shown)


src/main/resources/application-local.yml
```diff
@@ -88,6 +88,7 @@ management:
 
 plausible:
   site-id: local-no-analytics
```
JasminConterioSW, line 90: Not for this PR, but is it worth adding a profile so that the real plausible can be run locally? Similar to use-notify for sending real emails. It might be useful for testing?

JasminConterioSW (approved):
LGTM :)

src/main/resources/application-local.yml
Bill-Haigh: good point

PR #1467, opened 2026-06-18
Revert "PDJB-1033: Add metrics test cohorts to local and integration …

AEPR:
## Ticket number
PDJB-NONE

## Goal of change
Revert SQL change breaking deployments.


samyou-softwire (approved):
this reverts #1450

PR #1464, opened 2026-06-17
[PR 2 of 2] PDJB-318: deregister property emails

TomHanmer29:
## Ticket number

PDJB-318

## Goal of change

Adds the new deregister property emails for PDJB-318

## Description of main change(s)

Updates the email copy for the email to the landlord (only 1 left at this point)
Adds a new email that's sent to all landlords that have been invited, saying their invite is now void

## Anything you'd like to highlight to the reviewer?

I've added a few questions for Chris/Geetika on the ticket for this, but have implemented what I think will be the answer in the interest of time.
EDIT: These are now resolved

## Checklist

Delete any that are not applicable, and add explanation below for any that are applicable but haven't been done

- [ ] Screenshots of any UI changes have been added
- [x] Unit tests for new logic (e.g. new service methods) have been added
- [x] New email templates have been added to `/src/main/kotlin/resources/emails/emailTemplates.json`
- [x] Test suite has been run in full locally and is passing
- [x] Branch has been rebased onto main and run locally, with everything working as expected (both for your new feature
  and any related functionality)
- [ ] TODO comments referencing this JIRA ticket have been searched for and removed - if a future PR will address them,
  mention that here
- [ ] QA instructions have been added to the ticket (particularly if this is the last PR required to complete the ticket)

PR #1471, opened 2026-06-19
PDJB-308: Send a nudge email when the property has one landlord

samyou-softwire:
## Ticket number

[PDJB-308](https://mhclgdigital.atlassian.net/browse/PDJB-308)

## Goal of change

sends out an email when you become the only landlord on the property

preview of only landlord email

adds a new service to manage this and puts calls to it in appropriate places

adds a few new service calls to make a single entry point for cancelling invitations and removing landlords. this should make it so this keeps working if we add new ways to cancel invites / remove landlords from a property in the future

## Checklist

  mention that here
- [x] Any special release instructions (e.g. the database will need resetting) have been added as checklist items to a draft PR (merging
  `main` into `test`) for the next release


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationService.kt
```diff
@@ -43,6 +44,11 @@ class JointLandlordInvitationService(
         return Pair(pending, expired)
     }
 
+    fun getPendingInvitations(propertyOwnership: PropertyOwnership): List<JointLandlordInvitation> =
```
samyou-softwire: this has a conflict with #1464 but the call signatures are the same so should be fine

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/SwapToIndividualNudgeEmailService.kt
```diff
@@ -0,0 +1,47 @@
...
+import uk.gov.communities.prsdb.webapp.database.entity.PropertyOwnership
+import uk.gov.communities.prsdb.webapp.models.viewModels.emailModels.SwapToIndividualNudgeEmail
+
+@PrsdbFlip(name = JOINT_LANDLORDS, alterBean = "swap-to-individual-nudge-email-flag-on")
+interface SwapToIndividualNudgeEmailService {
+    fun sendNudgeEmailIfApplicable(propertyOwnership: PropertyOwnership)
+}
+
+@Primary
+@Service("swap-to-individual-nudge-email-flag-off")
```
samyou-softwire, line 18: this is uniqely a service as it needs to be callable from both the expire invitations scheduled task and from the web. are there any more suitable annotations we can give?

Travis-Softwire: We do already have a bunch of plain `@Service`'s for that exact reason :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationService.kt
```diff
@@ -233,9 +239,19 @@ class JointLandlordInvitationService(
         return invitation
     }
 
+    fun removeInvitation(invitation: JointLandlordInvitation) {
+        invitationRepository.delete(invitation)
+    }
+
+    /**
+     * Similar to removeInvitation but it will run checks if we need to send an email that the property now only has one landlord.
+     * Use in cases of where the deletion is an action (pressing cancel invitation) and not part of a mechanism (resending invites, deleting accepted invites).
```
Travis-Softwire: This suggests to me that either this logic should live in the relevant journey OR those other actions should also live on the service as atomic function calls.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/SwapToIndividualNudgeEmailService.kt
```diff
@@ -0,0 +1,54 @@
...
+import uk.gov.communities.prsdb.webapp.database.repository.JointLandlordInvitationRepository
+import uk.gov.communities.prsdb.webapp.models.viewModels.emailModels.SwapToIndividualNudgeEmail
+
+@PrsdbFlip(name = JOINT_LANDLORDS, alterBean = "swap-to-individual-nudge-email-flag-on")
+interface SwapToIndividualNudgeEmailService {
+    fun sendNudgeEmailIfApplicable(propertyOwnership: PropertyOwnership)
+}
+
+@Primary
+@Service("swap-to-individual-nudge-email-flag-off")
```
Travis-Softwire, line 18: What's the reasoning behind using it's own flag vs using the JL flag?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationExpiryEmailService.kt
```diff
@@ -45,6 +46,7 @@ class JointLandlordInvitationExpiryEmailServiceImplFlagOn(
                 invitation.markAsExpiredEmailSent()
                 invitationRepository.save(invitation)
                 expiredIds.add(invitation.id)
+                swapToIndividualNudgeEmailService.sendNudgeEmailIfApplicable(invitation.registeredOwnership)
```
Travis-Softwire, line 49: What happens if more than one invitation for the property expires in the same day?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -366,6 +367,14 @@ class PropertyOwnershipService(
         propertyOwnershipRepository.deleteAll(propertyOwnerships)
     }
 
+    fun removeLandlord(
```
Travis-Softwire, line 380: This is currently only called from one place inside a transaction, but I don't think we can guarantee that in the future. Calls to `propertyOwnership.removeLandlord` will need to be in a transaction because they mutate the join table. For defense in depth I think this function should probably be transactional

Travis-Softwire (approved):
A couple of Qs/ suggestions. LGTM once those are resolved but shout if you want to discuss

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/SwapToIndividualNudgeEmailService.kt
samyou-softwire: not sure what you mean? this is the service name, we use the JL flag above `@PrsdbFlip(name = JOINT_LANDLORDS, alterBean = "swap-to-individual-nudge-email-flag-on")`

unless I misunderstand how these work

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationExpiryEmailService.kt
samyou-softwire: I think this should be fine, if any invitations other than the last one expires we don't want anything to happen, and if the last one expires then there will be no more invitations in the list

checking for sending nudge emails should ensure there should be one nudge email sent & no more

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/SwapToIndividualNudgeEmailService.kt
Travis-Softwire: Yep, you're right, ignore me!

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationExpiryEmailService.kt
Travis-Softwire: Where do we check that it's the last expired invitation?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
samyou-softwire: good point, have done so

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationService.kt
samyou-softwire: have split it back out to calling the nudge service

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordInvitationExpiryEmailService.kt
samyou-softwire: ah I see your point; we just check that there's no pending invites so if two expire on the same day they'll both send nudge emails

I've added another check now that makes it not send the nudge email if there are any invitations that are expired but haven't had the email sent yet which should prevent double emails

PR #1472, opened 2026-06-19
PDJB-1033: Re-add metrics demo seed cohorts and multi-unit duration formatter

Bill-Haigh:
## Ticket number

PDJB-1033

## Goal of change

Re-add the deterministic "metrics demo" seed cohorts so the System Operator → Metrics page is populated locally and in integration, without crash-looping the integration container on boot (the reason the original cohorts were reverted).

- Re-adds two deterministic metrics demo cohorts (originally added in #1450, reverted by #1467) to `data-integration.sql` and a dedicated `data-metrics-local.sql`, populating the Metrics dashboard for the 2030 and 2028 reporting periods.
- The metrics duration formatter now shows **every** non-zero unit down to minutes (e.g. `1 day, 6 hours, 21 minutes`) instead of rounding down to a single unit.
- Extracts the local metrics seed into a self-contained `data-metrics-local.sql` (system operator + both cohorts), wired via explicit `sql.init` `data-locations`. `MetricsSinglePageTests` now seeds **only** this script, so the local demo data and the integration test share a single source of truth.

### This is the fix
- Makes the integration seed boot-safe: the `free_address` CTE now caps the candidate set with an inner `ORDER BY id LIMIT n` **before** `ROW_NUMBER()`, giving an index-driven plan instead of the full-table scan that crash-looped the container on every boot. All cohort inserts are idempotent (fixed ids + `ON CONFLICT DO NOTHING`) under `spring.sql.init mode: always`.

- Three commits, reviewable separately: (1) SQL seed cohorts, (2) duration formatter, (3) local-seed extraction into `data-metrics-local.sql`.
- `data-metrics-local.sql` and `data-integration.sql` deliberately differ **only** in address handling: local seeds its own addresses (no AddressBase data locally) while integration claims existing free addresses. Every metric-producing field is identical, so the dashboard numbers match. Verified live on a fresh local DB: 2030 → 121/97/101 and 50/90/95 days; 2028 → 120/72/100, median 22 minutes, p90 `1 day, 6 hours, 21 minutes`, p95 `2 days, 43 minutes`.

## Checklist

- [x] Test suite has been run in full locally and is passing (`testWithoutIntegration`, `MetricsSinglePageTests`, `ktlintCheck`)


Travis-Softwire (approved):
LGTM - I had copilot run some empirical tests locally w/ 5m addresses to check :)

PR #1475, opened 2026-06-22
PDJB-1031: fix edge case with has problems screen incorrectly appearing for expired EPC

TomHanmer29:
## Ticket number

PDJB-1031

## Goal of change

Previously the "has problems" screen in property registration appeared when the certificate was expired, but still in date when the tenancy began. This fixes that

Adds a check for the epc in date when tenacy began to the check.

On the compliance tab, we still show the expired banner in this case, which I think is correct but just wanted to flag. May be worth changing in this ticket if we disagree with how the behaviour is currently.

## Checklist

### DIFF
--- a/src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/HasMissingComplianceStepConfigTests.kt
+++ b/src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/HasMissingComplianceStepConfigTests.kt
@@ -336,6 +336,23 @@ class HasMissingComplianceStepConfigTests {
         }
 
         @Test
+        fun `returns true when accepted epc is expired with tenancy started before expiry but has low rating and no mees exemption`() {
+            whenever(mockState.hasEpcStep).thenReturn(mock<HasEpcStep>())
+            val mockEpc = mock<EpcDataModel>()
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/HasMissingComplianceStepConfigTests.kt
```diff
@@ -309,15 +309,32 @@ class HasMissingComplianceStepConfigTests {
...
             whenever(mockEpc.isPastExpiryDate()).thenReturn(true)
             whenever(mockState.acceptedEpcIfStillAccepted).thenReturn(mockEpc)
+            val mockTenancyStep = mock<EpcInDateAtStartOfTenancyCheckStep>()
+            whenever(mockTenancyStep.outcome).thenReturn(EpcInDateAtStartOfTenancyCheckMode.NOT_IN_DATE)
+            whenever(mockState.epcInDateAtStartOfTenancyCheckStep).thenReturn(mockTenancyStep)
 
             assertTrue(HasMissingComplianceStepConfig.isEpcInvalid(mockState))
         }
 
+        @Test
```
Travis-Softwire, line 324: Worth including a test for expired + occupied + tenency started before expiry + low epc + no MEES -> true?

Travis-Softwire (approved):
One suggestion, otherwise LGTM

PR #1476, opened 2026-06-22
Feat/pdjb 299 show multiple landlords on property record

JasminConterioSW:
## Ticket number

PDJB-299

## Goal of change

Display multiple landlords on a property record (landlord view only)

* Refactor PropertyDetailsController slightly to only check the JOINT_LANDLORDS feature flag in one place
* Update the landlord view only of the property record to show multiple landlords using a summary card and now only showing the landlords LRN and email address.

image

## Checklist

  mention that here


src/main/resources/data-local.sql
```diff
@@ -56,7 +56,8 @@ VALUES ('urn:fdc:gov.uk:2022:KLMNO', true, 1, '10/07/24', '10/07/24', 'Ford Pref
         'Team-PRSDB+laadmin@softwire.com', true),
        ('urn:fdc:gov.uk:2022:cgVX2oJWKHMwzm8Gzx25CSoVXixVS0rw32Sar4Om8vQ', false, 1, '10/15/24', '10/15/24',
         'PRSDB La User',
-        'Team-PRSDB+lauser@softwire.com', true);
+        'Team-PRSDB+lauser@softwire.com', true),
```
JasminConterioSW, line 59: When we run locally, the mock user is logged in with One-Login, so this makes it easier to automatically be able to see the local council pages

.secrets.baseline
```diff
@@ -276,6 +276,92 @@
         "is_verified": false,
         "line_number": 21
       }
+    ],
```
JasminConterioSW, line 269: This is to stop `detect-secrets-hook` complaining

src/main/resources/data-integration.sql
```diff
@@ -295,7 +295,9 @@ VALUES (1, 1, '2025-01-15'),
        (1, 45, '2025-01-15'),
        (1, 46, '2025-01-15'),
        (1, 47, '2025-01-15'),
-       (1, 48, '2025-01-15') ON CONFLICT DO NOTHING;
+       (1, 48, '2025-01-15'),
+       (6, 1, '2025-01-15'),
+       (7, 1, '2025-01-15') ON CONFLICT DO NOTHING;
```
JasminConterioSW, line 300: Seed multiple landlords onto one of the properties

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDetailsTests.kt
```diff
@@ -198,6 +181,50 @@ class PropertyDetailsTests : IntegrationTestWithImmutableData("data-local.sql")
                 )
             }
         }
+
+        @Test
+        fun `when joint landlords flag is enabled the landlord tab shows summary cards`(page: Page) {
```
Travis-Softwire: Worth a test to check multiple cards display correctly?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsLandlordViewModelBuilder.kt
```diff
@@ -50,5 +51,33 @@ class PropertyDetailsLandlordViewModelBuilder {
...
+
+        fun buildSummaryCards(
+            landlords: Set<Landlord>,
+            currentUserId: String,
+        ): List<SummaryCardViewModel> =
+            landlords
+                .sortedWith(compareByDescending<Landlord> { it.baseUser.id == currentUserId }.thenBy { it.name })
+                .map { landlord ->
+                    val isCurrentUser = landlord.baseUser.id == currentUserId
+                    val title = if (isCurrentUser) "${landlord.name} (you)" else landlord.name
```
Travis-Softwire: I don't think we generally assemble raw strings like this? Could we not either amend the view model so that this can be passed as a parameterised message key of `{0} (you)` or if that's going to be a pain we could use the the message source to resolve the parameterised message key instead?

src/main/resources/data-local.sql
Travis-Softwire: It depends which endpoint you load first in the session. If you navigate to `localhost:<port>/local-council` before any other endpoint then it will log you in with the mock IA service instead of the mock OL one

Travis-Softwire (approved):
A couple of minor points, but otherwise LGTM

src/main/resources/data-local.sql
JasminConterioSW: Ah ok. It's quite handy to have both options sometimes to compare landlord and local council version, I've just been modifying the db!

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsLandlordViewModelBuilder.kt
JasminConterioSW: Oops, yes that " (you)" definitely shouldn't be there

PR #1479, opened 2026-06-22
PDJB-307: Fix initial step in journey

samyou-softwire:
## Ticket number

[PDJB-307](https://mhclgdigital.atlassian.net/browse/PDJB-307)

## Goal of change

fix two issues on the journey

- give it a proper start step as we immediately need to redirect. based on `StartInviteJointLandlordStepConfig`
- fix the route to be consistent with the other prop details journeys `/landlord/property-details/<id>/switch-to-individual/<step>`

## Checklist


TomHanmer29 (approved):
LGTM!

PR #1481, opened 2026-06-23
PDJB-1131: Add org landlords feature flag

samyou-softwire:
## Ticket number

[PDJB-1131](https://mhclgdigital.atlassian.net/browse/PDJB-1131)

## Goal of change

adds a feature flag for org landlords which will enable the new journey

adds it so it is 
- disabled on prod
- disabled on nft
- disabled on test
- enabled on int
- enabled on local

## Checklist


TomHanmer29 (approved):
LGTM! Maybe worth making sure there's some consensus in #dev about the naming structure with the PR included but otherwise I'm happy

samyou-softwire:
I'm happy with it, we can always rename the flag in the future anyway

PR #1484, opened 2026-06-23
PDJB-1051: fix virus scan error

TomHanmer29:
## Ticket number

PDJB-1051

## Goal of change

Fixes the race condition bug with the virus scan

Adds pessimistic locking to findByObjectKeyAndVersionId, and adds a new method findWithLockById on fileUpload, to prevent the attempted writing of a field that no longer exists

Will clean up the old/orphaned records in a new PR / manually

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/FileUploadRepository.kt
```diff
@@ -1,9 +1,25 @@
...
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Lock
+import org.springframework.data.jpa.repository.QueryHints
 import uk.gov.communities.prsdb.webapp.database.entity.FileUpload
 
+// Bounds how long a locking finder waits for the file_upload row lock before failing. The scan-processor holds this
+// lock across non-transactional S3 work, so without a bound a hung S3 call could pin the row (and its database
+// connection) indefinitely. On PostgreSQL a query timeout cancels the whole SELECT ... FOR UPDATE - including the
+// lock wait - raising a QueryTimeoutException that rolls the transaction back.
+const val FILE_UPLOAD_LOCK_WAIT_TIMEOUT_MILLIS = "10000"
```
TomHanmer29: Have added a timeout for the locking which should go to the generic error page if exceeded, there may be a more graceful way to handle this but not sure 🤔

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/FileUploadRepositoryConcurrencyTests.kt
```diff
@@ -0,0 +1,225 @@
...
+        // Assert
+        assertTrue(errors.isEmpty(), "Unexpected errors on the lock holder: $errors")
+        val thrown =
+            assertNotNull(
+                contenderError.get(),
+                "Expected the contending finder to abort with a lock-wait timeout, but it returned normally after " +
+                    "${contenderNanos.get() / 1_000_000}ms. The query.timeout hint is not bounding the FOR UPDATE lock wait.",
+            )
+        assertTrue(
+            isLockWaitTimeout(thrown),
```
samyou-softwire: this feels an oddly permissive test; we should know exactly what kind of error is thrown here but this method does some odd stuff with checking the class name?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/FileUploadRepository.kt
```diff
@@ -1,9 +1,25 @@
...
+
 interface FileUploadRepository : JpaRepository<FileUpload, Long> {
+    @Lock(LockModeType.PESSIMISTIC_WRITE)
+    @QueryHints(QueryHint(name = "jakarta.persistence.query.timeout", value = FILE_UPLOAD_LOCK_WAIT_TIMEOUT_MILLIS))
+    fun findWithLockById(id: Long): FileUpload?
+
+    @Lock(LockModeType.PESSIMISTIC_WRITE)
+    @QueryHints(QueryHint(name = "jakarta.persistence.query.timeout", value = FILE_UPLOAD_LOCK_WAIT_TIMEOUT_MILLIS))
     fun findByObjectKeyAndVersionId(
```
samyou-softwire, line 7: should this method name also declare it acquires a lock?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -158,23 +159,30 @@ class PropertyComplianceService(
...
+        electricalSafetyCertUploadIds.forEach { uploadId ->
+            if (fileUploadIsStillQuarantined(uploadId)) {
+                virusScanCallbackService.deleteAllCallbacksForFileUpload(uploadId)
+                virusScanCallbackService.saveEmailToMonitoringTeam(propertyOwnershipId, uploadId, electricalCertType!!)
+                virusScanCallbackService.saveEmailToOwner(propertyOwnershipId, uploadId, electricalCertType)
+            }
         }
     }
 
+    private fun fileUploadIsStillQuarantined(fileUploadId: Long): Boolean =
```
samyou-softwire: I wonder if this method name is quite clear the goal of why we're doing this. it's kind of doing two things:
- blocking the thread till it can get a lock, so we know that the scan result thread isn't doing anything here
- after it gets the lock, check its still quarantined

samyou-softwire (approved):
this is quite a subtle fix we're putting in so I'm conscious that we name stuff as descriptively as we can. happy with the logic here

Travis-Softwire:
Hmm, can we really not eliminate the race condition rather than locking? I think I had in my head that we'd switch to updating the callback rows in place with a bulk update or something along those lines

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/VirusScanProcessingService.kt
```diff
@@ -23,12 +24,15 @@ class VirusScanProcessingService(
         locator: UploadedFileLocator,
         scanResultStatus: ScanResult,
     ) {
+        val fileUpload =
```
Travis-Softwire: Do we still need these changes the scan processor?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/VirusScanCallbackRepository.kt
```diff
@@ -13,6 +15,12 @@ interface VirusScanCallbackRepository : JpaRepository<VirusScanCallback, Long> {
     @Suppress("ktlint:standard:function-naming")
     fun findAllByFileUpload_Id(fileUploadId: Long): List<VirusScanCallback>
 
-    @Suppress("ktlint:standard:function-naming")
-    fun deleteAllByFileUpload_Id(fileUploadId: Long)
+    // Set-based update so that re-pointing a callback whose row a concurrent scan has already deleted is a harmless
+    // no-op rather than a stale-state failure (a loaded-entity save would fail Hibernate's row-count check on 0 rows).
+    @Modifying
```
Travis-Softwire, line 21: As this always needs to be called inside a transaction probably worth adding the annotation here as well in case anyone calls it in future from outside a transaction

Travis-Softwire (approved):
A couple of minor points, but LGTM once they're addressed :)

PR #1487, opened 2026-06-24
PDJB-1131: Part 2: Skeleton journey

samyou-softwire:
## Ticket number

[PDJB-1131](https://mhclgdigital.atlassian.net/browse/PDJB-1131)

## Goal of change

adds a skeleton journey for org landlords to unblock other tickets

## Description of main change(s)

adds a journey factory for the new flow after we turn the feature flag on

adds many new todo steps

## Anything you'd like to highlight to the reviewer?

I think the way we plug into the original LL registration flow could use some more work, I don't think it's quite explicit on what we want to do here. I raise this PR as an MVP that at least adds the skeleton journey and should unblock the other tickets surrounding it

I'll raise a part 3 that integrates this in a way I believe we can keep permanently (or something close to that)

## Checklist

Delete any that are not applicable, and add explanation below for any that are applicable but haven't been done

- [ ] Screenshots of any UI changes have been added
- [ ] Unit tests for new logic (e.g. new service methods) have been added (I've skipped testing for now as these are WIP skeleton pages)
- [ ] Controller tests for any new endpoints, including testing the relevant permissions
- [ ] Single page integration tests have been added for any unhappy-flow UI features, e.g. validation errors
- [ ] New journey steps have been added to the appropriate journey integration test(s)
- [ ] A new journey integration test has been added for any new journeys
- [ ] Test suite has been run in full locally and is passing
- [ ] Branch has been rebased onto main and run locally, with everything working as expected (both for your new feature
  and any related functionality) (not based on main)
- [x] TODO comments referencing this JIRA ticket have been searched for and removed - if a future PR will address them,
  mention that here

PR #1491, opened 2026-06-24
PDJB-1131: Part 3: Implement a better org landlord journey

samyou-softwire:
## Ticket number

[PDJB-1131](https://mhclgdigital.atlassian.net/browse/PDJB-1131)

## Goal of change

Implement a better structure for the org landlords journey

sets up a new landlord registration task. this task has the following steps:

- privacy notice (shared on all flows)
- IDV task
- branch point on if org landlords feature flag is enabled
- if it isn't, send to original registration task
- if it is, ask the org landlords question
  - if they say yes, send to new org landlords task
  - if they say no, send to original registration task

to suppose this, splits out all the questions on the landlord registration task into a new task that can be included where desired

updates a lot of the todos to be more accurate

`OrgLandlordFeatureGateStep` is a bit novel, it's a step that just checks the feature flag. I thought this seemed most sensible as we don't really have a precedent yet for feature flagged tasks. doing it this way gives a consistent single journey definition and a single state definition to handle that journey. feature flagging the journey itself is a bit finnicky as there are two journeys that contain landlord registration (normal registration, and JL invite accept), so this way they can just include LandlordRegistrationTask and worry no further

a lot of the files in #1487 were in a separate package to the landlord registration files. theyve been moved to share the same package

## Checklist

  mention that here
- [x] QA notes added


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
```diff
@@ -31,44 +32,50 @@ class LandlordRegistrationTask : Task<LandlordRegistrationState>() {
             }
             task(journey.identityTask) {
                 parents { journey.privacyNoticeStep.isComplete() }
-                nextStep { journey.emailStep }
+                nextStep { journey.orgLandlordFeatureGateStep }
```
TomHanmer29: Not sure how I feel about having a journey step that is just used for the purpose of feature flagging, it does seem pretty clean but I'm also not sure that it's immediately obvious what it does? I think elsewhere we've routed based on a feature flag using a strategy file (see jointLandlordsPropertyRegistrationStrategy), but not sure if that's super clean either tbf.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgLandlordCyaStepConfig.kt
```diff
@@ -7,6 +7,8 @@ import uk.gov.communities.prsdb.webapp.journeys.JourneyStep.RequestableStep
 import uk.gov.communities.prsdb.webapp.journeys.shared.Complete
 import uk.gov.communities.prsdb.webapp.models.requestModels.formModels.NoInputFormModel
 
+// TODO: PDJB-1168 this shouldn't really be a separate task at all, this should extend the existing landlord registration CYA page.
+// this was added as a separate page for easier scaffolding
```
TomHanmer29: Not 100% sure I follow how this makes the scaffolding easier? Just because we don't have anything to hook up to it?

TomHanmer29 (commented):
Not 100% on the flag step but other than that looks ok! If someone else is happy feel free to approve

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
JasminConterioSW: Hmm, the strategy pattern might be a better one to use in general for journeys with different behaviour depending on the feature flag status. For this particular journey, the different in behaviour all comes at one point, but that might not be the case on other journeys!

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
```diff
@@ -171,6 +187,12 @@ class AcceptOrRejectJointLandlordInvitationJourney(
     override val inviteUnavailableStep: InviteUnavailableStep,
     // Landlord registration task
     override val landlordRegistrationTask: LandlordRegistrationTask,
+    override val landlordRegistrationForNotOrgLandlordTask: LandlordRegistrationForNotOrgLandlordTask,
```
Travis-Softwire: Probably an IndividualLandlord rather not org landlord :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgLandlordCyaStepConfig.kt
Travis-Softwire: Came here to make a similar comment :D I think the relevant steps should just point at the existing LL registration CYA page?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgLandlordFeatureGateStepConfig.kt
```diff
@@ -0,0 +1,27 @@
...
+import uk.gov.communities.prsdb.webapp.config.managers.FeatureFlagManager
+import uk.gov.communities.prsdb.webapp.constants.ORGANISATION_LANDLORD_REGISTRATION
+import uk.gov.communities.prsdb.webapp.journeys.AbstractInternalStepConfig
+import uk.gov.communities.prsdb.webapp.journeys.JourneyState
+import uk.gov.communities.prsdb.webapp.journeys.JourneyStep.InternalStep
+
+enum class OrgLandlordFeatureGateMode { ENABLED, DISABLED }
+
+@JourneyFrameworkComponent
+class OrgLandlordFeatureGateStepConfig(
```
Travis-Softwire: Clever approach, but I'm a bit nervous that this means if someone misconfigures their new Org LL step and accidentally makes it reachable via a non-org-LL step (which isn't obvious unless you try and access it directly by typing in the url) then it will suddenly become live code - unless the intention that individual flags will also be feature flagged?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationForNotOrgLandlordTask.kt
```diff
@@ -0,0 +1,56 @@
...
+import uk.gov.communities.prsdb.webapp.journeys.landlordRegistration.states.LandlordRegistrationState
+import uk.gov.communities.prsdb.webapp.journeys.landlordRegistration.stepConfig.CountryOfResidenceMode
+import uk.gov.communities.prsdb.webapp.journeys.landlordRegistration.stepConfig.CountryOfResidenceStep
+import uk.gov.communities.prsdb.webapp.journeys.landlordRegistration.stepConfig.EmailStep
+import uk.gov.communities.prsdb.webapp.journeys.landlordRegistration.stepConfig.NonEnglandOrWalesAddressStep
+import uk.gov.communities.prsdb.webapp.journeys.landlordRegistration.stepConfig.PhoneNumberStep
+import uk.gov.communities.prsdb.webapp.journeys.shared.stepConfig.AbstractCheckYourAnswersStep
+
+@JourneyFrameworkComponent
+class LandlordRegistrationForNotOrgLandlordTask : Task<LandlordRegistrationState>() {
```
Travis-Softwire: As above, can we make this `LandlodRegistrationForIndividualLandlordTask` or maybe even `IndividualLandlordRegistrationTask` with the other one being just `OrgLandlordRegistrationTask`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/IndividualLandlordRegistrationTask.kt
```diff
@@ -0,0 +1,56 @@
...
+            step(journey.nonEnglandOrWalesAddressStep) {
+                routeSegment(NonEnglandOrWalesAddressStep.ROUTE_SEGMENT)
+                parents { journey.countryOfResidenceStep.hasOutcome(CountryOfResidenceMode.NON_ENGLAND_OR_WALES) }
+                noNextDestination()
+            }
+            task(journey.addressTask) {
+                parents { journey.countryOfResidenceStep.hasOutcome(CountryOfResidenceMode.ENGLAND_OR_WALES) }
+                nextStep { journey.cyaStep }
+            }
+            step(journey.cyaStep) {
```
Travis-Softwire, line 47: I'd pull this out of the task - we'll want one single CYA step I think

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
Travis-Softwire: Have made the same point elsewhere - I think there are some risks with this approach unless all the steps are also flagged off individually - if this is just for routing then you can include the FF logic in the relevant nextStep block

Travis-Softwire (changes requested):
I'm a bit nervous about FF gate approach, and would like a tweak to the task names but the rest LGTM so happy to approve once those are sorted :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
samyou-softwire: ok so I've switched the task definition to be the one to check the feature flag and return different sub journeys depending. this should make it crystal clear which tasks show when the feature flag is off and this can't be accidentally changed

I'm not sure we can use the strategy pattern here. to get access to `subJourney` we need to extend `Task<>` but `PrsdFlip` seems to want an interface to flip. so as far as I can see I can't just make a task flippable. I suspect some amount of restructuring effort would allow it to be used here but it doesn't feel worthwhile - most expressive to have an if condition in this task and named methods

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgLandlordCyaStepConfig.kt
samyou-softwire: I'll clarify this, I mean that if we just pop the CYA step here it'll likely not work right at all as it's missing a bunch of the questions, so to make the skeleton functional till the end this temp step will definitely work

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/AcceptOrRejectJointLandlordInvitationJourneyFactory.kt
samyou-softwire: sure, will rename

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgLandlordFeatureGateStepConfig.kt
samyou-softwire: have restructured - see above

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationForNotOrgLandlordTask.kt
samyou-softwire: ah that's cleaner yea, I'll rename

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/IndividualLandlordRegistrationTask.kt
samyou-softwire: see above, I've written it as a specific TODO to do this later on

Travis-Softwire (approved):
Thanks - happy with this :)

PR #1492, opened 2026-06-25
Feat/pdjb 311 no longer joint landlord

AEPR:
## Ticket number
PDJB-311

## Goal of change
Add the journey for landlords to leave a property

Adds the two entry points for the journey as well.

image

image

image

image

## Checklist

- [X] A new journey integration test has been added for any new journeys
- [X] TODO comments referencing this JIRA ticket have been searched for and removed - one additional TODO needs addressing, covering an email sent when deleting a landlord account. That will be a future PR


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -380,6 +381,8 @@ class PropertyOwnershipService(
         landlord: Landlord,
     ) {
         propertyOwnership.removeLandlord(landlord)
+
+        jointLandlordOtherLandlordLeftEmailService.sendNotificationToRemainingLandlords(propertyOwnership, landlord)
```
Travis-Softwire: Note the code comment above this function - do we have a check anywhere to see if this leaves someone else as the last LL?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsLandlordViewModelBuilder.kt
```diff
@@ -55,23 +56,32 @@ class PropertyDetailsLandlordViewModelBuilder {
...
+                    if (isCurrentUser) {
+                        SummaryCardViewModel(
+                            title =
+                                "propertyDetails.landlordDetails.registeredLandlords.currentUserCardTitle",
+                            cardNumber = landlord.name,
+                            summaryList = buildLandlordCardRows(landlord),
+                            actions =
+                                listOf(
+                                    SummaryCardActionViewModel(
+                                        "propertyDetails.landlordDetails.registeredLandlords.removeMe",
```
Travis-Softwire: Should we only be showing them this if they are not the only LL? Or should the leave property controller redirect them to the deregister property journey if they are the only LL? Atm I think they'll see this but get a 404 when they click it?

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/LeavePropertyControllerTests.kt
```diff
@@ -0,0 +1,202 @@
...
+    @MockitoBean
+    private lateinit var leavePropertyService: LeavePropertyService
+
+    @MockitoBean
+    private lateinit var mockStepLifecycleOrchestrator: StepLifecycleOrchestrator.VisitableStepLifecycleOrchestrator
+
+    private val testPropertyOwnershipId = 1L
+
+    @Test
+    fun `getJourneyStep returns a redirect for an unauthenticated user`() {
```
Travis-Softwire, line 49: I think we're missing controller tests for the POST endpoint?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -380,6 +381,8 @@ class PropertyOwnershipService(
         landlord: Landlord,
     ) {
         propertyOwnership.removeLandlord(landlord)
+
```
Travis-Softwire, line 383: Worth adding an assertion on this in the relevant unit test?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/JointLandlordOtherLandlordLeftEmailService.kt
```diff
@@ -0,0 +1,51 @@
...
+        // No-op: the joint-landlords feature is disabled.
+    }
+}
+
+@PrsdbWebService("joint-landlord-other-landlord-left-email-flag-on")
+class JointLandlordOtherLandlordLeftEmailServiceImplFlagOn(
+    private val absoluteUrlProvider: AbsoluteUrlProvider,
+    private val otherLandlordLeftEmailService: EmailNotificationService<JointLandlordOtherLandlordLeftNotification>,
+) : JointLandlordOtherLandlordLeftEmailService {
+    override fun sendNotificationToRemainingLandlords(
```
Travis-Softwire, line 35: Admittedly they might end up a little trivial, but worth some tests for this service?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LeavePropertyService.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.web.server.ResponseStatusException
+import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.PrsdbWebService
+import uk.gov.communities.prsdb.webapp.constants.PROPERTIES_LEFT_THIS_SESSION
+import uk.gov.communities.prsdb.webapp.database.entity.Landlord
+import uk.gov.communities.prsdb.webapp.database.entity.PropertyOwnership
+import uk.gov.communities.prsdb.webapp.models.viewModels.emailModels.JointLandlordYouLeftConfirmation
+import kotlin.String
+
+@PrsdbWebService
+class LeavePropertyService(
```
Travis-Softwire, line 15: I don't think we have any service layer tests for this?

Travis-Softwire (commented):
A few thoughts from me - happy to chat through any of them if I've gotten the wrong end of the stick anywhere though

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
AEPR: What do you think of moving the nudge into this function instead? If we are removing the penultimate landlord we surely always want to nudge?

Travis-Softwire: That's probably fine - but worth checking w/ @samyou-softwire if there's a reason we didn't do that (possibly to do with sending emails in the middle of a transaction that might roll back? But I might be thinking of a different issue)

samyou-softwire: yeah I think it was since
- we want to defer emails till the end of the transaction, which is hard to do if its baked into the service method
- there are some cases (though this was for invites) where we delete & then immediately recreate - we don't want to consider sending a nudge in that case

see #1471

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LeavePropertyService.kt
AEPR: There's the "LeavePropertyServiceTests"

Travis-Softwire: Sorry, that comment ended up in the wrong place. It was specifically `leavePropertyOwnership` I was commenting on

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -370,16 +373,17 @@ class PropertyOwnershipService(
...
-    /**
-     * Consider whether you need to also call SwapToIndividualNudgeEmailService#sendNudgeEmailIfApplicable.
-     * This would be in case this action can lead the property marked as JL but without any landlords.
-     */
     @Transactional
     fun removeLandlord(
         propertyOwnership: PropertyOwnership,
         landlord: Landlord,
     ) {
         propertyOwnership.removeLandlord(landlord)
```
Travis-Softwire, line 381: As discussed, missing explicit `save()` based on our conventions

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LeavePropertyJourneyTests.kt
```diff
@@ -0,0 +1,49 @@
...
+            assertPageIs(
+                page,
+                ConfirmationPageLeaveProperty::class,
+                mapOf("propertyOwnershipId" to jointPropertyOwnershipId.toString()),
+            )
+        BaseComponent.assertThat(confirmationPage.confirmationBanner)
+            .containsText("No longer registered as a landlord for 3 Imaginary Street")
+
+        confirmationPage.goToDashboardButton.clickAndWait()
+        assertPageIs(page, LandlordDashboardPage::class)
```
Travis-Softwire: Can we add e.g. a navigation back to the property record to show that the removed LL has been removed/ no longer has access?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LeavePropertyService.kt
```diff
@@ -0,0 +1,59 @@
...
+        }
+        return propertyOwnership
+    }
+
+    fun leavePropertyOwnership(
+        landlord: Landlord,
+        propertyOwnership: PropertyOwnership,
+    ) {
+        propertyOwnershipService.removeLandlord(propertyOwnership, landlord)
+        confirmationEmailSender.sendEmail(
```
Travis-Softwire: Should this also be in the transaction helper in case this function in case this function gets called from inside an outer transaction in the future?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
```diff
@@ -90,6 +92,7 @@ class PropertyDetailsController(
                 PropertyDetailsLandlordViewModelBuilder.buildSummaryCards(
                     propertyOwnership.landlords,
                     baseUserId,
+                    absoluteUrlProvider.buildLeavePropertyUri(propertyOwnershipId),
```
Travis-Softwire: Minor, but how come this needs to be an absolute URL rather than using the `getLeavePropertyPath` helper on the companion?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyDeregistration/stepConfig/CannotDeregisterStepConfig.kt
```diff
@@ -6,20 +6,21 @@ import uk.gov.communities.prsdb.webapp.journeys.JourneyStep.RequestableStep
...
 ) : AbstractRequestableStepConfig<Complete, NoInputFormModel, PropertyDeregistrationJourneyState>() {
     override val formModelClass = NoInputFormModel::class
 
     override fun getStepSpecificContent(state: PropertyDeregistrationJourneyState): Map<String, Any?> {
         val propertyOwnership = propertyOwnershipService.getPropertyOwnership(state.propertyOwnershipId)
         return mapOf(
             "addressLines" to propertyOwnership.address.toMultiLineAddress().split("\n"),
-            // TODO PDJB-311: Set noLongerALandlordUrl to the "remove self from property" journey URL
-            "noLongerALandlordUrl" to "#",
+            "leavePropertyUrl" to absoluteUrlProvider.buildLeavePropertyUri(state.propertyOwnershipId),
```
Travis-Softwire: Same point here RE using absolute url rather than relative from the companion helper? I might just be missing something though.

Travis-Softwire (approved):
A few minor comments, good to go once they're addressed though

PR #1496, opened 2026-06-25
Chore/pdjb 1069 remove primaryLandlord references from emails

JasminConterioSW:
## Ticket number

PDJB-1069

## Goal of change

Remove primaryLandlord references from notification emails

* Update VirusNotification emails so they go to all landlords on the property
* Update the existing compliance update emails to go to the logged in landlord (i.e. the landlord making the update) 

A new email should go to other joint landlords on compliance update but that will be a separate ticket.

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
@@ -375,6 +375,8 @@ class PropertyComplianceService(
                     "No landlord matching the logged in user $loggedInBaseUserId was found for property ${propertyOwnership.id}",
                 )
 
+        // TODO PDJB-1216 - send a different notification email to other landlords on the property
+
         complianceUpdateConfirmationSender.sendEmail(
             landlord.email,
             ComplianceUpdateConfirmationEmail(
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
@@ -5,6 +5,7 @@ import jakarta.transaction.Transactional
 import org.springframework.data.domain.Page
 import org.springframework.data.domain.PageImpl
 import org.springframework.data.domain.PageRequest
+import org.springframework.security.core.context.SecurityContextHolder
 import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.PrsdbWebService
 import uk.gov.communities.prsdb.webapp.constants.MAX_ENTRIES_IN_COMPLIANCE_ACTIONS_PAGE
 import uk.gov.communities.prsdb.webapp.constants.PROVIDE_LATER_DEADLINE_DAYS
@@ -15,6 +16,7 @@ import uk.gov.communities.prsdb.webapp.database.entity.PropertyCompliance
 import uk.gov.communities.prsdb.webapp.database.repository.FileUploadRepository
 import uk.gov.communities.prsdb.webapp.database.repository.PropertyComplianceRepository
 import uk.gov.communities.prsdb.webapp.database.repository.PropertyOwnershipRepository
+import uk.gov.communities.prsdb.webapp.exceptions.PrsdbWebException
 import uk.gov.communities.prsdb.webapp.exceptions.UpdateConflictException
 import uk.gov.communities.prsdb.webapp.models.dataModels.ComplianceStatusDataModel
 import uk.gov.communities.prsdb.webapp.models.dataModels.RegistrationNumberDataModel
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/VirusNotificationEmailHandler.kt
TomHanmer29: This should probably be renamed to owners?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/VirusNotificationEmailHandler.kt
```diff
@@ -38,8 +38,13 @@ class VirusNotificationEmailHandler(
 
         val email = buildAlertEmail(ownership, notification.certificateType)
 
-        // TODO PDJB-1069 - do not use primary landlord
-        emailNotificationService.sendEmail(emailAddress ?: ownership.primaryLandlord.email, email)
+        if (emailAddress != null) {
+            emailNotificationService.sendEmail(emailAddress, email)
+        } else {
+            ownership.landlords.forEach { landlord ->
+                emailNotificationService.sendEmail(landlord.email, email)
```
TomHanmer29, line 45: I wonder if this email should have different copy for joint landlords? There's a lot of "you should check this file" etc. inside VirusScanUnsuccessful.md

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/VirusNotificationEmailHandler.kt
```diff
@@ -38,8 +38,13 @@ class VirusNotificationEmailHandler(
 
         val email = buildAlertEmail(ownership, notification.certificateType)
 
-        // TODO PDJB-1069 - do not use primary landlord
-        emailNotificationService.sendEmail(emailAddress ?: ownership.primaryLandlord.email, email)
+        if (emailAddress != null) {
+            emailNotificationService.sendEmail(emailAddress, email)
+        } else {
```
TomHanmer29, line 43: If we're doing this, why do we even pass in an email address in the first place?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -366,11 +368,17 @@ class PropertyComplianceService(
...
 
-        // TODO PDJB-321 - do not use primary landlord
+        val loggedInBaseUserId = SecurityContextHolder.getContext().authentication.name
+        val landlord =
+            propertyOwnership.landlords.singleOrNull { it.baseUser.id == loggedInBaseUserId }
+                ?: throw PrsdbWebException(
+                    "No landlord matching the logged in user $loggedInBaseUserId was found for property ${propertyOwnership.id}",
+                )
+
         complianceUpdateConfirmationSender.sendEmail(
```
TomHanmer29, line 380: Could do with a TODO here for the new ticket to send emails to the JLs but I appreciate it might not have been made yet :)

TomHanmer29 (changes requested):
Few things I'm not sure about on this one, but the compliance update looks good!

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/VirusNotificationEmailHandler.kt
JasminConterioSW: I was going on the basis that _one_ of the landlords should go and sort it out at this point and it doesn't really matter which one...

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/VirusNotificationEmailHandler.kt
JasminConterioSW: `sendAlertToOwner(s)` also gets called by sendAlertToMonitoringTeam for a particular notification type, but with the `emailAddress` specified as `virusMonitoringEmail`

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/VirusNotificationEmailHandler.kt
TomHanmer29: I think it might be worth checking in with Chris on this one, I could see a user getting confused if their colleague uploads a file and they get an email titled "Your gas cert failed our virus scan"

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/VirusNotificationEmailHandler.kt
JasminConterioSW: So this keeps it as "use the emailAddress if supplied, otherwise email the landlords on the property"

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/VirusNotificationEmailHandler.kt
TomHanmer29: It's definitely not _too_ bad but we've changed copy for similar emails so I think we should be consistent

JasminConterioSW: I have just had a thought that we might not be able to identify which landlord added the file here (this doesn't get called directly by the logged in user doing something, it gets called by the virus scanner failing...)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
JasminConterioSW: It might be easier if I just make the ticket...

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/VirusNotificationEmailHandler.kt
JasminConterioSW: It might be a copy update if needed rather than anything else, I've made PDJB-1216 to track that and the new compliance update email

TomHanmer29 (approved):
Happy if you want to get this merged

PR #1497, opened 2026-06-25
PDJB-1069: Remove primaryLandlord references from local council endpoints

JasminConterioSW:
## Ticket number

PDJB-1069

## Goal of change

Remove primaryLandlord references from local council endpoints 

* Remove the "Last updated by" inset text on the local council view as we don't know this anymore so that multiple landlords can edit the property (screenshot has before on the left and after on the right)

image

* Remove "Registered landlord" from the property search results as there is more than one
image

## Checklist


TomHanmer29 (approved):
Happy with this if we're happy with the sentiment of the change - I think council users can still see the LL's on the property itself right?

JasminConterioSW:
> Happy with this if we're happy with the sentiment of the change - I think council users can still see the LL's on the property itself right?

Yes, they can click on the property record link and see them there

PR #1498, opened 2026-06-25
PDJB-1034: Count completed transactions using the Plausible Flow event

Bill-Haigh:
## Ticket number

PDJB-1034

## Goal of change

Add a "Total number of transactions" metric to the system-operator dashboard that counts completed journeys, including update journeys that have no dedicated confirmation page.

- Counts completed transactions from the Plausible `Flow` custom event (which carries the server-side `currentUrl` and `referrer` paths as props), via a single filtered query returning a scalar total.
- Counts 12 completion signatures: the 3 registration journeys, invite-joint-landlord, and the 8 check-answers property-update journeys.
- Deliberately excludes the 7 single-step update journeys, whose completion is indistinguishable from a "Back" click in the Flow props (would double-count).
- Renders the total as a new row on the metrics page; the local Plausible mock returns an aggregate count so the row is populated in local dev.

The 8 check-answers updates are matched by `and(referrer matches <check-answers-step>, currentUrl matches /landlord/property-details/<id>)`. All URL path segments are now derived from the relevant controller and step `ROUTE_SEGMENT` constants rather than literals, so counting won't silently break if a route changes. The metrics page is integration-tested by stubbing `PlausibleClient` (the same `@MockitoBean` approach used for the EPC API), asserting the completion-rate and transaction rows render.

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
@@ -5,6 +5,8 @@ import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.PrsdbWebServic
 import uk.gov.communities.prsdb.webapp.clients.PlausibleClient
 import uk.gov.communities.prsdb.webapp.constants.CONFIRMATION_PATH_SEGMENT
 import uk.gov.communities.prsdb.webapp.constants.PRIVACY_NOTICE_PATH_SEGMENT
+import uk.gov.communities.prsdb.webapp.controllers.InviteJointLandlordController.Companion.INVITE_JOINT_LANDLORD_ROUTE
+import uk.gov.communities.prsdb.webapp.controllers.PropertyDetailsController.Companion.LANDLORD_PROPERTY_DETAILS_ROUTE
 import uk.gov.communities.prsdb.webapp.controllers.RegisterLandlordController.Companion.LANDLORD_REGISTRATION_CONFIRMATION_ROUTE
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
```diff
@@ -100,5 +149,48 @@ class PlausibleMetricsService(
...
+        // The three registration journeys each finish on a fixed confirmation URL with no path variables.
+        private val REGISTRATION_CONFIRMATION_PAGES =
+            listOf(
+                LANDLORD_REGISTRATION_CONFIRMATION_ROUTE,
+                PROPERTY_REGISTRATION_CONFIRMATION_ROUTE,
+                LOCAL_COUNCIL_USER_REGISTRATION_CONFIRMATION_ROUTE,
+            )
+
+        // A landlord lands back on the property record (which contains a numeric property id) after completing an update.
+        private const val PROPERTY_RECORD_REGEX = "^/landlord/property-details/\\d+$"
```
Travis-Softwire: This and all the cases below should be using the existing path segments that live on the various controllers, otherwise this will silently stop counting transactions if those urls change.

Travis-Softwire:
We definitely can have integration tests for the page - see how we handle EPC API calls in the integration tests for example

Travis-Softwire (changes requested):
One change + some integration tests please :) But shoudl be good to go after that

Travis-Softwire (approved):
Happy w/ this - can we make sure we've tested to check that the regex's etc are working w/ real Plausible locally before merging?

PR #1501, opened 2026-06-26
PDJB-1132: Organisation name

samyou-softwire:
## Ticket number

[PDJB-1132](https://mhclgdigital.atlassian.net/browse/PDJB-1132)

## Goal of change

add the organisation name page
org name view

do we want to also rename nameForm to textForm? it seems a useful generic to keep around. do we know any other pages that could use this?

cherry picks a commit from [PDJB-1135](https://mhclgdigital.atlassian.net/browse/PDJB-1135) to allow for the label in a fieldset to be null

## Checklist

  mention that here


src/main/resources/templates/forms/nameForm.html
```diff
@@ -9,8 +9,10 @@
 <!--/*@thymesVar id="submitButtonText" type="java.lang.String"*/-->
 <!DOCTYPE html>
 <html id="form-content" th:replace="~{fragments/forms/questionPage :: questionPage(#{${title}}, ${formModel}, ~{::#form-content/content()}, ${backUrl}, ${sectionHeaderInfo})}">
-    <th:block id="fieldset-content" th:replace="~{fragments/forms/fieldSet :: fieldSet(~{::#fieldset-content/content()}, 'name', #{${fieldSetHeading}}, #{${fieldSetHint}})}">
-        <div th:replace="~{fragments/forms/basicTextInput :: textInput(#{${label}}, 'name', null)}"></div>
+    <th:block th:with="fieldSetHeadingLabelTo=${label == null ? 'name' : null}">
+        <th:block id="fieldset-content" th:replace="~{fragments/forms/fieldSet :: fieldSet(~{::#fieldset-content/content()}, 'name', #{${fieldSetHeading}}, #{${fieldSetHint}})}">
```
Travis-Softwire, line 13: Unfortunately I think the 'name' id means this is for names only (it's visible to the user in the jump link to the input). We could parameterise that, but I think that then becomes a bigger change - so I think we probably want to leave this specific to 'name' and put a ticket in the backlog for commonising (I actually think we might already have one)

Travis-Softwire (changes requested):
Happy with the substance of this, but I don't think we can do the name -> text rename just yet unfortunately because of the id and jump link. I think let's add a note to PDJB-357 to include basic text forms in that commonisation if/when we get there.

Travis-Softwire (approved):
👍

src/main/resources/templates/forms/nameForm.html
samyou-softwire: have reverted this

PR #1505, opened 2026-06-29
PDJB-1069: Remove remaining references to primaryLandlord

JasminConterioSW:
## Ticket number

PDJB-1069

## Goal of change

Remove remaining references to primaryLandlord

* Remove missed reference in PropertyDetailsController
* Replaced primaryLandlord reference in ReasonStepConfig with the definition that was on propertyOwnership. This is a temporary measure to be resolved in PDJB-319
* Removed the primaryLandlord getter from propertyOwnership

## Checklist

  mention that here

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
@@ -169,11 +169,6 @@ class PropertyDetailsController(
 
         val backUrlKey = backLinkStorageService.storeCurrentUrlReturningKey(LANDLORD_DETAILS_FRAGMENT)
 
-        val primaryLandlordDetailsUrl =
-            LandlordDetailsController
-                .getLandlordDetailsForLocalCouncilUserPath(propertyOwnership.primaryLandlord.id)
-                .overrideBackLinkForUrl(backUrlKey)
-
         val propertyCompliance = propertyComplianceService.getComplianceForPropertyOrNull(propertyOwnershipId)
 
         val propertyDetails =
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
```diff
@@ -197,6 +192,12 @@ class PropertyDetailsController(
             model.addAttribute("landlordSummaryCards", landlordSummaryCards)
             model.addAttribute("landlordCount", propertyOwnership.landlords.size)
         } else {
+            val primaryLandlord = propertyOwnership.landlords.first()
```
Travis-Softwire, line 195: Minor as this will be deleted once JLs goes live, but can we just say 'landlord' here, or maybe 'soleLandlord' just so we don't wonder at some point why we still have references to primary landlord?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyDeregistration/stepConfig/ReasonStepConfig.kt
```diff
@@ -35,8 +35,10 @@ class ReasonStepConfig(
     override fun afterStepDataIsAdded(state: PropertyDeregistrationJourneyState) {
         val propertyOwnership = propertyOwnershipService.getPropertyOwnership(state.propertyOwnershipId)
 
-        // TODO PDJB-319 - do not use primary landlord
-        val primaryLandlordEmailAddress = propertyOwnership.primaryLandlord.email
+        // TODO PDJB-319 - update this to use the relevant landlord(s).
```
Travis-Softwire: Is the old deregistration journey or the new one? If it's the old one then PDJB-319 deletes it rather than updating it. If it's the new one it should be updated as part of this ticket to use the correct Landlord (check w/ Tom)

Travis-Softwire (approved):
2 Qs/suggestions, LGTM once they're addressed

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyDeregistration/stepConfig/ReasonStepConfig.kt
TomHanmer29: Yep, this is the old one so this will all be deleted, the code change here just needs to make it backwards compatible and it does so all good :)

PR #1506, opened 2026-06-29
PDJB-1137: Add organisation type step to landlord registration

JasminConterioSW:
## Ticket number

PDJB-1137

## Goal of change

Replace the stubbed organisation type step with a real question asking which type(s) of organisation a landlord is registering as.

- Adds the "What type of organisation are you registering?" step to the organisation landlord registration journey, with multi-select checkboxes (Company, Charity, Trust) and a mutually-exclusive "None of these" option separated by an "or" divider.
- Validates that at least one option is selected and that "None of these" cannot be combined with another option, showing the design error message otherwise.
- Refactors `CheckboxViewModel` into a base type with `CheckboxButtonViewModel` and `CheckboxDividerViewModel` (mirroring `RadiosViewModel`) so checkbox lists can render an "or" divider, and updates the checkboxes fragment accordingly.
- Adds in a journey test covering a flow through org landlords (up to and including this step) with TODOs for intermediate steps

image

image

The `CheckboxViewModel` refactor touches existing checkbox usages (privacy notice, bills included, filter panels) — these are mechanical renames to `CheckboxButtonViewModel`.

## Checklist

- [x] Branch has been rebased onto main and run locally
- [ ] QA notes have been added to the ticket


Travis-Softwire:
Am I right that with JS turned on the component won't let you select 'None of these' and any of the others (i.e. the validation message is just for non-JS users?)

.run/local.run.xml
```diff
@@ -24,4 +24,4 @@
       <option name="RunConfigurationTask" enabled="true" run_configuration_name="flywayClean-local" run_configuration_type="GradleRunConfiguration" />
     </method>
   </configuration>
-</component>
+</component>
```
Travis-Softwire: Probably unintentional?

JasminConterioSW:
> Am I right that with JS turned on the component won't let you select 'None of these' and any of the others (i.e. the validation message is just for non-JS users?)

I don't think so, when I was running locally I was able to select both checkboxes then get the validation error (I hadn't switched js off)

Copilot says:

```
 I've checked. The answer to the reviewer's assumption is no — the validation isn't just for non-JS users.      
                                                                                                                                                                                                                              
  Findings:                                                                                                                                                                                                                                                                                                                                    
   - The "exclusive" checkbox behaviour in GOV.UK Frontend (where ticking "None of these" auto-unticks the others) requires each exclusive input to carry data-behaviour="exclusive".                                                                                                                                                         
   - The shared checkboxes fragment (fragments/forms/checkboxes.html) never renders that attribute, and CheckboxButtonViewModel has no field to set it. A grep for data-behaviour/exclusive across the templates returns nothing.                                                                                                             
   - The data-module="govuk-checkboxes" on the container only powers conditional reveals here; without data-behaviour="exclusive" on the NONE input, the JS does nothing to prevent mixed selections.                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                               
  So with JS on, the user can still tick "None of these" alongside Company/Charity/Trust. The server-side check in OrgTypeFormModel.isSelectionValid() (rejects NONE + others) is the only thing enforcing exclusivity, and it applies to all users, JS or not.     

  If the intent is for JS users to be prevented client-side, the fragment/view model would need to be extended to emit data-behaviour="exclusive" on the NONE option                                                                          ```

File: .run/local.run.xml (hunk 1 above)
Comment (TomHanmer29) [inline .run/local.run.xml]: This file probably shouldn't have changed?

**Review by TomHanmer29 — APPROVED** [review]
LGTM! Just 1 tiny file that's snuck in, plus the point Travis made (which I think you've addressed)

File: src/main/resources/templates/fragments/forms/checkboxes.html [hunk 2]
```diff
@@ -17,6 +19,7 @@
             <div th:if="${option.conditionalFragment}" class="govuk-checkboxes__conditional govuk-checkboxes__conditional--hidden" th:id="${conditionalFragmentID}" th:with="fragmentCall=|fragments/conditional/${option.conditionalFragment}|">
                 <th:block th:replace="~{__${fragmentCall}__ :: ${option.conditionalFragment}}"></th:block>
             </div>
+            </th:block>
```
Travis-Softwire, line 22: I think if we're able to apply the `data-behaviour="exclusive"` attribute to the 'None' checkbox the front end will enforce exclusivity for users with js active :)

Travis-Softwire (commented):
Quick additional point from me

src/main/resources/templates/fragments/forms/checkboxes.html
JasminConterioSW: (I'll have a look in another PR, ideally want to merge this one for the integration journey test)

PR #1516, opened 2026-06-30
PDJB-1034: Count transactions from the Plausible Transaction event after a cutover date

Bill-Haigh:
## Ticket number

PDJB-1034

## Goal of change

Count completed transactions on the System Operator metrics dashboard from a dedicated Plausible "Transaction" custom event, while still using the legacy Flow-event method for reporting periods that predate the new event.

The branch has two logical commits:

1. **Emit the `Transaction` event via a declarative button-press pattern** — a generic `plausibleTransactionEvent.js` handler fires `window.plausible('Transaction')` when the user presses a control tagged `data-plausible-event="Transaction"`. It fires on submit without calling `preventDefault`, so the native form submission is never blocked and the event is delivered via the Plausible beacon (and is a no-op if Plausible is unavailable). A `transactionSubmitButton` fragment (and warning-style `transactionWarningButton`) carries the tag. Dedicated completion templates adopt the fragment wholesale; shared templates (reused by non-completion steps) choose the fragment via a `submitButton` content property that names the button fragment and defaults to `primaryButton`, resolved with the existing Thymeleaf `__${...}__` fragment-preprocessing idiom. To add tracking to a future journey a developer just renders their commit button via the fragment.

2. **Split `getTransactionCounts` at a configurable cutover date** (`plausible.transaction-event-start-date`, env var `PLAUSIBLE_TRANSACTION_EVENT_START_DATE`): the part of a reporting period on or after the cutover is counted from the new `Transaction` event, the part before it from the legacy Flow event. A period straddling the cutover sums two day-granular sub-range queries (disjoint, no double-counting). The config default is a far-future date, so the metric stays on the proven Flow-event method until ops set the real go-live date. The local Plausible mock (`MockPlausibleController`) distinguishes the Flow vs Transaction queries by `event:name`.

- **Coverage:** all registration, deregistration, switch-to-individual and property/landlord update journeys are tagged, as is accepting a joint-landlord invitation (the confirm-you-are-a-landlord commit button). **Not tagged:** cancelling and rejecting joint-landlord invitations, which complete via buttonless redirects after an "are you sure?" radio page, so a plain Continue button would over-count. These are counted by the metric only once they gain a genuine commit button.
- **Deployment action required:** `PLAUSIBLE_TRANSACTION_EVENT_START_DATE` must be set to the real Transaction-event go-live date. Until then the dashboard continues to use the Flow-event count (the far-future default).

## Checklist

- [x] Special release instructions: `PLAUSIBLE_TRANSACTION_EVENT_START_DATE` must be set to the real go-live date on the release ticket

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
@@ -103,14 +103,14 @@ class PlausibleMetricsService(
     private fun flowEventCount(
         start: LocalDate,
         end: LocalDate,
-    ): Long = aggregateEvents(buildFlowTransactionQuery(start, end))
+    ): Long = queryEventCount(buildFlowTransactionQuery(start, end))
 
     private fun transactionEventCount(
         start: LocalDate,
         end: LocalDate,
-    ): Long = aggregateEvents(buildTransactionEventQuery(start, end))
+    ): Long = queryEventCount(buildTransactionEventQuery(start, end))
 
-    private fun aggregateEvents(query: PlausibleQuery): Long =
+    private fun queryEventCount(query: PlausibleQuery): Long =
         (plausibleClient.query(query).results.firstOrNull()?.metrics?.firstOrNull() ?: 0.0).toLong()
 
     private fun buildFlowTransactionQuery(
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
@@ -33,6 +33,7 @@ import uk.gov.communities.prsdb.webapp.models.dataModels.plausible.PlausibleQuer
 import java.math.BigDecimal
 import java.math.RoundingMode
 import java.time.Instant
+import java.time.LocalDate
 import java.time.ZoneId
 import java.time.format.DateTimeFormatter
 
@@ -40,7 +41,10 @@ import java.time.format.DateTimeFormatter
 class PlausibleMetricsService(
     private val plausibleClient: PlausibleClient,
     @Value("\${plausible.domain-id}") private val domainId: String,
+    @Value("\${plausible.transaction-event-start-date}") transactionEventStartDate: String,
 ) {
+    private val transactionEventStartDate: LocalDate = LocalDate.parse(transactionEventStartDate)
+
     fun getCompletionRates(period: ReportingPeriod): JourneyCompletionRatesDataModel =
         try {
             val response = plausibleClient.query(buildQuery(period))
@@ -80,20 +84,57 @@ class PlausibleMetricsService(
 
     fun getTransactionCounts(period: ReportingPeriod): Long =
         try {
-            val response = plausibleClient.query(buildTransactionQuery(period))
-            (response.results.firstOrNull()?.metrics?.firstOrNull() ?: 0.0).toLong()
+            val startDate = period.start.toUkLocalDate()
+            val endDate = period.end.toUkLocalDate()
+            when {
+                endDate.isBefore(transactionEventStartDate) ->
+                    flowEventCount(startDate, endDate)
+                !startDate.isBefore(transactionEventStartDate) ->
+                    transactionEventCount(startDate, endDate)
+                else ->
+                    flowEventCount(startDate, transactionEventStartDate.minusDays(1)) +
+                        transactionEventCount(transactionEventStartDate, endDate)
+            }
         } catch (e: Exception) {
             println("Failed to fetch transaction counts from Plausible: ${e.message}")
             0L
         }
 
-    private fun buildTransactionQuery(period: ReportingPeriod): PlausibleQuery =
+    private fun flowEventCount(
+        start: LocalDate,
+        end: LocalDate,
+    ): Long = aggregateEvents(buildFlowTransactionQuery(start, end))
+
+    private fun transactionEventCount(
+        start: LocalDate,
+        end: LocalDate,
+    ): Long = aggregateEvents(buildTransactionEventQuery(start, end))
+
+    private fun aggregateEvents(query: PlausibleQuery): Long =
+        (plausibleClient.query(query).results.firstOrNull()?.metrics?.firstOrNull() ?: 0.0).toLong()
+
+    private fun buildFlowTransactionQuery(
+        start: LocalDate,
+        end: LocalDate,
+    ): PlausibleQuery =
         PlausibleQuery(
             siteId = domainId,
-            dateRange = listOf(period.start.toUkDate(), period.end.toUkDate()),
+            dateRange = listOf(start.toString(), end.toString()),
             metrics = listOf("events"),
             dimensions = emptyList(),
             filters = listOf(transactionFilter()),
+        )
+
+    private fun buildTransactionEventQuery(
+        start: LocalDate,
+        end: LocalDate,
+    ): PlausibleQuery =
+        PlausibleQuery(
+            siteId = domainId,
+            dateRange = listOf(start.toString(), end.toString()),
+            metrics = listOf("events"),
+            dimensions = emptyList(),
+            filters = listOf(listOf("is", "event:name", listOf(TRANSACTION_EVENT_NAME))),
         )
 
     // A transaction is a completed journey, counted from the Plausible "Flow" custom event (which carries the
@@ -135,8 +176,13 @@ class PlausibleMetricsService(
 
     private fun Instant.toUkDate(): String = DateTimeFormatter.ISO_LOCAL_DATE.format(this.atZone(UK_ZONE).toLocalDate())
 
+    private fun Instant.toUkLocalDate(): LocalDate = this.atZone(UK_ZONE).toLocalDate()
+
     companion object {
         private val UK_ZONE = ZoneId.of("Europe/London")
+
+        // The name of the Plausible custom event fired on journey completion (see templates/fragments/layout.html).
+        const val TRANSACTION_EVENT_NAME = "Transaction"
 
         // Visitors are used for one-off-per-user journeys (landlord and local council registration); page views are
         // used for property registration, where a single landlord may register multiple properties.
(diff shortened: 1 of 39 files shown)


src/main/resources/application.yml
```diff
@@ -84,6 +84,9 @@ plausible:
   domain-id: ${PLAUSIBLE_DOMAIN_ID}
   base-url: https://plausible.io
   api-key: ${PLAUSIBLE_API_KEY}
+  # Date the Transaction custom event went live in production. Until set to the real go-live date, the
+  # far-future default keeps the total-transactions metric on the legacy Flow-event counting method.
+  transaction-event-start-date: ${PLAUSIBLE_TRANSACTION_EVENT_START_DATE:2099-01-01}
```
JasminConterioSW, line 89: Is this going into an infra PR as well? We should be able to set it to something real per environment in order to test

src/main/resources/templates/forms/phoneNumberForm.html
```diff
@@ -16,5 +16,5 @@
     <th:block th:if="${showWarning}">
         <div th:replace="~{fragments/warningText :: warningText(#{forms.update.warning})}">forms.update.warning</div>
     </th:block>
-    <button th:replace="~{fragments/buttons/primaryButton :: primaryButton(#{${submitButtonText}})}"></button>
+    <button th:replace="~{fragments/buttons/__${submitButton ?: 'primaryButton'}__ :: __${submitButton ?: 'primaryButton'}__(#{${submitButtonText}})}"></button>
```
JasminConterioSW, line 19: So this will normally be a normal submit button, unless it's at the end of an update journey.

src/main/resources/templates/forms/deregistrationReasonForm.html
```diff
@@ -13,6 +13,6 @@
         <div th:replace="~{fragments/forms/textArea :: textArea(null, 'reason', null)}">
         </div>
     </th:block>
-    <button th:replace="~{fragments/buttons/primaryButton :: primaryButton(#{${submitButtonText}})}"></button>
+    <button th:replace="~{fragments/buttons/__${submitButton ?: 'primaryButton'}__ :: __${submitButton ?: 'primaryButton'}__(#{${submitButtonText}})}"></button>
```
JasminConterioSW, line 16: Is this one always the last user-facing step of the journey - if so can it just be a transactionButton?

JasminConterioSW (commented):
I haven't finished looking yet, just some initial comments!

Do we want to count a landlord accepting a joint landlord invitation if we can? Can we add the event on the ConfirmYouAreALandlordForThisPropertyStep? (It won't add them if something goes wrong after the user submits like the token has expired, but that's going to be the case for all the button events...)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
```diff
@@ -80,22 +84,59 @@ class PlausibleMetricsService(
...
+        start: LocalDate,
+        end: LocalDate,
+    ): Long = aggregateEvents(buildFlowTransactionQuery(start, end))
+
+    private fun transactionEventCount(
+        start: LocalDate,
+        end: LocalDate,
+    ): Long = aggregateEvents(buildTransactionEventQuery(start, end))
+
+    private fun aggregateEvents(query: PlausibleQuery): Long =
```
JasminConterioSW: Minor, but `aggregate` sound more like it's combining results, which this one is more executing a query and extracting the single result from it - can you rename it?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/TransactionEventTests.kt
```diff
@@ -0,0 +1,127 @@
...
+import uk.gov.communities.prsdb.webapp.models.dataModels.VerifiedIdentityDataModel
+import uk.gov.communities.prsdb.webapp.models.viewModels.emailModels.LandlordRegistrationConfirmationEmail
+import uk.gov.communities.prsdb.webapp.services.AbsoluteUrlProvider
+import uk.gov.communities.prsdb.webapp.services.EmailNotificationService
+import java.net.URI
+import java.time.LocalDate
+
+private const val TAGGED_BUTTON_SELECTOR = "button[data-plausible-event='Transaction']"
+
+class LandlordRegistrationTransactionEventTests : IntegrationTestWithMutableData("data-mockuser-not-landlord.sql") {
```
JasminConterioSW: These don't go so far as to make database changes so can be `IntegrationTestWithImmutableData`

The Landlord Registration tests could be single page style tests (where you start it with a `navigator.skipToPage` call that sets up the session state) as they are only really checking the last page. 

Or you could have `only the landlord registration check answers commit button is tagged for the Plausible Transaction event` (or similar) where you check that there is no tagged button on each page until the check answers page (then you don't need the separate `a mid-journey landlord registration page is not tagged for the Plausible Transaction event` test )

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsServiceTests.kt
```diff
@@ -185,4 +199,68 @@ class PlausibleMetricsServiceTests {
         assertTrue(serialised.contains("check-gas-safety-answers"))
         assertTrue(!serialised.contains("update-bedrooms"))
     }
+
+    @Test
+    fun `getTransactionCounts uses only the Flow query when the period ends before the cutover`() {
+        stubByEventName(flowCount = 100.0, transactionCount = 999.0)
+
+        val count =
+            PlausibleMetricsService(plausibleClient, domainId, "2025-02-01")
```
JasminConterioSW: This (and the other new tests) can call the private service(...) method to instantiate the PlausibleMetricsService

JasminConterioSW (commented):
A couple more comments but generally looking good!

Bill-Haigh:
> I haven't finished looking yet, just some initial comments!
> 
> Do we want to count a landlord accepting a joint landlord invitation if we can? Can we add the event on the ConfirmYouAreALandlordForThisPropertyStep? (It won't add them if something goes wrong after the user submits like the token has expired, but that's going to be the case for all the button events...)

I have added the confirmYouAreALandlord button so this will be counted! Good spot. 

Yes the buttons have this problem but after considering adding something on the backend this was the best balance until / if we have full audit logging in the future

src/main/resources/application.yml
Bill-Haigh: yes this will need an infra PR also will raise one once this is approved

src/main/resources/templates/forms/phoneNumberForm.html
Bill-Haigh: yes exactly - defaults to `primaryButton` but content overrides when it is part of the update journey

src/main/resources/templates/forms/deregistrationReasonForm.html
Bill-Haigh: This one is shared between landlord and property deregistration but in property deregistration it isn't the final user step. This is why it uses the content override.

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsService.kt
Bill-Haigh: Agree, renamed to `queryEventCount`

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/TransactionEventTests.kt
Bill-Haigh: updated to immutable

and went with the second approach

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/PlausibleMetricsServiceTests.kt
Bill-Haigh: done

src/main/resources/templates/forms/deregistrationReasonForm.html
JasminConterioSW: It has just occurred to me that the user only sees this page when they have registered properties...

So we won't count deregistrations when the user doesn't have any registered properties.

```
step(journey.areYouSureStep) {
                routeSegment(AreYouSureStep.ROUTE_SEGMENT)
                initialStep()
                backUrl { LANDLORD_DETAILS_FOR_LANDLORD_ROUTE }
                nextDestination { mode ->
                    if (mode == AreYouSureMode.DOES_NOT_WANT_TO_PROCEED) {
                        Destination.ExternalUrl(LANDLORD_DETAILS_FOR_LANDLORD_ROUTE)
                    } else if (state.userHasRegisteredProperties) {
                        Destination(journey.reasonStep)
                    } else {
                        Destination(journey.deregisterStep)
                    }
                }
            }
            step(journey.reasonStep) {
                routeSegment(ReasonStep.ROUTE_SEGMENT)
                parents { journey.areYouSureStep.hasOutcome(AreYouSureMode.WANTS_TO_PROCEED) }
                nextDestination { Destination(journey.deregisterStep) }
            }
```

I don't think there's going to be an easy way around this. Are we counting property deregistrations? Or do we just leave off deregistration journeys?

JasminConterioSW: Or do we count deregistration journeys with the flow event?

JasminConterioSW: Oh no we do count property deregistrations in confirmPropertyDeregistrationForm

JasminConterioSW: As far as I can see, we can't count landlord deregistrations with a button click if they have no properties (unless we change the journey so they still see the reason page). Unless you can see a way around it I think we go with what we have but maybe flag it? Might be worth documenting what we're doing in the MetricsReadMe (this can be in a separate PR)

Bill-Haigh: I am less inclined to have a mixed method with using the flow event for this and think it might be cleaner to just accept not counting landlord with no properties de-registrations

JasminConterioSW (approved):
LGTM - landlords with no properties deregistering will get missed when counting button events (the last page they see is "Are you sure" with yes and no options, you can't add a transaction counter to that!) but I think we have to live with that and make sure it's documented

PR #1521, opened 2026-07-01
PDJB-1136: Add Organisation phone number step

danielledias-MT:
## Ticket number

PDJB-1142

## Goal of change

Add ability for organisation's to add a phone number as part of the landlord registration flow. This includes validation.

Adds new content to existing page

Adds validations

Page:
Screenshot 2026-07-01 at 11 13 58

Validation error for empty or invalid phone number:

Screenshot 2026-07-01 at 11 14 11

Regression test:
Landlord flow

image

## Checklist

  mention that here


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/OrganisationLandlordRegistrationSinglePageTests.kt
```diff
@@ -88,6 +88,25 @@ class OrganisationLandlordRegistrationSinglePageTests : IntegrationTestWithImmut
         }
     }
 
+    @Nested
+    inner class OrgPhoneNumberStep {
```
samyou-softwire, line 148: we normally (albeit it's not super consistent) have an initial test here that something simple eg the header renders here. see `the org email page renders the heading as a label` above

src/main/resources/templates/forms/phoneNumberForm.html
```diff
@@ -10,8 +10,8 @@
 <!--/*@thymesVar id="submitButtonText" type="java.lang.String"*/-->
 <!DOCTYPE html>
 <html id="form-content" th:replace="~{fragments/forms/questionPage :: questionPage(#{${title}}, ${formModel}, ~{::#form-content/content()}, ${backUrl}, ${sectionHeaderInfo})}">
-    <th:block id="fieldset-content" th:replace="~{fragments/forms/fieldSet :: fieldSet(~{::#fieldset-content/content()}, 'phonenumber', #{${fieldSetHeading}}, #{${fieldSetHint}})}">
-        <div th:replace="~{fragments/forms/phoneNumberInput :: phoneNumberInput(#{${label}}, 'phoneNumber', #{${hint}})}"></div>
+    <th:block id="fieldset-content" th:replace="~{fragments/forms/fieldSet :: fieldSet(~{::#fieldset-content/content()}, 'phonenumber', #{${fieldSetHeading}}, ${fieldSetHint != null ? #messages.msg(fieldSetHint) : null})}">
```
samyou-softwire: we need to be careful here - the phone number box does not have a label here as we've removed it as per the designs, but now the input has no label. see the `fieldSetHeadingLabelTo` we can pass to `fieldSet` and how the org email form handles this

samyou-softwire (approved):
two small comments on markup & an extra test we could consider including

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/OrganisationLandlordRegistrationSinglePageTests.kt
JasminConterioSW: We never used to do this and think we decided one dev catchup that we _didn't_ want to check all the content like that? I think copilot tends to enthusiastically check everything!

samyou-softwire: oh yea don't think we need to check all the content as a couple of these did, but worth having a test that verifies some important part of the page (eg the header) shows can be useful

JasminConterioSW: In the past we did this to check dynamic content. If the page hasn't rendered, there'll be no submit button to click so tests would fail at that point. I.e. having a test that interacts with the page will already check that it has rendered (and not just given you the error page)

JasminConterioSW: (The other consideration is that integration tests get very slow, so we ideally don't want to add more new tests than we need, which is why the happy path is usually testing with other steps in a journey test rather than adding a new test for submitting each one)

Travis-Softwire: I'm ok with an extra single page test (these are comparatively less expensive than journey tests because they share a context with the surrounding tests) but definitely agree we shouldn't be checking loads of content

Travis-Softwire (approved):
Agree w/ Sammy RE the input label, but otherwise LGTM as well :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/OrganisationLandlordRegistrationSinglePageTests.kt
danielledias-MT: Added a single happy path test

src/main/resources/templates/forms/phoneNumberForm.html
danielledias-MT: Updated

PR #1526, opened 2026-07-01
PDJB-1258: Limit passcode requirement to landlord registration journey

Travis-Softwire:
## Ticket number

PDJB-1258

## Goal of change

Let users invited as Joint Landlords (JLs) use the service without a passcode, by requiring a passcode only for the normal landlord registration journey.

- The passcode gate (active under the `require-passcode` profile) previously covered every `/landlord/**` route; it now applies only to the landlord registration journey (`/landlord/register-as-a-landlord` and its sub-paths).
- All other landlord areas — including the joint-landlord invitation acceptance flow — are now governed solely by the existing role-based authorization, so invited JLs are no longer redirected to passcode entry.

- Applied unconditionally (not behind the `joint-landlords` flag): safe because non-registration landlord routes are already protected by `@PreAuthorize`, and landlord roles are only granted via the passcode-gated registration or via JL acceptance. The JL invitation routes remain gated by the existing `joint-landlords` flag.
- Minor behaviour change: an authenticated user without a landlord role hitting a non-registration landlord route now gets a 403 (previously a redirect to passcode entry). Passcode claiming now happens within the registration journey (the normal flow authenticates mid-journey at ID verification).

## Checklist

- [x] Test suite has been run in full locally and is passing — targeted verification run instead (`PasscodeInterceptorTests` + the three passcode integration test classes + `ktlintCheck`), all passing; full suite not run as the change is localised to the passcode interceptor
- [x] Branch has been rebased onto main and run locally, with everything working as expected — branch created from the latest `origin/main`; targeted tests run locally and passing
- [x] TODO comments referencing this JIRA ticket have been searched for and removed — none found
- [x] QA instructions have been added to the ticket — suggested: with `require-passcode` active, confirm a joint-landlord invitation link reaches the accept/reject page without a passcode, and that `/landlord/register-as-a-landlord/start` still requires one


src/main/kotlin/uk/gov/communities/prsdb/webapp/config/interceptors/PasscodeInterceptor.kt
```diff
@@ -26,9 +26,9 @@ class PasscodeInterceptor(
             return handleAuthenticatedUserWithClaimedPasscode(request, response)
         }
 
-        // Only apply interceptor to non-passcode landlord routes
+        // Only require a passcode for the landlord registration journey
         val currentPath = request.requestURI
```
samyou-softwire: this var is probably now redundant

src/test/kotlin/uk/gov/communities/prsdb/webapp/config/interceptors/PasscodeInterceptorTests.kt
```diff
@@ -96,19 +106,19 @@ class PasscodeInterceptorTests {
 
     @Test
     fun `preHandle redirects unauthenticated users without a session passcode to entry page`() {
```
JasminConterioSW: Minor, but should the test name be updated to reflect that this applies on the registration route?

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/interceptors/PasscodeInterceptor.kt
```diff
@@ -26,9 +26,9 @@ class PasscodeInterceptor(
...
-        if (!currentPath.startsWith("/$LANDLORD_PATH_SEGMENT/") || currentPath in passcodeRoutes) {
```
samyou-softwire, line 31: ig to check in on the latter condition being removed here, as I see it this is redundant as if `currentPath in passcodeRoutes` was true it'd necessarily not start with landlord path segment and so it was redundant?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PasscodeEntryFlowTests.kt
```diff
@@ -148,4 +121,18 @@ class PasscodeEntryFlowTests : IntegrationTestWithMutableData("data-passcode.sql
         navigator.navigateToInvalidPasscodePage()
         assertPageIs(page, LandlordDashboardPage::class)
     }
+
+    private fun authenticateThenSubmitPasscodeOnRegistrationJourney(
+        page: Page,
+        passcode: String,
+    ) {
+        // The dashboard is no longer passcode-gated, so navigating to it simply authenticates the user
```
JasminConterioSW: Minor, but it is worth adding to the comment that this isn't a real flow? They won't really get to the dashboard before registration. Is this just so that the MockUser is authenticated for the test (a real user will be redirected to One-Login)

JasminConterioSW (approved):
Couple of minor comments, but LGTM :)

samyou-softwire (approved):
LGTM, some tiny comments but happy for this to be merged as-is

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/interceptors/PasscodeInterceptor.kt
Travis-Softwire: Good spot

src/test/kotlin/uk/gov/communities/prsdb/webapp/config/interceptors/PasscodeInterceptorTests.kt
Travis-Softwire: Good spot - amended

src/main/kotlin/uk/gov/communities/prsdb/webapp/config/interceptors/PasscodeInterceptor.kt
Travis-Softwire: Good question — this was because the two passcode routes do actually start with `/landlord/` (`PASSCODE_ENTRY_ROUTE` = `/landlord/passcode-entry`, `INVALID_PASSCODE_ROUTE` = `/landlord/invalid-passcode`). So this was preventing an infinite loop.

It is safe to drop now because those routes are not part of the registration journey.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PasscodeEntryFlowTests.kt
Travis-Softwire: Yep, good point - have added a comment to that effect

PR #1529, opened 2026-07-01
PDJB-NONE: Fix intellij tool name reference

samyou-softwire:
I'm not sure if `jetbrains-*` is valid in any case so maybe we can replace the reference outright


TomHanmer29 (approved):

samyou-softwire:
I've some PRs I'd like to get merged first, will merge this later

PR #1535, opened 2026-07-02
PDJB-NONE: email case validation normalisation

TomHanmer29:
## Ticket number

PDJB-NONE

## Goal of change

Currently, there are a few places in the codebase we weren't checking email case when validating. This sorts them out.

Removed case sensitivity from JL email checks.
Removed case sensitivity from isConfirmEmailSameAsEmail
Removed case sensitivity from landlord service
Extracts some email logic into a helper

Arguably we should sort out some kind of conversion to lowercase upon saving emails, but this would break existing emails / make peoples emails show differently to how they were entered.

I've decided that if we update a landlord email in **case only**, we should send 1 email to them rather than 1 to each address, since you can still register in a case sensitive way. This PR just fixes all validation comparisons.

## Checklist

  mention that here


Travis-Softwire (approved):
LGTM

danielledias-MT (approved):

PR #1537, opened 2026-07-02
PDJB-1155/PDJB-1156: Lead trustee address task

samyou-softwire:
## Ticket number

[PDJB-1155](https://mhclgdigital.atlassian.net/browse/PDJB-1155) [PDJB-1156](https://mhclgdigital.atlassian.net/browse/PDJB-1156)

## Goal of change

adds a copy of the address task into the org landlords flow for lead tenants

select lead tenant address page

manual lead tenant address page

adds a new task duplicating an existing address task

adds a state adapter to make the steps in this new state able to reference the other new steps in the state without having to rewrite the steps

adds design for pages

adds all the new steps & task into the relevant journey states

what do we think of the state adapter pattern? I think it's a reasonable compromise that lets us proceed with multiple of the same task in the same journey. it may be what we end up doing anyway, since task urls wouldn't solve this problem

in brief, the state adapter gets around the limitation that steps within the same task cannot know which task they are in, so when the address not found step needs to reference the postcode entered on the address search step, we need a way for this to 'just work'. the state adapter delegates all the state calls through a class to their actual (different) step on the object

I didn't add many tests for the address entry task itself - it already exists on other flows and it has the same behaviour as there

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/tasks/OrgLandlordLeadTrusteeAddressTask.kt
samyou-softwire, line 1: this is mostly a duplicate of `LandlordAddressTask` aside from the class at the end

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/tasks/OrgLandlordLeadTrusteeAddressTask.kt
```diff
@@ -0,0 +1,107 @@
...
+import uk.gov.communities.prsdb.webapp.journeys.shared.stepConfig.ManualAddressStep
+import uk.gov.communities.prsdb.webapp.journeys.shared.stepConfig.NoAddressFoundStep
+import uk.gov.communities.prsdb.webapp.journeys.shared.stepConfig.SelectAddressMode
+import uk.gov.communities.prsdb.webapp.journeys.shared.stepConfig.SelectAddressStep
+import uk.gov.communities.prsdb.webapp.models.dataModels.AddressDataModel
+
+@JourneyFrameworkComponent
+class OrgLandlordLeadTrusteeAddressTask : Task<LandlordRegistrationOrgLandlordState>() {
+    override fun makeSubJourney(state: LandlordRegistrationOrgLandlordState) =
+        subJourney(LeadTrusteeAddressStateAdapter(state)) {
```
samyou-softwire, line 22: this is what wires it together, transforming a state into our delegated state

Travis-Softwire:
I think this is a reasonable approach while we work on something that allows full as-is reuse (copilot has a draft of something that looks promising, but Alex is now picking it up properly)

Travis-Softwire (approved):
LGTM - as per my other comment, suspect we'll replace soon with a more reusable solution but for now this keeps us moving forwards :)

PR #1545, opened 2026-07-06
PDJB-1134: Updated error message for missing address line one

bnjn-mt:
## Ticket number

PDJB-1134

   • Changes the missing address line 1 error message                                                                                                                                                                                                                                  
   • Updates the affected single-page integration test assertions to match


src/main/resources/messages/forms.yml
```diff
@@ -240,7 +240,7 @@ manualAddress:
   addressLineOne:
     label: Address line 1
     error:
-      missing: 'Enter the first line of an address, typically the building and street'
+      missing: 'Enter address line 1, typically the building and street'
```
samyou-softwire: I think this is a bigger change than we want - looks like this would update the error message everywhere where the manual page is used.

if we're concerned about the error message being inconsistent on the service, we should flag this with design. I can ping them on the ticket if you'd like? I think we should probably use the same error everywhere, probably easiest for the designers of this journey to use the one we use elsewhere

samyou-softwire: in any case we don't need the quotation marks here, for instance

```suggestion
      missing: Enter address line 1, typically the building and street
```
will work

samyou-softwire (changes requested):
we should be careful around making these changes - let me know if you want me to post anything on jira on your behalf

src/main/resources/messages/forms.yml
bnjn-mt: That makes sense to me. Yes, could you add a comment on the ticket for me? Thanks

Travis-Softwire: I suspect this is just a mix up from the designers - they did agree that in general we could use our existing validation messages rather than the ones in the designs if they were appropriate for the org landlords work.

samyou-softwire: yeah they've agreed to this on jira too. let's leave the error message as-is which matches the rest of the site. I'll approve QA on [PDJB-1134](https://mhclgdigital.atlassian.net/browse/PDJB-1134) now

Travis-Softwire:
It sounds like we've agreed this PR is no longer needed? If so can we close it off please? Thanks both for resolving this

PR #1547, opened 2026-07-06
PDJB-1053: missing dashboard nav link

TomHanmer29:
## Ticket number

PDJB-1053

## Goal of change

Adds dashboard links to the nav bar

image
image
image

Adds a link to the LC dashboard for LC users, landlord dash for landlords and SO dash for SO's
Also fixes some LC manage users behaviour

This doesn't show the dashboard on an error page - should it? It's a bit more complex to implement if so but may be worth doing?

## Checklist

  mention that here

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LocalCouncilDashboardController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LocalCouncilDashboardController.kt
@@ -34,21 +34,23 @@ class LocalCouncilDashboardController(
 
         val isAdmin = userRolesService.getHasLocalCouncilAdminRole(principal.name)
 
+        val navLinks =
+            mutableListOf(
+                NavigationLinkViewModel(LOCAL_COUNCIL_DASHBOARD_URL, "navLink.dashboard.title", isActive = true),
+            )
         if (isAdmin) {
-            model.addAttribute(
-                "navLinks",
-                listOf(
-                    NavigationLinkViewModel(
-                        ManageLocalCouncilUsersController.getManageUsersRoute(
-                            localCouncilUser.localCouncil.id,
-                            ManageUsersViewType.LocalAuthorityView,
-                        ),
-                        "navLink.manageUsers.title",
-                        false,
+            navLinks.add(
+                NavigationLinkViewModel(
+                    ManageLocalCouncilUsersController.getManageUsersRoute(
+                        localCouncilUser.localCouncil.id,
+                        ManageUsersViewType.LocalAuthorityView,
                     ),
+                    "navLink.manageUsers.title",
+                    false,
                 ),
             )
         }
+        model.addAttribute("navLinks", navLinks)
 
         model.addAttribute("userName", localCouncilUser.name)
         model.addAttribute("localCouncil", localCouncilUser.localCouncil.name)
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LocalCouncilDashboardController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LocalCouncilDashboardController.kt
@@ -8,9 +8,9 @@ import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.PrsdbControlle
 import uk.gov.communities.prsdb.webapp.constants.DASHBOARD_PATH_SEGMENT
 import uk.gov.communities.prsdb.webapp.constants.LOCAL_COUNCIL_DASHBOARD_SURVEY_URL
 import uk.gov.communities.prsdb.webapp.constants.LOCAL_COUNCIL_PATH_SEGMENT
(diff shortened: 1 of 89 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/controllerAdvice/GlobalModelAttributes.kt
```diff
@@ -67,15 +78,68 @@ class GlobalModelAttributes(
...
         model.addAttribute("serviceName", serviceName)
         if (isCustomServiceName) {
             model.addAttribute("isCustomServiceName", true)
         }
+
+        // Dashboard nav link — one per service, chosen by path prefix, shown only to a matching role.
+        // Local council admins also get a Manage users link on every local council page.
+        val navLinks: List<NavigationLinkViewModel> =
+            when {
+                uri.isServicePage(LANDLORD_PATH_SEGMENT) && request.isUserInRole(ROLE_LANDLORD.toBareRole()) ->
```
TomHanmer29: This is a simple way of linking back to the relevant dashboard if a user has more than 1 role, but has the disadvantage that for e.g. error pages without a landlord route etc. then the nav link won't display there. If we want that, i can refactor this

Travis-Softwire: Elsewhere we're using the database as the source of truth for what roles the user has. I can see the advantage of using the session for something inconsequential like this, but if it's going to be a pattern then I think I'd still want it in the user roles service with a nice interface over it.

Travis-Softwire: A logged in user should never have more than one role :) We've already baked that assumption in elsewhere

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LocalCouncilDashboardController.kt
```diff
@@ -59,7 +43,20 @@ class LocalCouncilDashboardController(
         return "localCouncilDashboard"
     }
 
+    @PreAuthorize("hasRole('LOCAL_COUNCIL_ADMIN')")
```
Travis-Softwire: This feels like an odd change. I think I'd rather avoid it if we can - surely we can append/ prepend (depending on when global attributes runs) rather than __having__ to add all the nav links there?

Travis-Softwire (changes requested):
A couple of minor changes needed I think to make checking roles a bit nicer, and to avoid having to add the council manage users link from the global attributes.

Travis-Softwire (approved):
Nice, LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/controllerAdvice/GlobalModelAttributes.kt
```diff
@@ -67,15 +73,27 @@ class GlobalModelAttributes(
...
+            uri.isServicePage(LOCAL_COUNCIL_PATH_SEGMENT) || uri.isServicePage(SYSTEM_OPERATOR_PATH_SEGMENT)
         val serviceNameKey = if (isCustomServiceName) "localCouncilServiceName" else "serviceName"
         val serviceName = messageSource.getMessage(serviceNameKey, null, serviceNameKey, Locale.getDefault())
         model.addAttribute("serviceName", serviceName)
         if (isCustomServiceName) {
             model.addAttribute("isCustomServiceName", true)
         }
+
+        // Dashboard nav link — a user has a single role, so their dashboard is determined by that role
+        // and the link is shown on every authenticated page (including pages without a service-specific route).
```
samyou-softwire: this feels like a copilot comment - is it really explaining anything?

samyou-softwire (approved):
happy with this - there's maybe an argument to be made that UserRolesService.kt should be feature flagged (or maybe dashboard urls could be a new service), to save us from having to make every controller be featured flagged, but that's not really much of a worry. what do you think?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/controllerAdvice/GlobalModelAttributes.kt
TomHanmer29: Yeah, think I agree, I'll remove. I'd kept it in initially as we had some other comments doing stuff like this in that file but I think they're more niche than this. The only thing potentially worth keeping is saying that we use the assumption each user only has 1 role but if so I'll move that onto the userRolesService

TomHanmer29:
> happy with this - there's maybe an argument to be made that UserRolesService.kt should be feature flagged (or maybe dashboard urls could be a new service), to save us from having to make every controller be featured flagged, but that's not really much of a worry. what do you think?

Yeah, it's not super clean having to import it into all the tests and check the flag on every request. I think I'll move it into it's own new service and use a @PRSDFlip to toggle :)

PR #1556, opened 2026-07-08
PDJB-941: property registration restructure feature flag

TomHanmer29:
## Ticket number

PDJB-941

## Goal of change

Adds a flag for the restructure of the property registration journey

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/FeatureFlagNames.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/FeatureFlagNames.kt
@@ -18,7 +18,7 @@ const val COMPLIANCE_ACTIONS_MAY2026_REDESIGN = "compliance-actions-may2026-rede
 
 const val ORGANISATION_LANDLORD_REGISTRATION = "pdjb-1097-organisation-landlord-registration"
 
-const val ALLOW_SKIPPING_PROPERTY_REGISTRATION_FIELDS = "pdjb-939-allow-skipping-property-registration-fields"
+const val PROPERTY_REGISTRATION_RESTRUCTURE_AND_SKIPPING = "pdjb-939-property-registration-restructure-and-skipping"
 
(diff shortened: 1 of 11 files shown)


src/test/resources/application-integration.yml
```diff
@@ -19,6 +19,9 @@ features:
     - name: "pdjb-939-allow-skipping-property-registration-fields"
       enabled: true
       expiry-date: "2026-12-31"
+    - name: "pdjb-941-property-registration-restructure"
+      enabled: false
```
danielledias-MT: Do we want to enable this for testing purposes? The allow skipping one above is enabled

src/main/kotlin/uk/gov/communities/prsdb/webapp/constants/FeatureFlagNames.kt
```diff
@@ -20,6 +20,8 @@ const val ORGANISATION_LANDLORD_REGISTRATION = "pdjb-1097-organisation-landlord-
 
 const val ALLOW_SKIPPING_PROPERTY_REGISTRATION_FIELDS = "pdjb-939-allow-skipping-property-registration-fields"
 
+const val PROPERTY_REGISTRATION_RESTRUCTURE = "pdjb-941-property-registration-restructure"
```
samyou-softwire: do we need this to be a distinct feature flag to PDJB-939 above which covers the epic?

danielledias-MT: that's a good idea so they are in sync - can't think of a situation where we need one but not the other?

TomHanmer29: Yep, agree with both, will update this PR to widen the scope of the PDJB-939 flag via a rename!

danielledias-MT (approved):

PR #1560, opened 2026-07-08
PDJB-1169: Part 1: Allow for types of landlord

samyou-softwire:
## Ticket number

[PDJB-1169](https://mhclgdigital.atlassian.net/browse/PDJB-1169)

## Goal of change

update the database to allow for us to have different types of landlord. for now individual landlord is the only implementor, to be joined by organisation landlord later

makes Landlord an abstract class. it's implemented by the new IndividualLandlord. uses a single table and the discriminator JPA methods to allow JPA to figure out what kind of landlord is in the table

this is different than the design we originally decided on, as it doesn't split into two separate tables. I went this way as splitting into two tables had some tricky consequences (landlords no longer had a definitive ID). I think we can still get a clean application divide this way, which I'll explore further in the next PR to add org landlords

for now updates the code to just assume that all landlords are individual landlords, adds a series of TODOs that need to be resolved to make org landlords work. most of these tickets are under the [PDJB-1207](https://mhclgdigital.atlassian.net/browse/PDJB-1207) epic

this cannot be gated behind a feature flag as it includes database changes. however, there shouldn't be any logic changes included in the PR, just fixups to class types to make it still compile, except where commented otherwise

## Checklist

  mention that here

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/application/IncompletePropertiesReminderTaskApplicationRunner.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/application/IncompletePropertiesReminderTaskApplicationRunner.kt
@@ -10,6 +10,7 @@ import org.springframework.context.ApplicationContext
 import uk.gov.communities.prsdb.webapp.annotations.taskAnnotations.PrsdbScheduledTask
 import uk.gov.communities.prsdb.webapp.annotations.taskAnnotations.PrsdbTaskService
 import uk.gov.communities.prsdb.webapp.constants.INCOMPLETE_PROPERTY_AGE_WHEN_REMINDER_EMAIL_DUE_IN_DAYS
+import uk.gov.communities.prsdb.webapp.database.entity.IndividualLandlord
 import uk.gov.communities.prsdb.webapp.exceptions.PersistentEmailSendException
 import uk.gov.communities.prsdb.webapp.exceptions.TrackEmailSentException
 import uk.gov.communities.prsdb.webapp.exceptions.TransientEmailSentException
@@ -60,9 +61,12 @@ class IncompletePropertiesReminderTaskLogic(
         for (page in 0..<pagesOfProperties) {
             val incompleteProperties = incompletePropertiesService.getIncompletePropertiesDueReminderPage(cutoffDate, page)
             incompleteProperties.forEach { property ->
+                // TODO: PDJB-1274: Update emails to account for org landlord
+                val landlord = property.landlord
+                require(landlord is IndividualLandlord)
                 try {
                     emailSender.sendEmail(
-                        property.landlord.email,
+                        landlord.email,
                         IncompletePropertyReminderEmail(
                             singleLineAddress =
                                 property.savedJourneyState.getPropertyRegistrationSingleLineAddress(),
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/application/IncompletePropertiesReminderTaskApplicationRunner.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/application/IncompletePropertiesReminderTaskApplicationRunner.kt
@@ -63,7 +63,7 @@ class IncompletePropertiesReminderTaskLogic(
             incompleteProperties.forEach { property ->
                 // TODO: PDJB-1274: Update emails to account for org landlord
                 val landlord = property.landlord
-                require(landlord is IndividualLandlord)
+                check(landlord is IndividualLandlord)
                 try {
                     emailSender.sendEmail(
                         landlord.email,
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/IndividualLandlord.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/IndividualLandlord.kt
@@ -1,0 +1,104 @@
+package uk.gov.communities.prsdb.webapp.database.entity
+
+import jakarta.persistence.Column
+import jakarta.persistence.DiscriminatorValue
+import jakarta.persistence.Entity
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.ManyToOne
+import jakarta.persistence.OneToMany
+import jakarta.persistence.OneToOne
+import uk.gov.communities.prsdb.webapp.constants.ENGLAND_OR_WALES
+import java.time.LocalDate
+
+@Entity
+@DiscriminatorValue("0")
+class IndividualLandlord() : Landlord() {
+    @OneToOne(optional = false)
+    @JoinColumn(name = "subject_identifier", nullable = false, unique = true)
+    lateinit var baseUser: PrsdbUser
+        private set
+
+    @Column(nullable = false)
+    lateinit var name: String
+
+    @Column(nullable = false)
+    lateinit var email: String
+
+    @Column(nullable = false)
+    lateinit var phoneNumber: String
+
+    @ManyToOne(optional = false)
+    @JoinColumn(name = "address_id", nullable = false)
+    lateinit var address: Address
+
+    @Column(nullable = false)
+    lateinit var countryOfResidence: String
+        private set
+
+    @Column(length = 1000)
+    var nonEnglandOrWalesAddress: String? = null
+        private set
+
+    var dateOfBirth: LocalDate? = null
+
+    @Column(nullable = false)
+    var isActive: Boolean = false
+        private set
+
+    @Column(nullable = false)
+    var isVerified: Boolean = false
+        private set
+
+    @Column(nullable = false)
+    var hasAcceptedPrivacyNotice: Boolean = false
+        private set
+
+    @Column(nullable = false)
+    var hasRespondedToFeedback: Boolean = false
+
+    @OneToMany(mappedBy = "landlord", orphanRemoval = true)
+    private var ownershipLinks: MutableSet<OwnershipLink> = mutableSetOf()
+
+    val landlordships: Set<PropertyOwnership> get() = ownershipLinks.map { it.propertyOwnership }.toSet()
+
+    @OneToMany(
+        mappedBy = "landlord",
+        orphanRemoval = true,
+    )
+    private lateinit var landlordIncompleteProperties: MutableSet<LandlordIncompleteProperties>
+
+    val incompleteProperties: List<SavedJourneyState>
+        get() = landlordIncompleteProperties.map { it.savedJourneyState }.toList()
+
+    constructor(
+        baseUser: PrsdbUser,
+        name: String,
+        email: String,
+        phoneNumber: String,
+        address: Address,
+        registrationNumber: RegistrationNumber,
+        countryOfResidence: String,
+        isVerified: Boolean,
+        hasAcceptedPrivacyNotice: Boolean,
+        nonEnglandOrWalesAddress: String?,
+        dateOfBirth: LocalDate?,
+    ) : this() {
+        this.baseUser = baseUser
+        this.name = name
+        this.email = email
+        this.phoneNumber = phoneNumber
+        this.address = address
+        this.registrationNumber = registrationNumber
+        this.countryOfResidence = countryOfResidence
+        this.isVerified = isVerified
+        this.hasAcceptedPrivacyNotice = hasAcceptedPrivacyNotice
+        this.nonEnglandOrWalesAddress = nonEnglandOrWalesAddress
+        this.dateOfBirth = dateOfBirth
+        this.isActive = true
+    }
+
+    fun isEnglandOrWalesResident(): Boolean = countryOfResidence == ENGLAND_OR_WALES
+
+    val shouldSeeFeedback: Boolean
+        get() = !hasRespondedToFeedback
+}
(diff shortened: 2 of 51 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/LandlordIncompletePropertiesRepository.kt
```diff
@@ -17,9 +19,15 @@ interface LandlordIncompletePropertiesRepository : JpaRepository<LandlordIncompl
     @Suppress("ktlint:standard:function-naming")
     fun countBySavedJourneyState_CreatedDateBefore(cutoffDate: Instant): Long
 
+    // TODO: PDJB-1275: Update assumption one base user per landlord
```
samyou-softwire, line 22: the auto generated methods won't work properly as they only see a link to `Landlord`. adds this temporary SQL that we should be able to remove eventually

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/LandlordRepository.kt
```diff
@@ -1,17 +1,20 @@
...
-import uk.gov.communities.prsdb.webapp.database.entity.Landlord
+import uk.gov.communities.prsdb.webapp.database.entity.IndividualLandlord
 import java.time.Instant
 
+// TODO: PDJB-1275: Update checks for landlord users to account for multiple users representing a landlord
 // The underscore tells JPA to access fields relating to the referenced table
 @Suppress("ktlint:standard:function-naming")
 interface LandlordRepository :
-    JpaRepository<Landlord?, Long?>,
+    JpaRepository<IndividualLandlord?, Long?>,
```
samyou-softwire, line 11: this will definitely need to be undone, but for now this fixes the compilation errors

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/DateOfBirthFormModel.kt
```diff
@@ -200,13 +199,4 @@ class DateOfBirthFormModel : DateFormModel() {
         toLocalDateOrNull() ?: throw NotNullFormModelValueIsNullException(
             "DateOfBirthFormModel date fields are null or invalid when a valid date was expected",
         )
-
-    companion object {
```
samyou-softwire, line 204: these don't seem to have been used

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/LandlordTypeFormModel.kt
```diff
@@ -17,8 +18,3 @@ class LandlordTypeFormModel(
     )
     var landlordType: LandlordType? = null,
 ) : FormModel
-
-enum class LandlordType {
```
samyou-softwire, line 21: moved to its own enum class

JasminConterioSW (commented):
So the actual table in the database has all of the individual and organisational landlords in it, but spring looks at two different entities. This is a new pattern I think (where our application entities look different to our database), are we sure this is what we want to go for?

Something we've done before is to create a view which lets us query the two tables together - this is the LocalCouncilUserOrInvitation view (combing local council users and local council invitations).  Would that pattern work here?

samyou-softwire:
> So the actual table in the database has all of the individual and organisational landlords in it, but spring looks at two different entities. This is a new pattern I think (where our application entities look different to our database), are we sure this is what we want to go for?

yeah, I think in any case our application code will have to look a little different from the database, since the database can never have Landlord, IndividualLandlord and OrganisationLandlord as completely distinct entities. with some good application service code I think it can be made intuitive enough

> Something we've done before is to create a view which lets us query the two tables together - this is the LocalCouncilUserOrInvitation view (combing local council users and local council invitations). Would that pattern work here?

I think a view would help, but a problem that we need to solve is to still have a unique ID per landlord (regardless of its type), which two disparate tables I don't think can give us. we can solve this with another table but I thought the single table view simpler to reason with

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/IndividualLandlord.kt
```diff
@@ -0,0 +1,99 @@
...
+    @Column(nullable = false)
+    var isVerified: Boolean = false
+        private set
+
+    @Column(nullable = false)
+    var hasAcceptedPrivacyNotice: Boolean = false
+        private set
+
+    @Column(nullable = false)
+    var hasRespondedToFeedback: Boolean = false
```
Travis-Softwire, line 58: I think this is no longer used - can we have a ticket to remove it?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/IndividualLandlord.kt
```diff
@@ -0,0 +1,99 @@
...
+import jakarta.persistence.Entity
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.ManyToOne
+import jakarta.persistence.OneToMany
+import jakarta.persistence.OneToOne
+import uk.gov.communities.prsdb.webapp.constants.ENGLAND_OR_WALES
+import java.time.LocalDate
+
+@Entity
+@DiscriminatorValue("0")
```
Travis-Softwire, line 14: Could we have a compile-time constant instead? And potentially use it for the value of our enum values in the new enum as well...

src/main/kotlin/uk/gov/communities/prsdb/webapp/application/IncompletePropertiesReminderTaskApplicationRunner.kt
```diff
@@ -60,9 +61,12 @@ class IncompletePropertiesReminderTaskLogic(
         for (page in 0..<pagesOfProperties) {
             val incompleteProperties = incompletePropertiesService.getIncompletePropertiesDueReminderPage(cutoffDate, page)
             incompleteProperties.forEach { property ->
+                // TODO: PDJB-1274: Update emails to account for org landlord
+                val landlord = property.landlord
+                require(landlord is IndividualLandlord)
```
Travis-Softwire: Minor, but I think this should probably be `check` which throws `IllegalStateException` rather than `require` which throws `IllegalArgumentException`

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/steps/ConfirmYouAreALandlordForThisPropertyStepConfig.kt
```diff
@@ -107,7 +107,9 @@ class ConfirmYouAreALandlordForThisPropertyStepConfig(
 
         propertyOwnership.landlords
             .filter { it.id != acceptingLandlord.id }
+            // TODO: PDJB-1274: Update emails to account for org landlord
             .forEach { landlord ->
+                require(landlord is IndividualLandlord)
```
Travis-Softwire: Same point here RE `require` vs `check` and anywhere else we're using this pattern.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/LandlordViewModel.kt
```diff
@@ -18,10 +19,13 @@ import uk.gov.communities.prsdb.webapp.journeys.shared.stepConfig.NameStep
...
 class LandlordViewModel(
-    private val landlord: Landlord,
+    baseLandlord: Landlord,
     private val withChangeLinks: Boolean = true,
 ) {
-    private val isEnglandOrWalesResident = landlord.isEnglandOrWalesResident()
+    // TODO: PDJB-1251: Update landlord details view for org landlord
+    private val landlord = baseLandlord as IndividualLandlord
+
+    private val isEnglandOrWalesResident = this.landlord.isEnglandOrWalesResident()
```
Travis-Softwire: Why is the `this` required here?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordService.kt
```diff
@@ -249,7 +254,10 @@ class LandlordService(
         }
     }
 
-    fun getLandlordUserShouldSeeFeedbackPages(baseUserId: String) =
-        retrieveLandlordByBaseUserId(baseUserId)?.shouldSeeFeedback
-            ?: throw PrsdbWebException("User with id $baseUserId was not found in the Landlord repository")
+    fun getLandlordUserShouldSeeFeedbackPages(baseUserId: String): Boolean {
```
Travis-Softwire, line 259: Is this actually used?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -365,11 +366,17 @@ class PropertyComplianceService(
...
-            propertyOwnership.landlords.singleOrNull { it.baseUser.id == loggedInBaseUserId }
+            propertyOwnership.landlords.singleOrNull { landlord ->
+                require(landlord is IndividualLandlord)
+                landlord.baseUser.id == loggedInBaseUserId
+            }
                 ?: throw PrsdbWebException(
                     "No landlord matching the logged in user $loggedInBaseUserId was found for property ${propertyOwnership.id}",
                 )
+        // TODO: PDJB-1274: Update emails to account for org landlord
+        require(landlord is IndividualLandlord)
```
Travis-Softwire: Why do we need this check again? is it not covered in the assignment?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
```diff
@@ -387,8 +394,15 @@ class PropertyComplianceService(
...
-        val otherLandlords = propertyOwnership.landlords.filter { it.baseUser.id != loggedInBaseUserId }
+        // TODO: PDJB-1275: Update authorisation checks to account for org landlords
+        val otherLandlords =
+            propertyOwnership.landlords.filter { otherLandlord ->
+                require(otherLandlord is IndividualLandlord)
+                otherLandlord.baseUser.id != loggedInBaseUserId
+            }
+        // TODO: PDJB-1274: Update emails to account for org landlord
         otherLandlords.forEach { otherLandlord ->
+            require(otherLandlord is IndividualLandlord)
```
Travis-Softwire: And here - we already check this when we create otherLandlords , is it needed here as well?

Travis-Softwire (approved):
LGTM overall - a few minor points but good to go once they're sorted

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/IndividualLandlord.kt
samyou-softwire: [PDJB-1294](https://mhclgdigital.atlassian.net/browse/PDJB-1294), will added references here to remove it

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/IndividualLandlord.kt
samyou-softwire: it seems not unfortunately, but this is a good idea to make sure the link between these two is clear. I've added a note to LandlordType.kt to make sure people are careful when reordering these

src/main/kotlin/uk/gov/communities/prsdb/webapp/application/IncompletePropertiesReminderTaskApplicationRunner.kt
samyou-softwire: ah good point, will replace

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/acceptOrRejectJointLandlordInvitation/steps/ConfirmYouAreALandlordForThisPropertyStepConfig.kt
samyou-softwire: see above

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/LandlordViewModel.kt
samyou-softwire: not sure - have removed

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordService.kt
samyou-softwire: have marked with a TODO to [PDJB-1294](https://mhclgdigital.atlassian.net/browse/PDJB-1294), I think they're all related

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
samyou-softwire: I guess not - landlords.singleOrNull must return a list of Landlord. I've made it map instead so we don't have that second unneeded check

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyComplianceService.kt
samyou-softwire: not really, have removed it in the same refactor as above

PR #1561, opened 2026-07-08
Feat/pdjb 896 routeable tasks

AEPR:
## Ticket number
PDJB-896

## Goal of change
Allow tasks to have their own routes prepended to step route segments, including nesting tasks. 

All tasks now can carry a route which is applied to all their children. Duplicable tasks can have multiple of the same task in a single journey.

## Checklist

- [ ] TODO comments referencing this JIRA ticket have been searched for and removed - one outstanding. Will be covered in a future PR.


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/builders/SubJourneyBuilder.kt
```diff
@@ -83,6 +83,23 @@ abstract class AbstractJourneyBuilder<TState : JourneyState>(
...
+    override fun <TTaskState, T> routableTask(
+        uninitialisedTask: T,
+        routeSegment: String?,
+        init: TaskInitialiser<TTaskState>.() -> Unit,
+    ) where TTaskState : JourneyState, T : Task<TTaskState>, T : JourneyState {
+        // The task IS its own state, so build its sub-journey against the task itself. A non-null `routeSegment`
+        // prefixes each step's URL, giving "<routeSegment>/<step>", and (via the task's own route-scoped data
+        // keys, applied by TaskInitialiser.build -> task.bindRoute) its stored data; null keeps them bare. No
+        // external state delegate is needed - the task sources JourneyState from its own journeyStateService.
+        @Suppress("UNCHECKED_CAST")
```
AEPR: I am going to look at a better way of resolving this.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
```diff
@@ -128,7 +127,11 @@ class LandlordRegistrationTask(
                     }
 
                     LookupAddressStep.ROUTE_SEGMENT -> {
-                        checkAnswerTask(journey.addressTask)
+                        routableTask(journey.addressTask) {
```
Travis-Softwire: Good to still be able to call `checkAnswerTask` or if necessary `checkAnswerRoutableTask` - appreciate it's just DSL but feels like a gotcha otherwise

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/states/CheckYourAnswersJourneyState.kt
```diff
@@ -77,8 +77,12 @@ interface CheckYourAnswersJourneyState : JourneyState {
         get() = journeyMetadata.baseJourneyId ?: journeyId
 
     companion object {
-        fun <T : CheckYourAnswersJourneyState> JourneyBuilder<T>.checkAnswerTask(task: Task<T>) {
+        fun <T : CheckYourAnswersJourneyState> JourneyBuilder<T>.checkAnswerTask(
+            task: Task<T>,
+            route: String? = null,
+        ) {
             task(task) {
+                route?.let { routeSegment(it) }
```
Travis-Softwire, line 87: We're calling `task` here rather than `routableTask` - I've probably missed something but is that something that it's valid to do?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/tasks/AddressTask.kt
```diff
@@ -0,0 +1,90 @@
...
+    override val manualAddressStep: ManualAddressStep,
+) : SelfStatedRoutableTask<AddressState>(journeyStateService),
+    AddressState {
+    override var cachedAddresses: List<AddressDataModel>? by delegateProvider.nullableDelegate("cachedAddresses")
+    override var cachedSelectedAddress: String? by delegateProvider.nullableDelegate("cachedSelectedAddress")
+    override var isAddressAlreadyRegistered: Boolean? by delegateProvider.nullableDelegate("isAddressAlreadyRegistered")
+
+    // Field-set content for the address steps, supplied by content-specific subclasses (e.g. LandlordAddressTask,
+    // TrusteeAddressTask) and applied to the relevant steps in makeSubJourney.
+    protected abstract val lookupAddressContentProperties: Map<String, Any?>
```
Travis-Softwire, line 40: Is content the only reason for the base class? If we had the same content would this just be a concrete class we could reuse multiple times? Or will each reused task always have to follow this pattern even if the content didn't change?

samyou-softwire:
I haven't looked very deeply at this, wanted to double check if anything's been done here to help out multiple tasks on the same journey?

one issue I came across whilst doing this for multiple address tasks is ambiguity around where task has xyzStep, the journey state itself may have xyzStep1 and xyzStep2. when the task calls state.xyzStep, which step in the state will it refer to?

see #1537 for what I had to do, solving it with a 'state adapater' that redirected queries on steps to the right actual step on the journey.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/tasks/AddressTask.kt
AEPR: This is abstract purely because of content - otherwise you end up repeating step overrides in the DSL. I made it abstract so that when you use an address task you are forced to decide what content you want.

Travis-Softwire: Yep, that makes sense, just checking there wasn't a hidden requirement that the task instances have to be distinct concrete classes. If we can decide on a case-by-case basis how to structure them depending on what makes the most sense (abstract class, DSL overrides etc) then that's great.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
AEPR: I'm going to go with making a separate extension. It will make things simpler now and shouldn't be needed long term.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/states/CheckYourAnswersJourneyState.kt
AEPR: Yes, because it's in the basic CYA journey it's totally valid and setting a route on a task that isn't selfStatedRoutable means that the URL is updated for all the steps on the journey the steps themselves will not conflict. The issue is if the tasks have any non-step state then those will still conflict!

Travis-Softwire: Ah ok, so every task is 'routable' in that it can have a route associated with it, but only selfStatedRoutable tasks can be re-used in a journey? Or is that also not quite right? If it is so then we'll probably want to tweak the names?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/AcceptOrRejectJointLandlordInvitationController.kt
```diff
@@ -50,42 +51,40 @@ class AcceptOrRejectJointLandlordInvitationController(
...
-        try {
-            val journeyMap = journeyFactory.createJourneySteps()
-            return journeyMap[stepRouteSegment]?.getStepModelAndView()
-                ?: throw ResponseStatusException(HttpStatus.NOT_FOUND, "Step not found")
-        } catch (_: NoSuchJourneyException) {
-            return ModelAndView("redirect:$ACCEPT_OR_REJECT_JOINT_LANDLORD_INVITATION_ROUTE")
-        }
-    }
+        @PathVariable stepPath: String,
+    ): ModelAndView = dispatchJourneyStep(stepPath) { getStepModelAndView() }
```
Travis-Softwire, line 58: Copilot is concerned about the initialisation happening in the start endpoint outside of the dispatcher, and that meaning that the base path isn't set when the journey is initialised

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/RegistersDelegateKeys.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.communities.prsdb.webapp.journeys
+
+// Implemented by anything that owns delegate keys and can be attached to a journey-build-wide DelegateKeyRegistry.
+// On binding, the implementer registers its already-collected keys (resolved to their final route-scoped form) and
+// forwards any keys registered afterwards.
+interface RegistersDelegateKeys {
```
Travis-Softwire: Minor, but this naming convention (describing what it does rather than what it is) doesn't really fit with the rester of the codebase? I think the noun version would be `DelegateKeyRegistrar`?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/states/CheckYourAnswersJourneyState.kt
```diff
@@ -77,8 +78,25 @@ interface CheckYourAnswersJourneyState : JourneyState {
...
+                nextStep { journey.finishCyaStep }
+            }
+        }
+
+        @Suppress("ktlint:standard:max-line-length")
+        fun <TJourneyState : CheckYourAnswersJourneyState, TTaskState : JourneyState> JourneyBuilder<TJourneyState>.duplicableCheckAnswerTask(
+            task: DuplicableTask<TTaskState>,
+            route: String? = null,
+        ) {
+            duplicableTask(task) {
```
Travis-Softwire, line 99: Do we also need to look at how we create CYA journey IDs? I think we base these just off the step's route segment atm?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/builders/TaskInitialiser.kt
```diff
@@ -52,7 +84,45 @@ class TaskInitialiser<TStateInit : JourneyState>(
...
+
+        val builtSteps = taskSubJourney.build(registry)
+
+        // Prefix every requestable step in this task with the task route, so its URL path becomes
+        // "<taskRoute>/<routeSegment>" (internal steps have no URL). Prepending rather than overwriting lets
+        // nested routed tasks compose to "<outer>/<inner>/<routeSegment>", as inner tasks build first.
+        // Then append a landing step keyed by the bare task route that redirects to the task's first step,
+        // so a request to "<taskRoute>" resolves to the task's genuine first step (internal or requestable).
+        val stepsWithLanding =
+            taskRoute?.let { route ->
```
Travis-Softwire, line 99: Not currently a requirement, but does this work if two different tasks have steps w/ the same route segment?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/landlordRegistrationJourneyPages/LeadTrusteeAddressFormPageLandlordRegistration.kt
```diff
@@ -0,0 +1,16 @@
+package uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.landlordRegistrationJourneyPages
+
+import com.microsoft.playwright.Page
+import uk.gov.communities.prsdb.webapp.controllers.RegisterLandlordController.Companion.LANDLORD_REGISTRATION_ROUTE
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.basePages.LookupAddressFormPage
+import uk.gov.communities.prsdb.webapp.journeys.shared.stepConfig.LookupAddressStep
+import uk.gov.communities.prsdb.webapp.journeys.shared.tasks.TrusteeAddressTask
+
+// The lead trustee's address reuses the shared address-lookup task as a second, routed instance, so its first
```
Travis-Softwire: Minor, but not sure we need this comment? I think the ones inside the journey framework are sensible as the behaviour is quite abstract, but this feels easy enough to follow from the code?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/Navigator.kt
```diff
@@ -378,6 +379,14 @@ class Navigator(
         return createValidPage(page, LeadTrusteePhoneFormPageLandlordRegistration::class)
     }
 
+    // Navigates to a routed task's bare URL (no step segment). The dispatcher resolves this via the task's landing
```
Travis-Softwire: Similarly, I think this is just going to be the way tasks work, so I don't think we need this comment here?

Travis-Softwire (approved):
A few thoughts and questions, one or two of which might need to be resolved before merge if I'm correct, but no major blockers. Looking forward to all the tasks using it!

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/AcceptOrRejectJointLandlordInvitationController.kt
AEPR: The journey's initialisation is actually just generating an id, so there's nothing in it and nothing added to it. I'll add it to the explicit QA instructions to make sure.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/RegistersDelegateKeys.kt
AEPR: I've gone with owner -simpler language and also true.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/states/CheckYourAnswersJourneyState.kt
AEPR: Yes - fixed.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/builders/TaskInitialiser.kt
AEPR: Should be fine as long as the tasks have different routes.

PR #1568, opened 2026-07-09
PDJB-1048: Update property record layout to include "provide later" 

JasminConterioSW:
## Ticket number

PDJB-1048

## Goal of change

Add a read-only "Property and tenancy details" property record layout (gated behind the
`PROPERTY_REGISTRATION_RESTRUCTURE_AND_SKIPPING` flag) that handles properties registered with skipped tenancy/licensing
fields.

- Adds the new-layout property record showing property, tenancy and licensing details for both the landlord and local
  council views.
- For occupied properties where tenancy/licensing was skipped, shows a "Provide this later (before &lt;date&gt;)" prompt,
  where the deadline is `last_occupied_date + 28 days` — matching the existing compliance-tab behaviour (both use
  `PropertyOwnership.isOccupied` for the occupancy check and throw on a missing occupied date).
- Adds unit, controller and integration test coverage for the occupancy/licensing scenarios (see mapping below).
- Seeds the scenarios in the local, integration and test environments, and adds a reusable `available_addresses(n)`
  SQL helper the deployed seeds use to claim distinct free addresses without scanning the whole address table.

### Occupied, tenancy and licencing skipped
 Landlord view
image

Local council view
image

### Unoccupied, (tenancy and) licencing skipped
Landlord view
image

Local council view
image

### Occupied, all fields filled in
Landlord view
image

Local council view

image

All user-facing behaviour is behind `PROPERTY_REGISTRATION_RESTRUCTURE_AND_SKIPPING`. The non-flag-gated changes are
seed/test data (`data-local.sql`, `data-integration.sql`, `data-test.sql`, `data-joint-landlord-invitation.sql`,
`MockLandlordData`) and a `.gitignore`/`.DS_Store` tidy-up, none of which can be flag-gated.

**Follow-up:**
- A `// TODO PDJB-1048` remains in `NftDataSeeder.kt` (file not touched by this PR); the NFT seeder may need the same
  provide-later handling as the SQL seeds in a later change.
- Notification banners will be in PDJB-1305

## Checklist

- [x] Test suite has been run in full locally and is passing — the property record unit, controller and integration
  tests pass; the full suite has not been re-run since merging main
- [ ] TODO comments referencing this JIRA ticket have been searched for and removed — one remains in `NftDataSeeder.kt`
  (see follow-up above)


JasminConterioSW:
### Manual QA seed properties (landlord 1 — "PRSD Landlord")

Seeded provide-later property records for manual QA on the **integration** (`data-integration.sql`) and **test** (`data-test.sql`) environments. Both files seed the same 6 scenarios; each property claims a distinct free AddressBase address and occupied ones set `last_occupied_date` so the "provide within 28 days" deadline renders.

| Scenario | Integration id | Test id | Setup | Expected on Property record tab |
|---|---|---|---|---|
| Combined | 49 | 18 | Occupied, licensing + tenancy skipped, **no** compliance record | COMBINED banner (missing details + compliance) |
| Control (all provided) | 50 | 19 | Occupied, licence held, tenancy filled, fully compliant | **No** banner |
| Tenancy skipped | 51 | 20 | Occupied, licence held, tenancy skipped, compliant | TENANCY banner |
| Licensing skipped | 52 | 21 | Occupied, tenancy filled, no licence, compliant | LICENSING banner |
| Both skipped | 53 | 22 | Occupied, no licence, tenancy skipped, compliant | BOTH banner |
| Unoccupied, licensing skipped | 54 | 23 | Unoccupied, no licence | **No** banner; licensing "Provide this later" row (no date) |

Notes:
- These load only on the integration/test deployed environments — they are **not** exercised by the automated test suite (which uses `data-local.sql`).
- Property rows are separate `INSERT … SELECT` statements (not plain VALUES rows) because they must dynamically claim an unused address. Registration numbers, `ownership_link`, and compliance rows were folded into the existing VALUES blocks.

.DS_Store
JasminConterioSW, line 1: I was confused about this, but copilot says it's a macOS Finder metadata file that was committed to the repo by mistake. It's suggesting adding it to the .gitignore

src/main/resources/data-integration.sql
```diff
@@ -249,6 +274,59 @@ SELECT setval(pg_get_serial_sequence('property_ownership', 'id'), (SELECT MAX(id
...
+ON CONFLICT DO NOTHING;
+
+SELECT setval(pg_get_serial_sequence('property_ownership', 'id'), (SELECT MAX(id) FROM property_ownership));
+
+-- PDJB-1048 provide-later flags for the QA properties above: set to true where licensing/tenancy
+-- details were skipped during registration.
+--   49 both skipped, 52 licensing skipped, 53 both skipped, 54 unoccupied licensing skipped
+--   49 both skipped, 51 tenancy skipped, 53 both skipped
+UPDATE property_ownership SET license_provide_later = true WHERE id IN (49, 52, 53, 54);
+UPDATE property_ownership SET tenancy_provide_later = true WHERE id IN (49, 51, 53);
```
JasminConterioSW: These should be set in the insert statement - they are all for newly inserted properties anyway

src/main/resources/data-test.sql
```diff
@@ -226,6 +251,59 @@ SELECT setval(pg_get_serial_sequence('property_ownership', 'id'), (SELECT MAX(id
...
+ON CONFLICT DO NOTHING;
+
+SELECT setval(pg_get_serial_sequence('property_ownership', 'id'), (SELECT MAX(id) FROM property_ownership));
+
+-- PDJB-1048 provide-later flags for the QA properties above: set to true where licensing/tenancy
+-- details were skipped during registration.
+--   18 both skipped, 21 licensing skipped, 22 both skipped, 23 unoccupied licensing skipped
+--   18 both skipped, 20 tenancy skipped, 22 both skipped
+UPDATE property_ownership SET license_provide_later = true WHERE id IN (18, 21, 22, 23);
+UPDATE property_ownership SET tenancy_provide_later = true WHERE id IN (18, 20, 22);
```
JasminConterioSW: As for data-integration.sql, these values can be included in the INSERT INTO property_ownership statement

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsViewModel.kt
```diff
@@ -16,181 +19,416 @@ import uk.gov.communities.prsdb.webapp.helpers.BillsIncludedHelper
...
-                    withChangeLinks,
-                )
-                if (propertyOwnership.license != null && propertyOwnership.license!!.licenseType != LicensingType.NO_LICENSING) {
-                    addRow(
-                        "propertyDetails.propertyRecord.licensingInformation.licensingNumber",
-                        propertyOwnership.license!!.licenseNumber,
-                    )
+    val isTenancyProvideLater: Boolean = propertyOwnership.tenancyProvideLater == true
+
+    val showTenancySection: Boolean = !isLandlordView || isOccupied
```
JasminConterioSW: Design update - this should change to only show if the property is occupied, even on the local council view

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDetailsUpdateJourneyTests.kt
```diff
@@ -680,15 +681,15 @@ class PropertyDetailsUpdateJourneyTests : IntegrationTestWithMutableData("data-l
             fun `A property can have its ownership type updated`(page: Page) {
                 // Details page
                 var propertyDetailsPage = navigator.goToPropertyDetailsLandlordView(propertyOwnershipId)
-                propertyDetailsPage.propertyDetailsSummaryList.ownershipTypeRow.clickFirstActionLinkAndWait()
+                propertyDetailsPage.beforePdjb939SummaryList.ownershipTypeRow.clickFirstActionLinkAndWait()
```
JasminConterioSW, line 722: Note these are all in a RestructureAndSkippingDisabled subclass

src/test/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsViewModelTests.kt
```diff
@@ -14,469 +21,970 @@ import uk.gov.communities.prsdb.webapp.testHelpers.mockObjects.MockLandlordData.
...
-        val propertyOwnership =
-            createPropertyOwnership(
-                address = createAddress(uprn = 1234.toLong()),
+    private val featureFlagManager =
+        mock<FeatureFlagManager> {
+            on { checkFeature(PROPERTY_REGISTRATION_RESTRUCTURE_AND_SKIPPING) } doReturn true
+        }
+
+    @Nested
+    inner class BeforePdjb939Layout {
```
JasminConterioSW: This inner class has all the flag-off tests in it (the whole class can be deleted with the feature flag)
The only changes are:

* Message names
* Feature flag manager is now passed into the view model
* Some of the test names at the end (marked)

src/test/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsViewModelTests.kt
```diff
@@ -14,469 +21,936 @@ import uk.gov.communities.prsdb.webapp.testHelpers.mockObjects.MockLandlordData.
...
+            val tenancyInformationChangeLinkCount = viewModel.beforePdjb939TenancyAndRentalInformation.count { it.actions.isNotEmpty() }
+
+            val totalChangeLinkCount =
+                propertyRecordChangeLinkCount + licensingInformationChangeLinkCount + tenancyInformationChangeLinkCount
+
+            assertEquals(0, totalChangeLinkCount)
+        }
     }
 
     @Test
```
JasminConterioSW, line 21: Tests for the new layout start here

src/test/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsViewModelTests.kt
```diff
@@ -14,469 +21,936 @@ import uk.gov.communities.prsdb.webapp.testHelpers.mockObjects.MockLandlordData.
...
+
+            val licensingInformationChangeLinkCount = viewModel.beforePdjb939LicensingInformation.count { it.actions.isNotEmpty() }
+
+            val tenancyInformationChangeLinkCount = viewModel.beforePdjb939TenancyAndRentalInformation.count { it.actions.isNotEmpty() }
+
+            val totalChangeLinkCount =
+                propertyRecordChangeLinkCount + licensingInformationChangeLinkCount + tenancyInformationChangeLinkCount
+
+            assertEquals(0, totalChangeLinkCount)
+        }
```
JasminConterioSW: These test names have been updated to reflect that the behaviour depends on whether it is a landlord or local council view

.DS_Store
TomHanmer29: I think that's sensible, I've seen this pop up once before and now with 3 devs on a mac it's more likely it would get committed by accident again if we dont ignore

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsViewModel.kt
```diff
@@ -16,181 +19,414 @@ import uk.gov.communities.prsdb.webapp.helpers.BillsIncludedHelper
...
+                getProvideLaterDeadlineText("propertyDetails.propertyRecord.licensing.councilOccupied")
+            } else {
+                messageSource.getMessageForKey("propertyDetails.propertyRecord.licensing.councilNotProvided")
+            }
+        } else {
+            null
+        }
+    }
+
+    val tenancySection: List<SummaryListRowViewModel> by lazy {
```
TomHanmer29: How come we're getting all of these lazily now?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsViewModel.kt
```diff
@@ -16,181 +19,414 @@ import uk.gov.communities.prsdb.webapp.helpers.BillsIncludedHelper
...
+        when {
+            !showTenancySection -> emptyList()
+            isTenancyProvideLater && isLandlordView -> listOf(tenancyProvideLaterRow())
+            isTenancyProvideLater && !isLandlordView -> emptyList()
+            else ->
+                buildList {
+                    add(householdsRow())
+                    add(tenantsRow())
+                    add(rentFrequencyRow(withoutBottomBorder = false))
```
TomHanmer29: I think this defaults to false? (& a couple other bits in this file)

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsViewModel.kt
```diff
@@ -16,181 +19,414 @@ import uk.gov.communities.prsdb.webapp.helpers.BillsIncludedHelper
...
+            )
+        } else {
+            row("propertyDetails.propertyRecord.tenancyAndRentalInformation.billsIncluded", value, withActionLink = false)
+        }
+    }
+
+    private fun furnishedStatusRow(): SummaryListRowViewModel =
+        row(
+            "propertyDetails.propertyRecord.tenancyAndRentalInformation.furnishedStatus",
+            // TODO PDJB-548 remove not-null assertion !! once occupancy is embedded in PropertyOwnership
```
TomHanmer29: Just to check, this is a very old ticket and we already have an isOccupied column on propertyOwnership now, what does making it embedded actually do?

TomHanmer29 (commented):
Just adding some comments now pre-looking at the tests & SQL :)

src/main/resources/data-integration.sql
```diff
@@ -249,6 +274,55 @@ SELECT setval(pg_get_serial_sequence('property_ownership', 'id'), (SELECT MAX(id
...
+-- =============================================================================
+-- PDJB-1048 provide-later property record QA properties (landlord 1), ids 49-54.
+-- For manual QA of the new-layout notification banners and "Provide this later"
+-- rows behind PROPERTY_REGISTRATION_RESTRUCTURE_AND_SKIPPING. Like the rest of
+-- this seed, addresses are NOT inserted: each property claims a distinct existing
+-- active AddressBase/NGD address not already used by an active property (the
+-- combined insert below picks them via the available_addresses(n) function). Occupied
+-- properties set last_occupied_date so the "within 28 days" deadline renders.
+-- Fixed ids + ON CONFLICT DO NOTHING keep this idempotent under sql.init mode: always.
+--   49  occupied, licensing + tenancy skipped, no compliance   -> COMBINED banner
```
JasminConterioSW: Some of this seed data might need an update - it shouldn't be _no_ compliance

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsControllerTests.kt
```diff
@@ -109,6 +118,24 @@ class PropertyDetailsControllerTests(
             verify(jointLandlordInvitationService).getPendingAndExpiredInvitations(propertyOwnership)
         }
 
+        @ParameterizedTest(name = "when the provide later feature is {0}")
+        @ValueSource(booleans = [true, false])
+        @WithMockUser(roles = ["LANDLORD"])
+        fun `getPropertyDetails adds provideLaterEnabled to the model reflecting the provide later feature`(featureEnabled: Boolean) {
```
TomHanmer29: nit: I'd be tempted to call this isFeatureEnabled as when I first read it inside the `thenReturn(featureEnabled)` I thought you were returning that the feature was enabled

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDetailsUpdateJourneyTests.kt
TomHanmer29: Are we adding new tests for the equivalent of these or is that a new ticket?

TomHanmer29 (commented):
Couple more little bits on the tests, not looked at the sql yet still :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDetailsUpdateJourneyTests.kt
JasminConterioSW: Yes, there's a `RestructureAndSkippedEnabled` class which contains all the flag on versions already, there's just been a bit of renaming so for example `propertyDetailsPage.propertyDetailsSummaryList.ownershipTypeRow.clickFirstActionLinkAndWait()` looks for the new version..

When the flag is deleted, who whole of the `RestructureAndSkippingDisabled ` subclass should be able to be deleted.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsViewModel.kt
JasminConterioSW: It looks like the idea is to embed something like tenancyDetails on the propertyOwnership (the "occupancy" name I think is misleading, it's called occupation on the ticket but tenancyDetails is more consistent with what we have now).

TenancyDetails could then have non-nullable fields, so that  tenancyDetails is either entirely absent (for an unoccupied or provide this later property) or fully present (all fields non-null).

I'll update the comment and the PDJB-548 ticket

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsViewModel.kt
JasminConterioSW: Copilot says
` These became per-section lists (from the old single eager  propertyRecord ) because the view model currently serves two layouts behind  PROPERTY_REGISTRATION_RESTRUCTURE_AND_SKIPPING , and each request only renders one of them.  by lazy  avoids building the section set for the layout that isn't rendered (e.g. we never compute the  beforePdjb939*  record on a flag-on page`

But actually that might not be the right thing to do as we don't want to have to re-write it when we take out the old version, I'll get it to take another look

JasminConterioSW: It's still using `by lazy` for fields on the flag-off version (so the flag-on version doesn't try to load them) but not for the flag-on (new standard) version.
So the remaining `by lazy` ones will get deleting with flag pdjb939

src/main/resources/data-integration.sql
```diff
@@ -1,3 +1,17 @@
...
+-- cohorts below reference the AddressBase/NGD addresses already present rather than
+-- inserting their own; each claims distinct free addresses by joining this function
+-- on rn. The inner LIMIT n keeps callers from ranking the whole address table.
+-- Defined with a single-quoted SQL body (no dollar-quoting) because the spring.sql.init
+-- runner splits scripts on ';' and cannot parse dollar-quoted function blocks.
+-- =============================================================================
+CREATE OR REPLACE FUNCTION available_addresses(n integer)
+    RETURNS TABLE (address_id bigint, rn bigint)
+    LANGUAGE sql
+AS 'SELECT id, ROW_NUMBER() OVER (ORDER BY id) FROM (SELECT a.id FROM address a WHERE a.is_active AND NOT EXISTS (SELECT 1 FROM property_ownership po WHERE po.is_active AND po.address_id = a.id) ORDER BY a.id LIMIT $1) limited';
```
JasminConterioSW, line 13: This is the approach used previously to get available addresses when seeding property ownerships to test the metrics data - this is just to make it re-usable (and use it for generating property ownerships for this QA)

Travis-Softwire:
Sorry to come to this late, but given how much is different with the feature flag on vs off, for the view model I actually think we want to lift the choice of which view model/ template we want to use right up to the controller level and have two of each below that so that we have two clean implementations all the way down so that we can just delete the old one easily and only have to parse the feature flag logic once right at the start.

Travis-Softwire (changes requested):
Sorry, I think we've maybe gone in slightly the wrong direction with how we're gating the feature behind the feature flag on this occassion! Hopefully not too much of a pain to split it into two different view model classes and two different templates/ fragments though

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsBeforePdjb939ViewModel.kt
```diff
@@ -0,0 +1,80 @@
...
+    propertyOwnership: PropertyOwnership,
+    isLandlordView: Boolean = true,
+    messageSource: MessageSource,
+) : PropertyDetailsViewModelBase(propertyOwnership, isLandlordView, messageSource) {
+    // Property state may have been created while the flag was on, so an occupied property can be in a
+    // provide-later state with no tenancy details. This view cannot render those, so it treats such a
+    // property as unoccupied and hides the tenancy section.
+    private val isTenancyProvideLater = propertyOwnership.tenancyProvideLater == true
+
+    val effectivelyOccupied = isOccupied && !isTenancyProvideLater
```
JasminConterioSW, line 17: This has been updated to cope with properties which have been added with the flag on with provide this later (otherwise you get a null pointer exception if you later need to switch the flag off and view the properties)

In this case I think we just show it as unoccupied

JasminConterioSW:
> Sorry, I think we've maybe gone in slightly the wrong direction with how we're gating the feature behind the feature flag on this occassion! Hopefully not too much of a pain to split it into two different view model classes and two different templates/ fragments though

It's now refactored into separate templates and view models, also extracted the landlord details and compliance details tab contents into their own fragments (they are shared between the two templates)

src/main/resources/messages/propertyDetails.yml
```diff
@@ -164,6 +176,17 @@ propertyRecord:
       rowName: When rent is paid
       changeLinkAriaLabel: Change rent frequency or rent amount
     rentAmount: Rent amount
+  beforePdjb939:
+    address: Property address
+    uprn: UPRN
+    'uprn.unavailable': Not available
+    ownershipType: Ownership type
+    licensingInformation:
+      heading: Licensing information
```
Travis-Softwire: Copilot thinks that propertyRegistrationCheckAnswersForm still references this without the beforePdjb939 element, where it no longer exists

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsBeforePdjb939ViewModel.kt
Travis-Softwire: Should this also cover licensing?

src/main/resources/templates/fragments/compliance.html
```diff
@@ -0,0 +1,16 @@
+<!--/*@thymesVar id="complianceInfoTabId" type="java.lang.String"*/-->
```
Travis-Softwire, line 1: Should this (and the other fragement templates) be in the fragments folder?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsViewModelBase.kt
```diff
@@ -0,0 +1,278 @@
...
+        )
+
+    protected fun registrationDateRow(): SummaryListRowViewModel =
+        row(
+            "propertyDetails.propertyRecord.registrationDate",
+            DateTimeHelper.getDateInUK(propertyOwnership.createdDate.toKotlinInstant()),
+            withActionLink = false,
+        )
+
+    protected fun addressRow(): SummaryListRowViewModel =
```
Travis-Softwire, line 68: It looks like we were previously showing the UPRN to council users - did we deliberately drop that from the new version?

src/main/resources/data-integration.sql
```diff
@@ -1,3 +1,17 @@
+-- =============================================================================
+-- available_addresses(n): returns the first n existing active addresses not already
+-- used by an active property, ranked 1..n (rn). Like the rest of this seed, the
+-- cohorts below reference the AddressBase/NGD addresses already present rather than
+-- inserting their own; each claims distinct free addresses by joining this function
+-- on rn. The inner LIMIT n keeps callers from ranking the whole address table.
+-- Defined with a single-quoted SQL body (no dollar-quoting) because the spring.sql.init
+-- runner splits scripts on ';' and cannot parse dollar-quoted function blocks.
+-- =============================================================================
+CREATE OR REPLACE FUNCTION available_addresses(n integer)
```
Travis-Softwire, line 10: How certain are we that this will be performant against an address table of 35m rows?

Travis-Softwire (commented):
A few more Qs from me

src/main/resources/data-integration.sql
JasminConterioSW: It _should_ be the same as what Bill used previously for seeding some test data testing the metrics (this is the block from line 466-477 which has been replaced with the function call)

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsViewModelBase.kt
JasminConterioSW: Apparently the designers thought we'd dropped it ages ago and it hasn't been in designs for a while (but was never explicitly dropped in an update so we never changed the build).

I checked with Chris and he said take it out

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsBeforePdjb939ViewModel.kt
JasminConterioSW: @C:\Users\JasCon\AppData\Local\Temp\reply2.md

JasminConterioSW: Oh copilot, when I said you could post the reply, that's not quite what I meant... 🙄

JasminConterioSW: The  effectivelyOccupied  guard hides the tenancy section only because  furnishedStatusRow() / rentFrequencyRow() / rentAmountRow()  use  !!  on nullable fields — they throw when tenancy is deferred. The licensing rows are already null-safe ( license?.let{} ?: 
   NO_LICENSING , and  licensingNumberRow  guards  license != null ), so a deferred licence can't crash — it just renders "No licensing". So licensing does not need the same treatment.

src/main/resources/messages/propertyDetails.yml
JasminConterioSW: Should be added back in now!

Travis-Softwire (approved):
LGTM!

PR #1570, opened 2026-07-09
PDJB-1269: Remove compliance-actions-may2026-redesign feature flag dead code

JasminConterioSW:
## Ticket number

PDJB-1269

## Goal of change

Remove dead code that was only reachable when the `compliance-actions-may2026-redesign` feature flag was off, now that the redesign is permanent behaviour on `main`.

This is the first of three stacked PRs for PDJB-1269 (removing the `private-beta-release-2` flags and their dead code). It removes the flag-off code path for the compliance-actions redesign only.

- Removes the flag-off compliance actions page and its `Old`/`Legacy` view-model services, keeping the redesign path as the single implementation
- Collapses `ComplianceStatusDataModel`, `ComplianceActionViewModelBuilder` and `PropertyComplianceService` to the redesign variants and drops the now-neutral `May26Redesign` naming
- Simplifies `LandlordController` to always serve the paginated compliance actions page (removes the `FeatureFlagManager` branch)
- Merges the `NotificationBannerViewModelService` flip interface into a single concrete service
- Flattens the `complianceActions` message keys and removes an orphaned `hintText` key

The `COMPLIANCE_ACTIONS_MAY2026_REDESIGN` flag constant and its YAML config are intentionally retained; they are removed in the final stacked PR along with `joint-landlords`.

This PR is stacked: PR2 removes `joint-landlords` dead code, PR3 removes both flag constants and the release config. Reviewing in order is easiest.

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyCompliance.kt
```diff
@@ -83,9 +83,6 @@ class PropertyCompliance() : ModifiableAuditableEntity() {
     val isEpcExpired: Boolean?
         get() = epcExpiryDate?.isBefore(LocalDate.now())
 
-    val isEpcNonCompliantDueToExpiry: Boolean
```
JasminConterioSW, line 86: Copilot says this was unused...

Checking on another branch, this was only `ComplianceStatusDataModel` `PropertyCompliance.epcStatusOld` so should be fine to delete

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/PropertyComplianceViewModelFactory.kt
```diff
@@ -13,10 +13,10 @@ private const val VIEW_FULL_EPC_KEY = "propertyCompliance.epcTask.checkEpcAnswer
 
 @PrsdbWebService
 class PropertyComplianceViewModelFactory(
-    private val gasSafetyViewModelService: GasSafetyViewModelService,
-    private val electricalSafetyViewModelService: ElectricalSafetyViewModelService,
-    private val epcViewModelService: EpcViewModelService,
-    private val notificationBannerViewModelService: NotificationBannerViewModelService,
+    private val gasSafetyViewModelFactory: GasSafetyViewModelFactory,
```
samyou-softwire, line 16: I do wonder a little why these are annotated as factories, they feel more like plain services now that return view models. though it looks standard at least in this folder to call them factories so maybe this is fine. what do you think?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyComplianceTabFeatureFlagTests.kt
```diff
@@ -1,102 +0,0 @@
...
-        fun `notification banner is visible when certs are expired`(page: Page) {
-            // Property 9: unoccupied, gas expired, EPC expired
-            val detailsPage = navigator.goToPropertyDetailsLandlordView(9)
-            detailsPage.tabs.goToComplianceInformation()
-
-            assertThat(detailsPage.notificationBanner).isVisible()
-        }
-
-        @Test
-        fun `gas safety card has certificate status row`(page: Page) {
```
samyou-softwire, line 76: just to check we're happy that nothing of value is being lost in removing these tests?

samyou-softwire (approved):
happy with these deletions. some small questions

> This PR is stacked: PR2 removes joint-landlords dead code,

to double check, #1577 PR is around to remove some code that is now no longer needed due to joint-landlords flag being turned on. will there be a conflict here?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/PropertyComplianceViewModelFactory.kt
JasminConterioSW: I think you're right it could be either, I'll leave as-is for now. I guess the main purpose of them is to create the view model from the entity so factory is ok.

Before, there was a factory (this) and a service (which supplied different values to the factory depending on the flag status), so the factory name may also have been out of necessity to distinguish them!

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyComplianceTabFeatureFlagTests.kt
JasminConterioSW: Ooh I'll have a look, the flag on ones might need to be moved...

JasminConterioSW: Yes we should keep the flag on ones, I'll move them to PropertyDetailsTests

PR #1571, opened 2026-07-10
PDJB-1169: Part 3: Add org landlord access table

samyou-softwire:
## Ticket number

PDJB-1269

## Goal of change

add a table for managing access between onelogin users and landlords

creates the table

doesn't yet provide code for using it. I couldn't find a neat way to slot it in right now - we could half do it but I'd want to remove `individual_subject_identifier` at the same time and this felt like too large of a change for right now. I've left access as a whole to be assessed in [PDJB-1275](https://mhclgdigital.atlassian.net/browse/PDJB-1275)

again database design is the crux of the work here

## Checklist

  and any related functionality) (not based on main right now)
  mention that here
- Seed data has been updated as needed for your feature to be tested without having to e.g. register a new property (better served when we can add OL to the database)
- `NftDataSeeder` has been updated to reflect any changes to the database schema (same as above)


src/main/resources/db/migrations/V1_38_0__create_landlord_access_link_table.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE landlord_access_link (
```
Travis-Softwire, line 1: Is the plan to backfill this in a later migration?

Travis-Softwire (approved):
One Q but LGTM

src/main/resources/db/migrations/V1_38_0__create_landlord_access_link_table.sql
samyou-softwire: yes, for now it can only contain OL access links but there are none of those yet. we'll fill it with IL links when we migrate IL to use this table

samyou-softwire:
rebase onto main

PR #1578, opened 2026-07-13
PDJB-941: remove headers from mini cya pages for compliance

TomHanmer29:
## Ticket number

PDJB-941

## Goal of change

For the mini CYA pages on compliance, we want to remove the mini CYA pages. This is because the page title is the same as the section, so we shouldn't show a section.

image

image

image

Removes the section header from gas, electric and EPC mini CYA pages.

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/PropertyRegistrationJourneyFactory.kt
```diff
@@ -436,6 +436,15 @@ class PropertyRegistrationJourneyFactory(
                         )
                 }
             }
+            listOf(
+                journey.checkGasSafetyAnswersStep,
+                journey.checkElectricalSafetyAnswersStep,
+                journey.checkEpcAnswersStep,
+            ).forEach { checkAnswersStep ->
+                configureStep(checkAnswersStep) {
+                    withAdditionalContentProperty { "sectionHeaderInfo" to null }
```
TomHanmer29, line 446: We can either add the override here, or add it to the individual pages in the config. The config is shared between the update pages and this, but I don't think we have designs for the update journeys, so didn't want to change any behaviour there.

samyou-softwire: I think we should raise a ticket so this can be raised with design, and we can add a TODO here to review. it does seem a bit odd right now for this to be different between the two usages

samyou-softwire (approved):

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/PropertyRegistrationJourneyFactory.kt
TomHanmer29: My issue with this is that currently, we control whether we show a header from within the relevant journey factory by adding sections. I think adding a section override to an individual task just makes it more obtuse if we did ever want to add section headers to the CYA page (e.g. if we updated section headers in the future) - in my mind a section header override should just live on the same level as the section.

samyou-softwire: suppose so, but I think that's a question we should answer when we come round to reviewing this, as it'll depend what design want. happy to have a TODO and leave it for now

TomHanmer29: I'm just not sure that a TODO really solves the issue here - if we review it, decide that yep, the update pages don't need a section header, and then put the section header to null on the CYA pages, & then later down the line go back to a version where we do have the section header on the update journey then I think it would be easy to forget the CYA pages have a bit of code in them that nulls their header. But if you really do want a ticket, I can raise it and add a TODO here :)

samyou-softwire: oh I mostly want a ticket so that design decision is tracked; we can decide what is the most practical way at that time

TomHanmer29: I've just had a check through the existing update journeys and no page on an update journey has a section header (since it's not in a section, no update journeys are multi step). Is there any point in raising a design ticket to review this if the behaviour we have currently is consistent for update journeys? If we're still unsure about where the section header override should live it can always be raised in dev catchup

samyou-softwire: oh interesting, so this change actually removes the only case where we were showing section headers? okay I think this is fine then. thanks for looking into this further

maybe could be worth an explanatory comment as it's a bit of surprising line to come across. it looks to be a niche reason why we don't have the header on this specific step, but agree it's more surprising for some arbitrary page to not support headers

TomHanmer29: Will add a comment :)

PR #1580, opened 2026-07-13
PDJB-1293: Add stored is_occupied column to property_ownership

JasminConterioSW:
## Ticket number

PDJB-1293

## Goal of change

Make a property''s occupancy an independently stored value rather than deriving it from the current tenant count.

- Adds a stored `is_occupied` column to `property_ownership`; the migration backfills existing rows from `current_num_tenants > 0`.
- `PropertyOwnership.isOccupied` is now a persisted property instead of a derived getter, so occupancy is the user-answered source of truth and no longer tied to tenant count.
- Property registration and the update-occupancy journey now persist `is_occupied` from the user''s answer.
- Seed data (`data-*.sql`) and `NftDataSeeder` updated to populate `is_occupied`.

`is_occupied` is deliberately decoupled from tenant count going forward. The one-off migration backfill mirrors the previous derived semantics (`current_num_tenants > 0`) so existing data is unchanged.

## Checklist


src/main/resources/db/migrations/V1_39_0__add_is_occupied_to_property_ownership.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE property_ownership ADD COLUMN is_occupied BOOLEAN NOT NULL DEFAULT FALSE;
```
TomHanmer29, line 1: Are we happy with this always being true or false? We won't ever be able to skip this question?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/occupancy/UpdateOccupancyCyaConfig.kt
TomHanmer29, line 56: I'm assuming it'll still be correct to reset this on unoccupancy? Seems to make sense but just wanted to flag :)

TomHanmer29 (approved):
LGTM, a couple bits we should think extra hard on 🧠 but I think it's all good as is :)

src/main/resources/db/migrations/V1_39_0__add_is_occupied_to_property_ownership.sql
JasminConterioSW: Never say never...

At the moment though I think this is a required question before all the bits than can be delegated

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/occupancy/UpdateOccupancyCyaConfig.kt
JasminConterioSW: This one I think is fine, but thinking about it I was wondering whether number of bedrooms should not be nullified when resetting occupancy...

JasminConterioSW: Ok there's a ticket for the new occupancy update journey so I might add a TODO for that to check which fields should be updated

TomHanmer29: Yes, definitely number of bedrooms should NOT be affected by occupancy, we ask it before that now. Maybe that should have been updated in PDJB-941, but it's probably fine to do it in the new occupancy update journey?

PR #1582, opened 2026-07-13
PDJB-1034: Fix drifting CloudFront error-rate metrics

Bill-Haigh:
## Ticket number

PDJB-1034

## Goal of change

Fix the incorrect, refresh-to-refresh drifting 2dp CloudFront error-rate (and average-memory) percentages on the system operator metrics page for a fixed historic date range.

- Volume-weight the CloudWatch `Average` statistic (`Σ average×sampleCount / Σ sampleCount`) instead of taking an unweighted average-of-averages, so the value matches the AWS console and is invariant to how CloudWatch buckets the data. Returns "No data" when the total sample count is zero or the result is non-finite.
- Request `SampleCount` alongside `Average` (only for the average case) so the weighting has sample counts to work with.
- Make the query period age-aware: round the bucket size up to a valid CloudWatch resolution based on the age of the range (60s ≤15 days, 300s ≤63 days, else 3600s, 60s floor), so an aged fixed range no longer drifts as CloudWatch rolls data into coarser resolutions.
- Isolate each metric fetch in its own try/catch so one failing metric only blanks its own row instead of collapsing all six to "No data".
- Update `docs/MetricsReadMe.md` to describe the new behaviour.

A `Clock` is injected into `AwsCloudWatchMetricsClient` (default `Clock.systemUTC()`) purely for test determinism. There is no `Clock` bean; this relies on Spring supporting Kotlin optional constructor parameters (falls back to the default), consistent with the existing `DateTimeHelper`/`ReportingPeriod` pattern.

Only the two affected unit test classes were run (targeted), not the full suite.

## Checklist

- [x] TODO comments referencing this JIRA ticket have been searched for and removed - none were found


src/main/kotlin/uk/gov/communities/prsdb/webapp/clients/AwsCloudWatchMetricsClient.kt
```diff
@@ -34,27 +40,62 @@ class AwsCloudWatchMetricsClient(
...
+        return (weightedSum / totalSamples).takeIf { it.isFinite() }
+    }
+
     private fun periodSeconds(period: ReportingPeriod): Int {
         val rangeSeconds = period.end.epochSecond - period.start.epochSecond
-        val granularity = ((rangeSeconds / GRANULARITY_DIVISIONS) / SECONDS_PER_MINUTE) * SECONDS_PER_MINUTE
-        return max(granularity, MIN_PERIOD_SECONDS).toInt()
+        val rawBucketSeconds = rangeSeconds / GRANULARITY_DIVISIONS
+        val resolution = validResolutionSeconds(period.start)
+        val roundedUp = ceil(rawBucketSeconds.toDouble() / resolution).toLong() * resolution
```
JasminConterioSW: Do we need the .toDouble() conversion or can everything be kept as longs?

JasminConterioSW: To keep it a longs, you can use 
`Math.ceilDiv(rawBucketSeconds, resolution) * resolution`

It may not matter as everything is getting turned into an int at the end anyway, just saves some converting!

JasminConterioSW (approved):
One minor comment but LGTM

PR #1583, opened 2026-07-13
PDJB-1269: Remove joint-landlords dead code

JasminConterioSW:
## Ticket number

PDJB-1269

## Goal of change

Remove the dead code that only ran when the `joint-landlords` feature flag was disabled, now that the flag is permanently enabled in `main`.

- Collapsed four `@PrsdbFlip` joint-landlord email services (deleted the no-op flag-off implementations and their interfaces, promoted the flag-on implementations to the base names): invitation deletion, invitation expiry, other-landlord-left, and swap-to-individual nudge.
- Removed `@AvailableWhenFeatureEnabled("joint-landlords")` from the joint-landlord controller endpoints (invite, accept/reject, cancel, leave property, switch to individual) so they are always available.
- Removed the `checkFeature(JOINT_LANDLORDS)` gating (and the corresponding Thymeleaf conditionals) from the landlord privacy notice and property details controllers/templates so the redesigned behaviour is always shown.
- Inlined and deleted `JointLandlordsPropertyRegistrationStrategy`; the joint-landlords task, CYA row, and save step are now always part of the property registration journey.
- Removed the now-unused `PropertyDetailsLandlordViewModelBuilder.fromEntity` (the old single-landlord property-details view, replaced by the always-on summary cards) and its tests.
- Updated all affected unit, controller, and integration tests to the single (flag-on) behaviour, removing tests that only asserted the now-removed flag-off behaviour.

- **The `JOINT_LANDLORDS` constant and `featureFlagNames` entry are intentionally retained** so the flag-config validation stays in sync until PR3.

## Checklist

  mention that here


src/main/resources/templates/landlordPrivacyNotice.html
```diff
@@ -33,7 +33,7 @@ <h3 class="govuk-heading-s" th:text="#{privacyNotice.section.two.heading}">priva
                 <li th:text="#{privacyNotice.section.two.paragraph.one.bullet.six}">privacyNotice.section.two.paragraph.one.bullet.six</li>
                 <li th:text="#{privacyNotice.section.two.paragraph.one.bullet.seven}">privacyNotice.section.two.paragraph.one.bullet.seven</li>
             </ul>
-            <th:block th:if="${jointLandlordsEnabled}">
+            <th:block>
```
Travis-Softwire: Do we still need the blocks if we're dropping the if statement?

Travis-Softwire (approved):
One minor point, plus apparently the Feature flags readme uses the JL flag as an example so we might want to up date it (or maybe it should be a fictious one?). But otherwise LGTM!

src/main/resources/templates/landlordPrivacyNotice.html
JasminConterioSW: That's a point, no we don't!

PR #1593, opened 2026-07-15
PDJB-990: Add licenseProvideLater and tenancyProvideLater field to PropertyOwnership entity

bnjn-mt:
## Ticket number

PDJB-990

## Goal of change

Add a database field to track whether a landlord has chosen to provide their licence later.
Add a database field to track whether a landlord has chosen to provide their tenancy later.

- Adds `licenseProvideLater` boolean field  to `PropertyOwnership` entity
- Adds `tenancyProvideLater` boolean field  to `PropertyOwnership` entity
- Adds DB migration to add `license_provide_later BOOLEAN` and `tenancy_provide_later BOOLEAN` columns to `property_ownership` table
- Update test data sql files and seeder

## Checklist

Delete any that are not applicable, and add an explanation below for any that are applicable but haven't been done


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyOwnership.kt
```diff
@@ -104,6 +104,9 @@ class PropertyOwnership() : ModifiableAuditableEntity() {
 
     var lastOccupiedDate: LocalDate? = null
 
+    @Column(nullable = false)
+    var licenseProvideLater: Boolean = false
```
JasminConterioSW: The compliance versions I think _are_ nullable and default to null, we can do the same here.
That way the migration doesn't need to back-fill any existing records

JasminConterioSW:
We'll need to make sure to update the nft seeder (it's one of the checklist items on the PR template) for the new fields (copilot should sort it I think). It doesn't necessarily need to be in this PR though.

For context: the NFT seeder can be run to generate a "realistic" data set to add to the nft environment, which can then be used for testing. But it needs to know all the fields that need populating and what to populate them with

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyOwnership.kt
```diff
@@ -104,6 +104,8 @@ class PropertyOwnership() : ModifiableAuditableEntity() {
 
     var lastOccupiedDate: LocalDate? = null
 
+    var licenseProvideLater: Boolean? = null
+
```
JasminConterioSW, line 108: `tenancyProvideLater` needs to be added to this entity too

bnjn-mt: Good spot. I've added it in this [commit](https://github.com/communitiesuk/prsdb-webapp/pull/1593/changes/ed85c72f5cfcb605aa9c4ff1e28103a2c6226de7)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyOwnership.kt
```diff
@@ -125,6 +129,8 @@ class PropertyOwnership() : ModifiableAuditableEntity() {
         customPropertyType: String? = null,
         lastOccupiedDate: LocalDate? = null,
         markedJointLandlord: Boolean = false,
+        licenseProvideLater: Boolean = false,
+        tenancyProvideLater: Boolean = false,
```
JasminConterioSW: These can also be nullable booleans defaulting to null I think

JasminConterioSW (approved):
One minor comment but LGTM

Have you tried building locally to make sure it's happy? That'll test running the migration, and if you're setup to look at the database locally you can check that the new columns appear :)

src/main/resources/db/migrations/V1_40_0__add_provide_later_fields_to_property_ownership.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE property_ownership ADD COLUMN license_provide_later BOOLEAN;
```
Travis-Softwire, line 1: The new entity constructor has us setting `false` by default rather than null - so we'll have a mismatch between new records (always not null in practice) and existing ones (will all be null). I think we either want to make the default null, or set a default of false here

src/main/resources/data-integration.sql
```diff
@@ -248,6 +248,9 @@ VALUES (1, true, 1, 1, 2, 9, 1, '2024-10-15 00:00:00+00', null, 1,
 SELECT setval(pg_get_serial_sequence('property_ownership', 'id'), (SELECT MAX(id) FROM property_ownership));
 
 UPDATE property_ownership SET marked_joint_landlord = true WHERE id = 1;
+-- Properties with no license where the landlord has indicated they will provide it later:
+-- id=1 is occupied (license currently required), id=2 is unoccupied (license required once occupied)
+UPDATE property_ownership SET license_provide_later = true WHERE id IN (1, 2);
```
Travis-Softwire: Hadn't spotted we were doing this - what happens to this (and the existing one above) if someone has deleted that property from integration? Same point for the other envs

Travis-Softwire (approved):
A couple of comments from me. Happy with this once they're addressed though 👍

src/main/resources/data-integration.sql
JasminConterioSW: Running it locally, it seems to just do nothing if the property doesn't exist.

I'm likely to add some new properties covering the provide this later cases for qa in PDJB-1048, so the update statement there could maybe come out unless Ben is using it for testing?

bnjn-mt: I'm not using this for testing currently, but I will be in a future PR. I'm happy to drop it if that's easier?

JasminConterioSW: I don't mind either way, although slightly surprised that you need properties to be seeded as providing later as PDJB-990 is for making new properties?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/PropertyOwnership.kt
bnjn-mt: I've made this change and tested it locally

src/main/resources/db/migrations/V1_40_0__add_provide_later_fields_to_property_ownership.sql
bnjn-mt: I've made licenseProvideLater and tenancyProvideLater nullable with a default of null in PropertyOwnership

[Commit](https://github.com/communitiesuk/prsdb-webapp/pull/1593/commits/8760838ed94309297319eac389d09dc5b6b0c999)

src/main/resources/data-integration.sql
bnjn-mt: Good point. I've removed these statements

PR #1594, opened 2026-07-15
PDJB-896: Migrate landlord registration tasks to composition model

AEPR:
## Ticket number
PDJB-896

## Goal of change
Migrate the landlord-registration journey tasks from the inheritance-based state model to the new composition-based `DuplicableTask` model.

- Each landlord-registration task (`IdentityTask`, `IndividualLandlordRegistrationTask`, `OrgLandlordRegistrationTask`, and the root `LandlordRegistrationTask`) is now "self-stated": it extends `DuplicableTask` and implements its own narrow state interface, owning its steps/sub-tasks by composition rather than inheriting an aggregate state interface.
- Tasks that need another task''s data now compose that task and read `state.someTask.member` instead of sharing a union state interface.
- The cross-cutting CYA machinery is hoisted onto the root task; the two journey state classes (`LandlordRegistrationJourney`, `AcceptOrRejectJointLandlordInvitationJourney`) are trimmed to compose the root task and forward `checkingAnswersFor`.
- Behaviour-preserving: tasks are bound with a null route prefix so all delegate storage keys and URLs are unchanged.

This is the first of several PRs migrating all journeys to composition. Highest-risk area is the CYA flow, especially its reuse from the Accept/Reject joint-landlord-invitation journey - covered by the passing integration suites.

## Checklist
- [x] Test suite has been run in full locally and is passing (landlord-registration + accept/reject suites: 151/151)

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
@@ -1,12 +1,15 @@
 package uk.gov.communities.prsdb.webapp.journeys.landlordRegistration.tasks
 
+import kotlinx.datetime.Instant
+import org.springframework.beans.factory.ObjectFactory
 import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.JourneyFrameworkComponent
 import uk.gov.communities.prsdb.webapp.config.managers.FeatureFlagManager
 import uk.gov.communities.prsdb.webapp.constants.ORGANISATION_LANDLORD_REGISTRATION
 import uk.gov.communities.prsdb.webapp.journeys.Destination
+import uk.gov.communities.prsdb.webapp.journeys.DuplicableTask
+import uk.gov.communities.prsdb.webapp.journeys.JourneyStateService
 import uk.gov.communities.prsdb.webapp.journeys.OrParents
 import uk.gov.communities.prsdb.webapp.journeys.StepLifecycleOrchestrator
-import uk.gov.communities.prsdb.webapp.journeys.Task
 import uk.gov.communities.prsdb.webapp.journeys.builders.JourneyBuilder.Companion.journey
 import uk.gov.communities.prsdb.webapp.journeys.builders.SubJourneyBuilder
 import uk.gov.communities.prsdb.webapp.journeys.hasOutcome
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/OrgLandlordRegistrationTask.kt
```diff
@@ -41,7 +42,41 @@ import uk.gov.communities.prsdb.webapp.journeys.shared.tasks.TrusteeAddressTask
...
+    override val orgGovBodyDetailsStep: OrgGovBodyDetailsStep,
+    override val orgGovBodyMustProvideInfoStep: OrgGovBodyMustProvideInfoStep,
+    override val orgGovBodyWhoToProvideStep: OrgGovBodyWhoToProvideStep,
+    override val orgGovBodyMemberNameStep: OrgGovBodyMemberNameStep,
+    override val orgGovBodyMemberDobStep: OrgGovBodyMemberDobStep,
+    override val orgGovBodyMemberAddressStep: OrgGovBodyMemberAddressStep,
+    override val orgGovBodyMemberListStep: OrgGovBodyMemberListStep,
+    override val orgMainContactStep: OrgMainContactStep,
+    override val orgLandlordCyaStep: OrgLandlordCyaStep,
+) : DuplicableTask<LandlordRegistrationOrgLandlordState>(journeyStateService),
```
AEPR, line 76: Not one for the migration, but this could probably do with being split into tasks itself!

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
```diff
@@ -41,16 +64,21 @@ class LandlordRegistrationTask(
...
-            task(journey.individualLandlordRegistrationTask) {
+            duplicableTask(journey.individualLandlordRegistrationTask) {
                 parents { journey.identityTask.isComplete() }
+                nextStep { journey.cyaStep }
+            }
+            step(journey.cyaStep) {
+                routeSegment(AbstractCheckYourAnswersStep.ROUTE_SEGMENT)
+                parents { journey.individualLandlordRegistrationTask.isComplete() }
                 nextStep { exitStep }
             }
```
AEPR, line 79: The CYA step gets pulled up because it refers to both the identity task and individualLandlordRegistration task. I'm not entirely sure about the reason for the split of tasks here though!

Travis-Softwire (approved):
LGTM!

PR #1595, opened 2026-07-15
PDJB-896: Migrate invite joint landlord task to composition model

AEPR:
## Ticket number
PDJB-896

## Goal of change
Migrate the shared "invite joint landlord" task, and the two journeys that use it, from the inheritance-based state model to the composition-based `DuplicableTask` model.

- The shared invite-joint-landlord task is now "self-stated": it extends `DuplicableTask` and implements its own narrow state interface, owning its four invite steps and its invited-emails/member-id state, rather than inheriting an aggregate state interface.
- Journey-specific email context (logged-in landlord email; pre-existing invited/co-landlord emails looked up by property) is supplied by two concrete subclasses, mirroring the `AddressTask` content-subclass pattern.
- A narrow `LandlordInvitingState` is introduced so the single shared "has any joint landlords invited" step serves both flows without being split; consuming journeys now compose the task and read `state.inviteJointLandlordsTask.*` instead of sharing a union state interface.
- Behaviour-preserving: the task is added with bare delegate keys, so all storage keys and URLs are unchanged.

Second of several PRs migrating all journeys to composition (stacked on the PDJB-896 landlord-registration PR). The two deferred `startInviteJointLandlordStep` TODOs in the update journey factory are intentionally left for a later PR.

## Checklist
- [x] Test suite has been run in full locally and is passing (invite-joint-landlord + joint-landlord + property-registration suites)


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/builders/TaskInitialiser.kt
```diff
@@ -35,12 +41,36 @@ class TaskInitialiser<TStateInit : JourneyState>(
...
     override fun build(registry: DelegateKeyRegistry): List<JourneyStep<*, *, *>> {
+        // Bind the enclosing state (if this task declares a dependency contract) before its sub-journey builds.
+        // dependencies is only read at runtime, so ordering relative to bindRoute/bindKeyRegistry is immaterial.
+        if (task is DuplicableTaskWithDependencies<*, *>) {
+            @Suppress("UNCHECKED_CAST")
+            val dependencyTask = task as DuplicableTaskWithDependencies<TStateInit, TDependencies>
+            dependenciesProvider?.let { dependencyTask.bindDependencies(it()) }
+            if (dependencyTask.requiresDependencies && !dependencyTask.areDependenciesBound) {
+                throw JourneyInitialisationException(
+                    "Task ${dependencyTask::class.simpleName} requires dependencies but withDependencies { } was not called",
```
Travis-Softwire, line 69: Nice - hadn't considered that scenario

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/tasks/JointLandlordsPropertyRegistrationTask.kt
```diff
@@ -44,15 +61,16 @@ class JointLandlordsPropertyRegistrationTask : Task<InviteJointLandlordPropertyR
                 }
                 savable()
             }
-            task(journey.inviteJointLandlordsTask) {
+            duplicableTask(journey.inviteJointLandlordsTask) {
+                withDependencies { this@JointLandlordsPropertyRegistrationTask.dependencies }
```
Travis-Softwire, line 65: Ah I hadn't considered what this looks like for nested duplicable steps!

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/DuplicableTaskWithDependencies.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    // Whether this task must have its dependencies bound at the mount site. True here; DuplicableTask (which has no
+    // dependencies) overrides it to false so a bare duplicableTask(...) { } call needs no withDependencies { }.
+    open val requiresDependencies: Boolean = true
+
+    // Nullable backing field rather than lateinit so that DuplicableTask (TDependencies = Nothing) is legal: for a
+    // plain task this is always null and `dependencies` is never read.
+    private var boundDependencies: TDependencies? = null
+
+    val areDependenciesBound: Boolean get() = boundDependencies != null
```
Travis-Softwire, line 46: Could boundDependencies be a nullable type?

Travis-Softwire (approved):
LGTM! One small comment where we might need to future proof but that's it.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/DuplicableTaskWithDependencies.kt
AEPR: We should just enforce that it isn't I think. I can make this extend `Any`, which enforces non-nullable.

I think that makes sense in this case - in the current implementation you must provide dependencies. If we ever want to allow optional dependencies we can change it.

PR #1596, opened 2026-07-15
PDJB-1269: Remove joint-landlords and compliance-redesign feature flag config [3 of 3]

JasminConterioSW:
## Ticket number

PDJB-1269

## Goal of change

Remove the `joint-landlords` and `compliance-actions-may2026-redesign` feature flags now that their behaviour is
permanently live in production, completing the flag clean-up for PDJB-1269.

This is the final PR (3 of 3) in the PDJB-1269 flag removal, stacked on top of the joint-landlords dead-code
removal PR (#1583). PR1 (#1570) removed the compliance-redesign dead code; PR2 (#1583) removed the joint-landlords
dead code. With no remaining references to either flag in application logic, this PR removes the configuration
itself:

- Removed the `JOINT_LANDLORDS` and `COMPLIANCE_ACTIONS_MAY2026_REDESIGN` constants (and their `featureFlagNames`
  entries) from `FeatureFlagNames.kt`.
- Removed the `PRIVATE_BETA_RELEASE_2` constant from `FeatureFlagReleaseNames.kt`. As it was the only release,
  `featureFlagReleaseNames` is now an empty list.
- Removed both flag entries and the entire `releases:` block from all seven `application*.yml` profiles
  (main + test resources).
- Updated `FeatureFlagConfigTests` to drop the removed flags/release from its expectations and assert that the
  releases list is now empty.

The strict startup validation in `FeatureFlagConfig` treats empty flag/release lists as no-ops, so the configuration
stays in sync.

- **This PR is stacked on #1583** (base branch `chore/PDJB-1269-remove-joint-landlords-dead-code`), not `main`. On
  `main` the `JOINT_LANDLORDS` constant is still referenced by joint-landlords code that only #1583 removes, so this
  must merge after #1583. Please review #1570 and #1583 first.
- The `releases` mechanism is now completely unused in config, but the `FeatureFlagConfig` release-validation code and
  its unit tests (which use mock names) are retained for future use.

## Checklist

- [x] Test suite has been run in full locally and is passing (compile, `FeatureFlagConfigTests`,
  `FeatureFlagConsistencyTests`, `FeatureFlagManagerTests`, `PrsdbTaskApplicationTests`, ktlintCheck)
- [x] This feature is not behind a feature flag. This PR removes now-permanent flags; the behaviour is already live
  in production, so there is no functional change.


.secrets.baseline
```diff
@@ -205,7 +205,7 @@
         "filename": "src\\main\\resources\\application-local.yml",
         "hashed_secret": "ce8c74ce64b55dc94183540aac1aee68fd6d27ad",
         "is_verified": false,
-        "line_number": 138,
+        "line_number": 137,
```
TomHanmer29, line 208: Why has this changed?

TomHanmer29 (approved):
LGTM! 1 QQ

.secrets.baseline
JasminConterioSW: The position of the no-secret (it's `client-secret: local-dev-secret`) has moved because lines above have been deleted.

What looks odd is that it's only changed by one!
Copilot thinks this is actually correct because this baseline is old (from 25th June) and more feature flags have been added since the baseline was generated, so overall the position of that secret has only changed by one line

PR #1599, opened 2026-07-16
PDJB-1269: Remove orphaned compliance form templates, form models and message keys

JasminConterioSW:
## Ticket number

PDJB-1269

## Goal of change

Remove dead code left behind by the compliance journey redesign (PDJB-812): old form templates, orphaned form model classes, and message keys that are no longer referenced anywhere.

This is a pure dead-code cleanup — no functional change.

- **Deletes orphaned compliance/EPC form templates** - these are all old compliance templates, we made new ones when moving the compliance tasks to the property registration journey as the content had changed a lot.
- **Deletes orphaned form model classes** that no live journey wires up
- **Splits `CheckMatchedEpcFormModel.kt`**: the file also defined the still-used `MatchedEpcFormModel` interface (extended by the live `ConfirmEpcDetailsFrom*` models). The dead `CheckMatchedEpcFormModel` class is removed; the interface is preserved in its own `MatchedEpcFormModel.kt`.
- **Removes unreferenced message keys** from `messages/forms.yml`, including whole dead namespaces (`checkComplianceAnswers`, `declaration`, `epcExemptionReason`, `update.*`, etc.). Every removed key was verified to have zero references anywhere in `src`.
- **Deletes unused page objects**

## Checklist

- [x] Test suite has been run in full locally and is passing (relevant integration + unit compile + ktlint)

_No UI, endpoints, journeys, emails, seed data or schema are affected — this is a dead-code removal only._


src/main/resources/messages/forms.yml
```diff
@@ -9,36 +9,20 @@ buttons:
...
-  saveAndContinueToEICR: Save and continue to EICR
-  saveAndContinueToEPC: Save and continue to EPC
-  findProperty: Find property
-  searchAgain: Search again
-  continueToCheckEpc: Continue to check EPC
-  continueToSearch: Continue to search
-  confirmAndSubmitCompliance: Confirm and submit compliance information
-  returnToPropertyRecord: Return to property record
   uploadAndContinue: Upload and continue
-  addAnother: Add another
```
JasminConterioSW, line 29: I found it surprising that this isn't used but it looks legitimate - we use text like "Add another landlord" and "Upload another file" rather than a generic "Add another"

danielledias-MT (approved):

PR #1601, opened 2026-07-16
PDJB-1305: Add property record provide-later notification banners

JasminConterioSW:
## Ticket number

PDJB-1305

## Goal of change

Add "provide this later" notification banners to the property and tenancy details record, shown when a landlord has skipped licensing and/or tenancy details, combined with compliance-certificate warnings where relevant.

- Introduces a single, data-driven `PropertyDetailsNotificationBannerViewModel` for the property record banner. It exposes `messages: List<NotificationMessage>`, where each message is an optional lead-in text followed by ordered links (each with its own before/after text). This replaces the previous two-mechanism approach (a hardcoded provide-later variant banner plus a separate compliance banner coordinated by a suppress flag).
- Shows the appropriate banner based on the record's state: licensing-only, tenancy-only, both skipped, a combined provide-later + compliance sentence (single message, two links), or compliance-only. Banners are only shown for occupied properties in the provide-later cases.
- Uses landlord- vs local-council-specific copy throughout, and splits the "missing compliance certificates" copy by view.
- Names the specific certificate in the expired-certificate banner (adds `linkText` for gas, electrical and EPC).
- Moves compliance banner message construction into `NotificationBannerViewModelService` and lifts the feature-flag decision up to the controller / compliance factory, so the flag is parsed once per request.
- Wires the banner into `PropertyDetailsController` for both the landlord and local-council views.
- Updates the controller to throw if it can't find a compliance record as we always create one anyway - **applies to flag off as well as flag on**

Flag-off (legacy) behaviour is preserved: a parallel `PropertyDetailsBeforePdjb939NotificationBannerViewModel` produces the same banner shape from compliance messages only (using the legacy copy), so the flag-off template renders identically to before. All flag-off banner code is grouped behind the `beforePdjb939` naming so it can be deleted in one pass when the provide-later flag is permanently enabled.

### Provide everything later OR expired compliance and provide property details later
Landlord view
image

Local council view
image

### Provide just tenancy later
Landlord view
image

Local council view
image

### Provide just licensinglater
Landlord view
image

Local council view
image

### Provide licensing and tenancy later
Landlord view
image

Local council view
image

### Provide a compliance certificate later
Landlord view
image

Local council view
image

### Once compliance certificate expired
Landlord and local council view
image

### Multiple compliance certificates expired
Landlord and local council view

image

- The banner deliberately keeps two view models (flag-on `PropertyDetailsNotificationBannerViewModel` and flag-off `PropertyDetailsBeforePdjb939NotificationBannerViewModel`) that both produce the same `messages` shape, so both templates share the same rendering markup and the flag-off variant is trivially removable later.

## Checklist

- [x Screenshots of any UI changes have been added
- [x] Single page integration tests have been added for any unhappy-flow UI features, e.g. validation errors — 
- [x] Branch has been rebased onto main and run locally 
- [x] Seed data has been updated as needed for your feature to be tested 


src/main/resources/templates/propertyDetailsView.html
```diff
@@ -14,8 +14,50 @@
...
+                            th:text="#{${combinedPrefix} + 'middleText'}" th:remove="tag"></span><a
+                            class="govuk-notification-banner__link" th:href="'#' + ${complianceInfoTabId}"
+                            th:text="#{${combinedPrefix} + 'complianceLinkText'}"></a><span
+                            th:text="#{propertyDetails.propertyRecord.notificationBanner.afterLinkText}" th:remove="tag"></span>
+                    </th:block>
+                    <th:block th:case="'BOTH'">
+                        <th:block th:if="${isLandlordView}"
+                                  th:with="bothPrefix='propertyDetails.propertyRecord.notificationBanner.landlord.both.'">
+                            <span th:text="#{${bothPrefix} + 'beforeLinkText'}" th:remove="tag"></span><a
+                                class="govuk-notification-banner__link" th:href="'#' + ${propertyDetailsFragment}"
```
JasminConterioSW: Can these links be made more specific, so actually going to the licencing / tenancy sections rather than the whole tab? What does the compliance version do?

JasminConterioSW: Actually, copilot says you can't link to e.g. the tenancy section if you are currently looking at the compliance tab (and the banner is visible for all tabs). It might be better to keep it linking to the tab.

src/main/resources/messages/propertyDetails.yml
```diff
@@ -111,16 +111,22 @@ complianceInformation:
...
     epc:
       expired:
         mainText: This property’s EPC has expired.
+        linkText: View EPC
     multipleExpired:
       mainText: Multiple compliance certificates for this property have expired.
     missing:
-      mainText: You must add compliance certificates for this property.
+      landlord:
+        mainText: You must finish adding this property’s compliance certificates for this property.
```
JasminConterioSW: This matches the design, but is it right? Added a question to designers on the ticket.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDetailsTests.kt
```diff
@@ -539,18 +559,26 @@ class PropertyDetailsTests : IntegrationTestWithImmutableData("data-local.sql")
...
                     detailsPage.bodyParagraph("These details have not been provided yet"),
                 ).hasCount(1)
             }
+
+            @Test
+            fun `landlord view does not show a provide-later notification banner for an unoccupied property`(page: Page) {
+                navigator.goToPropertyDetailsLandlordView(9)
+
+                assertThat(page.getByText("You must finish adding")).isHidden()
+                assertThat(page.getByText("You must finish providing")).isHidden()
```
JasminConterioSW: Can this assert that detailsPage.notificationBanner is missing instead

src/test/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/NotificationBannerViewModelServiceTests.kt
```diff
@@ -0,0 +1,396 @@
...
+                listOf(
+                    NotificationMessage(
+                        mainText = expectedMissingMainText,
+                        links = expectedLinks,
+                    ),
+                )
+
+            assertEquals(expected, service.getComplianceNotificationMessageKeys(propertyCompliance, landlordView))
+        }
+    }
```
JasminConterioSW, line 286: These are essentially moved from PropertyComplianceViewModelFactoryTests

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/propertyComplianceViewModels/NotificationBannerViewModelService.kt
```diff
@@ -5,42 +5,83 @@ import uk.gov.communities.prsdb.webapp.constants.COMPLIANCE_INFO_FRAGMENT
...
                 }
 
                 statusModel.expiredCertificateCount > 1 -> {
-                    "$NOTIFICATION_KEY_PREFIX.multipleExpired.mainText"
+                    "$NOTIFICATION_KEY_PREFIX.multipleExpired.mainText" to VIEW_COMPLIANCE_CERTIFICATES_KEY
                 }
 
                 isGasExpired -> {
-                    "$NOTIFICATION_KEY_PREFIX.gasCert.expired.mainText"
+                    "$NOTIFICATION_KEY_PREFIX.gasCert.expired.mainText" to "$NOTIFICATION_KEY_PREFIX.gasCert.expired.linkText"
```
Travis-Softwire, line 68: Should these changes also be behind the feature flag?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
```diff
@@ -56,23 +60,29 @@ class PropertyDetailsController(
         val baseUserId = SecurityContextHolder.getContext().authentication.name
         val propertyOwnership = propertyOwnershipService.getPropertyOwnershipIfAuthorizedUser(propertyOwnershipId, baseUserId)
 
-        val propertyCompliance = propertyComplianceService.getComplianceForPropertyOrNull(propertyOwnershipId)
+        val propertyCompliance =
+            propertyComplianceService.getComplianceForPropertyOrNull(propertyOwnershipId)
```
Travis-Softwire, line 62: I think this (and the equivalent change in the POST below) means that we've got a bunch of code in the banner view model + service + tests that we don't need any more?

Travis-Softwire, line 62: Also, do we need to do anything to make sure all existing properties have a compliance record? That should be the case on prod and nft, but might not be on integration and test - could just be a manual script we run rather than anything committed?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
```diff
@@ -170,32 +184,51 @@ class PropertyDetailsController(
...
 
-    // Parse the provide-later feature flag exactly once and select the matching view model + template.
-    // TODO PDJB-939: remove the flag-off branch (and PropertyDetailsBeforePdjb939ViewModel /
-    // propertyDetailsViewBeforePdjb939.html) when the flag is permanently on.
+    private fun buildNotificationBanner(
+        provideLaterEnabled: Boolean,
+        isLandlordView: Boolean,
+        propertyDetails: PropertyDetailsViewModelBase,
+        propertyCompliance: PropertyCompliance,
+    ): Any =
```
Travis-Softwire: Does this need to be an Any type? Could it not be a list of notification messages?

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDetailsTests.kt
```diff
@@ -322,27 +326,27 @@ class PropertyDetailsTests : IntegrationTestWithImmutableData("data-local.sql")
...
             }
 
             @Test
-            fun `is not visible when certs are expired`(page: Page) {
+            fun `is visible and includes correct messages when all certs are expired`(page: Page) {
                 val propertyOwnershipId = 9
                 val detailsPage = navigator.goToPropertyDetailsLocalCouncilView(propertyOwnershipId.toLong())
 
-                assertThat(detailsPage.notificationBanner).isHidden()
+                assertThat(detailsPage.notificationBanner).containsText("Multiple compliance certificates for this property have expired")
```
Travis-Softwire, line 358: I think this isn't actually asserting it's visible - you might want to add another assertion for that

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsNotificationBannerViewModel.kt
```diff
@@ -0,0 +1,117 @@
...
+
+        fun fromState(
+            isLandlordView: Boolean,
+            isOccupied: Boolean,
+            isLicensingProvideLater: Boolean,
+            isTenancyProvideLater: Boolean,
+            complianceMessages: List<NotificationMessage>,
+        ): PropertyDetailsNotificationBannerViewModel {
+            val hasComplianceIssue = complianceMessages.isNotEmpty()
+            val hasPropertyProvideLater = isOccupied && (isLicensingProvideLater || isTenancyProvideLater)
```
Travis-Softwire, line 35: Tracing this through I think this ends up saying that they're missing licensing AND tenancy when they're missing one or the other plus compliance?

Travis-Softwire (approved):
A few minor points, but LGTM once they're sorted :)

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDetailsTests.kt
JasminConterioSW: In our case it should fail if the banner isn't visible, because the banner wouldn't be rendered at all (we don't render it and hide it, we just don't add it). But no harm in adding in another assertion here!

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsNotificationBannerViewModel.kt
JasminConterioSW: I'll double check which message is shown, but the designs show a message referring to "property and tenancy details and valid compliance certificates" if they are missing one or the other (or both) plus compliance.

The "property and tenancy details" matches the tab name.

image

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsController.kt
JasminConterioSW: Running a script against integration and test sounds like a good idea, I'll do that before this merges!

JasminConterioSW: Actually I should check nft too...

The seeder _wasn't_ always adding a compliance record when a property_ownership was added (probably because that wasn't the original behaviour) - I'm updating the NftDataSeeder now

PR #1604, opened 2026-07-16
PDJB-1290: Add another functionality for governing body

samyou-softwire:
## Ticket number

[PDJB-1290](https://mhclgdigital.atlassian.net/browse/PDJB-1290)

## Goal of change

make the edit, remove, add another buttons work for governing body for organisation landlord

adds some internal steps with routes for deleting gov members and setting up the form fields to edit an existing gov member

sets the back button urls as required

links these buttons back up

since the add gov member journey normally clears the previous questions by default the 'add another' button already just works

for the edit steps, we override `resolvePageContent` to prefill content from the state map. this is what JL flow does as well. the alternative would be setting up the state directly and relying on the existing function that prefills forms from the state. this way is a bit more verbose but requires less manipulation of the state

check that the config for the surrounding task looks correct and I've wired the steps properly together

tests are failing right now, due to a known issue with duplicable tasks. will run again when the fix is merged in which should resolve pipeline. happy to have this reviewed in the meantime

## Checklist

- Single page integration tests have been added for any unhappy-flow UI features, e.g. validation errors (no changes to single pages, all journey tests)
  mention that here

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgGovBodyMemberListStepConfig.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgGovBodyMemberListStepConfig.kt
@@ -29,6 +29,12 @@ class OrgGovBodyMemberListStepConfig(
             "addAnotherUrl" to Destination(state.orgGovBodyWhoToProvideStep).toUrlStringOrNull(),
         )
 
+    override fun afterStepIsReached(state: LandlordRegistrationOrgLandlordState) {
+        // ensure that if you ever get to this page we reset any state that is used by one of the buttons.
+        // this means we can be certain all the buttons will always work even if you use the browser back buttons.
+        state.editingGovBodyMemberId = null
+    }
+
     private fun getMemberRows(state: LandlordRegistrationOrgLandlordState): List<SummaryListRowViewModel> {
         val membersMap = state.governingBodyMembersMap ?: emptyMap()
         return membersMap
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgGovBodyMemberListStepConfig.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgGovBodyMemberListStepConfig.kt
@@ -1,17 +1,21 @@
 package uk.gov.communities.prsdb.webapp.journeys.landlordRegistration.stepConfig
 
 import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.JourneyFrameworkComponent
+import uk.gov.communities.prsdb.webapp.exceptions.PrsdbWebException
 import uk.gov.communities.prsdb.webapp.journeys.AbstractRequestableStepConfig
+import uk.gov.communities.prsdb.webapp.journeys.Destination
 import uk.gov.communities.prsdb.webapp.journeys.JourneyStep.RequestableStep
 import uk.gov.communities.prsdb.webapp.journeys.landlordRegistration.states.LandlordRegistrationOrgLandlordState
 import uk.gov.communities.prsdb.webapp.journeys.shared.Complete
 import uk.gov.communities.prsdb.webapp.models.requestModels.formModels.NoInputFormModel
 import uk.gov.communities.prsdb.webapp.models.viewModels.summaryModels.SummaryListRowActionsViewModel
 import uk.gov.communities.prsdb.webapp.models.viewModels.summaryModels.SummaryListRowViewModel
+import uk.gov.communities.prsdb.webapp.services.CollectionKeyParameterService
 
 @JourneyFrameworkComponent
-class OrgGovBodyMemberListStepConfig :
-    AbstractRequestableStepConfig<Complete, NoInputFormModel, LandlordRegistrationOrgLandlordState>() {
+class OrgGovBodyMemberListStepConfig(
+    private val urlParameterService: CollectionKeyParameterService,
+) : AbstractRequestableStepConfig<Complete, NoInputFormModel, LandlordRegistrationOrgLandlordState>() {
     override val formModelClass = NoInputFormModel::class
 
     override fun getStepSpecificContent(state: LandlordRegistrationOrgLandlordState) =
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/LoadGovBodyMemberForEditStepConfig.kt
```diff
@@ -0,0 +1,71 @@
...
+    override fun afterStepIsReached(state: LandlordRegistrationOrgLandlordState) {
+        val keyToEdit = collectionKeyParameterService.getParameterOrNull() ?: return
+        val member = state.governingBodyMembersMap?.get(keyToEdit) ?: return
+
+        state.editingGovBodyMemberId = keyToEdit
+
+        state.addStepData(
+            OrgGovBodyWhoToProvideStep.ROUTE_SEGMENT,
+            mapOf("whoToProvide" to member.type),
+        )
```
samyou-softwire: I was surprised that this seems to be the usual way, is there no way to update the step directly?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/SetStateForGovBodyMemberEditStepConfig.kt
samyou-softwire, line 1: this step used to do a lot more to prefill state but now it just saves the ID to state (after verifying it's a valid ID)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/SaveGovBodyMemberStepConfig.kt
```diff
@@ -42,9 +48,9 @@ class SaveGovBodyMemberStepConfig : AbstractInternalStepConfig<Complete, Landlor
                 address = address,
             )
 
-        currentMap[nextKey] = member
+        currentMap[targetKey] = member
         state.governingBodyMembersMap = currentMap
-        state.nextGoverningBodyMemberId = nextKey + 1
+        state.editingGovBodyMemberId = null
```
Travis-Softwire, line 53: Is this the only place we clear the member Id? What happens if a user uses the back button from the list page? I wonder if we should clear this when we load the list page to be on the safe side? And have the back link from the list page to to the page before the start of the loop, as it does for JLs and compliance certs?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgGovBodyMemberListStepConfig.kt
```diff
@@ -23,26 +27,41 @@ class OrgGovBodyMemberListStepConfig :
...
-                            SummaryListRowActionsViewModel(text = "forms.links.change", url = "#"),
-                            // TODO: PDJB-1290 - Replace with real remove URL
-                            SummaryListRowActionsViewModel(text = "forms.links.remove", url = "#"),
+                            SummaryListRowActionsViewModel(
+                                text = "forms.links.change",
+                                url =
+                                    Destination(state.setStateForGovBodyMemberEditStep)
+                                        .withUrlParameter(urlParameterService.createParameterPair(internalIndex))
+                                        .toUrlStringOrNull()
+                                        ?: throw PrsdbWebException(
```
Travis-Softwire: I don't think we normally throw from here when the URL is null? Don't we normally just not display the link by using the SummaryListRowActionsInputWithDestination?

Travis-Softwire (changes requested):
One behavioural point that I think needs fixing, plus one minor point for consistency with our other add another loops.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/SaveGovBodyMemberStepConfig.kt
samyou-softwire: > And have the back link from the list page to to the page before the start of the loop

this is how it'll work right now, the only way to get to the add another steps is to press change, press add another, or have 0 people added. no back links are be able to enter those steps

you can confuse it a little using the browser back button, but I think this'll be very difficult to fix

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgGovBodyMemberListStepConfig.kt
samyou-softwire: have updated this

samyou-softwire:
rebase onto main

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/SaveGovBodyMemberStepConfig.kt
Travis-Softwire: Hmm, I'm not sure that's right. If you had an entry, hit the change link, then backed out with the back link to the list page, then hit 'add another' then I think you'll be overwriting the entry that you hit the change link of rather than creating a new entry. I don't think this is hard to solve though, including for the browser back button - I think it just needs to clear the editing member Id when you arrive on the list page?

samyou-softwire: sure, I've added a method for this. I get a little nervous about adding these state updaters in places that may be surprising, so I've included a comment on what we're doing here

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgGovBodyMemberListStepConfig.kt
```diff
@@ -23,27 +26,40 @@ class OrgGovBodyMemberListStepConfig :
...
             "summaryListData" to getMemberRows(state),
-            // TODO: PDJB-1290 - Replace with real "add another" URL once implemented
-            "addAnotherUrl" to "#",
+            "addAnotherUrl" to Destination(state.orgGovBodyWhoToProvideStep).toUrlStringOrNull(),
         )
 
+    override fun afterStepIsReached(state: LandlordRegistrationOrgLandlordState) {
+        // ensure that if you ever get to this page we reset any state that is used by one of the buttons.
+        // this means we can be certain all the buttons will always work even if you use the browser back buttons.
+        state.editingGovBodyMemberId = null
```
Travis-Softwire, line 35: Worth a test for this as well?

Travis-Softwire (approved):
One suggestion for adding a test that the ID is reset when landing on the list page, but otherwise LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgGovBodyMemberListStepConfig.kt
samyou-softwire: have done so

PR #1607, opened 2026-07-16
PDJB-990: Skip licensing details during property registration

bnjn-mt:
## Ticket number

PDJB-990

## Goal of change

Allow landlords to defer providing licensing details during the property registration journey via a "Provide this later" option, mirroring the existing skip behaviour for compliance certificates.

Licensing page with Provide this later button
Provide licensing later confirmation
Provide licensing later variant
CYA showing Provide this later

- Adds a "Provide this later" secondary button to the licensing type step, behind the `PROPERTY_REGISTRATION_RESTRUCTURE_AND_SKIPPING` feature flag.
- Adds a new provided confirmation page with two variants — occupied and unoccupied
- Adds message keys in `forms.yml` and `registerProperty.yml`.
- Adds/updates unit tests for the new step configs, the licensing helper and the registration service.

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyRegistrationService.kt
```diff
@@ -151,7 +154,7 @@ class PropertyRegistrationService(
         val address = addressService.findOrCreateAddress(addressModel)
 
         val license =
-            if (licenseType != LicensingType.NO_LICENSING) {
```
samyou-softwire, line 150: what happened to no_licensing?

bnjn-mt: If the user clicks the provide this later button on the licensing type page, `licenseType` is null and this variable will also be null. This state means the user still needs to provide this information and will see the "Provide this later" text on other pages (CYA/update etc.)

We can create a license with every value of the `LicensingType` enum.

My reasoning for this is so we can differentiate between `NO_LICENSING` and null (provide this later).

samyou-softwire: makes sense, though I worry that the previous code signature here suggests that a license type of NO_LICENSING used to be stored as null. also, in the LicenseService there is this snippet in the license updating code
```kotlin
if (updateLicenceType == LicensingType.NO_LICENSING) {
    license?.let { licenseRepository.delete(license) }
    null
}
```
so we may need more work to update the assumption that 'provide later' => null & 'no_licensing' is stored explicitly

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LicenseService.kt
```diff
@@ -36,4 +36,13 @@ class LicenseService(
...
             license
         }
+
+    // A licence is only stored when the landlord selects a licence type other than NO_LICENSING and has not opted to
+    // provide the licensing details later. NO_LICENSING and "provide this later" are both represented by a null licence.
+    companion object {
+        fun licenceShouldBeStored(
+            licenceType: LicensingType?,
+            licenseProvideLater: Boolean = false,
+        ): Boolean = licenceType != null && licenceType != LicensingType.NO_LICENSING && !licenseProvideLater
```
samyou-softwire: to check; when would licenseType be null?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyRegistrationService.kt
bnjn-mt: I've reverted this change and created a [LicenseService.licenceShouldBeStored method](https://github.com/communitiesuk/prsdb-webapp/pull/1607/commits/fddb0d8bf4695aa584cd7765a26c9a2b0ff33a5c) that allows us to keep the existing code signature.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/LicensingTypeFormModel.kt
```diff
@@ -1,15 +1,29 @@
...
+                validatorType = DelegatedPropertyConstraintValidator::class,
+                targetMethod = "licensingTypeIsValidForAction",
+            ),
+        ],
+    )
     var licensingType: LicensingType? = null
 
+    var action: String? = null
+
+    fun licensingTypeIsValidForAction(): Boolean = action == "provideThisLater" || licensingType != null
```
Travis-Softwire: Should we be using the constant here rather than a hard coded string?

src/main/resources/messages/forms.yml
```diff
@@ -661,6 +661,11 @@ checkPropertyAnswers:
     ownership: Ownership type
     licensingType: Licensing type
     noLicensing: None
+    licensingProvideLater: Provide this later
+  additionalLandlords:
```
Travis-Softwire: These headings are a bit suspect...

Travis-Softwire (approved):
A couple of minor points, but otherwise LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LicenseService.kt
bnjn-mt: When the user clicks the 'Provide this later' button during the property registration.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/requestModels/formModels/LicensingTypeFormModel.kt
bnjn-mt: Good spot. I've updated this.

src/main/resources/messages/forms.yml
bnjn-mt: Would something like: `forms.checkPropertyAnswers.propertyDetails.licensing.licensingProvideLater` be better?

Travis-Softwire: Sorry it was the new lines below that were suspect - they don't seem related to this ticket!

bnjn-mt:
This is ready to merge if all looks good to you @samyou-softwire :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LicenseService.kt
samyou-softwire: if `licenceProvideLater` is false and licenceType is not `NO_LICENSING` those two would imply that the license is not null?

thinking about it, it feels like the reason this is getting hard to reason about is that it's surprising that `NO_LICENSING` is a license type registered in the enum but `PROVIDE_LATER` isn't. it leads to code with some quite surprising conditions where we 'just can't' check licenseType

if the database has been laid out this way that's fine, but I think our service layer is suffering. there are only 5 possible states for a license to be in
- selective
- HMO mandatory
- HMO additional
- no license
- provide later

so I think it'll make this code a lot more clearer if we can make the enum reflect that. it's fine if we don't actually end up putting them all in the database (`NO_LICENSING` was already stored as null), but I think this'll help us out now and down the road when it comes to interpreting the database

then the signature of this method could become 
```kotlin
fun licenceShouldBeStored(
            licenceType: LicensingType
        )
```

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/LicenseServiceTests.kt
```diff
@@ -78,4 +80,34 @@ class LicenseServiceTests {
...
+    }
+
+    @Test
+    fun `licenceShouldBeStored returns false for a null licence type`() {
+        assertFalse(LicenseService.licenceShouldBeStored(null))
+    }
+
+    @Test
+    fun `licenceShouldBeStored returns false when the licensing details are being provided later`() {
+        assertFalse(LicenseService.licenceShouldBeStored(null, licenseProvideLater = true))
```
Travis-Softwire: I think this is passing because you're passing in `null` rather than testing the actual functionality? Should it be e.g. `assertFalse(LicenseService.licenceShouldBeStored(LicensingType.SELECTIVE_LICENCE, licenseProvideLater = true))` instead?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LicenseService.kt
Travis-Softwire: Agree it would be nicer to just add an extra enum value here

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/tasks/LicensingTask.kt
```diff
@@ -19,7 +20,6 @@ class LicensingTask(
 ) : Task<LicensingState>() {
     override fun makeSubJourney(state: LicensingState) =
         subJourney(state) {
-            // TODO(PDJB-990): route to the 'provide details about licensing later' page when 'Provide this later' is selected behind the FF: pdjb-939-property-registration-restructure-and-skipping/PROPERTY_REGISTRATION_RESTRUCTURE_AND_SKIPPING
```
Travis-Softwire, line 31: Looks like we're not using the feature flag manager in this class - should we be? Or should it be removed?

src/main/resources/messages/forms.yml
Travis-Softwire: My previous comment has been marked as resolved, but these message keys that don't seem related to the ticket are still here?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfig.kt
```diff
@@ -71,7 +70,7 @@ class SavePropertyRegistrationDataStepConfig(
                 } else {
                     null
                 },
-            licenseType = state.licensingTypeStep.formModel.notNullValue(LicensingTypeFormModel::licensingType),
+            licenseType = state.licensingTypeStep.formModelOrNull?.licensingType,
```
Travis-Softwire: The suggestion RE making provide later a licensing type would also mean that we can keep the notNullValue here which guards against us registering a property with a licensing type of null

Travis-Softwire (changes requested):
A few more comments from me, plus seconding Sammy's earlier suggestion for adding provide later as a licensing type enum value to make some of the checking logic simpler

bnjn-mt:
I've made the suggested changes in this [commit](https://github.com/communitiesuk/prsdb-webapp/pull/1607/changes/b0d8b30910b90dc0475f026e7f612fc3693c0d0e) @Travis-Softwire @samyou-softwire

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/tasks/LicensingTask.kt
```diff
@@ -46,13 +44,20 @@ class LicensingTask(
                 parents { journey.licensingTypeStep.hasOutcome(LicensingTypeMode.HMO_ADDITIONAL_LICENCE) }
                 nextStep { exitStep }
             }
+            step(journey.provideLicensingLaterStep) {
+                routeSegment(ProvideLicensingLaterStep.ROUTE_SEGMENT)
+                parents { journey.licensingTypeStep.hasOutcome(LicensingTypeMode.PROVIDE_LATER) }
+                nextStep { exitStep }
+                savable()
```
Travis-Softwire, line 71: What's the reason for this step being savable but the others aren't?

Travis-Softwire (approved):
This is looking great now, thanks for persevering! I've left one Q, and it looks like there are some merge conflicts to resolve. Assuming those are relatively simple then this all LGTM (if they turn out to require some more substantive changes then shout and I can quickly re-review)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/tasks/LicensingTask.kt
bnjn-mt: Good spot. I've added `savable()` to the other steps

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/tasks/LicensingTask.kt
bnjn-mt: This has been removed

src/main/resources/messages/forms.yml
bnjn-mt: I've made this change

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/steps/SavePropertyRegistrationDataStepConfig.kt
bnjn-mt: I've made this change

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/LicenseServiceTests.kt
bnjn-mt: I've made this change. The `licenceShouldBeStored` no longer requires the boolean arg

src/main/resources/messages/forms.yml
bnjn-mt: I've removed these

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LicenseService.kt
bnjn-mt: We're using a `PROVIDE_LATER` enum value now

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/updateLicensing/UpdatePropertyLicensingJourneyFactory.kt
```diff
@@ -117,9 +119,13 @@ class UpdateLicensingJourney(
 
     override var originalJourneyUpdated: Instant? by delegateProvider.nullableDelegate("originalJourneyUpdated")
     override var cyaUrlPath: String? by delegateProvider.nullableDelegate("cyaRouteSegment")
+
+    override val allowProvideLicensingLaterRoute: Boolean = false
```
Travis-Softwire, line 123: As well as this, do we need to unset the provide later flag in the DB if they update their licensing?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/tasks/LicensingTask.kt
```diff
@@ -12,23 +11,28 @@ import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.HmoAd
...
 class LicensingTask(
     journeyStateService: JourneyStateService,
-    private val featureFlagManager: FeatureFlagManager,
     override val licensingTypeStep: LicensingTypeStep,
     override val selectiveLicenceStep: SelectiveLicenceStep,
     override val hmoMandatoryLicenceStep: HmoMandatoryLicenceStep,
     override val hmoAdditionalLicenceStep: HmoAdditionalLicenceStep,
-) : DuplicableTask<LicensingState>(journeyStateService),
+    override val provideLicensingLaterStep: ProvideLicensingLaterStep,
+) : DuplicableTaskWithDependencies<LicensingState, LicensingDependencies>(journeyStateService),
```
Travis-Softwire: Annoyingly I think some of this has been renamed on main, you might need to rebase

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyRegistrationService.kt
```diff
@@ -38,7 +38,7 @@ class PropertyRegistrationService(
     fun registerProperty(
         addressModel: AddressDataModel,
         propertyType: PropertyType,
-        licenseType: LicensingType,
+        licenseType: LicensingType?,
```
Travis-Softwire: Just realised that this probably doesn't need to be nullable anymore now that provide it later has been added as an enum value of this

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/helpers/LicensingDetailsHelper.kt
```diff
@@ -18,7 +20,18 @@ class LicensingDetailsHelper(
...
+            return listOf(
+                SummaryListRowViewModel.forCheckYourAnswersPage(
+                    "forms.checkPropertyAnswers.propertyDetails.licensingType",
+                    "forms.checkPropertyAnswers.propertyDetails.licensingProvideLater",
+                    Destination.VisitableStep(licensingState.licensingTypeStep, state.getCyaJourneyId(licensingState.licensingTypeStep)),
+                ),
+            )
+        }
+
         return licensingState.licensingTypeStep.formModel.notNullValue(LicensingTypeFormModel::licensingType).let { licensingType ->
```
Travis-Softwire: I think if you use `getLicensingType()` here instead of getting it from the form model you can omit the special case above because the message key convertor will now handle the `PROVIDE_LATER` case for the licensing type?

Travis-Softwire (approved):
One small gap that needs fixing and a few other minor points, but happy for this to be merged once they're addressed, I don't need to see it again :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LicenseService.kt
```diff
@@ -25,7 +25,7 @@ class LicenseService(
         updateLicenceType: LicensingType?,
```
samyou-softwire: can `updateLicenceType` ever actually be null? it makes the code a bit messy below to account for this

samyou-softwire (approved):
happy to approve this, though I do still think it'd be useful to get a method on say, the `PropertyOwnership` entity. it'd be called something similar to `licenseType` which returns the correct enum value by checking the provide later col & the license entity.

likely also worth a comment on the `License` entity too that explains that its own `licenseType` only holds certain values and it shouldn't be used for checking the state of the license

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LicenseService.kt
bnjn-mt: I've made this change to clean up the code and added the suggestions in the parent comment [here](https://github.com/communitiesuk/prsdb-webapp/pull/1607/commits/af65ccf37d316e1ffa0208ea5fb1097766225732)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyRegistrationService.kt
bnjn-mt: I've removed the nullable type

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/updateLicensing/UpdatePropertyLicensingJourneyFactory.kt
bnjn-mt: `PropertyOwnershipService.updateLicensing` now unsets the flag. I've added test coverage

PR #1608, opened 2026-07-16
PDJB-1312: fix healthcheck failure issue

TomHanmer29:
## Ticket number

PDJB-1312

## Goal of change

Fixes the healthcheck failure issue

Does this by early returning for healthcheck instead of trying to render the dashboard.
Also replaces PRSDB flip with flag manager.

## Checklist


Travis-Softwire (approved):
LGTM

PR #1609, opened 2026-07-16
PDJB-1169: Fix landlord model error

samyou-softwire:
## Ticket number

[PDJB-1169](https://mhclgdigital.atlassian.net/browse/PDJB-1169)

## Goal of change

resolves the following warning on app boot

```
Getter methods of lazy classes cannot be final: uk.gov.communities.prsdb.webapp.database.entity.AuditableEntity#getCreatedDate
```

the issue is that hibernate wants to make a proxy object for landlord to allow for lazy loading for a generic landlord reference. it cannot do this if:
- there are final methods
- the class is sealed

this PR removes the following from the landlord objects

since this removes the sealed annotation, I've added a getter to a landlord type to return an enum, this means you can still discriminate on landlord types exhaustively

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/IndividualLandlord.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/IndividualLandlord.kt
@@ -7,12 +7,18 @@ import jakarta.persistence.JoinColumn
 import jakarta.persistence.ManyToOne
 import jakarta.persistence.OneToMany
 import jakarta.persistence.OneToOne
+import jakarta.persistence.Transient
 import uk.gov.communities.prsdb.webapp.constants.ENGLAND_OR_WALES
+import uk.gov.communities.prsdb.webapp.constants.enums.LandlordType
 import java.time.LocalDate
 
 @Entity
 @DiscriminatorValue("0")
 class IndividualLandlord() : Landlord() {
+    @get:Transient
+    override val landlordType: LandlordType
+        get() = LandlordType.INDIVIDUAL
+
     @OneToOne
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/IndividualLandlord.kt
```diff
@@ -7,12 +7,18 @@ import jakarta.persistence.JoinColumn
...
 import uk.gov.communities.prsdb.webapp.constants.ENGLAND_OR_WALES
+import uk.gov.communities.prsdb.webapp.constants.enums.LandlordType
 import java.time.LocalDate
 
 @Entity
 @DiscriminatorValue("0")
 class IndividualLandlord() : Landlord() {
+    @get:Transient
+    override val landlordType: LandlordType
+        get() = LandlordType.INDIVIDUAL
```
TomHanmer29, line 20: May have misunderstood something but do we actually need these getters if we're not using them anywhere currently?

TomHanmer29 (approved):
LGTM! Just 1 small question

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/IndividualLandlord.kt
samyou-softwire: it's for future proofing; I'm fairly certain we will want to discriminate a landlord based on its type, so this enum allows for you to write checks like
- what type of landlord?
  - individual
  - org

and be certain you've covered all cases. sealed used to provide this guarantee

PR #1614, opened 2026-07-17
PDJB-NONE: Keep journey redirects absolute when the journey has no stored base path

Travis-Softwire:
## Ticket number

PDJB-NONE

## Goal of change

Fix broken journey URLs (a duplicated path segment, leading to 404s) that appeared when continuing a registration journey whose session had expired but whose id was still in the URL.

- Journey step URLs are built by prefixing the step path with a base path stored against the journey. When no base path was stored (e.g. the journey had expired from the session but its id was still in the URL), the URL was emitted **relative**, so the browser resolved it against the current path and duplicated the task route segment — e.g. `.../lead-trustee-address/lead-trustee-address/select-address` — which 404s.
- `resolvePathAgainstJourneyBase` now falls back to the base path the journey dispatcher already derives for the current request, so these redirects stay absolute.
- Added a unit test for the fallback, and updated one `RegisterLandlordControllerTests` assertion to expect the (now correct) absolute redirect URL.

- This is shared journey-framework code. Behaviour is unchanged for journeys that already have a stored base path, and for legacy controllers that do not set the request base-path attribute (those stay relative, exactly as before).
- This also fixes a failing integration test on the PDJB-1290 branch (the governing-body "add another" member-address flow), which hits the same root cause because the test seeds journey state into the session without a stored base path.

## Checklist

- [x] Full unit test suite (`testWithoutIntegration`) and `ktlintCheck` pass locally (integration suite runs in CI)
- [x] Branch is based on the latest `main`

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/JourneyStateService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/JourneyStateService.kt
@@ -198,11 +198,11 @@ class JourneyStateService(
             journeyId: String,
         ): String {
             if (path.startsWith("/")) return path
-            val basePath = journeyBasePath(journeyId)
-            // Fallback for journeys whose controller has not yet been ported to the journey dispatcher: no base
-            // path is stored, so the URL stays relative and behaves exactly as before. This fallback (and the
-            // relative-URL branch below) can be removed once every journey controller uses the dispatcher, at
-            // which point a base path is always stored and all journey URLs become absolute.
+            // Prefer the base path stored against the journey. When none is stored - for example the journey has
+            // expired from the session but its id is still in the URL - fall back to the base path the dispatcher
+            // derived for the current request. This keeps redirects absolute: a relative URL would be resolved by
+            // the browser against the current path and duplicate the task route segment.
+            val basePath = journeyBasePath(journeyId) ?: currentRequestBasePath()
             return if (!basePath.isNullOrEmpty()) "$basePath/$path" else path
         }
 
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/JourneyStateService.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/JourneyStateService.kt
@@ -200,8 +200,10 @@ class JourneyStateService(
             if (path.startsWith("/")) return path
             // Prefer the base path stored against the journey. When none is stored - for example the journey has
             // expired from the session but its id is still in the URL - fall back to the base path the dispatcher
-            // derived for the current request. This keeps redirects absolute: a relative URL would be resolved by
-            // the browser against the current path and duplicate the task route segment.
+            // derived for the current request, so the redirect stays absolute rather than relative (a relative URL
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLandlordControllerTests.kt
```diff
@@ -86,7 +85,7 @@ class RegisterLandlordControllerTests(
             .get("$LANDLORD_REGISTRATION_ROUTE/${PrivacyNoticeStep.ROUTE_SEGMENT}")
             .andExpect {
                 status { is3xxRedirection() }
-                redirectedUrl(JourneyStateService.urlWithJourneyState(PrivacyNoticeStep.ROUTE_SEGMENT, journeyId))
+                redirectedUrl("$LANDLORD_REGISTRATION_ROUTE/${PrivacyNoticeStep.ROUTE_SEGMENT}?journeyId=$journeyId")
```
TomHanmer29, line 88: Why have we updated this path?

TomHanmer29 (approved):
LGTM! 1 tiny (probably unrelated) thing

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/JourneyStateService.kt
```diff
@@ -198,11 +198,11 @@ class JourneyStateService(
...
-            // Fallback for journeys whose controller has not yet been ported to the journey dispatcher: no base
-            // path is stored, so the URL stays relative and behaves exactly as before. This fallback (and the
-            // relative-URL branch below) can be removed once every journey controller uses the dispatcher, at
-            // which point a base path is always stored and all journey URLs become absolute.
+            // Prefer the base path stored against the journey. When none is stored - for example the journey has
+            // expired from the session but its id is still in the URL - fall back to the base path the dispatcher
+            // derived for the current request. This keeps redirects absolute: a relative URL would be resolved by
+            // the browser against the current path and duplicate the task route segment.
+            val basePath = journeyBasePath(journeyId) ?: currentRequestBasePath()
             return if (!basePath.isNullOrEmpty()) "$basePath/$path" else path
```
samyou-softwire, line 208: something feels incomplete about this comment; we do still have the logic that would return a relative URL (the `path` only branch below I think?) but I feel like this comment implies we'd never use a relative path

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLandlordControllerTests.kt
Travis-Softwire: This was because the helper is called outside the context of a journey (with the basepath set in session) which means it returns a relative URL now in the test, but the actual URL is now absolute.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/JourneyStateService.kt
Travis-Softwire: Yep, good point. The fallback to `path` is only there for controllers that haven't been migrated to use the dispatcher yet, but agree it's a bit misleading as-is.

PR #1615, opened 2026-07-19
PDJB-1344: Add .wiz file to ignore restrictive-license false positives

Travis-Softwire:
## Ticket number

PDJB-1344

## Goal of change

Stop Wiz Code raising false-positive restrictive-license findings against our dependencies during org-level scans.

Adds a root `.wiz` Wiz Code configuration file that excludes the built-in "Code library with Restrictive license" rule (`wsscr-1`) for this repository's dependency files (`build.gradle.kts` and the `package.json` / `package-lock.json` files). This is Wiz's documented repository-level exception method. The tenant-level "Enable repository configuration file (.wiz)" setting is already enabled, so the file will take effect when the (currently disabled) rule is re-enabled.

- The exclusion is scoped to dependency files only (not repo-wide) and uses `reason: FALSE_POSITIVE`.
- `wsscr-1` was confirmed from the Wiz portal (Policies > Software Management Rules) as the built-in "Code library with Restrictive license" rule.
- No `expires` is set; it can be added if we prefer to time-box the exception.

## Checklist

N/A - configuration-only change. No application code, endpoints, journeys, email templates, or database schema changes.

### DIFF
--- a/.wiz
+++ b/.wiz
@@ -1,0 +1,22 @@
+# Wiz Code scan exclusions for this repository.
+# Docs: https://docs.wiz.io/docs/exception-management-with-wiz-code
+# Note: the "Repository configuration file (.wiz)" setting must be enabled for this
+# repo/connector/tenant in the Wiz portal, otherwise this file has no effect.
+ignore:
+  software_management:
+    # wsscr-1 = built-in "Code library with Restrictive license" rule (false positives)
+    wsscr-1:
+      "/build.gradle.kts":
+        reason: FALSE_POSITIVE
+      "/package.json":
+        reason: FALSE_POSITIVE
+      "/package-lock.json":
+        reason: FALSE_POSITIVE
+      "/scripts/package.json":
+        reason: FALSE_POSITIVE
+      "/scripts/package-lock.json":
+        reason: FALSE_POSITIVE
+      "/scripts/plausible/package.json":
+        reason: FALSE_POSITIVE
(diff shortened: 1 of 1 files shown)


samyou-softwire:
it looks like this would ignore all code license warnings. this could maybe leave us vulnerable if we add a new library and miss out on an important warning

is there a way to scope this denylist a couple specific licenses? then it becomes a job of adding any new libraries we've confirmed fine to that list when we add a new one

Travis-Softwire:
> it looks like this would ignore all code license warnings. this could maybe leave us vulnerable if we add a new library and miss out on an important warning
> 
> is there a way to scope this denylist a couple specific licenses? then it becomes a job of adding any new libraries we've confirmed fine to that list when we add a new one

You're right that it does, but unfortunately no, there's no way to narrow the scope. There is with 'ignore rules' but they don't seem to currently support ignoring comments on PR scans (we're raising it with Wiz support). Given that there are very few types of license that we couldn't use for this project given it's fully OSS we're happy to disable this one in full.

samyou-softwire (approved):
ah alas but happy if we're willing to accept the risk

TomHanmer29 (approved):
Fix as is LGTM. Is there a small risk if we did add a package with a genuinely restrictive license we'd no longer pick it up? Think it's fine to merge as is but we should possibly keep an eye out if we add a new package

PR #1616, opened 2026-07-20
PDJB-1168: org landlord CYA page

TomHanmer29:
## Ticket number

PDJB-1168

## Goal of change

Adds the org landlord registration journey CYA page

image

image

image

Adds the page skeleton with some data that's not been added currently mocked out
Doesn't implement the change links yet

The ticket mentions extending the existing CYA page for this journey, but I decided to separate them out as they need to use different templates, and they don't have any similar fields. It makes the journey in the DSL slightly uglier but otherwise I think it's the right approach.

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
```diff
@@ -115,13 +117,13 @@ class LandlordRegistrationTask(
                 parents { journey.individualLandlordRegistrationTask.isComplete() }
                 nextStep { exitStep }
             }
+            step(journey.orgCyaStep) {
```
TomHanmer29, line 120: This section looking a little ugly is the main downside to using separate steps for the new CYA page, so I can merge them if the reviewer would prefer.

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/AddressDataModel.kt
```diff
@@ -17,6 +17,26 @@ data class AddressDataModel(
     val townName: String? = null,
     val postcode: String? = null,
 ) {
+    fun toMultiLineAddress(): String =
```
TomHanmer29, line 20: This duplicates the address functionality inside of `Address.kt` (we don't multiline an address before saving it to the DB anywhere else)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgLandlordRegistrationCyaStepConfig.kt
```diff
@@ -0,0 +1,415 @@
...
+                    if (verified) {
+                        Destination.Nowhere()
+                    } else {
+                        Destination.VisitableStep(
+                            state.identityTask.dateOfBirthStep,
+                            state.getCyaJourneyId(state.identityTask.dateOfBirthStep),
+                        )
+                    },
+                ),
+                // TODO: PDJB-1172 - replace these dummy email and phone rows with the user's real contact details once collected.
```
samyou-softwire: PDJB-1172 is now a 'won't do'. this is now covered by [PDJB-1282](https://mhclgdigital.atlassian.net/browse/PDJB-1282)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgLandlordRegistrationCyaStepConfig.kt
samyou-softwire, line 1: I'm not super clear with how we deal with CYA pages on the project, but it does concern me a little that all of the conditional logic on questions is re-expressed in this file. (eg do we show charity number?). we'll also need to express the logic again when we do saving to database, which could lead us writing the same logic 3 times!

is there a more manageable way we can tie the conditional logic we have in the journey framework and this CYA page together? I think they are ultimately accomplishing the same task. the answer may be no but I think worth checking if we haven't already

maybe a method that returns a data model that is effectively the same as an Org Landlord? OrgLandlordRegistrationAnswersDataModel? maybe even just LandlordRegistrationAnswersDataModel? what do you think?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
samyou-softwire: merge was the original intention I had. the separate step for the org CYA page was intended as temporary and in the end we'd merge into having a single CYA page. you may need to do some splitting into files in some way so it doesn't end up too large

samyou-softwire (commented):
some comments on the broad approach

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/AddressDataModel.kt
Travis-Softwire: Wait - why are we saving a multi-line address to the DB?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
Travis-Softwire: I don't think this bit is particularly ugly (although we should probably rename the cyaStep). If there's no real duplication I'm fine with keeping them separate. If there's some duplication then possibly a case for a shared base class

Travis-Softwire (commented):
Would like to understand about the multiline address, but the rest LGTM - will leave for Sammy to approve though as he had some other comments

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/AddressDataModel.kt
TomHanmer29: Sorry, didn't make this clear in my original comment. We're grabbing the address from state to show it on the CYA page (as the CYA comes before we save to the DB), but the designs call for the address to be multiline in the CYA page - and the methods to display an address as multiline currently only exist on the Address model which we use after we've saved to the DB. We're not saving a multiline address to the DB :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
TomHanmer29: The only shared bit is overriding `resolveNextDestination` to make it not delete the journey (e.g. if it's in an invite JL journey). I don't think that warrants a whole abstract class, so I'll just rename 👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgLandlordRegistrationCyaStepConfig.kt
TomHanmer29: Yes, I agree it would be nicer to do something like this, but it's not a pattern we have on any other CYA page at the moment. I think it would be nice, but should probably be a wider change with an established convention. Something for the backlog?

samyou-softwire: have raised https://mhclgdigital.atlassian.net/browse/PDJB-1352

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
samyou-softwire: if we're happy to have them as separate steps then I think they should become part of orgLandlordRegistrationTask or individualLandlordRegistrationTask respectively

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgLandlordRegistrationCyaStepConfig.kt
```diff
@@ -0,0 +1,415 @@
...
+    // Overrides AbstractCheckYourAnswersStepConfig, which deleted the journey
+    // We don't want to delete the journey at this stage when this page is included within another journey,
+    // such as accepting a joint landlord invitation
+    override fun resolveNextDestination(
+        state: LandlordRegistrationState,
+        defaultDestination: Destination,
+    ): Destination = defaultDestination
+
+    private fun getYourDetailsCard(state: LandlordRegistrationState): SummaryCardViewModel {
+        val verified = state.identityTask.getIsIdentityVerified()
```
samyou-softwire, line 62: this may change depending on if OL can be verified at all - could we TODO this to [PDJB-1282](https://mhclgdigital.atlassian.net/browse/PDJB-1282) to review then?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgLandlordRegistrationCyaStepConfig.kt
```diff
@@ -0,0 +1,415 @@
...
+                        SummaryListRowViewModel.forCheckYourAnswersPage(
+                            "registerAsALandlord.orgCheckAnswers.landlordDetails.charityCommission",
+                            regulatorMessageKey(regulator),
+                            Destination.VisitableStep(
+                                org.orgCharityRegisteredWithStep,
+                                state.getCyaJourneyId(org.orgCharityRegisteredWithStep),
+                            ),
+                        ),
+                    )
+                    charityNumberRow(state, regulator)?.let { add(it) }
```
samyou-softwire: this method is only called once and obscures the branching a little. I think some vals like `showCharityRegulator` and `showCharityNumber` or similar would it really clear what cases can happen here (no charity, yes charity but no number, yes charity and yes number)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgLandlordRegistrationCyaStepConfig.kt
```diff
@@ -0,0 +1,415 @@
...
+
+            val isCharity = org.orgCharityStep.formModel.notNullValue(OrgCharityFormModel::charity)
+            add(
+                SummaryListRowViewModel.forCheckYourAnswersPage(
+                    "registerAsALandlord.orgCheckAnswers.landlordDetails.registeredCharity",
+                    isCharity,
+                    Destination.VisitableStep(org.orgCharityStep, state.getCyaJourneyId(org.orgCharityStep)),
+                ),
+            )
+            if (isCharity) {
```
samyou-softwire: bit of a nit and we've not been 100% consistent on this, but this check is not `isCharity` but `isRegisteredCharity`, like companies house below

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgLandlordRegistrationCyaStepConfig.kt
```diff
@@ -0,0 +1,415 @@
...
+    private fun dummyRow(
+        headingKey: String,
+        value: Any?,
+    ) = SummaryListRowViewModel(
+        fieldHeading = headingKey,
+        fieldValue = value,
+        actions = listOf(SummaryListRowActionsViewModel("forms.links.change", PLACEHOLDER_CHANGE_URL)),
+    )
+
+    private fun orgCardChangeAction(
```
samyou-softwire: something seems fishy about this method, it doesn't seem very specific to org at all. is this some universal functionality we're reimplementing?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/AddressDataModel.kt
samyou-softwire: this seems a bit odd; what about the OL registration journey would require multiline addresses? is it consistent with IL registration?

samyou-softwire (commented):
took a deeper look, a couple qs about the logic of the CYA page

I notice there aren't any single page tests. this CYA page is quite complicated so these would be appreciated

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgLandlordRegistrationCyaStepConfig.kt
```diff
@@ -0,0 +1,415 @@
...
+            "submitButtonText" to "registerAsALandlord.orgCheckAnswers.submitButton",
+            "yourDetailsCard" to getYourDetailsCard(state),
+            "landlordDetails" to getLandlordDetailsRows(state),
+            "leadTrusteeCard" to getLeadTrusteeCard(state),
+            "governingBodyMemberCards" to getGovBodyMemberCards(state),
+            "mainContactCard" to getMainContactCard(state),
+        )
+
+    override fun afterStepDataIsAdded(state: LandlordRegistrationState) {
+        // TODO: PDJB-1180 - persist the organisation landlord on submit.
```
samyou-softwire, line 47: to check also; on other CYA pages we have a check that the data the user submits == their session. do we have that here? feels like it'd be in scope for this ticket

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/AddressDataModel.kt
TomHanmer29: Good point, we do use single line everywhere else. However, having had a look through the service, the designs for 28 days later have also changed to a multiline address in their CYA so I think this is something the designers want

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgLandlordRegistrationCyaStepConfig.kt
TomHanmer29: This is inherited from the Abstract Cya page (and not overriden)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/OrgLandlordRegistrationCyaStepConfig.kt
TomHanmer29: It's not universal functionality per-se but it's not org specific - it's just that we only use it here since this is the first time we have an org card. We use a similar thing for Epc / electrical / gas so I could extract them out and move them into the summarCard?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
TomHanmer29: I don't think we can do this without a more substantial refactor of the states, as the CYA steps also need to access data from the landlordTypeStep (and I think individual landlord accesses privacyNoticeStep?) which are outside the tasks

samyou-softwire: that's a shame. but makes sense. if we're to have two CYA pages then can the original one be named to say, individualCyaStep?

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/AddressDataModel.kt
samyou-softwire: sounds good; I think we should check then whether there's an update to be done for the IL registration / other addresses on PRSDB CYA pages. doesn't have to be now but I think we should have a ticket for it

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/OrganisationLandlordRegistrationSinglePageTests.kt
```diff
@@ -852,4 +852,39 @@ class OrganisationLandlordRegistrationSinglePageTests : IntegrationTestWithImmut
                 address = AddressDataModel(singleLineAddress = "Test Address"),
             )
     }
+
+    @Nested
+    inner class CheckAnswersStep {
```
samyou-softwire, line 860: I do think for the amount of logic that's part of the org CYA step it does deserve some reasonably thorough tests, especially for the conditional hiding/showing of stuff

samyou-softwire (approved):
one last comment about scope of tests, happy with this overall

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/dataModels/AddressDataModel.kt
TomHanmer29: Have raised [PDJB-1393](https://mhclgdigital.atlassian.net/browse/PDJB-1393) for this

PR #1633, opened 2026-07-22
Chore/pdjb 896 migrate property registration without compliance

AEPR:
## Ticket number

PDJB-896

## Goal of change

Migrate the remaining property registration leaf-tasks (not counting compliance) to the self-stated duplicable-task composition model so the aggregate journey-state inheritance can be dismantled.

Pure refactor - no user-facing behaviour change in either journey variant.

- Migrates the property registration leaf tasks (rent frequency and amount, rent includes bills, households and tenants, licensing, address, occupancy) to self-stated `DuplicableTask`s.
- Updates both the property registration and update-occupancy journey factories to the `duplicableTask` / `duplicableCheckAnswerTask` mount DSL.

- The addition of the `EmbedBuilder` and `embed` function. This essentially allows a step within a task to be added directly to the journey map without the default configuration provided by the task. This is used here to add "checkYourAnswerSteps" where a single step is added to CYA journeys rather than a whole task. This wasn't needed before because all steps also existed on the top level journey.

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/occupancy/UpdateOccupancyJourneyFactory.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/occupancy/UpdateOccupancyJourneyFactory.kt
@@ -31,7 +31,6 @@ import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.tasks.RentI
 import uk.gov.communities.prsdb.webapp.journeys.shared.states.CheckYourAnswersJourneyState
 import uk.gov.communities.prsdb.webapp.journeys.shared.states.CheckYourAnswersJourneyState.Companion.checkAnswerStep
 import uk.gov.communities.prsdb.webapp.journeys.shared.states.CheckYourAnswersJourneyState.Companion.checkAnswerTask
-import uk.gov.communities.prsdb.webapp.journeys.shared.states.CheckYourAnswersJourneyState.Companion.duplicableCheckAnswerStep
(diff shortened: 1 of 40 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/update/occupancy/UpdateOccupancyJourneyFactory.kt
```diff
@@ -101,14 +102,39 @@ class UpdateOccupancyJourneyFactory(
...
+                BedroomsStep.ROUTE_SEGMENT -> {
+                    checkAnswerStep(journey.bedrooms, BedroomsStep.ROUTE_SEGMENT)
+                }
+
+                RentIncludesBillsStep.ROUTE_SEGMENT -> {
+                    duplicableCheckAnswerTask(journey.rentIncludesBillsTask)
+                }
+
+                BillsIncludedStep.ROUTE_SEGMENT -> {
+                    embed(journey.rentIncludesBillsTask) {
```
Travis-Softwire: Appreciate that `embed` describes what this is doing in the framework, but it doesn't feel very descriptive from the DSL pov. I wonder whether it should be something like `fromTask` or `withinTask` and then rename the argument in the builder `journey` -> `task` (or `subJourney` if we think we'll ever have non-task subjourneys) so that it's then clear from the inner DSL that you're accessing the step from the task?

Travis-Softwire (approved):
One naming suggestion to make the embedding part of the DSL nicer to ready, otherwise all LGTM :)

PR #1634, opened 2026-07-22
PDJB-1180: Part 2: Save org landlords to database

samyou-softwire:
## Ticket number

PDJB-1180

## Goal of change

add methods for allowing org landlords to be saved to the database

add a new LandlordRgistrationService, parity with PropertyRegistrationService

the registrant details were missing for org landlords so adds these

this one takes an entire form submission for registration and saves the correct object to the database

renames ownership_links to organisation_users

makes the remaining individual landlord fields nullable so org landlords can have the booleans as null

thoughts on renameing ownership_links to organisation_users? I think from this perspective it makes the database easier to read and it allows us to use this table to store details about the users in the org. the service layer will pick up the slack for interpreting the database and ensuring you can still tell to which landlord a user ID resides

## Checklist

- TODO comments referencing this JIRA ticket have been searched for and removed - if a future PR will address them,
  mention that here **one TODO left for making sure the confirmation page works, but can't do this until [PDJB-1168](https://mhclgdigital.atlassian.net/browse/PDJB-1168) is in**


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlordUser.kt
```diff
@@ -11,24 +11,24 @@ import jakarta.persistence.UniqueConstraint
...
+class OrganisationUser() : AuditableEntity() {
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     val id: Long = 0
 
     @ManyToOne(optional = false)
-    @JoinColumn(name = "landlord_id", nullable = false)
-    lateinit var landlord: Landlord
+    @JoinColumn(name = "organisation_landlord_id", nullable = false)
+    lateinit var organisationLandlord: OrganisationLandlord
```
Travis-Softwire, line 24: Were we only using this for org landlords then, never individual landlords?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationUser.kt
```diff
@@ -11,24 +11,24 @@ import jakarta.persistence.UniqueConstraint
 
 @Entity
 @Table(
-    name = "landlord_access_link",
-    uniqueConstraints = [UniqueConstraint(columnNames = ["landlord_id", "subject_identifier"])],
+    name = "organisation_users",
```
Travis-Softwire: I'd include 'landlord' in the name as we have/will have other types of organisations

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/OrganisationUserRepository.kt
```diff
@@ -0,0 +1,15 @@
+package uk.gov.communities.prsdb.webapp.database.repository
+
+import org.springframework.data.jpa.repository.JpaRepository
+import uk.gov.communities.prsdb.webapp.database.entity.OrganisationUser
+
+// TODO: PDJB-1275: Move org landlord access checking to use this table
+@Suppress("ktlint:standard:function-naming")
+interface OrganisationUserRepository : JpaRepository<OrganisationUser, Long> {
```
Travis-Softwire: Same point here, would keep landlord in the name

src/main/resources/db/migrations/V1_43_0__add_organisation_registrant_columns.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE landlord
+    ADD COLUMN organisation_registrant_name VARCHAR(255),
+    ADD COLUMN organisation_registrant_date_of_birth DATE,
```
Travis-Softwire, line 3: Just checking we usually type DoBs as Dates? I haven't checked, if we do then all good

src/main/resources/db/migrations/V1_42_0__rename_landlord_access_link_to_organisation_users.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE landlord_access_link RENAME TO organisation_users;
```
Travis-Softwire: Minor, but we usually use the singluar for the table names, in this case `organisation_user` instead of `organisation_users` - although again I'd include `landlord` in that as we'll have other types of organisation

src/main/resources/db/migrations/V1_42_0__rename_landlord_access_link_to_organisation_users.sql
```diff
@@ -0,0 +1,11 @@
...
+
+ALTER TABLE organisation_users RENAME COLUMN landlord_id TO organisation_landlord_id;
+
+-- Rename indexes to match new table and column names
+ALTER INDEX idx_landlord_access_link_subject RENAME TO idx_organisation_users_subject;
+ALTER INDEX idx_landlord_access_link_landlord RENAME TO idx_organisation_users_organisation_landlord;
+
+-- Rename the unique constraint
+ALTER TABLE organisation_users RENAME CONSTRAINT landlord_access_link_landlord_id_subject_identifier_key
+    TO organisation_users_organisation_landlord_id_subject_identifier_key;
```
Travis-Softwire: Should be fine, but think there's a max size on constraint names after which they're truncated. As we don't reference it explicitly in the code though it shouldn't be an issue

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordRegistrationService.kt
```diff
@@ -0,0 +1,105 @@
...
+                    baseUserId = baseUserId,
+                    name = individualName!!,
+                    email = individualEmail!!,
+                    phoneNumber = individualPhoneNumber!!,
+                    addressDataModel = individualAddress!!,
+                    countryOfResidence = individualCountryOfResidence!!,
+                    isVerified = individualIsVerified!!,
+                    hasAcceptedPrivacyNotice = individualHasAcceptedPrivacyNotice!!,
+                    nonEnglandOrWalesAddress = individualNonEnglandOrWalesAddress,
+                    dateOfBirth = individualDateOfBirth,
```
Travis-Softwire: And if we are usuing not-null assertions, why can this be null?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordRegistrationService.kt
```diff
@@ -0,0 +1,105 @@
...
+        organisationRegistrantPhoneNumber: String? = null,
+        organisationGoverningBodyMembers: List<GoverningBodyMemberDataModel> = emptyList(),
+    ): Landlord =
+        when (landlordType) {
+            LandlordType.INDIVIDUAL -> {
+                landlordService.createIndividualLandlord(
+                    baseUserId = baseUserId,
+                    name = individualName!!,
+                    email = individualEmail!!,
+                    phoneNumber = individualPhoneNumber!!,
```
Travis-Softwire: Do we normally use not null assertions when creating entities (genuine question, I haven't looked)

Travis-Softwire (approved):
A few minor points, including some on the naming. The main point I want to understand is whether we were using the access link table for indLLs? If not then LGTM once the rest of the points have been addressed :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlordUser.kt
samyou-softwire: yeah I did end up changing my mind on this one. it felt more straightforward to have the service layer figure out the user ID -> landlord system and manage the two tables rather than having a unified table for it

it also lines this table up nicely to be the one we'll use for the 'other members of an org' feature set that's coming soon

Travis-Softwire: Ah I hadn't clocked that. But we still have all of the information we need to search on for landlords in the one single table right? Because that's going to be essential for searching to be performant - joins/unions/multiple searches won't cut it for the volumes we're going to have.

samyou-softwire: ah good point; I think this should be fine. organisation_user only really stores the access links between onelogin users and landlords. landlords themselves still have all their info in the landlords table so search should be fine

or another way, I don't think search will ever need to look at the organisation_user table as afaik it doesn't need to know which users associate with a landlord, though let me know if this assumption isn't the case

samyou-softwire:
rebase onto main

src/main/resources/db/migrations/V1_43_0__add_organisation_registrant_columns.sql
samyou-softwire: I did double check this yes, this is what for example the individual landlord date of birth uses
```
    date_of_birth                DATE,
```

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlordUser.kt
Travis-Softwire: Yes, that sounds right - I was conflating the landlord table where still have just the one with the users table! Phew

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationUser.kt
samyou-softwire: ah good spot, will fix this + make it singular as noted below

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/OrganisationUserRepository.kt
samyou-softwire: have renamed

src/main/resources/db/migrations/V1_42_0__rename_landlord_access_link_to_organisation_users.sql
samyou-softwire: have renamed

src/main/resources/db/migrations/V1_42_0__rename_landlord_access_link_to_organisation_users.sql
samyou-softwire: good spot, have made the name a little bit shorter now

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordRegistrationService.kt
samyou-softwire: agree this seems a bit odd - as we've more of a consensus on how [PDJB-1168](https://mhclgdigital.atlassian.net/browse/PDJB-1168) I'm going to take another pass at the services for registering here and the method signatures

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordRegistrationService.kt
samyou-softwire: we don't on IL registration as we just assume all properties aren't null (nothing is optional)

on property registration it looks like we don't, and just same them as nullable to the database, but this feels a bit unsafe to me. I'd want to be sure we're only saving valid entities to the database. what do you think?

samyou-softwire:
I wasn't very happy with how the services were laid out (registration service was doing much too little, landlord service was doing a lot of registration work) so I've restructured this a bit to confine the landlordservice (and other similar services) to just handle the call to save the data and any necessary sub-entities

there's a lot of changes in this last commit so I think worth a re-review

samyou-softwire:
also as [PDJB-1168](https://mhclgdigital.atlassian.net/browse/PDJB-1168) is now merging; I'll integrate this into that change once it's in

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/OrganisationGoverningBodyMemberService.kt
```diff
@@ -0,0 +1,32 @@
...
+import uk.gov.communities.prsdb.webapp.database.entity.OrganisationLandlord
+import uk.gov.communities.prsdb.webapp.database.repository.OrganisationGoverningBodyMemberRepository
+import uk.gov.communities.prsdb.webapp.models.dataModels.GoverningBodyMemberDataModel
+
+@PrsdbWebService
+class OrganisationGoverningBodyMemberService(
+    private val organisationGoverningBodyMemberRepository: OrganisationGoverningBodyMemberRepository,
+    private val addressService: AddressService,
+) {
+    fun createGoverningBodyMembers(
```
Travis-Softwire, line 17: Belt and braces, but should this be transactional?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/OrganisationLandlordUserService.kt
```diff
@@ -0,0 +1,17 @@
...
+import uk.gov.communities.prsdb.webapp.database.entity.OrganisationLandlord
+import uk.gov.communities.prsdb.webapp.database.entity.OrganisationLandlordUser
+import uk.gov.communities.prsdb.webapp.database.entity.PrsdbUser
+import uk.gov.communities.prsdb.webapp.database.repository.OrganisationLandlordUserRepository
+
+@PrsdbWebService
+class OrganisationLandlordUserService(
+    private val organisationLandlordUserRepository: OrganisationLandlordUserRepository,
+) {
+    fun createOrganisationLandlordUser(
```
Travis-Softwire, line 15: Same point here

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/OrganisationGoverningBodyMemberService.kt
```diff
@@ -0,0 +1,32 @@
...
+        organisationLandlord: OrganisationLandlord,
+        members: List<GoverningBodyMemberDataModel>,
+    ) {
+        members.forEach { member ->
```
Travis-Softwire, line 21: Do we have any tests covering this?

Travis-Softwire (approved):
A couple more minor points but happy w/ this once they're addressed

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/OrganisationGoverningBodyMemberService.kt
samyou-softwire: ah yeah, the address creating may fail

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/OrganisationLandlordUserService.kt
samyou-softwire: I've added one here too - I'm not sure if it'll do much (this would probably end up as a single create statement?) though should be useful in case this grows in the future

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/OrganisationGoverningBodyMemberService.kt
samyou-softwire: have added some

PR #1644, opened 2026-07-24
Chore/pdjb 896 migrate property registration containers

AEPR:
## Ticket number
PDJB-896 

## Goal of change
Migrates the tasks (not compliance) in the new restructured journey to duplicable tasks.

* Sections have been refactored to use the same structure as embedded journeys - this means they play nicely together

## Checklist

- [X] Migration ongoing

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/tasks/OwnershipAndLandlordsTask.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/tasks/OwnershipAndLandlordsTask.kt
@@ -1,13 +1,24 @@
 package uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.tasks
 
 import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.JourneyFrameworkComponent
-import uk.gov.communities.prsdb.webapp.journeys.Task
+import uk.gov.communities.prsdb.webapp.journeys.DuplicableTaskWithDependencies
+import uk.gov.communities.prsdb.webapp.journeys.JourneyStateService
 import uk.gov.communities.prsdb.webapp.journeys.isComplete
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/tasks/OwnershipAndLandlordsTask.kt
```diff
@@ -16,7 +27,7 @@ class OwnershipAndLandlordsTask : Task<OwnershipAndLandlordsState>() {
                 savable()
             }
             duplicableTask(journey.jointLandlordsTask) {
-                withDependencies { journey }
+                withDependencies { this@OwnershipAndLandlordsTask.dependencies }
```
Travis-Softwire, line 30: Now that we're doing this, should we stop `OwnershipAndLandlordsState` from inheriting from `InviteJointLandlordsTaskDependencies`? So that if someone did replace this with `journey` it would fail to build instead of compiling and using the defaults from the interface definition?

src/test/kotlin/uk/gov/communities/prsdb/webapp/journeys/builders/JourneyBuilderTest.kt
```diff
@@ -765,6 +765,22 @@ class JourneyBuilderTest {
...
+            // Act & Assert
+            assertThrows<JourneyInitialisationException> {
+                jb.section {
+                    withHeadingMessageKey("section.outer")
+                    section {
+                        withHeadingMessageKey("section.inner")
+                    }
+                }
+            }
+        }
```
Travis-Softwire, line 782: Worth also adding in tests to make sure:
- a step inside a section added from `fromTask` gets a section header
- that `configure { withAdditionalContentProperty } }` still works on steps inside a section?

Travis-Softwire (approved):
Two suggestions, otherwise LGTM!

PR #1650, opened 2026-07-27
PDJB-896: Migrate remaining journey controllers to JourneyStepDispatcher

AEPR:
## Ticket number

PDJB-896

## Goal of change

Migrate all remaining journey controllers to `JourneyStepDispatcher` so that dispatching a journey request (routing-map lookup, journey initialisation on missing state, and complex `stepPath` support) is handled by a single shared component instead of being repeated inline in every controller.

- Migrated 24 controllers (one commit each) from the hand-rolled `try { routingMap[stepName]?.… } catch (NoSuchJourneyException) { … }` pattern to `JourneyStepDispatcher.handleInitialisableRequest` (or `handleUninitialisableRequest`).
- Changed journey-step mappings from `"/{stepName}"` to `"/{*stepPath}"` so nested step paths are supported.
- Preserved every controller-specific concern: pre-dispatch authorisation checks, extra path variables (`propertyOwnershipId`, `invitationId`), confirmation/resume/invalid-link endpoints, multipart file-upload POST overloads, and any additional exceptions that should trigger a new-journey redirect (e.g. `PropertyOwnershipMismatchException` on `DeregisterPropertyController`, `LeavePropertyController`, `SwitchToIndividualController`; `InvalidInvitationException` on `RegisterLocalCouncilUserController`).
- Removed now-dead `initializeAndRedirect(...)` / `getJourneySteps(...)` helpers where no longer referenced.
- Updated 3 test files (`DeregisterLandlordControllerTests`, `DeregisterPropertyControllerTests`, `LeavePropertyControllerTests`) whose redirect-URL assertions previously used the relative `JourneyStateService.urlWithJourneyState(...)`; the dispatcher now sets `JOURNEY_BASE_PATH_ATTRIBUTE` on the request so generated URLs are absolute, matching the migrated exemplars.

- `CancelJointLandlordInvitationController` previously prepended `$invitationId` manually in its redirect helper. Post-migration this is redundant because the dispatcher stores the base path on the request and journey URLs are generated absolute; the manual prefix has been removed and the existing tests still pass.
- Two `Update*` controllers had no dedicated test class (`UpdateRentFrequencyAndAmountController`, and no direct test for `SwitchToIndividualController`). Compilation was verified via `./gradlew compileKotlin compileTestKotlin`; the full `testWithoutIntegration` suite passes.
- Three `PDJB-896` TODO comments (`OccupationTask.kt:26`, `InviteJointLandlordJourneyFactory.kt:74` and `:113`) are unrelated concerns (DuplicableTaskWithDependencies / addressable tasks) and are intentionally out of scope for this PR.

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/CancelJointLandlordInvitationController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/CancelJointLandlordInvitationController.kt
@@ -1,13 +1,11 @@
 package uk.gov.communities.prsdb.webapp.controllers
 
-import org.springframework.http.HttpStatus
 import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RequestParam
-import org.springframework.web.server.ResponseStatusException
 import org.springframework.web.servlet.ModelAndView
 import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.PrsdbController
 import uk.gov.communities.prsdb.webapp.constants.CANCEL_JOINT_LANDLORD_INVITATION_JOURNEY_URL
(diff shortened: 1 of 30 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/UpdateLandlordEmailController.kt
```diff
@@ -25,36 +23,30 @@ import java.security.Principal
 class UpdateLandlordEmailController(
     private val journeyFactory: UpdateEmailJourneyFactory,
 ) {
-    @GetMapping("{stepName}")
+    @GetMapping("/{*stepPath}")
```
Travis-Softwire, line 26: Copilot has pointed out that if this is called with just the journey's base path, if there's no matching endpoint it'll match here, and while it will still return a 404 as before, it'll initialise a journey before doing so. It's suggested checking if the step path is empty in the dispatcher after we've trimmed leading `/`s and explicitly throwing a 404 from there (if there's an explicit root endpoint it'll match in preference to the `/{*stepPath}` one, so that should be fine)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/CancelJointLandlordInvitationController.kt
```diff
@@ -1,13 +1,11 @@
 package uk.gov.communities.prsdb.webapp.controllers
 
-import org.springframework.http.HttpStatus
 import org.springframework.security.access.prepost.PreAuthorize
```
Travis-Softwire, line 3: Copilot thinks the tests for this file are still using the old `urlWithJourneyState` function to build the expected redirect, which you've fixed in the other relevant tests.

Travis-Softwire (approved):
One point that we should probably fix if it's correct (although can be separate PR) and one minor comment, otherwise all LGTM

PR #1654, opened 2026-07-28
PDJB-1275: Part 2: Fix other assumptions a landlord has a base user ID

samyou-softwire:
## Ticket number

[PDJB-1275](https://mhclgdigital.atlassian.net/browse/PDJB-1275)

## Goal of change

remove all general assumptions from the codebase that a landlord has a single base user ID

fixes many controllers & services to use the general methods for getting the current landlord, and uses the landlord to compare base user IDs

this does not include any changes in function around org landlords. all of these changes should be stuff that would be expected to 'just work' with org landlords. any cases where we may want function to change (emails, page copy etc) are covered by specific tickets, referenced in TODO

most of the changes in this PR are uninteresting, removing the principal param from controllers etc. of particular note are the services which will now base their logic off of landlords directly

## Checklist

- Branch has been rebased onto main and run locally, with everything working as expected (both for your new feature
  and any related functionality) **not currently based on main, but does work locally right now**
  mention that here
- QA instructions have been added to the ticket (particularly if this is the last PR required to complete the ticket) **I don't think there's much to QA here. I presume we'll do a healthy amount of mobbing before release**


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/OrganisationLandlordUserRepository.kt
```diff
@@ -3,7 +3,6 @@ package uk.gov.communities.prsdb.webapp.database.repository
 import org.springframework.data.jpa.repository.JpaRepository
 import uk.gov.communities.prsdb.webapp.database.entity.OrganisationLandlordUser
 
-// TODO: PDJB-1275: Move org landlord access checking to use this table
```
samyou-softwire, line 6: we now do

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/AddressAvailabilityService.kt
```diff
@@ -8,9 +8,4 @@ class AddressAvailabilityService(
     private val propertyOwnershipRepository: PropertyOwnershipRepository,
 ) {
     fun isAddressOwned(uprn: Long): Boolean = propertyOwnershipRepository.existsByIsActiveTrueAndAddress_Uprn(uprn)
-
-    fun isAddressOwnedByUser(
```
samyou-softwire, line 12: this was unused

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterLandlordController.kt
```diff
@@ -20,59 +20,46 @@ import uk.gov.communities.prsdb.webapp.journeys.StepLifecycleOrchestrator
...
-    fun getConfirmation(principal: Principal): String {
+    fun getConfirmation(): String {
         if (!landlordDeregistrationService.hasLandlordDeregisteredInThisSession()) {
             throw ResponseStatusException(
                 HttpStatus.NOT_FOUND,
                 "Landlord deregistration has not been performed in this session",
             )
         }
 
-        if (landlordService.retrieveLandlordByBaseUserId(principal.name) != null) {
```
Travis-Softwire, line 69: It looks like we've just removed this check altogether? Should we not instead check if the current user is associated with a landlord? This still feels like a failure mode worth checking

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordController.kt
```diff
@@ -17,22 +17,22 @@ import uk.gov.communities.prsdb.webapp.constants.REGISTERED_PROPERTIES_FRAGMENT
...
 import uk.gov.communities.prsdb.webapp.services.PropertyOwnershipService
+import uk.gov.communities.prsdb.webapp.services.UserToLandlordService
 import java.security.Principal
 
 @PreAuthorize("hasAnyRole('LANDLORD')")
 @PrsdbController
 @RequestMapping(LANDLORD_BASE_URL, "/")
 class LandlordController(
-    private val landlordService: LandlordService,
+    private val userToLandlordService: UserToLandlordService,
```
Travis-Softwire, line 34: Not for this PR, but the heavy use of this has got me thinking about whether we should cache the landlord ID in session on login, which would not only save at least one DB roundtrip per call for most LL endpoints, but would also potentially make our system operator functionality MUCH simpler to implement when we get there.

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/UpdateBedroomsController.kt
```diff
@@ -63,14 +63,11 @@ class UpdateBedroomsController(
...
-        baseUserId: String,
-        propertyOwnershipId: Long,
-    ) {
-        if (!propertyOwnershipService.getIsAuthorizedToEditRecord(propertyOwnershipId, baseUserId)) {
+    private fun throwErrorIfUserIsNotAuthorized(propertyOwnershipId: Long) {
+        if (!propertyOwnershipService.getCurrentUserIsAuthorizedToEditRecord(propertyOwnershipId)) {
             throw ResponseStatusException(
                 HttpStatus.NOT_FOUND,
-                "User $baseUserId is not authorized to update property ownership $propertyOwnershipId",
+                "Current user is not authorized to update property ownership $propertyOwnershipId",
```
Travis-Softwire: This is a loss of fidelity in the logs that I don't think is necessary - we can still grab the base user Id to tell us who was trying to access something they shouldn't have been

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordService.kt
```diff
@@ -31,23 +32,15 @@ import kotlin.String
...
     private val organisationLandlordRepository: OrganisationLandlordRepository,
+    private val landlordRepository: LandlordRepository,
+    private val userToLandlordService: UserToLandlordService,
     private val addressService: AddressService,
     private val registrationNumberService: RegistrationNumberService,
     private val backLinkService: BackUrlStorageService,
     private val updateConfirmationSender: EmailNotificationService<LandlordUpdateConfirmation>,
     private val absoluteUrlProvider: AbsoluteUrlProvider,
 ) {
-    fun retrieveLandlordByRegNum(regNum: RegistrationNumberDataModel): Landlord? {
```
Travis-Softwire, line 40: Were we not using this? I thought we let LCs search by LRN?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -91,16 +92,13 @@ class PropertyOwnershipService(
...
     }
 
-    fun getPropertyOwnershipIfAuthorizedUser(
-        propertyOwnershipId: Long,
-        baseUserId: String,
-    ): PropertyOwnership {
+    fun getPropertyOwnershipIfCurrentUserAuthorized(propertyOwnershipId: Long): PropertyOwnership {
         val propertyOwnership = getPropertyOwnership(propertyOwnershipId)
+        val baseUserId = SecurityContextHolder.getContext().authentication.name
+        val landlord = userToLandlordService.getLandlordForBaseUserIdOrNull(baseUserId)
```
Travis-Softwire: Why this vs the `getCurrentLandlordForUser` you've used everywhere else?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -119,23 +117,19 @@ class PropertyOwnershipService(
...
+    fun getCurrentUserIsAuthorizedToEditRecord(propertyOwnershipId: Long): Boolean = isCurrentUserLandlord(propertyOwnershipId)
 
-    fun getIsLandlord(
-        propertyOwnershipId: Long,
-        baseUserId: String,
-        // TODO: PDJB-1275: Update authorisation checks to account for org landlords
-    ): Boolean =
-        getPropertyOwnership(propertyOwnershipId).landlords.any { (it as IndividualLandlord).baseUser.id == baseUserId }
+    fun isCurrentUserLandlord(propertyOwnershipId: Long): Boolean {
+        val baseUserId = SecurityContextHolder.getContext().authentication.name
```
Travis-Softwire: Same point here - feels odd to do the baseUserId retrieval sometimes in the userToLandlordService and sometimes in the calling code?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordDeregistration/stepConfig/DeregisterStepConfig.kt
```diff
@@ -39,6 +39,10 @@ class DeregisterStepConfig(
         landlordDeregistrationService.deregisterLandlord(baseUserId)
         landlordDeregistrationService.addLandlordHadActivePropertiesToSession(landlordHadActiveSoloProperties)
 
+        // TODO: PDJB-1274: Update emails to account for org landlord
+        check(landlord is IndividualLandlord)
```
Travis-Softwire: Shouldn't this run before we actually deregister them?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordService.kt
```diff
@@ -31,23 +32,15 @@ import kotlin.String
...
-            throw IllegalArgumentException("Invalid registration number type")
-        }
-        return individualLandlordRepository.findByRegistrationNumber_Number(regNum.number)
-    }
-
-    // TODO: PDJB-1275: Update checks for landlord users to account for multiple users representing a landlord
-    fun retrieveLandlordByBaseUserId(baseUserId: String): IndividualLandlord? = individualLandlordRepository.findByBaseUser_Id(baseUserId)
-
-    fun retrieveLandlordById(id: Long): Landlord? = individualLandlordRepository.findById(id).orElse(null)
+    fun retrieveLandlordById(id: Long): Landlord? = landlordRepository.findById(id).orElse(null)
```
Travis-Softwire, line 43: I think this means now if a LC user views the LL record of an org LL they'll get a 500 instead of a 404 until we update the LL record to handle Org LLs. Was that intentional?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyComplianceRepository.kt
```diff
@@ -1,26 +1,13 @@
...
-        "SELECT pc FROM PropertyCompliance pc " +
-            "JOIN pc.propertyOwnership po " +
-            "JOIN po.ownershipLinks ol " +
-            "JOIN TREAT(ol.landlord AS IndividualLandlord) l " +
-            "WHERE l.baseUser.id = :landlordBaseUserId",
-    )
-    fun findAllByPropertyOwnership_OwnershipLinks_Landlord_BaseUser_Id(
-        @Param("landlordBaseUserId") landlordBaseUserId: String,
-    ): List<PropertyCompliance>
+    fun findAllByPropertyOwnership_OwnershipLinks_Landlord_IdAndPropertyOwnership_IsActiveTrue(landlordId: Long): List<PropertyCompliance>
```
Travis-Softwire, line 10: Was adding the IsActiveTrue condition here intentional?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLandlordController.kt
```diff
@@ -64,13 +62,8 @@ class RegisterLandlordController(
...
     @GetMapping("/$CONFIRMATION_PATH_SEGMENT")
-    fun getConfirmation(
-        model: Model,
-        principal: Principal,
-    ): String {
-        val landlord =
-            landlordService.retrieveLandlordByBaseUserId(principal.name)
-                ?: throw ResponseStatusException(HttpStatus.BAD_REQUEST, "User ${principal.name} is not registered as a landlord")
+    fun getConfirmation(model: Model): String {
+        val landlord = userToLandlordService.getCurrentLandlordForUser()
```
Travis-Softwire, line 66: Again we've lost something userful in the logs here as we're no longer getting a 400. Although we could just update getCurrentLandlordForUser to throw a 400 instead of a 500?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
```diff
@@ -91,16 +92,13 @@ class PropertyOwnershipService(
...
+        val landlord = userToLandlordService.getLandlordForBaseUserIdOrNull(baseUserId)
 
         val isLocalCouncil = localCouncilDataService.getIsLocalCouncilUser(baseUserId)
-
-        // TODO: PDJB-1275: Update authorisation checks to account for org landlords
-        val isLandlord = propertyOwnership.landlords.any { (it as IndividualLandlord).baseUser.id == baseUserId }
+        val isLandlord = landlord != null && propertyOwnership.landlords.any { it.id == landlord.id }
 
         if (!isLocalCouncil && !isLandlord) {
```
Travis-Softwire: Minor, but we can cut down on DB call, esp for council pages, but checking these sequentially through nesting instead of evaluating both and then ANDing them

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordController.kt
```diff
@@ -89,8 +85,9 @@ class LandlordController(
         @RequestParam(value = "page", required = false) @Min(1) page: Int = 1,
```
Travis-Softwire, line 83: Are we still using the principal param?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/InviteJointLandlordController.kt
```diff
@@ -80,11 +80,9 @@ class InviteJointLandlordController(
         @PathVariable invitationId: Long,
```
Travis-Softwire, line 77: Still using the principal param?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/InviteJointLandlordController.kt
```diff
@@ -96,22 +94,19 @@ class InviteJointLandlordController(
         principal: Principal,
```
Travis-Softwire: Still using the principal param?

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyUpdateEmailServiceTests.kt
```diff
@@ -50,42 +45,39 @@ class PropertyUpdateEmailServiceTests {
...
     fun `sendUpdateEmails throws PrsdbWebException when the acting landlord is not found`() {
         val baseUserId = "unknown-user"
         val actor = MockLandlordData.createIndividualLandlord(baseUser = MockLandlordData.createPrsdbUser(baseUserId))
-        val propertyOwnership = MockLandlordData.createPropertyOwnership(id = propertyId, landlords = mutableSetOf(actor))
+        val propertyOwnership =
+            MockLandlordData.createPropertyOwnership(id = propertyId, landlords = mutableSetOf(actor))
         whenever(mockPropertyOwnershipService.getPropertyOwnership(propertyId)).thenReturn(propertyOwnership)
-        setMockPrincipal(baseUserId)
-        whenever(mockLandlordService.retrieveLandlordByBaseUserId(baseUserId)).thenReturn(null)
+        whenever(mockUserToLandlordService.getCurrentLandlordForUser()).thenThrow(PrsdbWebException("Landlord not found"))
```
Travis-Softwire: Is this test actually testing anything now or just asserting on the mock directly?

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyRegistrationServiceTests.kt
```diff
@@ -403,12 +365,11 @@ class PropertyRegistrationServiceTests {
             OwnershipType.FREEHOLD,
             true,
             2,
-            3,
-            "USER_ID",
             1,
```
Travis-Softwire: Careful here that we're using the correct arg in the correct place - might be worth using named arguments just in case these args change again in the future

Travis-Softwire (approved):
What a monster! I've left some comments but nothing blocking, happy for it to be merged once they've been addressed

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/DeregisterLandlordController.kt
samyou-softwire: ahh good point; I removed a lot of these checks as requesting a landlord will throw the same error, but missed this case where we only care to check if the landlord exists. I'll add a check back for this

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordController.kt
samyou-softwire: good idea; I've raised [PDJB-1477](https://mhclgdigital.atlassian.net/browse/PDJB-1477) and added a TODO here

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordService.kt
samyou-softwire: we weren't no, only usages of this method was in tests. searching uses different code paths at `searchForLandlords` in the same service which eventually ends up as a raw SQL query

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
samyou-softwire: not sure; might've been a little confusion as there isn't currently a method to get current landlord or null. I'll add one

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
samyou-softwire: this one is definitely a mistake, good spot thanks

for what it's worth we will eventually be removing these references since eventually we'll have to remove getLandlordForBaseUserId. though 100% we should be removing all the trivial cases now

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordDeregistration/stepConfig/DeregisterStepConfig.kt
samyou-softwire: sure; have added a TODO for the relevant ticket to this check() [PDJB-1435](https://mhclgdigital.atlassian.net/browse/PDJB-1435)

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordService.kt
samyou-softwire: after the communications this morning I'm going to do some further thinking on how org LLs should work for LC in the MVP, so will leave this as-is for now

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/repository/PropertyComplianceRepository.kt
samyou-softwire: yes - I don't think we use isActive at all yet but seemed sensible to start including it so it's consistent across all queries

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyOwnershipService.kt
samyou-softwire: sure; have reworked to return early

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/RegisterLandlordController.kt
samyou-softwire: I've done so, though left the case where the is >1 landlord per user as a 500 (something has seriously gone wrong if so)

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyUpdateEmailServiceTests.kt
samyou-softwire: good point, seems not. I've removed

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordController.kt
samyou-softwire: gone

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/InviteJointLandlordController.kt
samyou-softwire: gone

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/InviteJointLandlordController.kt
samyou-softwire: gone

src/test/kotlin/uk/gov/communities/prsdb/webapp/services/PropertyRegistrationServiceTests.kt
samyou-softwire: havw swapped to named args

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/UpdateBedroomsController.kt
samyou-softwire: agree, I've commonised this method as it's duplicated across all the update code paths and it now has a call to get the base user ID

PR #1659, opened 2026-07-29
PDJB-896: Migrate compliance tasks to Duplicable

AEPR:
## Ticket number

PDJB-896

## Goal of change

Migrate the three property-registration compliance sections (gas safety, electrical safety, EPC) — both container tasks and their inner details tasks — onto the `DuplicableTaskWithDependencies` composition model already used elsewhere in this ticket.

Six commits, one per task, applied in the same pattern for each section:

- **Container migration** (`GasSafetyTask`, `ElectricalSafetyTask`, `EpcTask`) — converts each container `Task` to a `DuplicableTaskWithDependencies<XxxState, XxxDependencies>`. Steps and delegates that used to live on `PropertyRegistrationJourney` (and on the update-journey concrete class) are pulled into the task's constructor and delegate storage. Journey factories now wire the task with `duplicableTask { withDependencies { journey } }`, and CYA sub-journeys use `fromTask(...) { checkAnswerTask(task.xxxDetailsTask) }`.
- **Details migration** (`GasSafetyDetailsTask`, `ElectricalSafetyDetailsTask`, `EpcDetailsTask`) — splits each state interface into a small container state (`XxxState`) and a wide detail state (`XxxDetailState`) and moves all step overrides, upload maps / EPC data-model delegates, and `isOccupied` / `allowProvideCertificateLaterRoute` overrides onto the details task. Step configs are re-typed against the new detail state, and the container's inner call switches to `duplicableTask(...) { withDependencies { dependencies } }`. Update journeys drop their per-step constructor params and instead extend `XxxDependencies`.
- Shared helpers (`ComplianceDetailsHelper.getXxxCyaContent`) now take an explicit `(cyaState, xxxState)` pair rather than a self-typed generic, mirroring the split-state design.
- `CombinedComplianceCheckState` no longer inherits the compliance states directly; it exposes `gasSafetyTask`, `electricalSafetyTask`, and `epcTask`, and `HasMissingComplianceStepConfig` reads through those.

Behaviour is unchanged; all existing tests were updated in place (routing stubs through the new `.xxxDetailsTask.*` accessors) and pass.

- The migration is best reviewed one commit at a time — each commit is a mechanical mirror of the previous section''s pattern.
- The `EmbedBuilder.checkAnswerTask` extension in `CheckYourAnswersJourneyState` is added and removed several times through the commit series as each pair of tasks changes; the final state (HEAD) is that it is removed, because once every compliance details task is duplicable the extension has no callers.

## Checklist

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/PropertyRegistrationJourneyFactory.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/PropertyRegistrationJourneyFactory.kt
@@ -215,7 +215,7 @@ class PropertyRegistrationJourneyFactory(
                 CheckElectricalCertUploadsStep.ROUTE_SEGMENT,
                 -> {
                     fromTask(journey.electricalSafetyTask, journey) {
-                        checkAnswerTask(task.electricalSafetyDetailsTask)
+                        duplicableCheckAnswerTask(task.electricalSafetyDetailsTask)
                     }
                 }
 
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/PropertyRegistrationJourneyFactory.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/PropertyRegistrationJourneyFactory.kt
@@ -26,24 +26,13 @@ import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.states.Occu
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.BedroomsStep
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.BillsIncludedStep
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.CheckElectricalCertUploadsStep
-import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.CheckEpcAnswersStep
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.CheckGasCertUploadsStep
-import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.ConfirmEpcDetailsRetrievedByCertificateNumberStep
-import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.ConfirmEpcRetrievedByUprnStep
 import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.ConfirmMissingComplianceCheckResult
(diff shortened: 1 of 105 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/PropertyRegistrationJourneyFactory.kt
```diff
@@ -219,14 +190,18 @@ class PropertyRegistrationJourneyFactory(
                 GasCertIssueDateStep.ROUTE_SEGMENT,
                 CheckGasCertUploadsStep.ROUTE_SEGMENT,
                 -> {
-                    checkAnswerTask(journey.gasSafetyDetailsTask)
+                    fromTask(journey.gasSafetyTask, journey) {
```
Travis-Softwire: Copilot thinks these 3 are a no-op because we don't have an `EmbedBuilder.duplicableCheckAnswerTaskEmbedBuilder.duplicableCheckAnswerTask` overload, so it's not really embedding the task (it does work, but the nested task is just bound to the CYA journey directly it seems).

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/tasks/ElectricalSafetyDetailsTask.kt
```diff
@@ -19,9 +22,32 @@ import uk.gov.communities.prsdb.webapp.journeys.propertyRegistration.steps.Remov
...
+    override val taskState: ElectricalSafetyDetailState
+        get() = this
+
+    override val isOccupied: Boolean
+        get() = dependencies.isOccupied
+    override val allowProvideCertificateLaterRoute: Boolean
+        get() = dependencies.allowProvideCertificateLaterRoute
+
+    override var electricalUploadMap: Map<Int, CertificateUpload> by delegateProvider.requiredDelegate("electricalUploadMap", mapOf())
+    override var highestAssignedElectricalMemberId: Int? by delegateProvider.nullableDelegate("highestAssignedElectricalMemberId")
```
Travis-Softwire, line 48: I think this was previously keyed as "highestElectricalUploadMemberId"? Will changing delegate keys not break any inflight journeys in production?

Travis-Softwire (approved):
Mechanically this all looked right to me. Copilot has flagged two points that are worth considering though. See if you agree - happy to go with your decision on them either way though.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/propertyRegistration/tasks/ElectricalSafetyDetailsTask.kt
AEPR: Nope! It was "highestAssignedElectricalMemberId" before as well. Line 705 in the old JourneyFactory file

PR #1661, opened 2026-07-29
PDJB-NONE: Stop deployed seed data scanning the address table

Travis-Softwire:
## Ticket number

PDJB-NONE (follow-up to PDJB-1048)

## Goal of change

The `available_addresses(n)` function added to the deployed seeds in #1568 scans the 35m-row `address` table on every boot, which takes too long.

- Removes `available_addresses(n)` from `data-test.sql` and `data-integration.sql`. The QA and metrics cohorts now use dedicated seed addresses at reserved ids (`9000000001` onwards) instead of claiming whichever AddressBase/NGD addresses were not yet used by an active property. This is the approach `data-local.sql` and `data-metrics-local.sql` already take.
- Pins the seeded properties to `local_council_id = 2`. Previously each property inherited the council of whichever address it happened to claim, so the cohorts landed in a different council in every environment and were not reliably visible to the seeded LC users.

Why the old approach was slow:

- `LIMIT $1` is a parameter, so it is never folded to a constant. The planner falls back to assuming 10% of the table will be returned — 3.5m rows at production size — and costs every join choice on that basis.
- The function has no volatility marker, so it is `VOLATILE` and can never be inlined. Inlining is what would have substituted the literal `9`/`101`/`100`.
- The cost is unbounded regardless of plan: it is an anti-join walking `address` in id order until it finds N unused active rows.

Measured end-to-end seed time (full `data-integration.sql`, Postgres restarted between runs to clear `shared_buffers`):

| `address` state | before | after |
|---|---|---|
| 2m rows, all active | 397 ms | 469 ms |
| 20m rows, all active | 542 ms | 420 ms |
| 20m rows, 5m inactive at low ids | 6,546 ms | 372 ms |
| 20m rows, 10m inactive at low ids | 7,477 ms | 404 ms |

A large inactive population is a normal state rather than an edge case: the data package is GB-wide (`add_gb_builtaddress`) and `NgdAddressLoader` marks anything outside England/Wales/Unassigned as `is_active = false`, with those rows only removed at the end of a full load. In that state a single `available_addresses(101)` call discards 10m rows and touches 20.1m buffers to return 101, and `data-integration.sql` called it three times per boot.

Note the before/after at 2m rows: in the best case the old version was fine and the new one is marginally slower, since it inserts 210 rows. The point of the change is that the new cost does not vary with the size or state of the address table.

**Interaction with the NGD loader.** The seeded addresses are inert to it, because it is delta-based and keyed on `uprn`:

- `uprn IS NULL`, so `ON CONFLICT (uprn) DO UPDATE` never matches them, and `updatePropertyOwnershipAddresses` (`WHERE a.uprn IN (...)`) never overwrites their properties
- `is_active = true`, so `deleteUnusedInactiveAddresses` (`WHERE a.is_active = false`) never considers them
- a NULL `uprn` also keeps them out of the address lookup, which requires `uprn IS NOT NULL`

I checked this by replaying the loader's own statements against the seeded data: all 210 addresses survived with `last_modified_date` still NULL, while 1,685 surrounding NGD rows were deleted.

**The seed must not `setval` the address sequence.** Deliberately omitted, otherwise the loader would start allocating ids above the reserved block. `data-local.sql` does call it, which is fine locally but would not be here.

**These two files have no automated test coverage.** They are selected by `spring.sql.init.platform: ${ENVIRONMENT_NAME}`, which is only set in deployed environments, so nothing in `src/test` loads them. I verified manually against a real Postgres: 47 Flyway migrations applied, each seed run twice to confirm idempotency under `mode: always`, then cohort addresses and trigger population asserted. Worth considering a smoke test that runs these seeds against a migrated testcontainer — it would have caught both #1657 and a `null::bigint` type bug I hit while writing this.

## Checklist

- [x] Branch is based on current `main` (`8a73da7d5`). The seeds were exercised directly against a migrated Postgres rather than by booting the app, since these files only load in deployed environments.
- [ ] Test suite has been run in full locally and is passing — ran `ktlintCheck` (passing) and `MetricsSinglePageTests` (10/10), the only test that loads any file changed here. The full suite was not run as no other test touches these seeds.

### DIFF
--- a/src/main/resources/data-integration.sql
+++ b/src/main/resources/data-integration.sql
@@ -1,16 +1,35 @@
 -- =============================================================================
--- available_addresses(n): returns the first n existing active addresses not already
--- used by an active property, ranked 1..n (rn). Like the rest of this seed, the
--- cohorts below reference the AddressBase/NGD addresses already present rather than
--- inserting their own; each claims distinct free addresses by joining this function
--- on rn. The inner LIMIT n keeps callers from ranking the whole address table.
--- Defined with a single-quoted SQL body (no dollar-quoting) because the spring.sql.init
--- runner splits scripts on ';' and cannot parse dollar-quoted function blocks.
+-- Addresses for the QA and metrics cohorts below, at reserved ids far above the range
+-- the AddressBase/NGD loader allocates from. The cohorts used to claim whichever
+-- existing addresses were not yet used by an active property, which meant an unbounded
+-- scan of the 35m-row address table on every boot and left each property in whichever
+-- council happened to own the address it claimed.
+--
+--   9000000001-9000000009  provide-later property record QA cohort (properties 49-57)
+--   9000001001-9000001101  metrics test cohort 1 (properties 1201-1301)
+--   9000002001-9000002100  metrics test cohort 2 (properties 1601-1700)
+--
+-- These rows are inert to the NGD loader, which is delta-based and keyed on uprn:
+--   * uprn IS NULL, so its ON CONFLICT (uprn) DO UPDATE never matches them, and its
+--     property_ownership refresh (WHERE a.uprn IN (...)) never overwrites them
+--   * is_active, so its "delete unused inactive addresses" pass never considers them
+--   * the address id sequence is deliberately NOT bumped past these ids, so the loader
+--     carries on allocating from where it left off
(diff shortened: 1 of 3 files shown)


TomHanmer29 (approved):
Changes LGTM. I wonder if it's worth documenting these seed data files somewhere additionally - they're becoming quite big and complicated, especially with now having reserved IDs?

PR #1663, opened 2026-07-30
PDJB-1180: Add an orgL to int seed data

samyou-softwire:
## Ticket number

[PDJB-1180](https://mhclgdigital.atlassian.net/browse/PDJB-1180)

## Goal of change

add an organisation landlord to integration seed data

landlord ID 11 does not currently exist so can be added fine

do we mind that the registration number is ID 900? it's a magic number & is not the largest on int right now

### DIFF
--- a/src/main/resources/data-integration.sql
+++ b/src/main/resources/data-integration.sql
@@ -207,6 +207,9 @@ VALUES (1, 1, 1, '2024-10-15 00:00:00+00', 'Team-PRSDB+landlord@softwire.com', n
         '07777777777', 'urn:fdc:gov.uk:2022:qw2_iN4-Be1BkbYb8y-KyMuPfG7F49W_1fsa_V6iX9w', '1990-01-01', 'England or Wales', true,
         true) ON CONFLICT DO NOTHING;
 
+SELECT setval(pg_get_serial_sequence('landlord', 'id'), (SELECT MAX(id) FROM landlord));
+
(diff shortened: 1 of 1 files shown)


src/main/resources/data-integration.sql
```diff
@@ -207,6 +207,41 @@ VALUES (1, 1, 1, '2024-10-15 00:00:00+00', 'Team-PRSDB+landlord@softwire.com', n
         '07777777777', 'urn:fdc:gov.uk:2022:qw2_iN4-Be1BkbYb8y-KyMuPfG7F49W_1fsa_V6iX9w', '1990-01-01', 'England or Wales', true,
         true) ON CONFLICT DO NOTHING;
 
+INSERT INTO prsdb_user (id, created_date)
```
Travis-Softwire, line 213: Why all in seperate statements rather than added to the existing ones with some end of line comments?

Travis-Softwire (approved):
Suggest also adding this to the local seed data (so that people can test their org landlord changes locally as well) and also checking that it does in fact work :). Otherwise just one minor comment 👍

src/main/resources/data-integration.sql
samyou-softwire: to me just felt a bit more digestible to have all the org landlord setup separate from the individual landlord stuff. I've added a comment to make this clearer

PR #1664, opened 2026-07-30
PDJB-1260: Send registration confirmation email for organisation landlords

Bill-Haigh:
## Ticket number

PDJB-1260

## Goal of change

Organisation landlords received no confirmation email on completing registration.

- Sends a confirmation email when an organisation landlord completes registration, with organisation-specific copy ("You have registered [organisation name] as a landlord")
- Adds the `ORGANISATIONAL_LANDLORD_REGISTRATION_CONFIRMATION_EMAIL` template, including its Notify template IDs
- Narrows the existing individual-landlord email helper to take an `IndividualLandlord`, removing its runtime type check
- Extends the organisation journey integration test to submit check your answers and assert the email is sent

- **The email goes to the registrant's email address.** This branch is rebased onto PDJB-1282, which added the registrant email step and made `OrgLandlordRegistrationCyaStepConfig` populate `organisationRegistrantEmail` for real. The second commit switches the recipient from the organisation address to the registrant's and removes the `// TODO: PDJB-1282` that was tracking it. The integration test submits a registrant email that differs from the organisation email, so the assertion proves which one is used.
- **Submitting check your answers still renders an error page for org landlords.** `RegisterLandlordController.getConfirmation` resolves the landlord via `retrieveLandlordByBaseUserId`, which only returns `IndividualLandlord`s. Pre-existing and out of scope here, so the integration test asserts the email and carries a `// TODO: PDJB-1392` where the confirmation page assertion belongs.
- **Naming.** Following review, the identifiers this PR introduces use `Organisational...`. The pre-existing `OrganisationLandlord` entity and the `ORGANISATION_LANDLORD_REGISTRATION` feature flag are deliberately untouched - a separate PR will rename those.
- The template was verified against the live Notify API via `NotifyEmailTemplateTests`, so the source-controlled markdown matches Notify exactly.

## Pre-release steps

`NotifyEmailTemplateTests` only checks the template ID for the Notify service its API key belongs to (non-prod, `test_id`), and is skipped when `EMAILNOTIFICATIONS_APIKEY` is unset - so CI does not cover any of the below.

- **test & nft** - check the Notify template ID for `ORGANISATIONAL_LANDLORD_REGISTRATION_CONFIRMATION_EMAIL` is correct
- **prod** - check the Notify template ID is correct
- **prod** - check the email template matches pre-prod

## Checklist

- [x] TODO comments referencing this JIRA ticket have been searched for and removed - two TODOs referencing future
  tickets remain: `PDJB-1274` (reassess the recipient address and name once there is a general way to email a landlord)
  and `PDJB-1392` (assert the confirmation page in the integration test)
- [x] Any special release instructions (e.g. the database will need resetting, emails to be added/removed) have been to the relevant release
  ticket on the Jira board, referencing your ticket number


src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/emailModels/OrganisationLandlordRegistrationConfirmationEmail.kt
```diff
@@ -0,0 +1,18 @@
...
+    val registrantName: String,
+    val organisationName: String,
+    val lrn: String,
+    val prsdURL: String,
+) : EmailTemplateModel {
+    override val template = EmailTemplate.ORGANISATION_LANDLORD_REGISTRATION_CONFIRMATION_EMAIL
+
+    override fun toHashMap() =
+        hashMapOf(
+            "registrant name" to registrantName,
```
danielledias-MT: **Minor:**

Looks like most of the other email models define private key constants (e.g.  private val `registrantNameKey = "registrant name"`) and reference them in the `hashMapOf`. Should be update this to be consistent? Appreciate it's not consistent in ALL places

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/emailModels/OrganisationLandlordRegistrationConfirmationEmail.kt
```diff
@@ -0,0 +1,18 @@
...
+
+data class OrganisationLandlordRegistrationConfirmationEmail(
+    val registrantName: String,
+    val organisationName: String,
+    val lrn: String,
+    val prsdURL: String,
+) : EmailTemplateModel {
+    override val template = EmailTemplate.ORGANISATION_LANDLORD_REGISTRATION_CONFIRMATION_EMAIL
+
+    override fun toHashMap() =
```
danielledias-MT: **Minor:**

Most of the other email models explicitly declare the return type as  `HashMap<String, String>` should we add that here for consistency?

danielledias-MT (approved):
Approved with minor suggestions

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordRegistrationService.kt
```diff
@@ -137,4 +136,16 @@ class LandlordRegistrationService(
             ),
         )
     }
+
+    private fun sendOrganisationRegistrationConfirmationEmail(landlord: OrganisationLandlord) {
```
samyou-softwire: not something for now, but when we start looking into [PDJB-1274](https://mhclgdigital.atlassian.net/browse/PDJB-1274) and [PDJB-1425](https://mhclgdigital.atlassian.net/browse/PDJB-1425) we're probably going to want to looking into some more general ways of sending an email 'to' a landlord (to what email? to which name?)

I think this is fine for now, but could there be a TODO-PDJB-1274 to come back and reassess this later?

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordRegistrationService.kt
```diff
@@ -137,4 +136,16 @@ class LandlordRegistrationService(
...
+        orgRegistrationConfirmationSender.sendEmail(
+            landlord.registrantEmail!!,
+            OrganisationLandlordRegistrationConfirmationEmail(
```
samyou-softwire: may need to be changed in a few places in this diff; it's OrgLandlord or OrganisationalLandlord. there is no OrganisationLandlord 😅

src/main/resources/emails/emailTemplates.json
samyou-softwire, line 1: if we're adding new emails, we should have a pre release step to 
- on test/nft & prod to check the notify email template IDs are correct 
- on prod the email template matches pre-prod

since our automated testing won't cover this

src/main/resources/emails/OrganisationalLandlordRegistrationConfirmation.md
```diff
@@ -0,0 +1,16 @@
...
+
+## What you need to do next
+
+You must register your rental properties.
+
+You need to include your Landlord Registration Number and Property Registration Number on any property advertisements.
+
+[Sign in to your account](((PRSD URL)))
+
+---
```
samyou-softwire, line 15: nit: we normally have a line break after the ---

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordRegistrationJourneyTests.kt
```diff
@@ -327,7 +335,26 @@ class LandlordRegistrationJourneyTests : IntegrationTestWithMutableData("data-mo
...
+            "registrant@example.com",
+            OrganisationLandlordRegistrationConfirmationEmail(
+                registrantName = "name",
+                organisationName = "Test Organisation Name",
+                lrn = createdOrgLandlordRegNum.toString(),
+                prsdURL = absoluteLandlordUrl,
+            ),
+        )
+
+        // TODO: PDJB-1180: assert the confirmation page renders here. It currently errors for org landlords because
```
samyou-softwire: [PDJB-1392](https://mhclgdigital.atlassian.net/browse/PDJB-1392) tracks this

samyou-softwire (approved):
LGTM, couple comments on namings/todos

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordRegistrationService.kt
samyou-softwire: looking at this I notice the class is called `OrganisationLandlord`. oops. I'll raise a PR to fix this. though there's still a couple places in this diff to update

src/main/resources/emails/OrganisationalLandlordRegistrationConfirmation.md
Bill-Haigh: We actually don't there is quite a split between templates where we do vs don't. I am going to add it to dev catchup, but leave this be for now as I would need to also update the template.

src/main/resources/emails/emailTemplates.json
Travis-Softwire: Do our notify tests not check prod notify? (Genuine question - I know they check our lower env notify)

samyou-softwire: I believe just that yeah - I don't think we give our tests a prod API key

Bill-Haigh: I think it is set depending on the base branch so when we merge to prod it should be using the prod api key for the tests

Bill-Haigh: it does make me think, should we have more test logging?

PR #1665, opened 2026-07-30
PDJB-896: Merge Task and DuplicableTaskWithDependencies

AEPR:
## Ticket number

PDJB-896

## Goal of change

Fold `DuplicableTaskWithDependencies` into `Task` now that every task in the codebase owns its own state, so the journey framework exposes a single concrete task base rather than three.

- `Task<TState, TDependencies>` becomes the merged base — always route-scoped, self-stated via `AbstractJourneyState`, with an optional typed `dependencies` reference to the enclosing state.
- The zero-dependencies convenience subclass `DuplicableTask` is renamed to `TaskWithoutDependencies` (after the merge, "duplicable" no longer distinguishes anything — every task is).
- Collapses the redundant DSL split: `duplicableTask { }` → `task { }` and `duplicableCheckAnswerTask` → `checkAnswerTask` (the previously unused bare-`Task` overloads are deleted).
- Updates ~45 call sites across `journeys/**` and 11 task-subclass supertypes; consolidates `DuplicableTaskTests` + `DuplicableTaskWithDependenciesTests` into `TaskTests` with three nested groupings; introduces `OccupationTask.inJourney(journey)` (called at the mount site) so the legacy task's `taskState get() = dependencies` binding is safe against eager evaluation by the DSL.
- Behaviour-preserving refactor. No user-visible change, no migrations, no feature flag.

`OccupationTask.inJourney(journey)` is a legacy shim needed because that task's `taskState get() = dependencies` would otherwise be evaluated by the DSL before `withDependencies { }` runs. It's called at the four mount sites and will be removed with `OccupationTask` itself under PDJB-1340.

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/Task.kt
```diff
@@ -7,12 +7,48 @@ import uk.gov.communities.prsdb.webapp.journeys.builders.ConfigurableElement
...
+// The single task base. A task owns its own JourneyState and namespaces its stored data behind a route prefix,
+// so the same task can be added to a journey more than once, each instance isolated by its route. A null route
+// keeps bare keys.
+//
+// Subclasses supply the task's steps, its task-specific state, and makeSubJourney; this base owns the route
+// binding and key-registry wiring (the actual scoping is done by the inherited delegateProvider).
+//
+// If a task needs typed access to the enclosing journey/sibling state, it declares a TDependencies contract; the
+// mount site binds the live state via withDependencies { }. Tasks with no such need use TaskWithoutDependencies
+// (TDependencies = Nothing, requiresDependencies = false).
```
JasminConterioSW, line 19: Copilot likes to add long explanatory comments but I don't think this is needed. It may be worth updating the relevant ReadMe though (doesn't have to be this PR)?

JasminConterioSW (approved):
LGTM :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/Task.kt
AEPR: OK, I will add another PR to strip out comments and update the readme, but I'll merge this as is now. I think it's worth including some of these comments in the Journey Framework because if you're looking into these files you probably want help understanding something about how to use it, but copilot does go overboard for sure.

PR #1666, opened 2026-07-30
PDJB-NONE: Improve org charity & company task names

samyou-softwire:
## Goal of change

the current names of these tasks (orgCharity and orgCompany) were not great - especially since they are not related to the is_charity / is_company fields in the database

renames tasks to make it clear they're a question of registration

adds some helper methods to org landlord to make it easier to reason with this

moves the question text out of forms.yml too


danielledias-MT (approved):
Nice one 💯

PR #1667, opened 2026-07-30
PDJB-1276: Org landlords on property details landlords tab

samyou-softwire:
## Ticket number

[PDJB-1276](https://mhclgdigital.atlassian.net/browse/PDJB-1276)

## Goal of change

make it so org landlords show up on the property details list of landlords tab

<details>
<summary>screenshots</summary>
my view
landlords tab viewing as me

someone elses view
landlords tab viewing as someone else
</details>

## Checklist

  mention that here

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
@@ -70,7 +70,7 @@ class LandlordDetailsController(
         return "landlordDetailsView"
     }
 
-    // TODO: PDJB-1276: Update skeleton page
+    // TODO: PDJB-1251: Update skeleton page
     private fun getOrgLandlordDetails(
         orgLandlord: OrganisationLandlord,
         model: Model,
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Landlord.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Landlord.kt
@@ -26,7 +26,7 @@ abstract class Landlord : ModifiableAuditableEntity() {
     abstract val displayName: String
 
     @get:Transient
-    abstract val contactEmailAddress: String
+    abstract val displayEmail: String
 
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Landlord.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Landlord.kt
@@ -22,6 +22,12 @@ abstract class Landlord : ModifiableAuditableEntity() {
     @get:Transient
     abstract val landlordType: LandlordType
 
+    @get:Transient
+    abstract val displayName: String
+
+    @get:Transient
+    abstract val contactEmailAddress: String
+
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     open val id: Long = 0
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Landlord.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Landlord.kt
@@ -23,10 +23,7 @@ abstract class Landlord : ModifiableAuditableEntity() {
     abstract val landlordType: LandlordType
 
     @get:Transient
-    abstract val displayName: String
-
-    @get:Transient
-    abstract val displayEmail: String
+    abstract val name: String
 
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlord.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlord.kt
@@ -23,7 +23,7 @@ class OrganisationLandlord() : Landlord() {
         get() = name
 
     @get:Transient
-    override val contactEmailAddress: String
+    override val displayEmail: String
         get() = email
 
     @Column(name = "organisation_landlord_name")
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlord.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlord.kt
@@ -18,6 +18,14 @@ class OrganisationLandlord() : Landlord() {
     override val landlordType: LandlordType
         get() = LandlordType.ORGANISATION
 
+    @get:Transient
+    override val displayName: String
+        get() = name
+
+    @get:Transient
+    override val contactEmailAddress: String
+        get() = email
+
     @Column(name = "organisation_landlord_name")
     lateinit var name: String
 
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlord.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlord.kt
@@ -19,41 +19,41 @@ class OrganisationLandlord() : Landlord() {
         get() = LandlordType.ORGANISATION
 
     @Column(name = "organisation_landlord_name")
-    var name: String? = null
+    lateinit var name: String
 
     @ManyToOne
     @JoinColumn(name = "organisation_address_id")
-    var address: Address? = null
+    lateinit var address: Address
 
     @Column(name = "organisation_email")
-    var email: String? = null
+    lateinit var email: String
 
     @Column(name = "organisation_phone_number")
-    var phoneNumber: String? = null
+    lateinit var phoneNumber: String
 
     @Column(name = "organisation_registrant_name")
-    var registrantName: String? = null
+    lateinit var registrantName: String
 
     @Column(name = "organisation_registrant_date_of_birth")
-    var registrantDateOfBirth: LocalDate? = null
+    lateinit var registrantDateOfBirth: LocalDate
 
     @Column(name = "organisation_registrant_email")
-    var registrantEmail: String? = null
+    lateinit var registrantEmail: String
 
     @Column(name = "organisation_registrant_phone_number")
-    var registrantPhoneNumber: String? = null
+    lateinit var registrantPhoneNumber: String
 
     // Note that this has no relation to the companyNumber nullable col
     @Column(name = "organisation_is_company")
-    var isCompany: Boolean? = null
+    var isCompany: Boolean = false
 
     // Note that this has no relation to the charityRegisteredWith & charityNumber nullable col
     @Column(name = "organisation_is_charity")
-    var isCharity: Boolean? = null
+    var isCharity: Boolean = false
 
     // Note that this determines whether the organisation needs a lead trustee
     @Column(name = "organisation_is_trust")
-    var isTrust: Boolean? = null
+    var isTrust: Boolean = false
 
     @Column(name = "organisation_company_number")
     var companyNumber: String? = null
@@ -81,13 +81,13 @@ class OrganisationLandlord() : Landlord() {
     var leadTrusteeAddress: Address? = null
 
     @Column(name = "organisation_main_contact_name")
-    var mainContactName: String? = null
+    lateinit var mainContactName: String
 
     @Column(name = "organisation_main_contact_email")
-    var mainContactEmail: String? = null
+    lateinit var mainContactEmail: String
 
     @Column(name = "organisation_main_contact_phone")
-    var mainContactPhone: String? = null
+    lateinit var mainContactPhone: String
 
     constructor(
         registrationNumber: RegistrationNumber,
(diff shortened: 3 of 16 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Landlord.kt
```diff
@@ -22,6 +22,12 @@ abstract class Landlord : ModifiableAuditableEntity() {
     @get:Transient
     abstract val landlordType: LandlordType
 
+    @get:Transient
+    abstract val displayName: String
```
samyou-softwire: is this a helpful distinction? or just `name` would be fine?

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
```diff
@@ -70,7 +70,7 @@ class LandlordDetailsController(
         return "landlordDetailsView"
     }
 
-    // TODO: PDJB-1276: Update skeleton page
+    // TODO: PDJB-1251: Update skeleton page
```
samyou-softwire, line 73: this TODO was pointing to the wrong ticket

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlord.kt
```diff
@@ -17,39 +17,47 @@ class OrganisationLandlord() : Landlord() {
...
+    override val displayName: String
+        get() = name
+
+    @get:Transient
+    override val contactEmailAddress: String
+        get() = email
+
     @Column(name = "organisation_landlord_name")
-    var name: String? = null
+    lateinit var name: String
```
samyou-softwire: this typing was unhelpful - for an org landlord the name is not nullable

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Landlord.kt
TomHanmer29: I don't think name can ever be different to the name we display right - so name should be fine? I think it's fine either way as both are correct so up to you what you prefer

src/main/resources/data-integration.sql
```diff
@@ -740,6 +740,10 @@ FROM generate_series(1, 100) AS s(i)
 JOIN property_ownership po ON po.id = 1600 + i
 ON CONFLICT DO NOTHING;
 
+-- PDJB-1276: Add org landlord (11) as joint landlord on property 1
```
TomHanmer29: nit: Do we need to prefix this with the ticket number? Is this a convention we should adopt more widely? I think in general these files are becoming pretty big and hard to follow, but we're talking about them in catchup soon anyway

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDetailsTests.kt
```diff
@@ -228,6 +228,21 @@ class PropertyDetailsTests : IntegrationTestWithImmutableData("data-local.sql")
...
             }
+
+            @Test
+            fun `joint property with org landlord shows org landlord card with LRN and email`(page: Page) {
+                val detailsPage = navigator.goToPropertyDetailsLandlordView(48)
+                detailsPage.tabs.goToLandlordDetails()
+
+                assertEquals(2, detailsPage.landlordSummaryCards.size)
+                val currentUserCard = detailsPage.landlordSummaryCards[0]
+                assertEquals("Alexander Smith (you)", currentUserCard.title.getText())
```
TomHanmer29, line 239: Can we add one of these to assert the (your organisation) text is working or is that a lot of setup for this PR?

TomHanmer29 (approved):
LGTM! Just a couple small thoughts :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/Landlord.kt
samyou-softwire: yeah the intent is that the 'name' of the landlord may change depending on context (emailing yourself vs including the name in emails). though yeah I think this is fine

although, I realise `contactEmailAddress` probably isn't so useful. internally the email should not be used for contacting the landlord. I'll change that one to `displayEmail`

src/main/resources/data-integration.sql
samyou-softwire: yeah I think we should audit this file a bit; I wonder how much of this data was just added for a QA

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyDetailsTests.kt
samyou-softwire: I'll look into this; not big right now but I can imagine tests down the road will need to test views like this

samyou-softwire: okay; the @WithOrgLandlordUser can be applied to an integration test class and will make you be logged in as an org landlord

samyou-softwire:
re requesting review on how the org landlord testing infrastructure looks

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/WithOrgLandlordUser.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.communities.prsdb.webapp.integration
+
+import org.springframework.test.context.ActiveProfiles
+
+@ActiveProfiles(profiles = ["local", "local-no-auth", "local-org-landlord"])
```
TomHanmer29: I think this is a fine way of doing it, it kinda sets the precedent that individual landlords are "normal" (as we have no annotation for them) and org landlords are "special" but I think that's true. The only similar thing to this is the @withMockUser annotation for landlords, LC and SO but I think these are separate enough concepts.

TomHanmer29 (approved):

samyou-softwire:
waiting on base branch #1654 to merge

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlord.kt
```diff
@@ -17,39 +17,47 @@ class OrganisationLandlord() : Landlord() {
     override val landlordType: LandlordType
         get() = LandlordType.ORGANISATION
 
+    @get:Transient
+    override val displayName: String
```
Travis-Softwire: I can live with this, but I think my preference would be to make the 'internal' names more specific and then use `name` and `email` as these accessor properties. Otherwise I think it's quite easy for someone to access the wrong property and get into trouble.

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/WithOrgLandlordUser.kt
```diff
@@ -0,0 +1,8 @@
...
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class WithOrgLandlordUser
```
Travis-Softwire: I'm not sure about this - local and local-no-auth are already applied to all integration tests, and everywhere else where we've added extra profiles we've added them explicitly. I think I'd prefer to keep it explicit if that's ok so it's easy to see what's going on - it would be easy to assume this works like @WithMockUser and is doing something directly to the security context - whereas someone could accidentally use this annotation in combination with some seed data that meant the user created doesn't actually become an org landlord.

src/main/resources/data-integration.sql
```diff
@@ -740,6 +740,10 @@ FROM generate_series(1, 100) AS s(i)
...
+INSERT INTO ownership_link (landlord_id, landlordship_id, created_date)
+VALUES (11, 1, '2026-07-31') ON CONFLICT DO NOTHING;
```
Travis-Softwire: I wonder if it's sensible to create some new propreties to add an org LL to rather than using one of the existing ones? It's less likely then that someone will come across it and trip a 500

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlord.kt
```diff
@@ -17,39 +17,47 @@ class OrganisationLandlord() : Landlord() {
...
+        get() = name
+
+    @get:Transient
+    override val displayEmail: String
+        get() = email
+
     @Column(name = "organisation_landlord_name")
-    var name: String? = null
+    lateinit var name: String
```
Travis-Softwire: Copilot is worried about the fact that these cols don't have a NOT NULL constraint on them, creating a type mismatch that apparently Hibernate can freak out about if it comes across one.

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsControllerTests.kt
```diff
@@ -50,6 +52,7 @@ class PropertyDetailsControllerTests(
     @BeforeEach
     fun setUpFeatureFlag() {
         whenever(featureFlagManager.checkFeature(PROPERTY_REGISTRATION_RESTRUCTURE_AND_SKIPPING)).thenReturn(false)
+        whenever(userToLandlordService.getLandlordForBaseUserId(any())).thenReturn(createIndividualLandlord())
```
Travis-Softwire: Is this actually used in practice?

Travis-Softwire (approved):
Some relatively small points, but overall LGTM once they're addressed

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlord.kt
samyou-softwire: I don't love this, though the issue is it previously implied that the name of an organisation landlord is nullable. technically this is true in the database but functionally I don't want everyone to have to write null checks as this should never be null

I'll take a look if there's a way we can handle this more gracefully than just forcing the types to be non-null

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlord.kt
samyou-softwire: yeah, after thinking about this I think we can leave name an email as-is. for now there is technically a difference since a one-user org LL (the only kind) can have both 
- a name we use if the org LL as a recipient (personal name)
- a name if used in most other contexts (org name)

though this distinction will only exist briefly, as soon an org will have multiple members the recipient system will all change again, so I think I'll revert this for now and just go back to `name` to save us restructuring everything twice. name can just be the org name

though I do think we will need the two email properties will need to exist, since these need to be treated differently
- "what is your email" -> we send emails here
- "what is your org's email" -> we show on this page

this will need to work correctly in org MVP so I don't think we can defer this, unless we're happy to use one email everywhere for this?

I think I'll leave the functional one as `email` and the display one as say `orgEmailAddress` with a comment to not use this email for email sending

src/test/kotlin/uk/gov/communities/prsdb/webapp/controllers/PropertyDetailsControllerTests.kt
samyou-softwire: ah good spot; no it isnt

src/main/resources/data-integration.sql
samyou-softwire: as mentioned in dev catchup I'll add this one-off data to integration manually

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/WithOrgLandlordUser.kt
samyou-softwire: ah yes I see this now. I still think this is useful as it's own annotation for less repetition, so I've renamed it to be more explicit on what is happening here `WithOrgLandlordProfile`. also removed the redundant profiles

samyou-softwire:
rebase onto main

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlord.kt
samyou-softwire: looking into it I can't see any specific recommendations on this, since in java it need not care about nullability https://docs.hibernate.org/orm/current/userguide/html_single/#entity-inheritance-discriminator

it looks like we could add some `CHECK()` constraints if we want to be more happy that this can't happen, though I wonder if this is necessary. what do you think?

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlord.kt
```diff
@@ -19,41 +19,44 @@ class OrganisationLandlord() : Landlord() {
...
     @JoinColumn(name = "organisation_address_id")
-    var address: Address? = null
+    lateinit var address: Address
 
+    /**
+     * This should not be used for sending emails, just for displaying the contact info for an organisation.
+     */
     @Column(name = "organisation_email")
-    var email: String? = null
+    lateinit var wholeOrgEmail: String
```
Travis-Softwire, line 32: Looks like the pre-rename version is still referenced in `orgLandlordDetailsView.html`?

Travis-Softwire (approved):
Just one more tiny point, otherwise good to go

src/main/kotlin/uk/gov/communities/prsdb/webapp/database/entity/OrganisationLandlord.kt
samyou-softwire: good spot, fixed

PR #1668, opened 2026-07-30
PDJB-1290: Auto-fill edit org gov body member address

samyou-softwire:
## Ticket number

[PDJB-1290](https://mhclgdigital.atlassian.net/browse/PDJB-1290)

## Goal of change

makes the edit button preserve the user's selected options for address editing

adds the specific options the user picked to the gov body member data model. these will not be saved to the database, just used temporarily in this journey

uses these fields to autofill the answers the user gave when editing the gov body member

I'm fairly certain this is new behaviour - whilst there have been other address edits journeys they:
- are their own journey, in which case they always start blank
- are on a CYA page, in which case they autofill by default what the user entered before

the new behaviour is allowing us to pass in any arbitrary info to show in the address task, and not necessarily what the user entered last (you can enter any governing body member)

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/OrgLandlordRegistrationTask.kt
```diff
@@ -277,15 +277,51 @@ class OrgLandlordRegistrationTask(
             duplicableTask(journey.govBodyMemberAddressTask, GovBodyMemberAddressTask.ROUTE_SEGMENT) {
                 parents { journey.orgGovBodyMemberDobStep.isComplete() }
                 nextStep { journey.saveGovBodyMemberStep }
-            }
-            step(journey.saveGovBodyMemberStep) {
-                parents { journey.govBodyMemberAddressTask.isComplete() }
-                nextStep { journey.orgGovBodyMemberListStep }
+                configureStep(journey.govBodyMemberAddressTask.lookupAddressStep) {
+                    withAdditionalContentProperties {
+                        val editingMember =
```
Travis-Softwire: I'd move the editingMember logic into a getter on the journey state.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/stepConfig/LookupAddressStepConfig.kt
```diff
@@ -25,6 +26,18 @@ class LookupAddressStepConfig(
 
     override fun chooseTemplate(state: AddressSearchState) = "forms/lookupAddressForm"
 
+    override fun resolvePageContent(
```
Travis-Softwire, line 29: It feels quite grim to pass the data in this way. Let's keep an eye on it - if we have another use case for this then we'll probably want a framework extension to allow us to inject data into the formModel from the DSL instead

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/SaveGovBodyMemberStepConfig.kt
```diff
@@ -30,6 +30,9 @@ class SaveGovBodyMemberStepConfig : AbstractInternalStepConfig<Complete, Landlor
                 ?: throw PrsdbWebException("Governing body member date of birth is invalid")
 
         val address = state.govBodyMemberAddressTask.getAddress()
+        val lookupFormModel = state.govBodyMemberAddressTask.lookupAddressStep.formModelOrNull
+        val selectFormModel = state.govBodyMemberAddressTask.selectAddressStep.formModelOrNull
+        val manualFormModel = state.govBodyMemberAddressTask.manualAddressStep.formModelOrNull
```
Travis-Softwire: I don't think we clear one or the other of these if the user changes their mind atm, so a user that enters a manual address, then backs out and selects an looked up address (or vice versa) will then find that abandoned branch has pre-populated the form if they edit it.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/stepConfig/LookupAddressStepConfig.kt
```diff
@@ -25,6 +26,18 @@ class LookupAddressStepConfig(
...
+        state: AddressSearchState,
+        defaultContent: Map<String, Any?>,
+    ): Map<String, Any?> {
+        val prefillPostcode = defaultContent["prefillPostcode"] as? String ?: return defaultContent
```
Travis-Softwire: Using the same key for both lookup and manual only works because you're setting them on the step. If they were set on the task it would populate both of them, which could be a gotcha. So you might want to make the keys unique (another indicator that this approach might not be the best - but maybe fine for now if we're careful)

Travis-Softwire (approved):
Some thoughts and suggestions from me. I don't love how we're passing in the pre-fill, but can live with it if we think addresses are a special case. LGTM once these are addressed

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/stepConfig/SaveGovBodyMemberStepConfig.kt
samyou-softwire: we can do this one way (the example you describe) to clear manual address if one is selected.

we can't do the other way since manual address is a question after select address (you "select" manual) so clearing the former makes the latter inaccessible

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/stepConfig/LookupAddressStepConfig.kt
samyou-softwire: I've renamed them to be specific. since they're specific to the step and not used in templates I've moved the strings to consts

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/OrgLandlordRegistrationTask.kt
samyou-softwire: have done so

PR #1669, opened 2026-07-31
PDJB-NONE: Speed up integration tests with a truncate-based database reset

Travis-Softwire:
## Ticket number

PDJB-NONE

## Goal of change

Speed up the integration test suite without losing any test coverage.

- Integration tests now reset the database by truncating every table in the `public` schema with `RESTART IDENTITY CASCADE`, rather than running `flyway.clean()` + `flyway.migrate()`. The old reset re-ran all 47 migrations (~1.3s) and happened roughly 263 times per suite run; the new one takes ~30ms. Migrations still run once per JVM at Spring context startup.
- `flyway_schema_history` and `local_council` are preserved, as the only two tables holding rows after a bare migrate. `PRESERVED_TABLES` and a note in `database.instructions.md` record that any future migration inserting reference data must be added there.
- Every test run now ends with a table of the slowest tests, so suite speed stays visible as we add scenarios. Length is configurable with `-PslowTestCount=N`.

Full local suite: 32.8 min before, 19.6–22.3 min across three runs after. The set of test names is identical before and after, verified by diffing the JUnit XML, plus four new tests covering the reset itself.

## Checklist

### DIFF
--- a/.github/instructions/database.instructions.md
+++ b/.github/instructions/database.instructions.md
@@ -101,6 +101,13 @@ CREATE TABLE example_table (
 CREATE INDEX idx_example_name ON example_table(name);
 ```
 
+### Reference Data in Migrations
+
+Integration tests reset the database by truncating every table except `flyway_schema_history` and `local_council`.
+If a migration inserts reference data that tests rely on, add that table to `PRESERVED_TABLES` in
+`src/test/kotlin/uk/gov/communities/prsdb/webapp/testHelpers/IntegrationTestHelper.kt`. Otherwise the data will be
+deleted before every test that resets the database.
+
 ## Search with Trigrams
 - Use `pg_trgm` extension for fuzzy text search
 - See existing search implementations for patterns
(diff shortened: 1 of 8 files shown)


Travis-Softwire:
Closing this in favour of a combined change.

Measured on CI over three runs of this branch against ten recent runs without it, using the `gradle check` step:

| | n | mean | median | range |
|---|---|---|---|---|
| With this change | 3 | 21.7 min | 21.4 | 21.0 - 22.7 |
| Without | 10 | 22.1 min | 21.9 | 19.9 - 27.1 |

A 1.8% difference against a baseline standard deviation of ~2.0 min, so no measurable CI gain. The reset it replaces costs ~1.3s per call on Docker Desktop but only a fraction of that on native Linux Docker, so there was little to reclaim on CI.

The change is still a large local win (full suite 32.8 -> 20.7 min on a dev machine) and it is a prerequisite for giving each test worker its own database cheaply. The branch is retained and the work continues on `chore/PDJB-NONE-test-parallelism-experiment`, which is based on it, so this will return as part of a change that actually moves CI time.

PR #1671, opened 2026-07-31
PDJB-NONE: [EXPERIMENT] Block external analytics requests in integration tests

Travis-Softwire:
**Experiment branch — not for merge.**

Measuring whether removing the external analytics request from the test critical path speeds up CI.

The layout renders `<script async src="https://plausible.io/js/pa-...js">` from a hardcoded constant (`PLAUSIBLE_URL`), not from the configurable `plausible.base-url`. So every page load in every integration test makes a real external HTTPS request. Being an `async` script it delays the window `load` event, which `ClickAndWaitable.clickAndWait` waits for — putting an internet round trip on the critical path of every interaction. Each test gets a fresh `BrowserContext`, so the browser cache never helps.

This branch aborts those requests via `page.route`, and temporarily uploads `build/test-results` so each run yields per-class timings rather than one noisy wall-clock number.

Control: the reset-only branch measured 21.0 / 22.7 / 21.4 min (mean 21.7) for the `gradle check` step over three runs of identical code.

### DIFF
--- a/.github/instructions/database.instructions.md
+++ b/.github/instructions/database.instructions.md
@@ -101,6 +101,13 @@ CREATE TABLE example_table (
 CREATE INDEX idx_example_name ON example_table(name);
 ```
 
+### Reference Data in Migrations
+
+Integration tests reset the database by truncating every table except `flyway_schema_history` and `local_council`.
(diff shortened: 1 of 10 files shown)


Travis-Softwire:
Experiment complete - findings carried into the stacked PRs #1675 / #1676. Closing.

PR #1675, opened 2026-08-01
PDJB-NONE: Shard CI tests across four parallel jobs

Travis-Softwire:
## Ticket number

PDJB-NONE

## Goal of change

Reduces the time the test suite takes on CI

- Runs the test suite as 4 parallel CI jobs instead of 1, each taking a quarter of the test classes
- Runs ktlint in its own job rather than repeating it in every shard
- Adds an aggregating job for the shards to report through, named `Build and test` to match the existing required status check
- Makes sharding opt-in via `-PshardIndex` and `-PshardCount`, so a local `./gradlew test` is unaffected

Takes CI from ~22 min to 12.0 min, and to 9.7 min with #1676 and #1677 on top. Coverage is unchanged: the shards run the same 2,869 tests as an unsharded run.

No branch protection change is needed, as the aggregating job keeps the existing `Build and test` name.

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -5,8 +5,12 @@ on:
 
 jobs:
   test:
-    name: Build and test
+    name: Test shard
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [0, 1, 2, 3]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -5,8 +5,12 @@ on:
 
 jobs:
   test:
-    name: Build and test
+    name: Test shard
```
TomHanmer29, line 8: nit: should this be called "Build and test shard"
Is there a way we could just do the build step once if we're sharding the tests into 4 bits or does it not matter if we're running them at the same time?

TomHanmer29 (approved):
Question on the workflow, but otherwise LGTM!

.github/workflows/build-and-test.yml
Travis-Softwire: Yeah it would slow it down as the build will run in parallel across each shard atm, so in real time it takes the same, plus you'd then have the overhead of uploading and downloading the build artefacts

PR #1676, opened 2026-08-01
PDJB-NONE: Split property registration single page tests by journey phase

Travis-Softwire:
## Ticket number

PDJB-NONE

## Goal of change

Splits `PropertyRegistrationSinglePageTests` so that CI sharding can distribute it

- Splits a single 3,207-line class into 20 classes grouped by journey phase
- Groups the tests that run with the property registration restructure flag off under a `PreRestructure` prefix, so that PDJB-1340 can delete those files whole when the flag is retired
- Leaves all test bodies unchanged: 347 tests before, 347 after

Sharding, added in #1675, distributes work per top-level class, so this class set a floor under it. Takes CI from 12.0 min to 10.7 min.

- `ProvideTenancyDetailsLaterStep` now runs with the restructure flag enabled. It does not set the flag itself, so it ran flag-off only by virtue of where it sat in the file. It still carries a `// TODO PDJB-942`.
- The `epcRegisterClient` and `savedJourneyStateRepository` overrides are declared only on the two classes that use them, to avoid forcing a second Spring test context.

## Checklist


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationOccupancySinglePageTests.kt
```diff
@@ -0,0 +1,176 @@
...
+            peoplePage.submitNumOfPeople(2)
+            assertThat(peoplePage.form.getErrorMessage())
+                .containsText(
+                    "The number of people in the property must be the same as or higher than the number of households in the property",
+                )
+        }
+    }
+
+    @Nested
+    inner class NumberOfBedroomsStep {
```
TomHanmer29: Bedrooms should probably move out of this section, now that it's no longer gated behind occupancy (maybe a hangover from another PR where this didn't happen?)

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationPreRestructureElectricalSafetySinglePageTests.kt
```diff
@@ -0,0 +1,129 @@
...
+import uk.gov.communities.prsdb.webapp.constants.PROPERTY_REGISTRATION_RESTRUCTURE_AND_SKIPPING
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.basePages.BasePage.Companion.assertPageIs
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.propertyRegistrationJourneyPages.CheckElectricalCertUploadsFormPagePropertyRegistration
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.propertyRegistrationJourneyPages.ElectricalCertExpiryDateFormPagePropertyRegistration
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.propertyRegistrationJourneyPages.HasElectricalCertFormPagePropertyRegistration
+import uk.gov.communities.prsdb.webapp.testHelpers.builders.PropertyStateSessionBuilder
+
+// Covers the property registration journey with the restructure-and-skipping feature flag OFF.
+// TODO PDJB-1340: delete every PropertyRegistrationPreRestructure*SinglePageTests file when
+// PROPERTY_REGISTRATION_RESTRUCTURE_AND_SKIPPING is removed. Every class here has an equivalent
```
TomHanmer29, line 19: I might be missing something (as the diff hasn't changed much, so I suspect these were here already, but) I'm surprised we need to duplicate these tests for gas, electrical, JLs and EPC as the flag shouldn't really touch them for single page tests. Maybe it's to do with the navigator? Taking a quick look at both the EPC files, the contents seem identical but with the flag the other way

TomHanmer29 (commented):
The amount of duplication here looks a bit fishy, but I think it may have been like this already. Probably something copilot can fix!

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationPreRestructureElectricalSafetySinglePageTests.kt
Travis-Softwire: Will double check. Suspect we might have just automatically had duplicate tests with the flag off and on to catch regressions

Travis-Softwire: Yes, the flag switches between two journeys, so although the tests are identical they're running the steps in the 'other' journey - I think we should keep them in for now as 28 days later should go live in the not-too-distant future

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/PropertyRegistrationOccupancySinglePageTests.kt
Travis-Softwire: Good spot, have moved it

TomHanmer29 (approved):

PR #1677, opened 2026-08-01
PDJB-NONE: Reset integration test database by truncation instead of Flyway clean

Travis-Softwire:
## Ticket number

PDJB-NONE

## Goal of change

Cuts the time a full test run takes on a developer machine by about a third

- Resets the test database by truncating the `public` schema rather than running `flyway.clean()` and `flyway.migrate()`. That reset runs around 263 times per suite run, and drops from ~1.3s to ~30ms. Migrations still run once per JVM at context startup
- Preserves `flyway_schema_history` and `local_council`, the only tables holding rows after a bare migrate
- Adds `IntegrationTestHelperTests` covering the reset

Measured over three runs, a full local run drops by around a third (32.8 min to 20.7 min when measured). Absolute times vary by up to 20% day to day on the same machine, so the proportion is the meaningful figure rather than the raw minutes. The effect on CI is within run-to-run variance, as the reset is I/O bound and Docker Desktop on Windows is much slower at it than CI's native Linux Docker.

Stacked on #1676.

Any future migration that inserts reference data must add its table to `PRESERVED_TABLES`, or that data will be truncated. This is noted in `database.instructions.md`.

## Checklist


docs/IntegrationTestReadMe.md
```diff
@@ -13,6 +13,18 @@ We set seed data by passing SQL script names into integration test class constru
...
+Resetting the database means truncating every table in the `public` schema with `RESTART IDENTITY CASCADE`, then
+running the seed scripts. Migrations run once per JVM at Spring context startup rather than being re-applied for every
+reset. Two tables are deliberately preserved:
+
+* `flyway_schema_history` - Flyway's own bookkeeping. Dropping it would make every migration re-run.
+* `local_council` - reference data inserted by `V1_0_0__la_and_address_tables.sql` (as `local_authority`, renamed by
+  `V1_6_0`). No seed script writes to it.
+
+These are the only two tables with rows after a bare migrate. If you add a migration that inserts reference data, add
+its table to `PRESERVED_TABLES` in `IntegrationTestHelper`, or the data will be wiped before every test. You can check
```
TomHanmer29: Worth adding a note to update the readme to include the PRESERVED_TABLE?
And also to add a new integration test helper test to check that table is preserved?

TomHanmer29 (approved):

PR #1679, opened 2026-08-01
PDJB-NONE: [TEMP] measure six shards

Travis-Softwire:
Throwaway branch to measure 6 shards against the 4-shard result of 9.8 min. Will be closed.


Travis-Softwire:
Experiment complete. Measured on CI: 4 shards + hash 9.8 min (current), 4 shards + packing 10.2, 6 shards + hash 11.4, 6 shards + packing 10.1. Packing by source file size improved the spread every time but never the critical path, because Gradle's suite timings exclude Spring context and container startup - roughly 19% of runtime, distributed by which contexts a shard needs rather than by test size. Keeping 4 shards with hashing.

PR #1681, opened 2026-08-03
PDJB-NONE: Allow test shards to run in parallel locally

Travis-Softwire:
## Ticket number

PDJB-NONE

## Goal of change

Lets a developer run the test suite as parallel shards locally, roughly halving the time a full run takes

- Adds a `Run Tests in Parallel` run configuration that builds once, then runs the four shards concurrently
- Gives each shard its own test output directories, as concurrent shards otherwise delete each other's results
- Numbers the shards from 1 rather than 0, so the names match on CI and locally
- Fixes the shard predicate so nested classes always run with their enclosing class

Measured back to back on the same machine: 26.4 min serial, 14.2 min across four shards. Coverage is unchanged, with the shards together running the same 766 suites and 3,823 tests as a serial run.

- The nested class fix is a bug in the sharding added by #1675. `path.substringBefore('$')` kept the `.class` suffix for a top-level class but stripped it for a nested one, so the two hashed differently and could land in different shards - `NgdAddressLoaderTests` and its `DataPackageProgressTests` were running in separate shards. No coverage was lost, but shard contents were incoherent. Verified against an unsharded baseline: 2,880 tests either way, none missing, none duplicated, no class split across shards.
- The orchestration is three configurations rather than one because IntelliJ only runs a Gradle before-launch task when the parent configuration is not itself a Gradle configuration. `Run Tests in Parallel` therefore runs a trivial npm script and does the real work in its before-launch tasks, which run in order.
- Renaming an existing run configuration is safer than changing its type, as IntelliJ does not reload a configuration whose type changes while the name stays the same.
- The shards run 40 to 66 containers on Docker Desktop. This passed on 8 cores and 32GB RAM, but two shards would be a safer default on a smaller machine.

## Checklist

### DIFF
--- a/.run/Test all shards.run.xml
+++ b/.run/Test all shards.run.xml
@@ -1,0 +1,9 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Parallel test suite" type="CompoundRunConfigurationType">
+    <toRun name="Test shard 0 of 4" type="GradleRunConfiguration" />
+    <toRun name="Test shard 1 of 4" type="GradleRunConfiguration" />
+    <toRun name="Test shard 2 of 4" type="GradleRunConfiguration" />
+    <toRun name="Test shard 3 of 4" type="GradleRunConfiguration" />
+    <method v="2" />
+  </configuration>
+</component>
--- a/.run/Test all shards.run.xml
+++ b/.run/Test all shards.run.xml
@@ -1,9 +1,9 @@
 <component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="Parallel test suite" type="CompoundRunConfigurationType">
-    <toRun name="Test shard 0 of 4" type="GradleRunConfiguration" />
+  <configuration default="false" name="Test all shards" type="CompoundRunConfigurationType">
     <toRun name="Test shard 1 of 4" type="GradleRunConfiguration" />
     <toRun name="Test shard 2 of 4" type="GradleRunConfiguration" />
     <toRun name="Test shard 3 of 4" type="GradleRunConfiguration" />
+    <toRun name="Test shard 4 of 4" type="GradleRunConfiguration" />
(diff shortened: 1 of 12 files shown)


.run/Test all shards.run.xml
```diff
@@ -0,0 +1,9 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Parallel test suite" type="CompoundRunConfigurationType">
+    <toRun name="Test shard 0 of 4" type="GradleRunConfiguration" />
+    <toRun name="Test shard 1 of 4" type="GradleRunConfiguration" />
+    <toRun name="Test shard 2 of 4" type="GradleRunConfiguration" />
+    <toRun name="Test shard 3 of 4" type="GradleRunConfiguration" />
```
TomHanmer29: nit: Does naming this 0 of 4 to 3 of 4 make that much sense? (it can never be 4 of 4, maybe 1 to 4 is better?)

.run/Test shard 4 of 4.run.xml
```diff
@@ -0,0 +1,24 @@
+<component name="ProjectRunConfigurationManager">
```
TomHanmer29, line 1: Maybe a silly question but why do these need to be 4 almost-identical files? Is there no way we can do this dynamically?

TomHanmer29 (approved):
Couple of very minor points, otherwise happy :)

.run/Test all shards.run.xml
Travis-Softwire: Yes, fair, have renamed them here and on the pipeline so they're 1 based

.run/Test shard 4 of 4.run.xml
Travis-Softwire: Sadly not, Intellij doesn't allow parameterised run configs

PR #1683, opened 2026-08-03
PDJB-1168: org LL cya page refactors and general improvements

TomHanmer29:
## Ticket number

PDJB-1168

## Goal of change

Adds some general PR fixes and comment resolutions from [a previous PR](https://github.com/communitiesuk/prsdb-webapp/pull/1628)

## Description of main change(s)

Adds a TODO for PDJB-1133
Makes destinationForSelectedLandlordType an internal step
use an else instead of early return and put refresh context call at the end
Back behaviour on governing bodies change page shouldn’t go to the “have you got governing body members”. Otherwise unchanged in the normal flow.

## Checklist

Delete any that are not applicable, and add explanation below for any that are applicable but haven't been done

- [x] Test suite has been run in full locally and is passing
- [x] Branch has been rebased onto main and run locally, with everything working as expected (both for your new feature
  and any related functionality)
- [x] TODO comments referencing this JIRA ticket have been searched for and removed - if a future PR will address them,
  mention that here
- [x] QA instructions have been added to the ticket (particularly if this is the last PR required to complete the ticket)

PR #1689, opened 2026-08-04
PDJB-1168: address org landlord cya review comments

TomHanmer29:
## Ticket number

PDJB-1168

## Goal of change

Addresses review comments added in https://github.com/communitiesuk/prsdb-webapp/pull/1683 after it merged

Adds comment about parentage
Adds checkAnswerTask overload that takes a configure block
Adds tests for the new internal step

## Checklist


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/states/CheckYourAnswersJourneyState.kt
```diff
@@ -107,6 +108,20 @@ interface CheckYourAnswersJourneyState : JourneyState {
             }
         }
 
+        fun <TJourneyState : CheckYourAnswersJourneyState, TTaskState : JourneyState> JourneyBuilder<TJourneyState>.checkAnswerTask(
```
TomHanmer29, line 111: I can split this out into a NONE pr if you'd prefer :)

Travis-Softwire (approved):
LGTM!

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/shared/states/CheckYourAnswersJourneyState.kt
Travis-Softwire: No, happy for it to be included here :)

PR #1687, opened 2026-08-03
Feat/pdjb 1251 landlord details for org landlords structure

TomHanmer29:
## Ticket number

PDJB-1251

## Goal of change

skeleton out org landlord details tab structure

## Description of main change(s)

image

image

image

Adds the tab structure in the designs, and a delete link
Moves the existing skeleton into the tabs
picks the page to show based on the landlord type in the controller

## Anything you'd like to highlight to the reviewer?

Include e.g. anything unusual about the PR, where there was some debate over how to implement it, or anywhere you were
unsure of the approach to take and would like specific feedback.

## Checklist

Delete any that are not applicable, and add explanation below for any that are applicable but haven't been done

- [x] Screenshots of any UI changes have been added
- [x] Controller tests for any new endpoints, including testing the relevant permissions
- [x] Test suite has been run in full locally and is passing
- [x] Branch has been rebased onto main and run locally, with everything working as expected (both for your new feature
  and any related functionality)
- [x] TODO comments referencing this JIRA ticket have been searched for and removed - if a future PR will address them,
  mention that here
- [ ] QA instructions have been added to the ticket (particularly if this is the last PR required to complete the ticket)
- [x] This feature is behind a feature flag. I've checked that there will be no change in function if the feature flag is disabled

PR #1691, opened 2026-08-04
PDJB-1475: Org landlord details -  organisation contacts tab

JasminConterioSW:
## Ticket number

PDJB-1475

## Goal of change

Implement the "Organisation contacts" tab on Landlord Details

Populates the Organisation contacts tab Renders an organisation landlord's contact information as GOV.UK summary cards on the Organisation for 
- Main contact, 
- Lead trustee (trusts only), 
- Governing body (one numbered card per member) - section can be empty
- Registration contact.

image
image

This is a stacked PR on top of #1687 (PDJB-1251), which introduces the 3-tab structure for the organisation landlord details page; the base is `feat/PDJB-1251-landlord-details-for-org-landlords-structure` and this PR should be retargeted to `main` once #1687 merges.

## Checklist

  mention that here
  - Not checked.
  - Uses the existing feature flag for the stacked organisation landlord details work; this PR does not add a new flag.


src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/LandlordDetailsOrganisationContactsPageTests.kt
```diff
@@ -0,0 +1,83 @@
...
+
+            val registrationContact = detailsPage.registrationContactCard
+            assertThat(registrationContact.title).containsText("Registration contact")
+            assertThat(registrationContact.summaryList.nameRow.value).containsText("Riya Registrant")
+            assertThat(registrationContact.summaryList.dateOfBirthRow.value).containsText("15 June 1985")
+            assertThat(registrationContact.summaryList.emailRow.value).containsText("riya.registrant@meadowbrook.example")
+            assertThat(registrationContact.summaryList.phoneNumberRow.value).containsText("01444444444")
+
+            assertThat(detailsPage.leadTrusteeCard).isHidden()
+            assertThat(detailsPage.governingBodyMembersLink).isHidden()
```
JasminConterioSW, line 33: Is this correct? Should we have the link if there are no governing body members?

JasminConterioSW: Looks like this is correct - members are only added if the organisation is not registered with companies' house, so would be updated by changing that (from the org details tab)

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/OrganisationalLandlordContactsViewModel.kt
```diff
@@ -0,0 +1,94 @@
...
+                            addRow("landlordDetails.org.leadTrusteePhone", orgLandlord.leadTrusteePhone)
+                            addRow(
+                                "landlordDetails.org.leadTrusteeAddress",
+                                orgLandlord.leadTrusteeAddress?.toMultiLineAddress()?.split("\n"),
+                            )
+                        }.toList(),
+            )
+        }
+
+    val showGoverningBody: Boolean = governingBodyMembers.isNotEmpty()
```
samyou-softwire: OrganisationLandlord has a method `hasGoverningBody`, so don't think we need this check too

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/converters/MessageKeyConverter.kt
```diff
@@ -46,6 +49,14 @@ class MessageKeyConverter {
...
             }
 
+        private fun convertGoverningBodyMemberType(type: GoverningBodyMemberType): String =
+            when (type) {
+                GoverningBodyMemberType.DIRECTOR -> "landlordDetails.org.governingBody.type.director"
+                GoverningBodyMemberType.TRUSTEE -> "landlordDetails.org.governingBody.type.trustee"
+                GoverningBodyMemberType.PARTNER -> "landlordDetails.org.governingBody.type.partner"
+                GoverningBodyMemberType.OTHER -> "landlordDetails.org.governingBody.type.other"
+            }
+
```
samyou-softwire, line 59: I suspect that this might be a case where we do actually want to put in `forms.`. these methods are meant to be called from many contexts

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
```diff
@@ -80,7 +81,7 @@ class LandlordDetailsController(
             organisationGoverningBodyMemberService.getGoverningBodyMembers(orgLandlord)
```
samyou-softwire: I found it a little surprising when looking at the code like this that you need to use a separate service to get the list of members. this is the first time we query the list of users for an org LL so may be worth checking if this is a useful pattern

I'm thinking of how for a landlord you can do `landlord.ownershipLinks` to join the tables up

src/test/resources/data-mockuser-org-landlord-basic.sql
samyou-softwire, line 1: maybe I'll check this in dev catchup; making a new .sql test files vs using pre-existing ones. we define a lot of data here and it feels like we could be using the `data.local.sql` for this

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/OrganisationalLandlordDetailsPageTests.kt
```diff
@@ -0,0 +1,83 @@
...
+            assertThat(detailsPage.leadTrusteeCard).isHidden()
+            assertThat(detailsPage.governingBodyMembersLink).isHidden()
+            assertEquals(0, detailsPage.governingBodyMemberCardCount())
+        }
+    }
+
+    @Nested
+    inner class TrustWithGoverningBodyMembers : NestedIntegrationTestWithImmutableData("data-mockuser-org-landlord-trust.sql") {
+        @Test
+        fun `organisation contacts cards show the correct rows for each contact type of a trust org landlord`() {
```
samyou-softwire: this name I feel is a little misleading - governing body isn't related to an org landlord being a trust. could we split the governing body checks into its own test?

samyou-softwire (approved):
LGTM, a couple small comments & one thought on how we can improve getting the org users for an org

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/converters/MessageKeyConverter.kt
JasminConterioSW: The place it's actually called at the moment is not in a form though...
I'd say it's more strongly linked to the landlordDetails (or organisationDetails if we wanted a new .yml file) than forms?

There's a lot that's in `forms.yml` that I think shouldn't be there. Some of the other methods in MessageKeyConverter also don't use messages from the forms file, so it wouldn't be a new pattern to leae these in landlordDetails.yml

src/test/resources/data-mockuser-org-landlord-basic.sql
JasminConterioSW: Because we are looking at the logged in user's landlord record, we need MockUser to be registered as the right kind of landlord - in `data-local.sql` MockUser is an individual landlord.

Here I'm adding 2 sql files because in one case I want MockUser to be an organisation with minimal contacts (to check that the others don't show) and in the other case I want it to have all the different types of contact.

We often do use `data-local.sql`, but sometimes it's better not to so that test data isn't changed by accident (e.g. there was a local council user test that failed every time we added someone to seed data on data-local.sql because it counted the number of users in some way, it might have been how many emails were sent!)

src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/LandlordDetailsController.kt
JasminConterioSW: I think you're right, copilot is looking now :)

src/main/kotlin/uk/gov/communities/prsdb/webapp/helpers/converters/MessageKeyConverter.kt
samyou-softwire: hm, I suppose the thought is that they're not specific to the landlord details page, but I guess they are details of a landlord? should be fine then

src/test/resources/data-mockuser-org-landlord-basic.sql
JasminConterioSW: Ooh unless `@WithOrgLandlordProfile` makes it Just Work...

JasminConterioSW: I think I can use that to get rid of `data-mockuser-org-landlord-basic.sql`, but want to keep `data-mockuser-org-landlord-trust.sql` which will show all the cards

PR #1696, opened 2026-08-04
PDJB-NONE: rename organisation landlord user to organisational landlord user

samyou-softwire:
as part of an ongoing effort to fix all references to 'organisationlandlord'

starting with the related organisation user entity, rename this to more simply `OrganisationUser`

separate PR to keep the diffs manageable here


Travis-Softwire:
I'm not sure about this - as mentioned previously we're going to have other organisations on the service over time (Local Councils, potentially lettings agents in the future, maybe more) so I think unless this is going to be generic to any organisation it should continue to reference landlords

samyou-softwire:
makes sense 👍, I'll rename to organisational_landlord_user then

JasminConterioSW (approved):
LGTM - can you update the PR name so we get the correct rename in release notes?

PR #1698, opened 2026-08-04
PDJB-NONE: Share one database and cache container across test contexts

Travis-Softwire:
## Ticket number

PDJB-NONE

## Goal of change

Stops the test suite starting a Postgres and Redis container for every Spring test context

- Holds one Postgres and one Redis container per JVM as shared instances, rather than creating a pair per Spring test context
- Raises `max_connections` on Postgres, as one database now serves every context's connection pool

A serial run previously started around 16 container pairs, and running the shards in parallel started enough to destabilise Docker Desktop.

Verified on a full local run: 800 suites, 3,943 tests, 0 failures.

**Expect the speed up to vary by platform.** Most of the gain comes from removing container pressure, so it is largest when running the shards in parallel on Docker Desktop for Windows. Serially it is roughly time neutral, and on macOS or CI it is likely to be closer to neutral than to the Windows result.

**The shared containers deliberately ignore `stop()` and `close()`.** Spring closes container beans when a test context is destroyed, which would take the containers away from every other cached context. Testcontainers' reaper still stops them when the JVM exits.

Spring does have a supported way to skip that shutdown, but it only applies when the container is marked for reuse *and* the environment supports reuse, which requires `testcontainers.reuse.enable=true` in each developer's `~/.testcontainers.properties`. That would work on some machines and silently not on others, so it is not relied on here.

## Checklist

### DIFF
--- a/src/test/kotlin/uk/gov/communities/prsdb/webapp/SharedContainers.kt
+++ b/src/test/kotlin/uk/gov/communities/prsdb/webapp/SharedContainers.kt
@@ -1,0 +1,39 @@
+package uk.gov.communities.prsdb.webapp
+
+import org.testcontainers.containers.GenericContainer
+import org.testcontainers.containers.PostgreSQLContainer
+import org.testcontainers.utility.DockerImageName
+
+// The containers are shared by every Spring test context in the JVM, so they must outlive any one context.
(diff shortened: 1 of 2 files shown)


TomHanmer29 (approved):
LGTM!

PR #1703, opened 2026-08-05
PDJB-896: Fix nested task delegate collision logic

AEPR:
## Ticket number
PDJB-896

## Goal of change
While url routes were being correctly applied to all elements in a task (including other tasks) the delegate key prefixes were not being correctly distributed to nested tasks. This resolves that issue by ensuring that during the build of a task the prefix is distributed to all its members fully.

## Checklist

- [X] TODO comments referencing this JIRA ticket have been searched for and removed - if a future PR will address them,
  mention that here
- [X] This feature is not behind a feature flag. I've checked that this is appropriate and we're happy with this code becoming live as soon as we release


src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/builders/TaskInitialiser.kt
```diff
@@ -115,26 +119,19 @@ class TaskInitialiser<TStateInit : JourneyState, TDependencies : Any>(
...
         // so a request to "<taskRoute>" resolves to the task's genuine first step (internal or requestable).
-        val stepsWithLanding =
-            taskRoute?.let { route ->
-                builtSteps.filterIsInstance<JourneyStep.RequestableStep<*, *, *>>().forEach { step ->
-                    val existingPrefix = step.stepConfig.urlPathPrefix
-                    step.stepConfig.urlPathPrefix = if (existingPrefix != null) "$route/$existingPrefix" else route
-                }
-                builtSteps + createLandingStep(route)
-            } ?: builtSteps
+        val landingStep = elementConfiguration.buildPath(taskSegment)?.let { createLandingStep(it) }
```
Travis-Softwire: What happens if we have a task with no route nested inside one that does have a route?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/JourneyStateDelegateProvider.kt
```diff
@@ -29,7 +29,7 @@ class JourneyStateDelegateProvider(
     private var routePrefix: String? = null
 
     fun bindRoutePrefix(routePrefix: String?) {
-        this.routePrefix = routePrefix
+        this.routePrefix = this.routePrefix?.let { "$it/$routePrefix" } ?: routePrefix
```
Travis-Softwire: If `routePrefix` (the arg) is null but `this.routePrefix` isn't, won't that set `this.routePrefix` to `<this.routePrefix value>/null`?

Travis-Softwire: Also don't 100% understand why we need this AND `buildPath` but I'm probably missing some nuance

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/builders/TaskInitialiser.kt
Travis-Softwire: Related to that, I think the comment above is out of date as we don't have a loop any more?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/builders/StepInitialiser.kt
```diff
@@ -10,6 +10,7 @@ import uk.gov.communities.prsdb.webapp.journeys.JourneyStep
 import uk.gov.communities.prsdb.webapp.journeys.NoParents
 import uk.gov.communities.prsdb.webapp.journeys.Parentage
 import uk.gov.communities.prsdb.webapp.journeys.StepInitialisationStage
+import java.util.Stack
```
Travis-Softwire: We don't actually use this as a stack do we? Don't we fold it in order? So it could just be a mutable list?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/JourneyStep.kt
```diff
@@ -15,24 +15,21 @@ sealed class JourneyStep<out TEnum : Enum<out TEnum>, TFormModel : FormModel, in
...
         stepConfig: AbstractRequestableStepConfig<TEnum, TFormModel, TState>,
     ) : JourneyStep<TEnum, TFormModel, TState>(stepConfig) {
-        val routeSegment: String get() = stepConfig.routeSegment
+        val urlPath: String get() = stepConfig.urlPath
 
-        // The route of the task this step sits under, if any. See the `urlPath` extension for how it composes.
-        val urlPathPrefix: String? get() = stepConfig.urlPathPrefix
-
-        override fun getRouteSegmentOrNull(): String? = stepConfig.routeSegment
+        override fun getRouteSegmentOrNull(): String? = stepConfig.urlPath
```
Travis-Softwire: Does this need to be renamed if it's now returning the full urlPath and not the route segment?

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/builders/StepInitialiser.kt
```diff
@@ -216,7 +232,7 @@ class StepInitialiser<TStep : AbstractStepConfig<TMode, *, TState>, in TState :
         checkForUninitialisedParents(parentage.potentialParents)
 
         step.initialize(
-            segment,
+            segment?.let { elementConfiguration.buildPath(segment) },
```
Travis-Softwire: Minor, think the second `segment` could be `it`?

Travis-Softwire (commented):
A few Qs/ points that I wanted to clarify before approving (can chat tomorrow before the away day starts if helpful)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/builders/TaskInitialiser.kt
AEPR: Yep, this is slightly wrong! Good catch, sorry about that.

Updating so that if a task does not have a taskSegment it doesn't have a landing step.

Travis-Softwire (approved):
👍

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/JourneyStateDelegateProvider.kt
AEPR: On the contrary, I think this was my initial attempt at a fix that revealed the wider issue and I missed that it was still in here.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/builders/TaskInitialiser.kt
AEPR: Yes, and it's been updated to not reference the implementation at all in the other PR (now merged).

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/builders/StepInitialiser.kt
AEPR: No, I'd assumed looping over a stack would be the correct FILO order. I've actually simplified to only allow a single prefix for now.

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/JourneyStep.kt
AEPR: I've removed all references to route in the journey step itself - it's now a concept during initialisation but the step is initialised with a path.

PR #1706, opened 2026-08-05
PDJB-NONE: Rename org landlord class

samyou-softwire:
renames to OrganisationalLandlord

doesn't touch any related services yet - just the class name and direct instantiations of the class


danielledias-MT (approved):
💯

PR #1710, opened 2026-08-07
PDJB-1237: User update org type

samyou-softwire:
## Ticket number

PDJB-1237

## Goal of change

allow for org type & if needed, lead trustee to be updated

adds a new controller & journey

updates the TODO in the existing CYA page for org type update

adds a shared state interface to allow for some sharing of common state between these two objects

uses step initialisation to allow for the steps to know whether this answer is a changed from the previous or not

adds tests

adds some common base page test files for pages reused between the landlord reg & update journeys

are we happy with the way the two journeys are set up?

## Checklist

  mention that here

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
@@ -5,10 +5,12 @@ import org.springframework.beans.factory.ObjectFactory
 import uk.gov.communities.prsdb.webapp.annotations.webAnnotations.JourneyFrameworkComponent
 import uk.gov.communities.prsdb.webapp.config.managers.FeatureFlagManager
 import uk.gov.communities.prsdb.webapp.constants.ORGANISATION_LANDLORD_REGISTRATION
+import uk.gov.communities.prsdb.webapp.constants.enums.OrgType
 import uk.gov.communities.prsdb.webapp.journeys.AndParents
 import uk.gov.communities.prsdb.webapp.journeys.Destination
 import uk.gov.communities.prsdb.webapp.journeys.JourneyStateService
 import uk.gov.communities.prsdb.webapp.journeys.OrParents
+import uk.gov.communities.prsdb.webapp.journeys.StepInitialisationStage
 import uk.gov.communities.prsdb.webapp.journeys.StepLifecycleOrchestrator
 import uk.gov.communities.prsdb.webapp.journeys.TaskWithoutDependencies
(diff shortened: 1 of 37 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordService.kt
```diff
@@ -208,6 +208,26 @@ class LandlordService(
         val landlordEntity = userToLandlordService.getCurrentOrganisationLandlordForUser()
 
         orgLandlordUpdate.name?.let { landlordEntity.name = it }
+        orgLandlordUpdate.isCompany?.let { landlordEntity.isCompany = it }
```
Travis-Softwire, line 211: Don't we need to clear CH number if this is set to null (similar to what we're doing w/ trusts below)? And ditto for Charity data? (If that's planned for a later ticket then probably just worth a TODO)

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
```diff
@@ -262,9 +277,53 @@ class LandlordRegistrationTask(
                     }
 
                     OrgTypeStep.ROUTE_SEGMENT -> {
-                        // TODO PDJB-1237 : replace this placeholder with the org type update journey
-                        checkAnswerStep(journey.orgLandlordRegistrationTask.updateDetailsTodoStep, OrgTypeStep.ROUTE_SEGMENT) {
-                            withAdditionalContentProperty { "todoComment" to "TODO PDJB-1237: Organisation type update journey" }
+                        step(journey.orgLandlordRegistrationTask.orgTypeStep) {
```
Travis-Softwire, line 280: Do we have an integration test for this? Looks like there might be a missed TODO to add some in `LandlordRegistrationJourneyTests`?

Travis-Softwire (approved):
A couple of small points, but otherwise LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordService.kt
samyou-softwire: don't think so; the isCompany and isCharity tickboxes don't relate to the charity or company number questions since we ask 'do you have one' separately

appreciate this is still unclear, may raise a PR at some point to rename these cols in kotlin to make this clearer

may also help when the actual column clearing for company & charity number is in

src/main/kotlin/uk/gov/communities/prsdb/webapp/journeys/landlordRegistration/tasks/LandlordRegistrationTask.kt
samyou-softwire: yes - I missed these when I initially put this up for review but they're added now

PR #1720, opened 2026-08-10
PDJB-1277: Show organisation landlords on local council property Landlords tab

JasminConterioSW:
## Ticket number

PDJB-1277

## Goal of change

Stop the local council view of a property record's Landlords tab from crashing when a landlord is an organisation, and show them in a card.

- `PropertyDetailsLandlordViewModelBuilder.buildLocalCouncilSummaryCards` now branches by landlord type instead of asserting every landlord is an individual.
- Organisation landlords render a card mirroring the landlord-facing org card: org name as the title, with Landlord Registration Number and (whole-org) email address.
- Org cards get a "View landlord record" link like individual cards. It currently points at the shared local council landlord-record path; a `TODO PDJB-1432` marks pointing it at the org landlord record once that page supports organisations.
- Individual landlord cards are unchanged (LRN, email, phone, contact address + "View landlord record" link).

- The org "View landlord record" link currently targets the shared local council landlord-record page, which does not yet render organisations (it casts to `IndividualLandlord`). Linking it correctly is deferred to PDJB-1432 (TODO in `PropertyDetailsLandlordViewModelBuilder.kt`).
- No screenshot attached — the new single-page integration test drives the exact local council page and asserts the org card rendering, so behaviour is verified end-to-end.

## Checklist

- [x] Single page integration test added covering an org landlord on the local council Landlords tab (uses existing seed property 48)
- [x] TODO comments referencing this JIRA ticket have been searched for and removed (`PropertyDetailsLandlordViewModelBuilder.kt`)
- [x] This feature is not behind a feature flag. This mirrors the landlord-facing card, which already renders org co-landlords unconditionally; org landlords only exist where the existing org-registration flag is/was enabled, and the change removes a crash rather than introducing new user-facing behaviour where none should be.

### DIFF
--- a/src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsLandlordViewModelBuilder.kt
+++ b/src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsLandlordViewModelBuilder.kt
@@ -91,6 +91,15 @@ class PropertyDetailsLandlordViewModelBuilder {
                             SummaryCardViewModel(
                                 title = landlord.name,
                                 summaryList = buildLocalCouncilOrgCardRows(landlord),
+                                actions =
+                                    listOf(
+                                        // TODO: PDJB-1432: Link this to the landlord record for an org landlord
+                                        SummaryCardActionViewModel(
+                                            text = "propertyDetails.landlordDetails.registeredLandlords.viewLandlordRecord",
+                                            url = landlordDetailsUrlProvider(landlord),
+                                            opensInNewTab = true,
+                                        ),
+                                    ),
                             )
 
                         else -> throw IllegalArgumentException("Unknown landlord type")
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsLandlordViewModelBuilder.kt
```diff
@@ -70,26 +70,45 @@ class PropertyDetailsLandlordViewModelBuilder {
...
-                                    url = landlordDetailsUrlProvider(landlord),
-                                    opensInNewTab = true,
-                                ),
-                            ),
-                    )
+                    when (landlord) {
+                        is IndividualLandlord ->
+                            SummaryCardViewModel(
+                                title = landlord.name,
+                                summaryList = buildLocalCouncilCardRows(landlord),
```
samyou-softwire: `buildLocalCouncilIndividualCardRows` would be a little clearer

samyou-softwire (approved):
LGTM

src/main/kotlin/uk/gov/communities/prsdb/webapp/models/viewModels/summaryModels/PropertyDetailsLandlordViewModelBuilder.kt
```diff
@@ -70,26 +70,45 @@ class PropertyDetailsLandlordViewModelBuilder {
...
-                    check(landlord is IndividualLandlord)
-                    landlord
-                }.sortedBy { it.name }
-                .map { landlord ->
-                    SummaryCardViewModel(
-                        title = landlord.name,
-                        summaryList = buildLocalCouncilCardRows(landlord),
-                        actions =
-                            listOf(
-                                SummaryCardActionViewModel(
```
samyou-softwire, line 84: or actually, I think this should be in scope. we won't add the page here but I think it's sensible to add a link to it in this ticket, or a TODO if we don't know what the link will be

[PDJB-1432](https://mhclgdigital.atlassian.net/browse/PDJB-1432) will add the page properly

PR #1716, opened 2026-08-10
PDJB-1469: Org landlord update lead trustee

samyou-softwire:
## Ticket number

[PDJB-1469](https://mhclgdigital.atlassian.net/browse/PDJB-1469)

## Goal of change

lead trustee update journey for org landlords

## Description of main change(s)

adds new controller for the journey

adds journey. this one is simple as it's just journey -> cya page -> confirmation step

adds skeleton CYA step for [PDJB-1470](https://mhclgdigital.atlassian.net/browse/PDJB-1470)

saves to database

adds relevant tests

## Anything you'd like to highlight to the reviewer?

merges into [PDJB-1237](https://mhclgdigital.atlassian.net/browse/PDJB-1237) #1710  as it reuses a lot of the same mechanism for updating lead trustee

## Checklist

Delete any that are not applicable, and add explanation below for any that are applicable but haven't been done

- [x] Controller tests for any new endpoints, including testing the relevant permissions
- [x] A new journey integration test has been added for any new journeys
- [x] Test suite has been run in full locally and is passing
- [x] Branch has been rebased onto main and run locally, with everything working as expected (both for your new feature
  and any related functionality)
- [x] TODO comments referencing this JIRA ticket have been searched for and removed - if a future PR will address them,
  mention that here
- [x] QA instructions have been added to the ticket (particularly if this is the last PR required to complete the ticket)
- [x] This feature is behind a feature flag. I've checked that there will be no change in function if the feature flag is disabled

PR #1726, opened 2026-08-11
PDJB-1239: Add organisation charity update journey for org landlords

Bill-Haigh:
## Ticket number

PDJB-1239

## Goal of change

Org landlords can start a journey to update their organisation's charity registration details from their landlord record.

- Activates the "Change" link on the Organisation charity row of the org landlord record
- Adds an update journey that walks the existing charity registration questions (is registered charity → charity registered with → charity number)
- The journey currently ends on a placeholder page — the check your answers page, persistence and confirmation email are deferred to PDJB-1463

- The journey reuses the registration `OrgCharityTask` via the journey builder's `task(...)` DSL rather than redefining the regulator branching, following the PDJB-1234 (update organisation name) precedent.
- No prefill of existing answers — that lands with the check your answers page in PDJB-1463.
- `checkUserIsOrganisationLandlord()` is now duplicated verbatim in this controller and `UpdateOrganisationLandlordNameController`. Left as-is for now; worth extracting once PDJB-1238/PDJB-1444 add the third and fourth copies.
- PDJB-1463 will need to add a `transactionSubmitButton` to the real commit step, otherwise this journey never registers as a transaction for metrics.
- The Figma design shows a "Landlord record" section caption above each question. This has been deliberately omitted as no other update journey uses one — flagging for the designer rather than diverging from the rest of the service.

## Checklist

  mention that here


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/UpdateOrganisationLandlordCharityController.kt
```diff
@@ -0,0 +1,77 @@
...
+
+@PrsdbController
+@RequestMapping(UPDATE_ORG_CHARITY_ROUTE)
+@PreAuthorize("hasRole('LANDLORD')")
+class UpdateOrganisationLandlordCharityController(
+    private val journeyFactory: UpdateOrganisationCharityJourneyFactory,
+    private val userToLandlordService: UserToLandlordService,
+) {
+    @GetMapping("/{*stepPath}")
+    fun getUpdateStep(
```
samyou-softwire, line 35: we should make sure these endpoints are blocked behind org landlords feature flag

samyou-softwire (changes requested):
ah, apologies that the ticket scope was unclear here. the intention is to add the full journey (including updating the details). we can leave a TODO for [PDJB-1463](https://mhclgdigital.atlassian.net/browse/PDJB-1463) but this is only a TODO to make the cya page actually show your answers. we can add the function now after you press 'continue' & commit the changes to the database

see https://github.com/communitiesuk/prsdb-webapp/pull/1716 for an example of this, instead with the org landlord lead trustee task rather than charity. happy to hop on a quick call if it'd be easier

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordService.kt
```diff
@@ -224,6 +224,11 @@ class LandlordService(
             }
         }
 
+        orgLandlordUpdate.isRegisteredCharity?.let { isRegisteredCharity ->
```
samyou-softwire: I'm not sure the additional mechanism is needed here: if isRegisteredCharity is false then the charityRegisteredWith will be null, meaning we could ?.let on charityRegisteredWith

I think it'd be a more useful API to have explicit public methods for the 3 options here (setting to not registered charity, registered charity no number, registered charity with number). then we can do away with these optional params and be confident we're only constructing the landlord update model with valid values. 

this way the actual update code can be simpler since it can be certain somebody didn't call with registered charity as false and then put in a charity number anyway

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/updateLandlordDetailsPages/OrgCharityNumberEnglandAndWalesFormPageUpdateLandlordDetails.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.microsoft.playwright.Page
+import uk.gov.communities.prsdb.webapp.controllers.UpdateOrganisationLandlordCharityController.Companion.UPDATE_ORG_CHARITY_ROUTE
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.components.PostForm
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.components.TextInput
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.basePages.BasePage
+import uk.gov.communities.prsdb.webapp.journeys.landlordRegistration.stepConfig.OrgCharityNumberEnglandAndWalesStep
+
+class OrgCharityNumberEnglandAndWalesFormPageUpdateLandlordDetails(
+    page: Page,
+) : BasePage(page, "$UPDATE_ORG_CHARITY_ROUTE/${OrgCharityNumberEnglandAndWalesStep.ROUTE_SEGMENT}") {
```
samyou-softwire: we should ideally make a abstract shared base page that can be used on both the reg & update journey; they are identical bar the URL

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/updateLandlordDetailsPages/OrgCharityRegisteredWithFormPageUpdateLandlordDetails.kt
```diff
@@ -0,0 +1,26 @@
...
+import uk.gov.communities.prsdb.webapp.constants.enums.CharityRegulator
+import uk.gov.communities.prsdb.webapp.controllers.UpdateOrganisationLandlordCharityController.Companion.UPDATE_ORG_CHARITY_ROUTE
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.components.PostForm
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.components.Radios
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.basePages.BasePage
+import uk.gov.communities.prsdb.webapp.journeys.landlordRegistration.stepConfig.OrgCharityRegisteredWithStep
+
+class OrgCharityRegisteredWithFormPageUpdateLandlordDetails(
+    page: Page,
+) : BasePage(page, "$UPDATE_ORG_CHARITY_ROUTE/${OrgCharityRegisteredWithStep.ROUTE_SEGMENT}") {
```
samyou-softwire: same here

src/test/kotlin/uk/gov/communities/prsdb/webapp/integration/pageObjects/pages/updateLandlordDetailsPages/OrgIsRegisteredCharityFormPageUpdateLandlordDetails.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.microsoft.playwright.Page
+import uk.gov.communities.prsdb.webapp.controllers.UpdateOrganisationLandlordCharityController.Companion.UPDATE_ORG_CHARITY_ROUTE
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.components.PostForm
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.components.Radios
+import uk.gov.communities.prsdb.webapp.integration.pageObjects.pages.basePages.BasePage
+import uk.gov.communities.prsdb.webapp.journeys.landlordRegistration.stepConfig.OrgIsRegisteredCharityStep
+
+class OrgIsRegisteredCharityFormPageUpdateLandlordDetails(
+    page: Page,
+) : BasePage(page, "$UPDATE_ORG_CHARITY_ROUTE/${OrgIsRegisteredCharityStep.ROUTE_SEGMENT}") {
```
samyou-softwire: same here

src/main/kotlin/uk/gov/communities/prsdb/webapp/services/LandlordService.kt
Bill-Haigh: So I have left the `?.let` as is as I think if we were to update from charity = yes to charity = no the regulator / number would persist? The isTrust uses the same pattern. 

Have updated the rest of the comments.

PR #1728, opened 2026-08-11
PDJB-1244: Org landlord update main contact

samyou-softwire:
## Ticket number

[PDJB-1244](https://mhclgdigital.atlassian.net/browse/PDJB-1244)

## Goal of change

page view

Summary of the changes made. These should focus on the functionality that you've changed rather than the actual code
changes - those will be clear from the PR.
E.g. Prefer "Adds new endpoint for uploading gas safety certificates to s3" to "Adds new `UploadGasSafetyCertificate`
controller that accepts a `UploadFileRequest` and uses the `fileUploadService` to upload the file to s3"

## Checklist

- [ ] Screenshots of any UI changes have been added
- [ ] A new journey integration test has been added for any new journeys
- [ ] New email templates have been added to `/src/main/kotlin/resources/emails/emailTemplates.json`
  mention that here
- [ ] Seed data has been updated as needed for your feature to be tested without having to e.g. register a new property
- [ ] `NftDataSeeder` has been updated to reflect any changes to the database schema
- [ ] Any special release instructions (e.g. the database will need resetting, emails to be added/removed) have been to the relevant release
  ticket on the Jira board, referencing your ticket number
- [ ] This feature is behind a feature flag. I've checked that there will be no change in function if the feature flag is disabled
- [ ] This feature is not behind a feature flag. I've checked that this is appropriate and we're happy with this code becoming live as soon as we release


src/main/kotlin/uk/gov/communities/prsdb/webapp/controllers/UpdateOrganisationMainContactController.kt
```diff
@@ -0,0 +1,81 @@
...
+    fun postUpdateStep(
+        principal: Principal,
+        @PathVariable stepPath: String,
+        @RequestParam formData: FormData,
+    ): ModelAndView {
+        checkUserIsOrganisationLandlord()
+        return dispatchJourneyStep(stepPath, principal) { postStepModelAndView(formData) }
+    }
+
+    private fun checkUserIsOrganisationLandlord() {
```
danielledias-MT, line 54: This permission check appears in multiple places. Could we extract it into a shared helper (e.g `getCurrentOrganisationLandlordOrThrow()`  on `UserToLandlordService` or a small auth helper service) and reuse it here?                                                                                      

That would avoid duplicating the type-check + forbidden response logic and keep behaviour/error messaging consistent.

What do you think?

danielledias-MT: Happy for this to be a separate tech debt story

samyou-softwire: I'm not so sure; in this simple case the method isn't very long, and it feels fine to have the controller in charge of throwing (and therefore its own error message)

I think I'll leave as-is. as mentioned in dev standup we may move to commonise the entire controller at some point

